PR #24, opened 2020-02-19
Restore testing on Windows via Github Actions

phargogh:
This PR restores automated testing functionality to InVEST via Github Actions.  The core of this is in the github actions workflow `.github/workflows/tests.yml`, but I also modified a few other files along the way.  Here's a summary of changes:

1. Tests are established across a build matrix of {32-bit python, 64-bit python}, {python 3.6, python 3.6} {model tests, UI tests} on Windows.  Appveyor still performs binary builds.
2. Before tests are run, a linux-based linting step is performed.  The build will fail if there are any sufficiently-serious linting errors.  (example: https://github.com/phargogh/invest/runs/456419021?check_suite_focus=true)
3. I've attempted to make effective use of caching in our builds, but this hasn't clearly affected runtime one way or another.
4. The Makefile has been updated to extract version strings and repo information from `git` rather than `hg`.
5. A few cython language levels have been defined in response to a build error.
6. Two tests were updated to address OS-specific line endings.
7. Parts of the InVEST fisheries model were updated to correctly compare strings.  This was in response to better reporting of linter errors.
8. Some python2 imports were removed in response to linter errors.


src/natcap/invest/ui/usage.py
```diff
@@ -11,14 +11,8 @@
...
 
-try:
-    from urllib.request import urlopen, Request
-    from urllib.parse import urlencode
-    string_types = (str,)
-except ImportError:
-    from urllib2 import urlopen, Request
-    from urllib import urlencode
-    string_types = (str, unicode)
+from urllib.request import urlopen, Request
```
dcdenu4: Good to get some Python 2.7 stuff out of there!

tests/test_reporting.py
```diff
@@ -106,10 +107,16 @@ def test_generate_report_dict_script(self):
 
         reporting.generate_report(args)
 
-        pygeoprocessing.testing.assert_text_equal(
-            args['out_uri'],
-            os.path.join(
-                REGRESSION_DATA, 'html_reports', 'report_dict_script.html'))
+        regression_path = os.path.join(
+            REGRESSION_DATA, 'html_reports', 'report_dict_script.html')
+        for source_line, regression_line in itertools.zip_longest(
```
dcdenu4: This is fun and a creative solution!

appveyor.yml
```diff
@@ -16,19 +16,6 @@ environment:
     # https://www.appveyor.com/docs/windows-images-software/#python
 
     - PYTHON: "C:\\Python37"
-      TARGET: "test"
```
dcdenu4: Hey James, I'm still unfamiliar with the appveyor stuff. Could you just mention why these lines are being pulled? Thanks!

davemfish:
> I did see that a check passed for appveyor builds in the conversation tab / thread @davemfish , if that's what you were hoping for? It is kind of hidden. I'll follow along with the comment that Dave left!

Ah yes, I see that now. Thanks for pointing it out!

src/natcap/invest/ui/usage.py
phargogh: 🎉

tests/test_reporting.py
phargogh: Gotta make sure we don't ignore any lines!

appveyor.yml
phargogh: Yeah, so the main idea behind pulling these lines is that this PR moves most of the test jobs to github actions.  The main reason for the move is that while AppVeyor offers free builds for open-source projects, they only allow one job to run at a time.  As a result, the full InVEST test suite within AppVeyor takes almost an hour, and if you have any other build jobs running at all they start queueing.

By moving the tests to github actions, we're able to run our _entire_ test suite (including GUI tests!) in parallel, all while allowing the appveyor-based builds to happen independently.

src/natcap/invest/ui/usage.py
phargogh: I'm sure there's more to remove, though :)

appveyor.yml
dcdenu4: Ahhh, I see, I see. Thanks!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,131 @@
...
+
+              # Taken from https://github.com/actions/cache/blob/master/examples.md#python---pip
+            - name: Restore pip cache
+              uses: actions/cache@v1
+              with:
+                  path: ~\AppData\Local\pip\Cache
+                  key: ${{ runner.os }}-py${{ matrix.python-version }}-${{ matrix.python-architecture }}-pip-${{ hashFiles('**/requirements*.txt') }}-model-tests
+
+            # NOTE: It takes twice as long to save the sample data cache
+            # as it does to do a fresh clone (almost 5 minutes vs. 2.5 minutes)
```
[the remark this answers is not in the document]
phargogh: Yep, you're absolutely right about the redundancy here.  I went with option 1 in rev b998e1d and moved the `make validate_sampledata` step to its own build job outside of the build matrix.  With this task able to happen in parallel with the other jobs, the total build time dropped 22% from about 14 minutes to about 11 minutes. 🎉 

It looks like we had the sampledata validation tests in this repo because they do a `from natcap.invest import datastack`.  We could, of course, move these validation tests over to the other repo, but I think it makes sense to keep them here and in sync with any changes we're making to validation itself.

phargogh:
OK!  @davemfish @dcdenu4 , Looks like tests are passing across AppVeyor and Github Actions.  @davemfish , I in particular responded (in prose and in code) to your comment about the sampledata validation, with the final result being a 22% reduction in build runtime and a new build job just for the sampledata validation.

Let me know what you think, and happy to take another look!

.github/workflows/tests.yml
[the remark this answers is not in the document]
davemfish: Cool @phargogh , I'm happy with this!

FYI I think marking the conversation as resolved hid it from the Conversation page, which is not so convenient.  Overall I wish github PRs had fewer bells & whistles like this. Maybe I'll start liking it more eventually.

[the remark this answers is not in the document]
phargogh: Ah, sorry @davemfish , that was my mistake marking the conversation as resolved.  Sorry about that!

phargogh:
@dcdenu4 , no rush on this at all, but if you approve of the changes in the PR, could you add your approval?  And then since you'd be the last person to approve, could you merge the PR if and when you approve?  Thanks!

PR #33, opened 2020-02-25
Bugfix/cli 25 newline in log

davemfish:
This PR makes the logging from `cli.py` identical to that from `model.py` when a model starts execution. This actually has implications for `datastack.extract_parameters_from_logfile`, which expects to find a line starting with "Arguments".

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,9 @@
 
 Unreleased Changes
 ------------------
+* Fixed a bug in CLI logging where logfiles created by the CLI were 
+  incompatible with the ``natcap.invest.datastack`` operation that
+  allows the UI to load model arguments from logfiles.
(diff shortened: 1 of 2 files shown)


davemfish:
> Thanks, Dave! This is a really important one and thanks for catching it. Since this is a user-facing change, would you be willing to add a note in `HISTORY.rst`? Looks great otherwise!

Yes, good call. I was thinking this change in the logfile was not worth noting for users, but actually the bugfix for loading parameters is very user-facing. Noted in HISTORY and updated PR!

phargogh:
I see that [one of the tests is failing](https://github.com/natcap/invest/pull/33/checks?check_run_id=468080170), but I'm going to chalk that up to a flaky UI issue and have created #36 so that we remember to address it someday and merge this.   Thanks, @davemfish !

Issue #37, opened 2020-02-26
Move the InVEST User's Guide to GitHub


phargogh:
The InVEST User's Guide has recently been migrated to GitHub (https://github.com/natcap/invest.users-guide) and we'll need to  make sure that this is reflected in the Makefile.

phargogh:
This is mostly ready to go now that the repo has been migrated and can be rebuilt on actions on demand.

phargogh:
this is a test mention to myself. @phargogh

phargogh:
Pending #37

phargogh:
Merged with PR #38

PR #38, opened 2020-02-26
Updating Makefile to point to new git home. #37

phargogh:
This PR updates the `Makefile` to use the new git home of the user's guide.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,7 +2,9 @@
 
 Unreleased Changes
 ------------------
-* Fixed a bug in CLI logging where logfiles created by the CLI were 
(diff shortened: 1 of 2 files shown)


phargogh:
@dcdenu4 yeah, I've been seeing those annotations as well.  From what I can tell, I think GitHub is picking up on the fact that the UI tests include some where an exception is being printed to the logfile as a part of the run (which is an expected part of the test itself).  The tests that are logging these exceptions are behaving as expected, so I think we can safely ignore them.

Come to think of it, it might be a better choice to only include the logging when there's a test failure and not print out everything!  That would prevent github actions from annotating a false positive.

PR #40, opened 2020-02-26
Bugfix/binary builds on GitHub actions

### LINKED ISSUE #39: PyInstaller can't find DLLs on Windows
When building the InVEST binary build on github actions for #26 , I noticed that about 7.5 minutes of the build time was being occupied by the pyinstaller part of the build, and, specifically, the part of the build where it is `Looking for dynamic libraries` and generating a pile of warnings that look like this:

```
278982 WARNING: lib not found: libbispeu.KX7AQLB2Z5NFTCADW76YV5UMLBQNDALD.gfortran-win32.dll dependency of C:\hostedtoolcache\windows\Python\3.7.6\x86\lib\site-packages\scipy\interpolate\_fitpack.cp37-win32.pyd
281314 WARNING: lib not found: libdfitpack.EM6D2ZEFEP35UVNTFVXVVRYRWM22ZXEY.gfortran-win32.dll dependency of C:\hostedtoolcache\windows\Python\3.7.6\x86\lib\site-packages\scipy\interpolate\dfitpack.cp37-win32.pyd
299317 WARNING: lib not found: libansari.Q4BAGRNANLWD2YZJOKYPOAUIOLXW2LXK.gfortran-win32.dll dependency of C:\hostedtoolcache\windows\Python\3.7.6\x86\lib\site-packages\scipy\stats\statlib.cp37-win32.pyd
```

Example build job with these warnings: https://github.com/phargogh/invest/runs/471001452?check_suite_focus=true

The build runs so this isn't exactly an error, but it sure would be nice to trim ~7.5 minutes off of our ~20 minute runtime for the build job.

Marking this with a complexity estimate of 13 to indicate the possible worst-case situation for this fix.  It's quite possible that just upgrading pyinstaller would fix the issue, or maybe it's an environment thing (the conda-based mac build doesn't have this issue), but it could also be something deeper.

phargogh:
This PR addresses binary builds by reimplementing them on github actions as well as addressing a few other things that were not included on AppVeyor:

1. AppVeyor-specific configuration has been removed.
2. Windows builds now happen on GitHub Actions
3. Configuration is also provided to build mac binaries on GitHub Actions
4. Once the Windows Build executes successfully, a new workflow will _also_ run the `invest-autotest.py` script on the built binaries.

It's also worth noting that the runtime of the windows binary build is still not great (about 20 minutes), and takes about twice as long to complete as the other build/test processes.  The root cause of this is somewhere within pyinstaller and has to do with pyinstaller not finding a bunch of the binaries that it would like to find.  I've created #39 to address this timing issue later on.  At least with GH Actions we're able to run all of these in parallel.

I was optimistic that we might be able to run the InVEST autotest script on this, but I've decided to postpone adding this until later, especially if we can resolve #39 .  The snippet at https://gist.github.com/phargogh/6c2ae5473298ac8082af7de438583ec0 has the YML I was using in the binary build YML file to attempt to run `invest-autotest.py`.

ALSO, to get these builds working on our forks, there are two secrets you'll need to add to your repo's settings: `GOOGLE_SERVICE_ACC_KEY` and `STANFORD_CERT_KEY_PASS`.  Let me know and I'll help you set them up!


phargogh:
Ah, yeah, those binary build failures look like they were because I had forgotten to add the required secrets (which has been done now in natcap/invest.  I've resolved the merge conflicts and we'll see how the builds/tests fare!

phargogh:
Aha!  @dcdenu4 , it looks like the issue here was that [secrets are not passed to workflows in a PR](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow)!  I've updated the workflow to only sign and deploy binaries if the `GOOGLE_SERVICE_ACC_KEY` is provided.  Either way, artifacts are still archived on github actions itself.

phargogh:
OK!  @dcdenu4 looks like all the builds passed except for AppVeyor, but that's because I've removed the appveyor configuration in this PR :)

phargogh:
That's a great point about the releases ... we're going to have to completely rethink how we do releases with this move for several reasons:

1. GH Actions doesn't allow for secrets to be included in PR builds (for obvious security reasons)
2. Git itself doesn't track tags in the same way in the source tree as mercurial, so the tagging process will have to happen outside of a PR.

Because of the second issue in particular, it's pretty clear that tagging the repository has to happen outside of the PR process, and could take place either through the creation of a GitHub release or by tagging the target rev and pushing the tags back.  And so because of this, we will need a way to trigger a release build outside of the PR process itself.

Issue #46, opened 2020-03-02
Binary builds failing in upstream


phargogh:
I recently noticed that there have been a few build failures when PRs were being merged into `master`.  Here are a few examples:

* https://github.com/natcap/invest/commit/883ab64468e035700f30b73205b274c4c6bc39ae
* https://github.com/natcap/invest/commit/4fd660cfd4d57b25e8dddb6bd599b647fadc51a0
* https://github.com/natcap/invest/commit/b699329450958ced12661dce384bc0df9f4e9927

These actions runs all seem to be failing on the binary build steps.

PR #47, opened 2020-03-02
Using 'w' write mode and setting encoding. RE#46

phargogh:
This PR attempts to address an issue (https://github.com/natcap/invest/issues/46) encountered in the upstream binary builds when generating the redirection links. I can confirm that the updated file fixes the issue on my workstation but I haven't tried it yet on actions since this will only run on the upstream repo.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -157,4 +157,5 @@ jobs:
                   python setup.py install
 
             - name: Run UI tests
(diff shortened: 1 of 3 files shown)


phargogh:
ok!  @davemfish I just pushed a change that will remove the updates to the JSON file per what we talked about in the office and is reflected in #49 .  Further updates to the script itself have been left off until later.  Let me know what you think!

Issue #14, opened 2020-02-13
 Design an InVEST automated release process 


phargogh:
The InVEST release process has been quite manual (documented here: https://bitbucket.org/natcap/invest/wiki/Release Checklist) and pretty involved, requiring a nontrivial amount of engineering time to become confident in our releases before we send it out.

In short, it would be a fantastic improvement to our team productivity if we could design a process that:

* Instills confidence in our releases by thoroughly testing the final executable. Related tasks for this are:
    * Testing of models via the InVEST user interface would be the most useful thing to do here
    * Building and testing the python packages directly across all supported architectures and versions
* Requires minimal (if any) human interactions in order to complete the release. Related tasks would be:
    * Tagging the UG repository with the correct revision found in Makefile (or possibly failing the build if the rev in Makefile is really out of date?)
    * Updating HISTORY.rst with the correct date
    * Automatically incrementing the InVEST release version number as part of the build process
    * Building the release, including:
        * Uploading artifacts to our storage bucket
        * Uploading python binaries to PyPI
    * Announcing the latest InVEST release to interested audiences via needed channels (email? slack? forums?)

Other related questions we’ll want to consider are:

* Whether the release process above should happen for all commits, or if a release should happen at most once per unit of time (once per day? once per week?)
* Should there be a separate branch (separate from develop) to indicate features that should be released via this automated release process?
* How will errors that come up during an automated release be handled?

phargogh:
I've started this process by working on a couple of scripts to do some of the text processing required for this sort of automation, all written (at present) to `ci/release/`.  Next steps are to figure out how to create a new release object in the repository via the github api during an actions run.

phargogh:
I've added a script to `ci/release`/tag-github-repo.py` to tag a target github repository.

The best thing to do at this point will be to work up a prototype on a sample repository on my own account.  This prototype should have two workflows:

1. A workflow that will do a 'build' (maybe zip something up or something like that). This should execute on every push, PR and Release.  If the triggering event is a release, the build should also upload its artifact to the correct release in the correct repository.
2. A workflow that will create a github release, but only if there are new commits on `master`.
3. I'm concerned that all of this will cause release to be automatically created on forks as well.  This prototype should only trigger the creation of releases IFF on the main repository, not on a fork.

phargogh:
- [ ] As an additional functional requirement, we should not proceed with a bugfix release if the builds/tests on `master` are failing.

phargogh:
[...]

PR #54, opened 2020-03-09
Update LS calculations to account for MFD flow

phargogh:
This PR updates the `LS` calculations to account for Multiple Flow Direction routing, replacing the `x` term (aspect, in radians) that was previously used.  The approach included here removes `x` from `LS` calculations, creates a new raster of the weighted mean of proportional flow for each pixel with flow, and then includes this raster in the `RKLS` calculations.

I've made an attempt to update the user's guide in a reasonable and accurate way, but I'd also very much appreciate a review of that as well.


src/natcap/invest/sdr/sdr.py
```diff
@@ -880,7 +896,8 @@ def rkls_function(ls_factor, erosivity, erodibility, stream):
     # aligned with LULCs
     pygeoprocessing.raster_calculator(
         [(path, 1) for path in [
-            ls_factor_path, erosivity_path, erodibility_path, stream_path]],
+            ls_factor_path, erosivity_path, erodibility_path, stream_path,
+            avg_flow_path]],
```
richpsharp: change to `avg_aspect`?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,141 @@ def calculate_sediment_deposition(
...
+    cdef int n_cols, n_rows
+    n_cols, n_rows = flow_direction_info['raster_size']
+
+    cdef _ManagedRaster mfd_flow_direction_raster = _ManagedRaster(
+        mfd_flow_direction_path, 1, False)
+
+    cdef _ManagedRaster average_flow_raster = _ManagedRaster(
+        target_average_flow_path, 1, True)
+
+    cdef int* neighbor_flows = [0, 0, 0, 0, 0, 0, 0, 0]
```
richpsharp: These aren't "flows" so much as "weights"... In other places in this module "flow" refers to the flux that's flowing through the pixel. Could we change the nomenclature here too to be consistent? Like weights if it's those 8 bit values from the 64 bit and then "aspect" when calculating the sin+cos values?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,141 @@ def calculate_sediment_deposition(
...
+                    neighbor_col = seed_col + ROW_OFFSETS[neighbor_index]
+                    if neighbor_col < 0 or neighbor_col >= n_cols:
+                        continue
+
+                    flow_in_direction = (seed_flow_value >> (
+                        neighbor_index * 4) & 0xF)
+                    flow_sum += flow_in_direction
+                    neighbor_flows[neighbor_index] = flow_in_direction
+
+                flow_value_weighted_average = -1.0
```
richpsharp: I'd call this `aspect_weighted_average`

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,141 @@ def calculate_sediment_deposition(
...
+        xoff = offset_dict['xoff']
+        yoff = offset_dict['yoff']
+
+        LOGGER.info('Average proportional flow %.2f%% complete', 100.0 * (
+            (seed_row * seed_col) / float(n_cols * n_rows)))
+
+        for row_index in range(win_ysize):
+            seed_row = yoff + row_index
+            for col_index in range(win_xsize):
+                flow_sum = 0
```
richpsharp: `weight_sum` perhaps?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,141 @@ def calculate_sediment_deposition(
...
+                        if neighbor_index % 2 == 0:
+                            flow_length = 1.0
+                        else:
+                            flow_length = <float>cmath.M_SQRT2
+
+                        # Flow_length will always be greater than 0 since it
+                        # can only be 1 or sqrt(2).
+                        with cython.cdivision(True):
+                            flow_value_weighted_average += <float>(
+                                (1.0 / flow_length) * proportional_flow)
```
richpsharp: Just to save a division here, I'd not calculate `proportional_flow` on line 685 but instead use `neighbor_flows[neighbor_index]` directly in line 699 below and then outside the loop that starts at line 674, divide `flow_value_weighted_average` by `flow_sum`. That'll save you 7 float divisions per cell and be a tich more numerically accurate.

Also changing all the "flow" to "weight" or "aspect" depending on what it is. :)

phargogh: Ah, good call here!  Patched in faeac4ee2d6e25c24793835d9664e5e86dbfd431

richpsharp:
Hey @phargogh, did you want me to look at this again? I can't tell if github can tell me when I need to re-review the review....

phargogh:
@richpsharp not quite yet, but thanks for checking in!  There are still a couple more things left to take a look at.  Should be ready today, though.  I'll ping you again and re-request changes when it's ready!

phargogh:
OK!  @richpsharp I just re-requested a review from you as I believe this is now ready.  The changes implemented here include:

1. I was able to further optimize the cython loop so that it now runs on the Gura sample data in about 41% of the time that it did at the beginning of this PR process.
2. Variables and everything have been renamed to use `aspect` rather than `flow`.
3. The cython-calculated average aspect is now used directly in the `LS` calculations, which addresses the issue around thresholding at 333m.

And sorry, there's also some noise in the commit history from attempting to address an issue with `zip` not being available through `choco` and failing the windows build.

phargogh:
@richpsharp , just hoping to ping you again on this.  No rush, but just wanted to be sure this PR didn't fall through any cracks!

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -392,7 +399,8 @@ def calculate_sediment_deposition(
         e_prime_path)['nodata'][0]
     cdef int col_index, row_index, win_xsize, win_ysize, xoff, yoff
     cdef int global_col, global_row, flat_index, j, k
-    cdef int seed_col, seed_row
+    cdef int seed_col = 0
+    cdef int seed_row = 0
```
richpsharp: Wow, how did this even work before?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+    LOGGER.info('100% complete')
     sediment_deposition_raster.close()
+
+
+def calculate_average_aspect(
+    mfd_flow_direction_path, target_average_aspect_path):
+    """Calculate the Average Proportional Flow from MFD.
+
+    Calculates the weighted average of proportional flow for every MFD flow
+    direction pixel provided.
```
richpsharp: Can you change this to "Calculate the average aspect ratio weighted by proportional flow direction." ?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+                if seed_flow_value == mfd_flow_direction_nodata:
+                    continue
+
+                weight_sum = 0
+                aspect_weighted_sum = 0
+                for neighbor_index in range(8):
+                    # We only need to check the neighbors relative to the
+                    # boundaries if we're in the first or last rows and
+                    # columns.
+                    if seed_row == 0 or seed_row == n_rows:
```
richpsharp: Should the second predicate be `seed_row == n_rows -1`?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+                        neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
+                        if neighbor_row < 0 or neighbor_row >= n_rows:
+                            continue
+
+                    if seed_col == 0 or seed_col == n_cols:
```
richpsharp: Same here `... == n_cols -1`?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+                weight_sum = 0
+                aspect_weighted_sum = 0
+                for neighbor_index in range(8):
+                    # We only need to check the neighbors relative to the
+                    # boundaries if we're in the first or last rows and
+                    # columns.
+                    if seed_row == 0 or seed_row == n_rows:
+                        neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
+                        if neighbor_row < 0 or neighbor_row >= n_rows:
+                            continue
```
richpsharp: Don't you need to check this even if you're not on the first or last row?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+
+                    if seed_col == 0 or seed_col == n_cols:
+                        neighbor_col = seed_col + ROW_OFFSETS[neighbor_index]
+                        if neighbor_col < 0 or neighbor_col >= n_cols:
+                            continue
```
richpsharp: Same here, don't you need to check this even if you're not on the first or last column?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+                    # We only need to check the neighbors relative to the
+                    # boundaries if we're in the first or last rows and
+                    # columns.
+                    if seed_row == 0 or seed_row == n_rows:
+                        neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
+                        if neighbor_row < 0 or neighbor_row >= n_rows:
+                            continue
+
+                    if seed_col == 0 or seed_col == n_cols:
+                        neighbor_col = seed_col + ROW_OFFSETS[neighbor_index]
```
richpsharp: Should this be `COL_OFFSETS`?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -557,4 +567,136 @@ def calculate_sediment_deposition(
...
+                        if neighbor_col < 0 or neighbor_col >= n_cols:
+                            continue
+
```
richpsharp: Can the whole section above be:
```
neighbor_col = seed_col + COL_OFFSETS[neighbor_index]
if neighbor_col < 0 or neighbor_col == n_cols:
    continue
neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
if neighbor_row < 0 or neighbor_row == n_rows:
    continue
```

src/natcap/invest/sdr/sdr.py
```diff
@@ -711,57 +726,62 @@ def execute(args):
...
 
     LS factor as Equation 3 from "Extension and validation
     of a geographic information system-based method for calculating the
     Revised Universal Soil Loss Equation length-slope factor for erosion
     risk assessments in large watersheds"
 
+    Note that to account for multiple possible flow directions, the LS factor
+    described above has been adapted.  The aspect term ``x`` now represents the
+    weighted average of proportional flow from pixel ``i``.  Values for ``x``
+    are created by ``sdr_core.calculate_average_aspect``.
```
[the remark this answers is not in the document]
phargogh: Patched in 990e3864

src/natcap/invest/sdr/sdr_core.pyx
phargogh: How do you mean?

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Patched in d239882b (which also removes some unused variables).

src/natcap/invest/sdr/sdr_core.pyx
richpsharp: Because I'd use it directly in a logging message below without it being initialized! It'd get set later but the first time through would have made some weird output that I never noticed.

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Yep! ff878776

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Also patched in ff878776

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Well that's a sloppy mistake!  Yes, it should be. 5247c349

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Of course it can be (and it was originally).  The additional checking came out of profiling the file line-by-line.  Having the extra conditional (`if seed_col == 0 or seed_col == n_cols:`), while a bit uglier, did seem to reduce the overall runtime.

So, I timed this out!  Here are the mean times for these different implementations running on the pit-filled (via pygeoprocessing) Willamette DEM.  Tests were run over 100 repetitions.  Times reported are the mean time per call to `sdr_core.calculate_average_aspect`:

```python
# Implementation 1: This is the implementation I had in the PR.
# Mean time per repetition: 0.32161s
if seed_row == 0 or seed_row == (n_rows -1):
    neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
    if neighbor_row < 0 or neighbor_row >= n_rows:
        continue
if seed_col == 0 or seed_col == (n_cols -1):
    neighbor_col = seed_col + COL_OFFSETS[neighbor_index]
    if neighbor_col < 0 or neighbor_col >= n_cols:
        continue
```

```python
# Implementation 2: This allows us to test whether the initial seed checks are really worth it.
# Mean time per repetition: 0.32169s
if seed_row == 0 or seed_row == (n_rows -1):
    neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
    if neighbor_row == -1 or neighbor_row == n_rows:
        continue
if seed_col == 0 or seed_col == (n_cols -1):
    neighbor_col = seed_col + COL_OFFSETS[neighbor_index]
    if neighbor_col == -1 or neighbor_col == n_cols:
        continue
```

```python
# Implementation 3: This is what was requested by @richpsharp 
# Mean time per repetition: 0.327s
neighbor_col = seed_col + COL_OFFSETS[neighbor_index]
if neighbor_col < 0 or neighbor_col == n_cols:
    continue
neighbor_row = seed_row + ROW_OFFSETS[neighbor_index]
if neighbor_row < 0 or neighbor_row == n_rows:
    continue
```

```python
# Implementation 4: Taking @richpsharp 's suggestion a step further
# Mean time per repetition: 0.32152s
neighbor_col = seed_col + COL_OFFSETS[neighbor_index]
if neighbor_col == -1 or neighbor_col == n_cols:
    continue
neighbor_row = seed_col + ROW_OFFSETS[neighbor_index]
if neighbor_row == -1 or neighbor_row == n_rows:
    continue
```
Interestingly, while the line-profiling might have indicated that implementations 1 or 2 should have been faster, they were not.  I would have expected implementations 2 and 4 to have been about the same runtime based on what line-profiler was telling me, but the timings here would indicate that this routine is fastest when we're only checking for exactly certain row/col indexes in the most concise way.

So, implementation 4 is what's included in 8ac25750.

phargogh:
@richpsharp, thanks for taking another look!  I believe I have addressed all of your comments, and tried timing out some possible solutions to the section comment you had in the inner cython loop.  I think what's in there now is probably about as fast as we'll get it.

richpsharp:
Thanks James! After I approved there was a merge conflict with `HISTORY.rst` which was really easy to take care of myself with github's inline web editor. Cool.

richpsharp:
Aww crud @phargogh , I just noticed this merged into master instead of `release/3.9`. Is that what you intended?

phargogh:
Thanks @richpsharp!  Yes, I had indeed intended on it being merged into `master`.  It's an important fix, but the updated SDR results aren't extremely different from the old SDR (~2.5% reduction in `sed_export` over the Gura sample data), so I thought including it in a bugfix release would be fine.

richpsharp:
@phargogh, I feel like I'd be nervous about merge conflicts/bugs if I was working on SDR in 3.9 and these changes weren't in there... I also feel like there's not a good plan yet about how to handle these things? Is it reasonable to merge this change into 3.9 as well?

phargogh:
@richpsharp , noted!  In the short-term, here's a PR for you to merge `master` into `release/3.9`: https://github.com/natcap/invest/pull/73

In the slightly longer-term, automated PRs in response to bugfix releases are a part of the automation plan (see ongoing comments on #14 ), and this would be straightforward to expand to include other pushes to `master` as well.

And also, you can always do this sort of thing by hand! For example, if you're working on a new feature in SDR (let's say that you're planning to merge it into `release/3.9`) and you know that there's a fix to SDR just merged into `master`, you can always just fetch from upstream and merge upstream/master into your feature branch.  If there were merge conflicts, that would be an opportunity to take care of them, even before a PR into `release/3.9` is created.

richpsharp:
Okay and if there's ever a case where there's some bugfix we merge into `master` that we don't want in the release it'll probably show up as a merge conflict and we'll edit it out right? I can't imagine what that'd be right now, but I'm sure it's gonna need a human to anticipate it if we're otherwise doing automatic PRs.

phargogh:
@richpsharp if there's ever a case like you're describing (where some bugfix in `master` is definitely _not_ wanted in a release branch), we still have flexibility ... automation should not trump flexibility!  Here are a few things to remember about the automation I'm thinking about:

1. **An automated PR is still that ... a PR**. 
    We will have an opportunity to approve or decline it if the merge would include something we don't want.  A PR gives us the opportunity to review changes before making a merge.

2. **We always have the option to handle special cases via the usual PR process**.
    Let's say that there's something merged into `master` that we do not want in `release/3.9`.  In that case, we'll need to handle the merge manually via a `feature` branch.  To do this, make a `feature` branch off of `release/3.9`, merge in `master` (handling the thing we want to avoid including) and then PR `feature` --> `release/3.9`.  After this is complete, any future merges from `master` into `release/3.9` will not include the patch we manually excluded.

3. **We can always reverse an accidental or erroneous merge**
    If something _does_ end up being included in a `release/*` branch that should not be there, we can always use the normal tools available to us including backing out of a revision, reverting a merge, or else fixing things by hand and merging in the fix via a PR process.

Again, the point of automation is to make it so that we don't have to be so involved in steps that are frequently repeated, but I don't want us to be so locked into a specific process that we cannot handle special cases.  I think the automation as planned will still allow us the flexibility to do what is needed when there are special cases.

So I think that our bases are all covered here ... but @richpsharp please let me know if I've missed something!

PR #57, opened 2020-03-12
Clean Python 2 code out of InVEST

### LINKED ISSUE #41: Remove python2 stuff from throughout natcap.invest
When `natcap.invest` supported both python2 and python3, we used the `future` package and a few other things (like conditionally importing `Queue` vs `queue`) for compatibility.  It would be great to trim this to whatever degree is reasonable.

Known parts to this:

- [x] mock --> unittest.mock
- [x] Remove `future` from `requirements.txt`
- [x] Remove `future` and future-defined imports (builtins, others?) throughout the package
- [x] Replace instances of `basestring` with `str` (hra.py, ... )

dcdenu4:
This PR fixes #41 in attempting to clean up Python 2 remnants from source and testing files. 

Things removed:
  - future calls
  - replacing `basetring` with `str`
  - removing `unicode` and using `str` where necessary
  - removing `pyqt4 checks`
  - removing `mock` and using `unittest.mock`
  - six usage
  - u'' and u"" casting
  - Queue for queue

Updated `api-docs/conf.py` to use `unittest.mock` and added `sdr.sdr_core` to be mocked

History has been updated to reflect clean up.

There is also some attempt to do some `pycodestyle` cleaning on testing files. 

Suggestions for more known Python 2 to remove are welcome.


src/natcap/invest/__init__.py
```diff
@@ -10,7 +10,7 @@
 
 LOGGER = logging.getLogger('natcap.invest')
 LOGGER.addHandler(logging.NullHandler())
-__all__ = ['local_dir',]
+__all__ = ['local_dir', ]
```
phargogh: This is trivial, but do we need that trailing comma and space at all?

tests/test_crop_production.py
```diff
@@ -333,4 +334,5 @@ def test_missing_keys_regression(self):
         expected_missing_keys = set(
             self.base_required_keys +
             ['fertilization_rate_table_path'])
-        self.assertEqual(invalid_keys, expected_missing_keys)
\ No newline at end of file
+        self.assertEqual(invalid_keys, expected_missing_keys)
+
```
phargogh: Hey sorry @dcdenu4 I think maybe this is at the end of the file?  If so, doesn't PEP8 call for no blank lines at the end?  Sorry again for the triviality of this!

src/natcap/invest/__init__.py
dcdenu4: I think I added that when I was trying to do some pycodestyle cleaning. I believe it recommended the space after comma.

tests/test_crop_production.py
dcdenu4: Great question, I'll check. Pycodestyle was sometimes suggesting a line at the end and sometimes not... I'll check PEP8

dcdenu4:
Oh geez, my `grep` must not have recursively gone into the sub folders of invest. Sorry about this.

dcdenu4:
@phargogh @davemfish OK, I realize my huge mistake. I already made most the changes you guys were suggesting, I just apparently was on my personal `master` branch... [ facepalm ]. So there's 4 commits sitting on my master branch that should be in here. I thought it odd I wouldn't have caught these. I'm going to try and do some `git` maneuvering to move these commits into this branch for the PR

dcdenu4:
Alright, I added the changes that were accidentally committed under my personal master branch. I *think* these include the suggestions you had. I still need to do some PEP8 updates.

src/natcap/invest/datastack.py
```diff
@@ -30,11 +30,6 @@
...
 import warnings
 
-try:
-    basestring
-except NameError:
-    basestring = str
-
 from osgeo import gdal
 from osgeo import ogr
 import six
```
phargogh: `six` is also a python 2/3 compatibility layer, and I'm sorry I didn't notice this before!  Could you also remove a uses of `six` here in favor of their python3 equivalent?  I suspect it'll just be a matter of an iterator or a string type.

dcdenu4:
@davemfish @phargogh I did a lot more spring cleaning for removal of Python 2 related things. Currently for some reason Mac Binaries are now failing, but Windows seems good. All other checks pass. If anyone has some Mac Binary debugging ideas I'm all ears.

doc/api-docs/conf.py
```diff
@@ -305,23 +305,23 @@
...
+    @classmethod
+    def __getattr__(cls, name):
+        return MagicMock()
 
 # Any time we run a sphinx operation, also generate the API documentation.
 # This should work OK, so long as the required modules (above) are sufficiently
 # mocked up.
-for name in autodoc_mock_imports:
-    sys.modules[name] = Mock()
+sys.modules.update((mod_name, Mock()) for mod_name in autodoc_mock_imports)
```
phargogh: Huh, makes sense that a generator expression would work here, but I had never seen nor tried it!  Very cool!

requirements-docs.txt
```diff
@@ -17,7 +17,7 @@ sphinx
 sphinx-rtd-theme
 nose
 Cython
-mock
+unittest
```
phargogh: I'm a little concerned about this being here.  Since `unittest` is part of the python standard library, I get this when I try to `pip install unittest`:
```
ERROR: Could not find a version that satisfies the requirement unittest (from versions: none)
ERROR: No matching distribution found for unittest
```

I'm concerned that merging this would cause our sphinx autodoc build to fail.

src/natcap/invest/fisheries/fisheries_io.py
```diff
@@ -14,11 +14,6 @@
 from .. import reporting
 from .. import utils
 
-try:
-    unicode
-except NameError:
-    # No unicode in Python 3
-    unicode = str
```
phargogh: Huh, I'm surprised we included this import in this file, despite the fact that `unicode` isn't used anywhere in this file!

requirements-docs.txt
dcdenu4: oops!

dcdenu4:
@phargogh Just pushed the removal of `unittest` from `requirements-docs.txt`. 

I agree with the Mac Build binary issue. I think it makes sense to open that in its own issue like you did.

phargogh:
@dcdenu4 on the subject of mac builds, I just noticed while debugging #63 that [adding a `conda upgrade libwebp` within the activated conda environment might do the trick](https://github.com/natcap/invest/issues/63#issuecomment-601397759)!  Could you try adding that to the mac workflow and see if that resolves the issue?

Or, alternatively @davemfish would you happen to know of a better/more appropriate way to specify that we want `libwebp` to be installed specifically from the `default` channel?

Issue #53, opened 2020-03-06
SDR's LS equation does not account for MFD


phargogh:
When we refactored SDR to use Multiple Flow Direction, we failed to update SDR's RKLS calculations to account for the fact that aspect (a component of the LS equation) should now account for the possibility that a pixel can flow into 8 possible neighbors.  LS does not currently account for this.

After talking about this with Rich, we decided on the following, corrected approach:

## Given
`FV` = "flow value", the packed 64-bit, 8-direction (4 bits per direction) value coming from MFD Flow Direction
`d` = The integer direction in the form:
```
3 2 1
4 x 0
5 6 7
```
`P_i(d)` = The proportion of flow in pixel `i` in direction `d`
`LS_i'` = A modified `LS` equation that removes the `x_d` term entirely
`x_d` = The length of the pixel from `x` to neighbor `d`.  If `d` is even, this will be `1`,  `sqrt(2)` otherwise.

## Calculate the following
0. Determine the proportional flow `P_i(d) = FV(d) / sum(FV(j) for j in {0..7})`
1. The weighted average of the aspect based on flow value: `FV_avg = sum(1/(x_d) * P_i(d) for d in {0..7})`
2. Calculate `LS_i'`
3. Calculate the modified RKLS: `RKLS_i = R_i * K_i * LS_i' * FV_avg`

## TODOs
- [x] Implement
- [x] Test
- [x] Update the User's Guide
- [x] Ask Rich to look over the updated UG
- [x] Note change in `HISTORY.rst`
- [x] PR this into InVEST

phargogh:
Working on this in my fork at branch `bugfix/update-sdr-ls-for-mfd-33`

phargogh:
Pending #54

phargogh:
Pending #54

Issue #72, opened 2020-04-03
UI Tests on master failing consistently


phargogh:
The UI tests on `master` seem to be failing reliably.  Since they keep failing, this means that we should (hopefully) be able to reproduce the issue and fix it.

phargogh:
In the interest of having most of our tests run, I've added `@unittest.skip` to `ui_tests/test_ui_inputs.py:ModelTests`.  Leaving this issue open for now.  See #90

PR #73, opened 2020-04-06
Merge the recent LS updates for MFD into release/3.9

phargogh:
This PR seeks to merge the latest changes on `master` (and the updates to LS from issue #53 in particular, but this will include other fixes as well ) into the InVEST 3.9 release branch.


richpsharp:
@phargogh, so what do we do when the UI tests fail? Looking at the results it looks like rando-thread exceptions like we get sometimes.

phargogh:
@richpsharp the UI tests have been failing _much_ more regularly of late, so much so that I created #72 for it.  I think we should ignore the UI tests for the time being.

The other builds that are failing are the binary builds on master, specifically on a small, late part of `make deploy` that is being addressed in #46 .

So, do we care about failing tests?  I'd say yes, but I think it's also outside the scope of this PR and being addressed in the issues mentioned above.

PR #86, opened 2020-04-20
Minor update to Urban Cooling helptext

phargogh:
This is just a minor correction to the Urban Cooling model UI's help text, plus some bonus updates to docstrings (`Parameters` --> `Args`, per google style conventions) and a few lines that were ever so slightly too long.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,8 @@
 
 Unreleased Changes
 ------------------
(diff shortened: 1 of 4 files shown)


phargogh:
Alrighty @dcdenu4, I just pushed e3935db0 that bumps the UG repo revision.  Turned out that a number of parameters had not been added to the UG chapter.

Issue #81, opened 2020-04-16
CV bug in how interpolate_shore_points deals with MultiPart geometries


davemfish:
This is a small bug in the logic that is supposed to decompose MultiLineStrings into LineStrings. Things break when MultiLineStrings are sent to a `shapely.ops.linemerge`, which expects a list of LineStrings. We are doing that decomposition correctly, but appending the original MultiLineString anyway.

Essentially a chain of `if...if...else` needs to be `if...elif...else`. 

This occurs rarely because the MultiLineStrings only come into existance after an intersection between a LineString and a Polygon, and usually that result is a LineString.

davemfish:
resolved in #89

Issue #80, opened 2020-04-16
CV should fail immediately if shore point creation yields 0 points


davemfish:
I've seen this occur a couple times I think. Here's one case:

https://community.naturalcapitalproject.org/t/coastal-vulnerability-valueerror-cannot-convert-float-nan-to-integer/940/4

There's no need for execution to continue if there are 0 shore points. This could happen if there were no landmass edges intersecting the AOI. I think that's a case which would get past the spatial overlap check during validation. Otherwise I would expect that validation check to prevent this eventuality.

davemfish:
resolved in #89

PR #89, opened 2020-04-21
Bugfix/cv 81 interpolate shore points multilinestrings

davemfish:
This PR resolves issue #81 and #80 . Please see those issues for more details.

### DIFF
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -218,7 +218,7 @@ def execute(args):
     Also quantify the role of habitats in reducing the hazard. Optionally
     summarize the population density in proximity to each shore point.
 
-    Arguments:
+    Args:
         args['workspace_dir'] (string): (required) a path to the directory that
             will write output and other temporary files during calculation.
         args['results_suffix'] (string): (optional) appended to any output
@@ -607,7 +607,7 @@ def prepare_landmass_line_index(
     and saves spatial index of lines for fast intersections. Also
     saves unioned geometry polygons for fast point-in-polygon checks.
 
-    Arguments:
+    Args:
         landmass_vector_path (string): path to polygon vector
         target_polygon_pickle_path (string): path to pickle
             storing shapely polygon geometry.
@@ -692,7 +692,7 @@ def interpolate_shore_points(
         target_vector_path):
     """Create points along a polyline geometry at a specified interval.
 
-    Arguments:
+    Args:
         aoi_vector_path (string): path to polygon vector used to define
             boundaries for adding points.
         landmass_lines_pickle_path (string): path to pickle file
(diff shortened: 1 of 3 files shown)


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -724,25 +724,21 @@ def interpolate_shore_points(
         shapely_line_list = pickle.load(lines_pickle_file)
     lines_in_aoi_list = []
     for shapely_geom in shapely_line_list:
-        # Already clipped landlines to AOI, with max-fetch extent
-        # still want to clip lines by actual AOI, for shore point creation
+        # Already clipped landlines to AOI plus max-fetch extent
+        # still want to clip lines by actual AOI for shore point creation
         if aoi_shapely_prepped.intersects(shapely_geom):
             intersected_shapely_geom = aoi_shapely.intersection(shapely_geom)
             if not intersected_shapely_geom.is_empty:
```
dcdenu4, line 731: Could this be empty since there is an intersection? Shapely documentation indicates that a Point( ) would return True, is that possible here?

tests/test_coastal_vulnerability.py
```diff
@@ -982,6 +982,49 @@ def test_shore_points_on_multi_polygon(self):
...
+        """CV: test RuntimeError raised if zero shore points are created."""
+
+        args = CoastalVulnerabilityTests.generate_base_args(self.workspace_dir)
+        aoi_path = os.path.join(self.workspace_dir, 'aoi.geojson')
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(26910)  # UTM Zone 10N
+        wkt = srs.ExportToWkt()
+        a = 100
+        sampledata.create_vector_on_disk(
+            [Polygon([(-a, -a), (a, -a), (a, a), (-a, a), (-a, -a)])],
```
dcdenu4, line 995: this is great

src/natcap/invest/coastal_vulnerability.py
davemfish: @dcdenu4, the intersection could be empty: `shapely_geom` could be outside of `aoi_shapely`. That's not so uncommon because the landmass lines at this point (`shapely_geom`) were clipped to a bounding box that is aoi extent + max-fetch-distance. 

The intersection could also be a point in the very rare case that `shapely_geom` is a tangent line to `aoi_shapely`. And that's not empty, but is handled below.

davemfish:
Thanks @dcdenu4 , I pushed some changes with the docstring formatting you suggested. And replied to your other comment. Back to yo!

davemfish:
Okay, @dcdenu4 , PR updated! Looking closer, not only was that `is_empty` check redundant, but an empty geom will have `.type` `GeometryCollection`, and so the subsequent logic checking for LineStrings, etc, will pass it through correctly even if an empty geom does somehow exist!  

Also kinda interesting: the `intersects` check is done with the `prepared` shapely geom, cuz that's super fast, but `prepared` geoms have no `intersection` method...so technically the intersection happens with a different object than the one we already check for intersects. They should represent identical shapes, but I'm glad the final solution still handles empty geoms, just in case.

dcdenu4:
Thanks Dave! That is interesting and sneaky about the prepared geoms.

PR #91, opened 2020-04-21
Task/tests 84 tests on macos

davemfish:
This PR adds github Actions jobs for model tests on Python 3.6 and Python 3.7 on macos-latest. 

There are a handful of failing tests right now, but I think I would like to address that in separate issues for those models, rather than adding fixes as part of this PR. Those failures all reproduce on ubuntu as well.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -70,7 +70,8 @@ jobs:
                   path: data/invest-test-data
                   key: git-lfs-testdata-${{ hashfiles('Makefile') }}
 
-            - name: Set up python
+            - name: Set up Windows python
+              if: matrix.os == 'windows-latest'
               uses: actions/setup-python@v1
               with:
                 python-version: ${{ matrix.python-version }}
@@ -87,7 +88,7 @@ jobs:
                   python -m pip install -r requirements.txt -r requirements-dev.txt
                   python setup.py install
 
-            - name: Install dependencies MacOS
+            - name: Set up MacOS python
               if: matrix.os != 'windows-latest'
               uses: goanpeca/setup-miniconda@v1.1.2
               with:
@@ -98,6 +99,9 @@ jobs:
                   python-version: ${{ matrix.python-version }}
                   conda-channels: defaults
                   channels: conda-forge, anaconda
+            
+            - name: Install dependencies MacOS
+              if: matrix.os != 'windows-latest'
               shell: bash -l {0}
               run: |
                   conda upgrade -y pip setuptools
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -104,8 +104,10 @@ jobs:
               shell: bash -l {0}
               run: |
                   conda upgrade -y pip setuptools
-                  conda install Cython requests
-                  make env
+                  conda install Cython requests numpy
+                  python ./scripts/convert-requirements-to-conda-yml.py requirements.txt requirements-dev.txt requirements-gui.txt > requirements-all.yml
+                  conda env update --file requirements-all.yml
+                  python setup.py install
 
             - name: Run model tests
               run: make test
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -96,7 +96,6 @@ jobs:
                   activate-environment: pyenv
                   auto-update-conda: true 
                   python-version: ${{ matrix.python-version }}
-                  python-version: ${{ matrix.python-version }}
                   conda-channels: defaults
                   channels: conda-forge, anaconda
             
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,6 +43,9 @@ jobs:
                 python-version: [3.6, 3.7]
                 python-architecture: [x86, x64]
                 os: [windows-latest, macos-latest]
+                exclude:
+                    - os: macos-latest
+                      python-arch: x86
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -42,7 +42,10 @@ jobs:
             matrix:
                 python-version: [3.6, 3.7]
                 python-architecture: [x86, x64]
-                os: [windows-latest]
+                os: [windows-latest, macos-latest]
```
davemfish, line 45: I added to the os matrix here, rather than making a separate job for mac tests. That lets us re-use the step of checking out test data, but then it also requires a bunch of conditionals checking the os is subsequent steps. I'm not sure if this was a good idea or not.

.github/workflows/build-and-test.yml
```diff
@@ -82,9 +87,34 @@ jobs:
...
+                  conda install Cython requests numpy
+                  python ./scripts/convert-requirements-to-conda-yml.py requirements.txt requirements-dev.txt requirements-gui.txt > requirements-all.yml
+                  conda env update --file requirements-all.yml
+                  python setup.py install
+                  echo ::set-env name=PYTHON::$(which python)
 
             - name: Run model tests
-              run: make test
+              run: |
+                make PYTHON=$PYTHON test
```
davemfish, line 116: In order to have both os tests share this run context, I'm trying to pass the os-specific python to make test. It seems to work, but I wonder if there was a better approach for setting this variable?

.github/workflows/build-and-test.yml
```diff
@@ -82,9 +87,34 @@ jobs:
                   python -m pip install --upgrade pip nose setuptools
                   python -m pip install -r requirements.txt -r requirements-dev.txt
                   python setup.py install
+                  echo ::set-env name=PYTHON::$(where python)
```
davemfish, line 90: Trying to define a global variable with the os-specific python. I wonder if there is a better way.

.github/workflows/build-and-test.yml
phargogh: I think the ideal case here would be that we're using the same setup and test configuration across all of our target OSes, and I think that's a good thing for us to strive for.  To that end, defining the OS as part of the build matrix makes very good sense to me.

.github/workflows/build-and-test.yml
phargogh: Does this set the `$PYTHON` environment variable?  I haven't seen this notation before ... neat if it does!

.github/workflows/build-and-test.yml
phargogh: I think the intent of the `setup-python` step was to do just that ... but I think we're in an interesting case here where we're trying to use the conda python and not just the system python, right?  If that's the case, does `setup-miniconda` set up a `python` binary that's accessible from other steps, much like `setup-python` sets up?

phargogh: I guess what I mean by this is that when we've used `setup-miniconda`, does that mean that when we call `python` in later steps, is that located within the conda environment?  If it is, it might be sufficient to call `make PYTHON=python`, or, if we need to be super specific about the path, we could specify the shell (bash is available on all OSes, I believe), and do a `make PYTHON=$(which python) test`.

davemfish: Good suggestions, it turns out a `setup-miniconda` system actually has both `python` and `python3` on the path. `python` being in the active conda env, and `python3` being `/usr/local/opt/python/bin/python3.7`. That fact, combined with the Makefile defaulting to `PYTHON=python3` on non-windows OS is what was causing problems. 

I pushed a change to get rid of the setting the environment variable, and calling `make PYTHON=python test`

.github/workflows/build-and-test.yml
```diff
@@ -83,8 +88,38 @@ jobs:
...
+            - name: Set up MacOS python
+              if: matrix.os != 'windows-latest'
+              uses: goanpeca/setup-miniconda@v1.1.2
+              with:
+                  update-conda: false
+                  activate-environment: env
+                  auto-update-conda: true 
+                  python-version: ${{ matrix.python-version }}
+                  conda-channels: defaults
+                  channels: conda-forge, anaconda
```
dcdenu4, line 100: Hey @davemfish , this looks likes some duplication between `setup-conda` (which we used to use) and `setup-miniconda` (which we've decided to use going forward). I believe `setup-conda` uses `conda-channels` and `setup-miniconda` uses `channels`. So you could probably drop `conda-channels` here.  Similarly I think that `update-conda` is from `setup-conda` and that `auto-update-conda` is `setup-miniconda`s equivalent. So `update-conda` could be removed as it's probably not being used...

Sorry if I'm missing something and feel free to ignore...

dcdenu4:
Hey @davemfish sorry for jumping in and adding myself to review here, but I just saw some chatter about `setup-miniconda` and thought I'd take a look since I just worked with it via `pygeoprocessing`. I also see this is a Draft! So maybe I shouldn't even be commenting yet... Sorry!

.github/workflows/build-and-test.yml
davemfish: Thanks, @dcdenu4 ! You're input always appreciated. Indeed, this section needs cleaning up.

davemfish:
I'm closing this PR because the branch contains mostly experimentation that didn't end up being a good idea. So I'll re-open a PR from a another branch with a much cleaner history. And I'll copy over some of the discussion so we can pick it back up.

PR #92, opened 2020-04-22
Adding MacOS to the matrix for running model tests.

davemfish:
This PR adds github Actions jobs for model tests on Python 3.6 and Python 3.7 on macos-latest.

There are a handful of failing tests right now, but I think I would like to address that in separate issues for those models, rather than adding fixes as part of this PR. Those failures all reproduce on ubuntu as well.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -109,6 +109,7 @@ jobs:
 
             - name: Run model tests on MacOS
               if: matrix.os != 'windows-latest'
+              shell: bash -l {0}
               run: |
                   which python
                   make PYTHON="/usr/local/miniconda/envs/env/bin/python" test
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -42,7 +42,10 @@ jobs:
             matrix:
                 python-version: [3.6, 3.7]
                 python-architecture: [x86, x64]
-                os: [windows-latest]
+                os: [windows-latest, macos-latest]
+                exclude:
+                    - os: macos-latest
+                      python-architecture: x86
 
         steps:
             - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


davemfish:
@phargogh , quoting you from convo started on the defunct PR:

"I guess what I mean by this is that when we've used setup-miniconda, does that mean that when we call python in later steps, is that located within the conda environment? If it is, it might be sufficient to call make PYTHON=python, or, if we need to be super specific about the path, we could specify the shell (bash is available on all OSes, I believe), and do a make PYTHON=$(which python) test."

.github/workflows/build-and-test.yml
```diff
@@ -83,8 +88,36 @@ jobs:
...
+            
+            - name: Install dependencies MacOS
+              if: matrix.os != 'windows-latest'
+              shell: bash -l {0}
+              run: |
+                  conda upgrade -y pip setuptools
+                  conda install Cython requests numpy
+                  python ./scripts/convert-requirements-to-conda-yml.py requirements.txt requirements-dev.txt requirements-gui.txt > requirements-all.yml
+                  conda env update --file requirements-all.yml
+                  python setup.py install
```
davemfish, line 108: Here, `python` is the conda environment python. But see below, that's not true in the next section.

.github/workflows/build-and-test.yml
```diff
@@ -83,8 +88,36 @@ jobs:
...
+
+            - name: Run model tests on MacOS
+              if: matrix.os != 'windows-latest'
+              run: |
+                  which python
+                  make PYTHON="/usr/local/miniconda/envs/env/bin/python" test
```
davemfish, line 114: Here, `python` is a default system python, hence the need to pass the conda python to `make test`. I don't understand why this changes from the previous section.

davemfish:
Okay @phargogh , I think this is ready for review, again. Thanks for bearing with me. 

The trick with getting `python` to refer to the conda python instead of another system python is in setting the shell in each section where we're calling `python`. And since that doesn't seem to interfere with the Windows setup, the two OS can share a `make test` section.

dcdenu4:
Hey Dave! Yea, the maintainer of setup-miniconda makes a comment about why setting the shell is needed: https://github.com/marketplace/actions/setup-miniconda

> Bash shells do not use ~/.profile or ~/.bashrc so these shells need to be explicitely declared as shell: bash -l {0} on steps that need to be properly activated. This is because bash shells are executed with bash --noprofile --norc -eo pipefail {0} thus ignoring updated on bash profile files made by conda init bash. See Github Actions Documentation and thread.

PR #100, opened 2020-05-01
Bugfix for several tests failing on macos

davemfish:
Hi @phargogh , here's a PR to make a few of our tests cross-platform compatible. Originally failing on Mac OS were a `utils.build_lookup_from_csv` test, a `wind_energy` test related to numerical precision, `pollination` tests related to filepath separators in a txt file of expected filepaths, and `scenario_generator_proximity` tests that are resolved with update test data.

Strangely, I'm still seeing the compilation error on the python 3.6 Mac VM even after merging your fix. I'm not sure how to explain that, but would be happy to update this PR before merging if we can figure it out. It  looks to happen on Mac 10.7, while the python 3.7 Mac VM is on Mac 10.9.

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -159,7 +159,7 @@ jobs:
             - name: Set up conda
               uses: goanpeca/setup-miniconda@v1.1.2
               with:
-                  activate-environment: mac-env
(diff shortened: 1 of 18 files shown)


davemfish:
Woops, just noticed something in here also broke the Mac binary builds. Moving this PR to draft.

PR #131, opened 2020-05-18
Adding workflow for merging master into release branches.

phargogh:
This PR adds a workflow for automatically merging `master` into `release/*` branches when new changes are added to `master`.  Like in pygeoprocessing, the PR will be assigned to the original PR's requester.

Fixes #119

PR #139, opened 2020-05-20
Task/119 propagate requester to release pr assignee

phargogh:
This is a follow-up PR from #131 because I forgot to add a needed file to the PR and that was causing the PR creation to fail.


dcdenu4:
Waiting for test completion before merge.

phargogh:
@dcdenu4 whew, I'm glad you waited!  Everything should be there now, just needed to sneak in an updated filepath.

Issue #144, opened 2020-05-23
Minor updates to bugfix automation


phargogh:
In recent discussions, we've decided to make a few minor changes to how release automation operates:

- [x] In addition to the other checks for whether release automation should trigger a release, the bugfix release automation should also check whether there are any user-facing changes to release, using `HISTORY.rst` as a proxy. If there are no user-facing changes, a release should not be triggered.
- [x] If the latest PR submitted is `github-actions`, it means that the PR last submitted was an automated PR and it should instead be assigned to `$RELEASE_MANAGER`.  See example failed build: https://github.com/natcap/invest/pull/164/checks?check_run_id=743226968
    - [x] If we're using `$RELEASE_MANAGER` across multiple workflows, let's find a way to have that defined in a single place.
- [x] Figure out how to format HISTORY so that `pandoc` can convert correctly.  The 3.8.4 release text was not formatted correctly when converted to Markdown.  Related to #133 
- [x] Use some new automation tag instead of including 'AUTO' in PR titles.
- [x] Make auto release program exit nicely (expected exit), but still send a slack alert with the verdict of whether to do a release
- [x] Although all of the steps for a bugfix release could be walked through by hand to do a manual release, it would be preferable if we could trigger the release automation on demand, either through a script or a web application of some sort to make the API call.  **The objective would be to put the team in control of decision making, not the automation.**
- [x] Update the Release checklist wiki page with these changes.
- [x] ~~1 week updates is too frequent.  We'll move this to every other week.~~  There does not seem to be an obvious way to do this.

phargogh:
Working on this in my fork on `task/144-minor-updates-to-bugfix-automation`

phargogh:
# Plan for improved release automation

## Objective
The software team should be in charge of deciding whether and when to do a release.  When we decide a release should take place, automation should be in place to make everything happen as expected.

## Planned Pause Points

The automation workflow should have a few pause points, opportunities to decide whether a release should proceed.

1. Before the release process starts, we would decide whether a release should take place at all.
2. After the repository has been tagged and binaries have been built, we should have an opportunity to decide whether to publish the release.

## Automation

### Initiate a Release

When we decide to initiate a release, the release manager will run a script locally to initiate the release process.

Assumptions:
* The `ref` that the local repo is checked out to will be used as the base for the `autorelease/3.x.x` branch that the script will create.

Invocation: `./scripts/release-initiate.sh <version>`

What the script will do:

[...]

PR #169, opened 2020-06-11
Task/144 minor updates to bugfix automation

phargogh:
This PR includes a couple of tweaks to the bugfix release automation:

* The `RELEASE_MANAGER` is now defined in a file rather than twice in YML workflow files.
* The bugfix release will now also check HISTORY to make sure that there have been changes since the last release.  If no changes have been made to HISTORY since the release, the release will not take place.
* The release object will now use strict markdown (which renders better on github).


.github/workflows/auto-release-bugfix.yml
```diff
@@ -38,8 +34,24 @@ jobs:
...
+                  # Get the shared environment variables.
+                  # Adds the RELEASE_MANAGER environment variable.
+                  source ./ci/release/RELEASE_MANAGER.env
+
+                  # Don't do a release if there are no changes on HISTORY since
+                  # the last release.  This is done by checking to see if
+                  # HISTORY has been modified between $LATEST_VERSION and
+                  # the commit this build is running on ($GITHUB_SHA).
+                  LATEST_VERSION=$(python setup.py --version | awk -F '.' '{print $1"."$2"."$3}')
+                  if git diff --name-only $LATEST_VERSION $GITHUB_SHA | grep HISTORY.rst > /dev/null
```
davemfish, line 47: I'm sure you tested this conditional and it works, but where would one go to find out what qualifies as truthy for github actions yml? Or is it the rules for the current shell that matter here?

.github/workflows/auto-pr-from-master-into-releases.yml
```diff
@@ -43,6 +43,13 @@ jobs:
                   PR_USERNAME=$(hub pr show -f "%au" $PR_NUM)
                   echo "Latest PR on master ($PR_NUM) was authored by $PR_USERNAME."
 
+                  if [ "$PR_USERNAME" = "github-actions" ]
```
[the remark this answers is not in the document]
phargogh: Yep!  There's only one circumstance that I'm aware of where this condition would be met: if the most recent PR was triggered by the `github-actions` bot, which only happens as part of the bugfix release automation.

.github/workflows/auto-release-bugfix.yml
phargogh: Yeah, it's completely determined by the rules of the current shell.  This is a bash script here, so bash truthiness rules apply.

I still have an outstanding TODO on #144 to refactor these build scripts out of the YML so that they can be more easily run independently of github actions, and I hope that'll help clarify the nature of the scripts as well.

phargogh:
@davemfish I'm not sure what the hangup was on the actions check, but it looks like everything passed!  Thanks for your approval here ... I'll merge the PR.

PR #175, opened 2020-06-19
Bugfix/168 ufrm missing vector unless infrastructure provided

### LINKED ISSUE #167: UFRM: missing infrastructure vector geometry causes crash
A user on the forums (link below) encountered an exception during Urban Flood Risk Mitigation Model when the infrastructure vector had a feature that was missing its geometry.  We should handle this case in a reasonable way.

https://community.naturalcapitalproject.org/t/attributeerror-nonetype-object-has-no-attribute-clone-urban-flood/1051

### LINKED ISSUE #168: UFRM doesn't write its summary vector unless the built_infrastructure vector is provided
When running UFRM without the built infrastructure inputs, the output vector (`flood_risk_services.shp`) is not included in the output directory, despite the stats that are summarized in this vector not being a part of the built infrastructure calculations.

> There should be a shapefile output with the two fields that are related to the supply: the retention index and retention vol (m3).

Send Perrine a development build when this is patched.

phargogh:
This PR updates UFRM to fix #167 and #168.  The two functional changes here are:

1. Updating the task graph flow to ensure that the output vector `flood_risk_service.shp` is always created on every run.  If the built infrastructure input is provided, two additional columns are added to the vector. ( #168 )
2. The built infrastructure vector processing step now handles cases where provided geometries are invalid or absent. ( #167 )

The biggest deal here was unravelling some tightly-coupled, complicated taskgraph tasks that were pretty nonorthogonal.  These are now much more neatly broken into various tasks, which also helps us avoid recomputation on a more granular level.  Note the number of lines deleted is greater than the number of lines added ... a sign of better design, I hope.

Fixes #167 
Fixes #168


tests/test_ufrm.py
```diff
@@ -61,6 +61,12 @@ def test_ufrm_regression(self):
             args['workspace_dir'], 'flood_risk_service_Test1.shp'),
             gdal.OF_VECTOR)
         result_layer = result_vector.GetLayer()
+
+        # Check that all four expected fields are there.
+        self.assertEqual(
+            set(('aff_bld', 'serv_bld', 'rnf_rt_idx', 'rnf_rt_m3')),
+            set(field.GetName() for field in result_layer.schema))
+
```
davemfish, line 69: This is a great addition to this test.

tests/test_ufrm.py
```diff
@@ -61,6 +61,12 @@ def test_ufrm_regression(self):
...
             gdal.OF_VECTOR)
         result_layer = result_vector.GetLayer()
+
+        # Check that all four expected fields are there.
+        self.assertEqual(
+            set(('aff_bld', 'serv_bld', 'rnf_rt_idx', 'rnf_rt_m3')),
+            set(field.GetName() for field in result_layer.schema))
+
         result_feature = next(result_layer)
         result_val = result_feature.GetField('serv_bld')
```
davemfish, line 71: What about checking other field values, like `aff_bld`?

tests/test_ufrm.py
```diff
@@ -93,6 +99,28 @@ def test_ufrm_regression_no_infrastructure(self):
...
+            gdal.OF_VECTOR)
+        result_layer = result_vector.GetLayer()
+        result_feature = next(result_layer)
+
+        # Check that only the two expected fields are there.
+        self.assertEqual(
+            set(('rnf_rt_idx', 'rnf_rt_m3')),
+            set(field.GetName() for field in result_layer.schema))
+
+        result_val = result_feature.GetField('rnf_rt_m3')
```
davemfish, line 113: Is it worth checking the other field's value also?

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+        # Determine flood_volume over the watershed
+        flood_volume_in_aoi_task = task_graph.add_task(
+            func=pygeoprocessing.zonal_statistics,
+            args=(
+                (runoff_retention_raster_path, 1),
+                reprojected_aoi_path),
+            dependent_task_list=[flood_vol_task],
+            task_name='zonal_statistics over the flood_volume raster')
+        flood_volume_stats = flood_volume_in_aoi_task.get()
+        damage_per_aoi_stats = damage_to_infrastructure_in_aoi_task.get()
```
davemfish, line 384: Ahhh yes, how nice to replace the pickling with a simple `get`!

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
     task_graph.join()
 
 
-def _pickle_zonal_stats(
-        base_vector_path, base_raster_path, target_pickle_path):
-    """Calculate Zonal Stats for a vector/raster pair and pickle result.
+def _write_summary_vector(
+        source_aoi_vector_path, target_vector_path, runoff_ret_stats=None,
+        runoff_ret_vol_stats=None, damage_per_aoi_stats=None,
+        flood_volume_stats=None):
```
davemfish, line 409: How about a docstring for this function?

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -290,7 +281,7 @@ def execute(args):
         task_name='generate runoff retention')
 
     # calculate runoff retention volumne
-    runoff_retention_ret_vol_raster_path = os.path.join(
+    runoff_retention_vol_raster_path = os.path.join(
         args['workspace_dir'], 'Runoff_retention_m3%s.tif' % file_suffix)
     runoff_retention_ret_vol_task = task_graph.add_task(
```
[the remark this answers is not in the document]
phargogh: Yes indeed!  I patched this and a few others in d9e52a77.

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -300,9 +291,9 @@ def execute(args):
             (float(args['rainfall_depth']), 'raw'),
             (abs(target_pixel_size[0]*target_pixel_size[1]), 'raw'),
             (runoff_retention_nodata, 'raw')], _runoff_retention_ret_vol_op,
```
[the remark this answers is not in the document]
phargogh: Patched in d9e52a77

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+                reproject_built_infrastructure_task],
+            task_name='calculate damage to infrastructure in aoi')
+        summary_tasks.append(damage_to_infrastructure_in_aoi_task)
+
+        # Determine flood_volume over the watershed
+        flood_volume_in_aoi_task = task_graph.add_task(
+            func=pygeoprocessing.zonal_statistics,
+            args=(
+                (runoff_retention_raster_path, 1),
+                reprojected_aoi_path),
```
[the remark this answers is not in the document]
phargogh: Yep! d81206a5

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+                  reprojected_structures_path,
+                  args['infrastructure_damage_loss_table_path']),
+            dependent_task_list=[
+                reprojected_aoi_task,
+                reproject_built_infrastructure_task],
+            task_name='calculate damage to infrastructure in aoi')
+        summary_tasks.append(damage_to_infrastructure_in_aoi_task)
+
+        # Determine flood_volume over the watershed
+        flood_volume_in_aoi_task = task_graph.add_task(
```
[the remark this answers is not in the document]
phargogh: This is an interesting one!  I added the tasks in f9533e26 for good measure, but also they shouldn't be strictly necessary.  Because all 4 of the tasks tracked in `summary_tasks` have their `.get()` method called, which will block until the task completes.  I added them anyways as a 'just in case' sort of condition, and also because I suppose that it might help with determining whether the task should be recomputed.

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+                mean_value = (
+                    runoff_ret_stats[feature_id]['sum'] / float(pixel_count))
+                target_feature.SetField('rnf_rt_idx', float(mean_value))
+
+        if feature_id in runoff_ret_vol_stats:
+            target_feature.SetField(
+                'rnf_rt_m3', float(
+                    runoff_ret_vol_stats[feature_id]['sum']))
+
+        if feature_id in damage_per_aoi_stats:
```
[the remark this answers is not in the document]
phargogh: > is it important to check that feature_id is in all of them?
I think we can safely assume that under normal execution of the model the feature ID will be present in the stats dict.  This is because this task and the `zonal_statistics` task that created the stats dict are both run with the same watersheds/aoi vector.  There could, of course, be a case that I'm not expecting here, but @davemfish what do you think about waiting to see what that case looks like before guarding against it?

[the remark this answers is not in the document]
phargogh: > Also I'm a little confused why the pixel count is taken from flood_volume_stats while the other stats are from other datasets.

Ha, good catch!  I've patched this in 14f02eb9 so that the pixel count is taken from another stats dict.  This also has the convenient side effect of doing one less `zonal_statistics`. 

> Is this function missing a block for writing flood_volume_stats?

Based on the UG, I can only guess at why the `flood_volume_stats` were being used in a previous version.  I'll confirm with Perrine that we should not be outputting some summary statistic of the flood volume.

And for future us, just back out of 14f02eb9 to restore the removed bits!

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+        target_vector_path))[0])
+    LOGGER.debug("creating layer %s", layer_name)
+    target_watershed_layer = target_watershed_vector.CreateLayer(
+        str(layer_name), source_srs, source_geom_type)
+
+    target_fields = ['rnf_rt_idx', 'rnf_rt_m3']
+    if not damage_per_aoi_stats:
+        damage_per_aoi_stats = {}
+    else:
+        target_fields += ['aff_bld', 'serv_bld']
```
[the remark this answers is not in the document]
phargogh: It is indeed, and has been patched in 91bbc64d.  That commit also improves the tests so that this oversight would have been caught had they been in place when I was refactoring the model.

tests/test_ufrm.py
phargogh: Yep!  As noted in another discussion thread, having these per-field checks would have helped prevent the issue with the missing field values for `aff_bld`.  Patched in 91bbc64d.

tests/test_ufrm.py
phargogh: 👍  91bbc64d

src/natcap/invest/urban_flood_risk_mitigation.py
phargogh: So much cleaner!  The logs are a little worse for it, unfortunately, but it is nicer to read this way in the source code.

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -322,107 +313,245 @@ def execute(args):
...
+
+    infrastructure_layer_defn = infrastructure_layer.GetLayerDefn()
+    for field_name in ['type', 'Type', 'TYPE']:
+        type_index = infrastructure_layer_defn.GetFieldIndex(field_name)
+        if type_index != -1:
+            break
+    if type_index == -1:
+        raise ValueError(
+            "Could not find field 'Type' in %s",
+            structures_vector_path)
```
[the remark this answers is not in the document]
phargogh: This case handling implementation bothered me, so I updated it to be completely insensitive to case in 0568c366.  In the same revision, I also added a call to `validate` to the `execute` function for good measure.

As for the missing-column error handling, I can't imagine how we'd end up in a case where we wouldn't be able to find the field index for 'type' given how things are anyways.  But if we were to, I figured the error message there would help us to debug it.

src/natcap/invest/urban_flood_risk_mitigation.py
phargogh: Can't believe I missed that.  Added in ebde3b71.

phargogh:
Thanks for the helpful review @davemfish !  I believe I responded to all of your comments and also resolved the merge conflict on`HISTORY.rst`.  I'll re-request your review.

src/natcap/invest/urban_flood_risk_mitigation.py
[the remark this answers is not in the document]
davemfish: Yeah, seems like a good idea to double-check. I don't know this model deeply at all, but now I wonder why the flood volume raster is calculated at all if it's not meant to be summarized and it doesn't appear to be an intermediate step towards something else. Maybe it's sometimes useful to the modeler though.

tests/test_ufrm.py
```diff
@@ -68,16 +68,20 @@ def test_ufrm_regression(self):
             set(field.GetName() for field in result_layer.schema))
 
         result_feature = next(result_layer)
-        result_val = result_feature.GetField('serv_bld')
+        for fieldname, expected_value in (
+                ('aff_bld', 187010830.32202843),
+                ('serv_bld', 13253546667257.65),
```
davemfish, line 73: That's a whole 'lotta 'serv_bld' right there.

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -458,9 +490,10 @@ def _write_summary_vector(
...
 
         if feature_id in damage_per_aoi_stats:
-            pixel_count = flood_volume_stats[feature_id]['count']
+            pixel_count = runoff_ret_vol_stats[feature_id]['count']
             if pixel_count > 0:
                 damage_sum = damage_per_aoi_stats[feature_id]
+                target_feature.SetField('aff_bld', damage_sum)
                 target_feature.SetField(
                     'serv_bld', damage_sum * float(
                         runoff_ret_vol_stats[feature_id]['sum']))
```
davemfish, line 499: Forgive me for going outside the bounds of the changes here, but this equation for `serv_bld` doesn't match the UG's `Service.built`, if I'm looking in the correct place. http://releases.naturalcapitalproject.org/invest-userguide/latest/urban_flood_mitigation.html#calculate-potential-service-optional

src/natcap/invest/urban_flood_risk_mitigation.py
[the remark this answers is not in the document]
davemfish: Oh cool, I did not know about that feature of `get()`.

src/natcap/invest/urban_flood_risk_mitigation.py
phargogh: Yep, you're right that this doesn't match up!  As far as I can tell, the `Service.built` variable noted in the UG is being created as the `Q_m3.tif` output raster.  So things look to be correct within that calculation, despite having misleading variable names.  I'll double-check with Perrine about the name mismatch and the intent of the output vector, in case there's something funky going on here.

Either way, I think we can safely merge this PR since it's consistent with what the model is currently doing, and then I'll follow up with what I find out from Perrine about what the model is supposed to be doing.

phargogh:
Thanks @davemfish !  I just resolved the merge conflicts from the recent PR into `master` and am waiting for the tests to pass.  After that, I think the thing to do here will be to wait to hear back from Perrine about the intent of the flood_volume/`service.built` outputs.

phargogh:
Well, the binary builds are still failing for Windows, so I've created https://github.com/natcap/invest/issues/180 so that we can debug.  Given that, I'm happy to merge.

Issue #122, opened 2020-05-14
Nodata masking bug in GLOBIO


dcdenu4:
There is a nodata masking bug that was brought to light from debugging a users data from this forum post: [natcap-forums:globio-unexpected-output](https://community.naturalcapitalproject.org/t/globio-unexpected-output/978/4)

The bug occurs in the following function:
```
def _collapse_infrastructure_op(*infrastructure_array_list):
        """For each pixel, create mask 1 if all valid, else set to nodata."""
        nodata_mask = (
            numpy.isclose(
                infrastructure_array_list[0], infrastructure_nodata_list[0]))
        infrastructure_result = infrastructure_array_list[0] > 0
        for index in range(1, len(infrastructure_array_list)):
            current_nodata = numpy.isclose(
                infrastructure_array_list[index],
                infrastructure_nodata_list[index])

            infrastructure_result = (
                infrastructure_result |
                ((infrastructure_array_list[index] > 0) & ~current_nodata))

            nodata_mask = (
                nodata_mask & current_nodata)

        infrastructure_result[nodata_mask] = infrastructure_nodata
        return infrastructure_result
```

When `*infrastructure_array_list` is of length 1, `infrastructure_result` is set to a boolean type. Then before returning the result is masked and attempted to be updated with an integer type of `-1`. This causes improper masking.

dcdenu4:
Helpful information from Rich (referring to this gist https://gist.github.com/dcdenu4/b7bcb9c317869ced37b46426bfb57da6):
>OH I thought the issue was that line 13 defines result as a Boolean array then in line 17 gets a non-boolean assignment to it, specifically -1. I don't read that as an error with a Boolean array indexing a Boolean array, but rather a type error. Since it's Python and numpy, the -1 will get cast to True (since `bool(-1)` is "True") so line 17 is the same as `result[nodata_mask] = True`. I suspect that is an unexpected result.

dcdenu4:
For myself to remember: I have a branch in my fork `bugfix/globio-numpy-bool-bug`

dcdenu4:
resolved in PR #186

PR #186, opened 2020-06-25
GLOBIO infrastructure bug fix

dcdenu4:
This PR fixes a bug that was uncovered when helping a user on the forums:
https://community.naturalcapitalproject.org/t/globio-unexpected-output/978

**The Bug**:
When an infrastructure directory is provided with 1 raster, the `_collapse_infrastructure_op` was improperly handling numpy integer and boolean types. This was causing positive nodata values to be set to `True` and thus 1. 

**The Fix**:
Create an integer numpy array upfront for the result and set initial values to 0. A 0 indicates where there was no infrastructure and no nodata value. Create a infrastructure mask that keeps track of where infrastructure has been indicated and a mask that keeps track of where nodata values have been indicated. Set the result to 1 using the infrastructure mask and set a nodata value from the nodata mask. 

Added a test to check this case.

Fixes Issue #122


src/natcap/invest/globio.py
```diff
@@ -939,22 +939,29 @@ def _collapse_infrastructure_layers(
...
         """For each pixel, create mask 1 if all valid, else set to nodata."""
-        nodata_mask = (
-            numpy.isclose(
-                infrastructure_array_list[0], infrastructure_nodata_list[0]))
-        infrastructure_result = infrastructure_array_list[0] > 0
+        infrastructure_result = numpy.empty_like(
+            infrastructure_array_list[0], dtype=numpy.int32)
+        # Zero's indicate where there's no infrastructure influence but not
+        # necessarily nodata
+        infrastructure_result[:] = 0
```
richpsharp, line 946: If you want to skip this two step, you can say `nodata.zeros(infrastructure_array_list[0].shape...` rather than make an empty array and then set it all to 0. Also I see you changed the type to an `int32` but I can't tell why. Can you talk a little more about that? 

Also typo on "Zero's"

src/natcap/invest/globio.py
```diff
@@ -939,22 +939,29 @@ def _collapse_infrastructure_layers(
...
-                infrastructure_result |
+            infrastructure_mask = (
+                infrastructure_mask |
                 ((infrastructure_array_list[index] > 0) & ~current_nodata))
 
-            nodata_mask = (
-                nodata_mask & current_nodata)
+            nodata_mask = (nodata_mask & current_nodata)
 
+        infrastructure_result[infrastructure_mask] = 1
```
richpsharp, line 964: Can you also explain why `infrastructure_mask` is not the result and needs to be separated into `infrastructure_result`. If I'm reading this right, `infrastructure_result` will be 0 where `infrastructure_mask` is false and 1 where it is true. Is that not right?

src/natcap/invest/globio.py
```diff
@@ -975,7 +982,7 @@ def _collapse_infrastructure_op(*infrastructure_array_list):
         (x, 1) for x in aligned_infrastructure_target_list]
     pygeoprocessing.raster_calculator(
         infra_filename_band_list, _collapse_infrastructure_op,
-        infrastructure_path, gdal.GDT_Byte, infrastructure_nodata)
+        infrastructure_path, gdal.GDT_Int32, infrastructure_nodata)
```
richpsharp, line 985: I commented on this above too, just checking why it's necessary to change this to an int32 when the result is a 0/1 mask?

src/natcap/invest/globio.py
```diff
@@ -939,22 +939,29 @@ def _collapse_infrastructure_layers(
 
     def _collapse_infrastructure_op(*infrastructure_array_list):
         """For each pixel, create mask 1 if all valid, else set to nodata."""
-        nodata_mask = (
-            numpy.isclose(
-                infrastructure_array_list[0], infrastructure_nodata_list[0]))
-        infrastructure_result = infrastructure_array_list[0] > 0
```
richpsharp, line 945: Hi, this comment is out of order, but it's the last one I'm doing after I parsed through the whole thing. If I understand the issue, would a single change of this line to `infrastructure_result = (infrastructure_array_list[0] > 0) & ~nodata_mask` fix the issue?

src/natcap/invest/globio.py
dcdenu4: That is exactly right, but I don't think we want to actually return a boolean array here. We can't properly set nodata values into a boolean array because the type conflict will just set the boolean array to either 0 or 1 depending on a positive or negative nodata value. Currently the model sets that nodata value to -1. 

I also think it's weird to return a boolean array from a function that writes a raster with GDAL which doesn't have a boolean type, but should be a Byte type.

src/natcap/invest/globio.py
dcdenu4: It's not, should stay a Byte since the out nodata is -1 and results will be 0 and 1.

src/natcap/invest/globio.py
dcdenu4: Dang, that's a great paren catch on a bug that I didn't realize was actually there. I actually just posted about this in slack before reading over your comments again.

src/natcap/invest/globio.py
richpsharp: Ahh got it, I missed that there is a nodata possibility there. I feel like making the initial raster a `numpy.uint8` type, then setting the nodata value to "2" or "255" is the solution that keeps everything in 8 bit space?

src/natcap/invest/globio.py
richpsharp: Ahh got it, I missed that there is a valid nodata case there. I mentioned below I think the solution would be to initialize `infrastructure_result` to a `numpy.uint8` and choose a nodata value of 2 or 255 would handle this. (p.s. if you choose 1 it turns into 255 anyway but best not to bit twiddle).

dcdenu4: haha, no one wants to be a bit twiddle. I like it!

src/natcap/invest/globio.py
dcdenu4: Now using numpy.zeros and dropped the int32 for uint8 here: https://github.com/natcap/invest/pull/186/commits/86dab80e89b22c452e5da5b2efec78decbfa62cb

src/natcap/invest/globio.py
dcdenu4: Nodata value is now 255 here: https://github.com/natcap/invest/pull/186/commits/86dab80e89b22c452e5da5b2efec78decbfa62cb

src/natcap/invest/globio.py
dcdenu4: This is patched here with proper parens: https://github.com/natcap/invest/pull/186/commits/86dab80e89b22c452e5da5b2efec78decbfa62cb

dcdenu4:
Hey @richpsharp and @davemfish I think I've touched on everything. Dave, I did refactor that test a bit to use a made up smaller example. However I am calling the outer function since the `op` function's scope resides in there.

HISTORY.rst
```diff
@@ -21,14 +21,6 @@ Unreleased Changes
...
     * Fix a bug that mishandled combining infrastructure data when only one
       infrastructure data was present.
-* Urban Flood Risk Mitigation
-    * Fixed an issue where the output vector ``flood_risk_service.shp`` would
-      only be created when the built infrastructure vector was provided.  Now,
-      the ``flood_risk_service.shp`` vector is always created, but the fields
-      created differ depending on whether the built infrastructure input is
-      present during the model run.
-    * Fixed an issue where the model would crash if an infrastructure geometry
-      were invalid or absent.  Such features are now skipped.
```
richpsharp: Does this PR really remove an update to HISTORY? Just checking in case there was some weird merge thing.

Issue #1, opened 2020-01-29
Support python 3.8


phargogh:
Python 3.8 is the most recently-released version of python and we should make sure that the package supports it.

- [x] Add tests for Python 3.8
- [x] Note support for Python 3.8 in setup.py

phargogh:
This is probably worth waiting to tackle until we finish migrated InVEST over to github.

phargogh:
This has been created on bitbucket at https://bitbucket.org/natcap/invest/issues/3948/support-python-38

phargogh:
We ended up doing the 3.8 release just with python 3.8 support.  See below for @dcdenu4 's comment on the original  bitbucket issue:

```
Unfortunately golhke does not have a built GDAL 2.4.1 with Python 3.8.
There is a GDAL 3.0 with Python 3.8, so without compiling our own version of
GDAL, it might make sense to add support GDAL 3.0 and then support Python
3.8 with that version of GDAL…
```

dcdenu4:
Python 3.8 is dependent on GDAL 3... so waiting for that to make it through the pipeline

dcdenu4:
It looks like Pyinstaller does not support Python 3.8... https://github.com/pyinstaller/pyinstaller/issues/4311.

phargogh:
@dcdenu4 do we need to upgrade the pyinstaller part of the build process to python 3.8 as well?  Since `release/3.9` requires GDAL3, couldn't we just stick with python 3.7 for the time being?  It looks like Gohlke has a GDAL build for python3.7/GDAL3+.

dcdenu4:
@phargogh Ohhh, interesting. Yes, definitely can stick with Python 3.7 for the builds. I confused myself in what was doing what.

dcdenu4:
I have been testing this out as of late and discovered the following:

Bump version as is:
- Python 3.7 and MacOS run and pass
- Python 3.8 Windows Errors:
```python
Collecting Shapely<1.7.0,>=1.6.4
  Downloading Shapely-1.6.4.post2.tar.gz (225 kB)
    ERROR: Command errored out with exit status 1:
[...]

PR #187, opened 2020-06-25
Python 3.8 Support

dcdenu4:
This PR adds Python 3.8 support and drops 3.6 from testing.

Removing restriction on `Shapely` version because `1.7.0` is needed for Python 3.8. 

Fixes Issue #1


dcdenu4:
Closing this draft PR as it looks like Pyinstaller does not yet support Python 3.8, which is a larger issue for moving forward with this.

PR #194, opened 2020-06-30
Experimental invest binary builds that include the python dependencies of invest-gui

davemfish:
This PR modifies the invest binary build process to generate the binaries needed by the invest-gui Electron app. That app needs `natcap.invest` binaries (including all dependencies) that we already build for invest, _plus_ `Flask` and `invest-gui/src/server.py` dependencies. Since `server.py` can share the frozen binaries that we already package for invest, it makes sense to do all that packaging here in `natcap/invest` rather than in `natcap/invest-gui`.

In addition to the modified PyInstaller script, there's also supporting changes to the Makefile, to clone the invest-gui repo and to deploy archives of the frozen python environments that the invest-gui build process expects to find on GCS. 

I added a build for ubuntu because that happens to be what I'm developing on right now. The deleted hook files were to appease the version of PyInstaller I'm using, which was telling me that conflicting hooks were present for those modules. 

I don't think we should merge this PR, but thought the diff would be useful to see while reviewing https://github.com/natcap/invest-gui/pull/27 . Sometime after we are confident about that project's build process, we can consider merging this PR. Probably not into `master`.


dcdenu4:
Hey Dave, I'm not too plugged in on the step by step build process but everything in here seems reasonable to me.

phargogh:
Hey @davemfish I mentioned this over in https://github.com/natcap/invest-gui/pull/27#issuecomment-655645930 but I'll include my comments here as well.

I think the changes proposed there look good to me and make sense for what we're trying to do  The PR is making me wonder about a few things in the InVEST build process overall (not specific to this PR or build process, just noting them here for my own memory):

* The deleted hooks make me wonder if they're just no longer needed or if they actually conflict with the latest version of pyinstaller.  Should we upgrade pyinstaller in our build process?  Worth looking into.
* The binary build step for ubuntu looks pretty similar to the mac binary build step, and that makes me wonder if we'd be able to 1) move the windows build over to conda in order to 2) consolidate all our binary builds into a single configuration.
* The hardcoding of environment paths in our pyinstaller specs (which we've done forever) made me wonder if there were a way to pull the conda environement dynamically, maybe from `$CONDA_PREFIX` or something?
* Related to the hardcoding of environment paths, I wondered if the `gui` repo path in `pyinstaller.spec` could be pulled from the Makefile, something along the lines of `make jprint-GIT_GUI_REPO_PATH`

davemfish:
See #211 

We'll revisit this PR (or more likely open a different one that merges into a release branch) sometime in the future.

PR #199, opened 2020-07-01
Bugfix/sdr 129 flow area cap

richpsharp:
This PR fixes a design error bug in SDR where the upstream area to the LS factor should be capped to 333m^2, previously the LS factor itself was capped to 333m. This fixes an issue where there were the occasional high spikes of SDR pixel export.

In turn this creates differences in the expected output of SDR. These are all refelected in the "expected output" components of the test suite.

Also increased the the amount of significant digits from 2 to 6 when testing against expected results.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -42,9 +42,9 @@ Unreleased Changes (3.9)
     * ``get_raster_info`` / ``get_vector_info`` keyword ``projection`` to
       ``projection_wkt``.
 * SDR:
-  * Fixing an issue where the LS factor should be capped to 333 m^2 upstream.
-    In previous versions the LS factor was erroniously capped to "333" leading
-    to high export spikes in some pixels.
+  * Fixing an issue where the LS factor should be capped to an upstream area of
+    333^2 m^2. In previous versions the LS factor was erroniously capped to
+    "333" leading to high export spikes in some pixels.
 
 Unreleased Changes
 ------------------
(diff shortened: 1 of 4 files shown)


src/natcap/invest/sdr/sdr.py
```diff
@@ -804,16 +804,18 @@ def ls_factor_function(percent_slope, flow_accumulation, avg_aspect):
             beta[big_slope_mask] / (1 + beta[big_slope_mask]))
         m_exp[~big_slope_mask] = m_table[m_indexes]
 
+        # from McCool paper: "as a final check against excessively long slope
+        # length calculations ... cap of 333m"
+        # from Rafa, this should really be the upstream area capped to
+        # "333m^2" because McCool is 1D
```
davemfish, line 810: Wait, should it be capped to `333**2 m^2` as in the code below or to `333 m^2` as in the comment above?

HISTORY.rst
```diff
@@ -41,6 +41,10 @@ Unreleased Changes (3.9)
     * ``convolve_2d`` keyword ``ignore_nodata`` to ``ignore_nodata_and_edges``.
     * ``get_raster_info`` / ``get_vector_info`` keyword ``projection`` to
       ``projection_wkt``.
+* SDR:
+  * Fixing an issue where the LS factor should be capped to 333 m^2 upstream.
```
davemfish, line 45: Here's another place to just confirm whether it should be 333 square meters or 333**2 square meters.

src/natcap/invest/sdr/sdr.py
richpsharp: Nice catch. It should be 333^2 m^2. That'd be a dammed confusing comment for a future engineer. Fixed in fc2cebe6bf3357e68b8f5383e3adbaa9c3fbef74

HISTORY.rst
richpsharp: Fixed in fc2cebe6bf3357e68b8f5383e3adbaa9c3fbef74

PR #212, opened 2020-07-09
Task/205 build and test py dists in tests

### LINKED ISSUE #205: Build and test python distributions (not eggs) in model tests
As part of our release processes, it would be nice to have confidence that the distributions we're shipping work as expected.  That is to say, that we want to be sure there aren't any functional issues with the packaging of source or wheel distributions.  To verify this, we should update our release automation to:

1. Build platform-specific wheels and a source distribution
2. Run tests on each.

As a part of this, we can also look into consolidating two of our workflows (the one for [building python distributions](https://github.com/natcap/invest/blob/master/.github/workflows/build-py-dists.yml), and the one for [running model tests](https://github.com/natcap/invest/blob/master/.github/workflows/build-and-test.yml)) into a single workflow.

phargogh:
This PR fixes #205 by making the following changes to our model build/test workflow:

1. The python distribution workflow `build-py-dists.yml` has been consolidated into `build-and-test.yml`, which is where the model tests are executed.  `build-py-dists.yml` has been deleted.
2. A new build matrix for the source distribution has been set up to ensure that `natcap.invest` can be compiled from source on mac and windows.  This is only tested in a very limited capacity, and the assumption is that issues with the models will be caught from the tests that build wheels.
3. The model tests are executed only after a wheel has been built and installed for the target python version, os and architecture.
4. In all cases where a python distribution is created, the distribution is saved as a build artifact.

The build step to upload the python distribution to a release object was not copied, in anticipation of a pipelined release process (#144)


.github/workflows/build-and-test.yml
```diff
@@ -99,21 +100,106 @@ jobs:
...
+                  python -m twine check dist/*
+
+            - name: Install runtime dependencies Windows
+              if: matrix.os == 'windows-latest'
+              shell: bash -l {0}
+              env:
+                  PIP_EXTRA_INDEX_URL: "http://pypi.naturalcapitalproject.org/simple/"
+                  PIP_TRUSTED_HOST: "pypi.naturalcapitalproject.org"
+                  PIP_PREFER_BINARY: 1
+              run: python -m pip install -r requirements.txt -r requirements-dev.txt
```
davemfish, line 124: I just noticed that `requirements-gui` is not included here. The UI tests seem to run okay though, despite the usual flakiness. How exactly does that work? :)

davemfish:
Hey @phargogh , I approved but then I went back and found something I didn't quite understand about the UI tests and installing their dependencies.

.github/workflows/build-and-test.yml
phargogh: Yes indeed!  The GUI tests are executed in a batch separately from the model tests (the `run-ui-tests` job in this configuration file, starting on line 238).  The idea with removing the `-r requirements-gui.txt` was simply to cut down the number of packages that would need to be installed while still being able to run the model tests successfully.  Come to think of it, I could probably remove the `-r requirements-dev.txt` as well, now that build dependencies are tracked in `pyproject.toml` and installed here in the job.

It's also worth noting that the UI tests themselves are still using `python setup.py install` for their installation and not a python artifact.

phargogh:
@davemfish thanks for taking a look at this!

It's hard to ensure that this will be truly 'safe', but I hope that we can at least be more confident that it works as expected, and we can have this confidence because our tests ran ok after installing the wheel.  And you're right: the goal is to not need to test the wheels after uploading them to PyPI.  I just updated the release checklist (https://github.com/natcap/invest/wiki/Release-Checklist) to include a note about the tests already using the wheels.

.github/workflows/build-and-test.yml
davemfish: Ah, right, thanks. I missed the ui tests section since nothing was changed there.

PR #240, opened 2020-07-22
CV: handle invalid geometry in AOI

### LINKED ISSUE #239: CV crashes over invalid geometry in AOI
As we do in other places throughout CV, `interpolate_shore_points` should check for valid geometry and try to fix invalid geometry using a `buffer(0)`. Failing that, we can either crash or skip the offending feature and issue a warning.

davemfish:
This PR fixes a bug where CV would crash on invalid geometry in the AOI. That's a bug because we could easily fix that geometry on the fly instead of crashing, as is already done in other CV operations.

It turned out there were three places in CV where the same vector processing and geometry checking were happening in order to get OGR geometries into shapely data structures. So that's now abstracted into a new function `_ogr_to_geometry_list`. That function is borrowed nearly verbatim from `recmodel_client.py`. 

Fixes #239

### DIFF
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -267,8 +267,8 @@ def execute(args):
             The radius in meters around each shore point in which to compute
             the population density.
         args['slr_vector_path'] (string): (optional) path to point vector
-            containing the field `args['slr_field']`.
-        args['slr_field'] (string): name of a field in `args['slr_vector_path']`
+            containing the field ``args['slr_field']``.
+        args['slr_field'] (string): name of a field in ``args['slr_vector_path']``
             containing numeric values.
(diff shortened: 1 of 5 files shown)


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2760,6 +2701,47 @@ def polygon_to_lines(geometry):
     return line_list
 
 
+def _ogr_to_geometry_list(vector_path):
+    """Convert an OGR type with one layer to a list of shapely geometry.
+
+    Iterates through the features in the `vector_path`'s first layer and
```
phargogh, line 2707: This is also pretty trivial, but would it be possible to use RST's double-backticks for inline code rather than mardown's single-backticks?

davemfish:
Okay thanks, @phargogh . I changed the docstring language and switched to the RST markup throughout two modules. Back to you!

PR #241, opened 2020-07-23
Change units in UCM to account for building area and improve handling of CC inputs

### LINKED ISSUE #28: UCM: some optional parameters are not handled well by validation / UI
Here's an error report: https://community.naturalcapitalproject.org/t/urban-cooling-model-error/824/2

The model expects the args in question to either have values that cast to float, or for the key to be missing from the args dict altogether. Not handled is the case where `args['cc_weight_shade'] = ''`. Ditto for the other two `cc_weight` args.

### LINKED ISSUE #138: Updates to UCM energy savings calculations
A [user on the forums](https://community.naturalcapitalproject.org/t/understanding-urban-cooling-energy-consumption-input/1006/3) recently asked some interesting questions about energy vs. power and about average building power consumption when there's substantial variability within a building class.

There are a few changes we need to make:

- [x] Power should indeed be power, not energy.  Change units of `kW` to `kWh` in `ARGS_SPEC` and in the UI.
- [x] Correct the units of power in the UG
- [x] Depending on what the UCM science team decides, see about implementing their changes to energy savings calculations to account for floor space. ~~(Currently waiting on a decision from the science team, will post back here when it's decided)~~ See below for changes to energy savings changes.
- [x] Document changes in the User's Guide
- [x] Document changes in HISTORY

Adding to the 3.9 milestone because this significantly changes the model's functionality.

## Changes to Energy Savings

After some discussion, we decided to change the energy savings calculations to:

```
savings = consumption_per_deg_per_area * (Tmax - Tair) * floor_area
```
Where
* `floor_area`: the area (in m^2) of the building footprint (from used-defined geometry).
* `consumption_per_deg_per_area`: the consumption per degree C per m^2 of this class of building.  This must account for the building type as well as the number of floors.

- [x] Implement the change in the model
- [x] Update the User's Guide chapter to reflect this

phargogh:
This PR updates the Urban Cooling model's energy consumption equation in response to a discussion that's been playing out on the side and in response to a [discussion on the forums](https://community.naturalcapitalproject.org/t/understanding-urban-cooling-energy-consumption-input/1006).

Summary of changes:

* Energy consumption inputs are now considered to be in units of kWh/Degrees C/m^2.  This requires that we include the building footprint area (in m^2) in the energy consumption calculations. (see #138)
  * The UG has been updated (and repo rev updated here) to include the above unit change
  * Tests updated to reflect the above unit change
* The model now very explicitly handles cases where the user does not provide an input value for the cooling coefficient inputs.  Cases where parameters are `""`, `None` and absent from `args` are all accounted for and default values assumed.  (see #28 )
* In debugging this, it turned out that there was an issue with which version of InVEST was being installed in these tests.  The model tests workflow has been updated to install exactly which file we want to install, rather than relying on pip to find it for us.
* PEP8/pylint updates in `test_ucm.py`.

Fixes #138 , fixes #28

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -133,7 +133,7 @@ jobs:
             - name: Run model tests
               shell: bash -l {0}
               run: |
-                  pip install --isolated --upgrade --only-binary natcap.invest --find-links=dist natcap.invest
+                  pip install $(find dist -name "natcap.invest*.whl")
                   make test
 
             - uses: actions/upload-artifact@v1
@@ -187,7 +187,7 @@ jobs:
                   rm -r natcap.invest.egg-info
 
                   # Install natcap.invest from the sdist in dist/
-                  pip install --isolated --find-links=dist natcap.invest
+                  pip install $(find dist -name "natcap.invest*")
 
                   # Model tests should cover model functionality, we just want
                   # to be sure that we can import `natcap.invest` here.
--- a/src/natcap/invest/urban_cooling_model.py
+++ b/src/natcap/invest/urban_cooling_model.py
@@ -303,9 +303,24 @@ def execute(args):
     # cast to float and calculate relative weights
     # Use default weights for shade, albedo, eti if the user didn't provide
     # weights.
-    cc_weight_shade_raw = float(args.get('cc_weight_shade', 0.6))
-    cc_weight_albedo_raw = float(args.get('cc_weight_albedo', 0.2))
-    cc_weight_eti_raw = float(args.get('cc_weight_eti', 0.2))
+    # TypeError when float(None)
+    # ValueError when float('')
+    # KeyError when the parameter is not present in the args dict.
+    try:
+        cc_weight_shade_raw = float(args['cc_weight_shade'])
+    except (ValueError, TypeError, KeyError):
+        cc_weight_shade_raw = 0.6
+
+    try:
+        cc_weight_albedo_raw = float(args['cc_weight_albedo'])
+    except (ValueError, TypeError, KeyError):
+        cc_weight_albedo_raw = 0.2
+
+    try:
+        cc_weight_eti_raw = float(args['cc_weight_eti'])
+    except (ValueError, TypeError, KeyError):
+        cc_weight_eti_raw = 0.2
+
     t_ref_raw = float(args['t_ref'])
     uhi_max_raw = float(args['uhi_max'])
     cc_weight_sum = sum(
--- a/src/natcap/invest/urban_cooling_model.py
+++ b/src/natcap/invest/urban_cooling_model.py
@@ -177,7 +177,7 @@ ARGS_SPEC = {
             },
             "about": (
                 "A CSV table containing information on energy consumption "
-                "for various types of buildings, in kWh/degC/m^2."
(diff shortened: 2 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             - name: Run model tests
               shell: bash -l {0}
               run: |
-                  pip install --isolated --upgrade --only-binary natcap.invest --find-links=dist natcap.invest
+                  pip install $(find dist -name "natcap.invest*.whl")
```
dcdenu4, line 136: I might not be remembering correctly, but didn't we already make this change in a prior PR? This update looks familiar to me is all.

src/natcap/invest/urban_cooling_model.py
```diff
@@ -177,7 +177,7 @@
             },
             "about": (
                 "A CSV table containing information on energy consumption "
-                "for various types of buildings, in kW/degC."
+                "for various types of buildings, in kWh/degC/m^2."
```
dcdenu4, line 180: Should there be a space between `deg` and `C`? I think you've written it that way until now.

src/natcap/invest/urban_cooling_model.py
```diff
@@ -935,6 +950,8 @@ def calculate_energy_savings(
     target_building_vector = gdal.OpenEx(
         target_building_vector_path, gdal.OF_VECTOR | gdal.GA_Update)
     target_building_layer = target_building_vector.GetLayer()
+    target_building_srs = target_building_layer.GetSpatialRef()
+    target_building_square_units = target_building_srs.GetLinearUnits() ** 2
```
dcdenu4, line 954: I think I saw that the building vectors are reprojected to the LULC. But just wanted to make sure, otherwise `GetLinearUnits` on a Geographic vector could get weird.

.github/workflows/build-and-test.yml
phargogh: Yeah, I had remembered the same thing and was surprised that I needed to do it once again.  Looking back through the PR queue, looks like #222 did something related, but only for sdists, I guess, and even then only in `release/3.9`.  This PR uses the same thinking for wheels as well as as sdists.  Pip's approach _should_ work, but the `bash`/`find` approach is just so much more expressive AND it works as we expect!

src/natcap/invest/urban_cooling_model.py
phargogh: Yes!  Way better to err on the side of consistency.  Patched in 686d91a8.

src/natcap/invest/urban_cooling_model.py
phargogh: Yep!  This is a really important point to make, and you're right that the buildings vector is reprojected to the projection of the LULC (task `intermediate_building_vector_task`), which validation enforces as being projected in `m`.  I've added notes to the docstrings of `execute` (about the LULC being in a linearly-projected CRS) and to `calculate_energy_savings` (about the need for the buildings vector being in a linearly-projected CRS) to clarify the contract.  Patched in 05a0b834

phargogh:
Thanks for the review @dcdenu4 !  I've responded to your comments, patched a few things in response and re-requested your review.

PR #244, opened 2020-07-27
Handle non projected inputs nicely in HRA

### LINKED ISSUE #79: HRA failing on valid projections
I wanted to start documenting an issue a user was having from the forums here:

https://community.naturalcapitalproject.org/t/habitat-risk-assessment-output-error-visualization-fails/926/2

Given the users input there was something about the projections of the AOI and habitat and stressor layers that were causing any downstream created raster to have a bad projection. I did  a quick fix for the user by projecting all of the users input to : *EPSG:32724
WGS 84 / UTM zone 24S*.

This might warrant further investigation to ensure that there is not a bug.

dcdenu4:
This PR fixes a bug in HRA where habitat and stressor inputs were not checked to see if they were projected. The model uses the projected AOI as the base but does not reproject geographic inputs before using them in an operation. Now if spatial inputs are not projected a ValueError is raised with an error message. The users guide has been updated to reflect these inputs should be projected. Added a test to validate the proper error is raised.

Closes #79

### DIFF
--- a/tests/test_hra.py
+++ b/tests/test_hra.py
@@ -744,7 +744,7 @@ class HraUnitTests(unittest.TestCase):
             1E-6)
 
     def test_simplify_geometry_points(self):
-        """HRA: test _simplify_geometry does't alter geometry given points."""
+        """HRA: test _simplify_geometry does not alter point geometries."""
         from natcap.invest.hra import _simplify_geometry
 
         srs = osr.SpatialReference()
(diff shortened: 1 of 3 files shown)


tests/test_hra.py
```diff
@@ -973,6 +986,49 @@ def test_layer_no_projection(self):
         actual_message = str(cm.exception)
         self.assertTrue(expected_message in actual_message, actual_message)
 
+    def test_layer_geographic(self):
```
davemfish, line 989: Do you mind naming this `test_layer_geographic_projection` so that it's kind of linked by name to the very similar test named `test_layer_no_projection`?

dcdenu4:
Hey @davemfish thanks for the comments and recommendations. I think I have them all covered now. I ended up combining those two projection, spatial reference checks in one. I kind of forgot that Python does nice order of operations where if the first if conditional fails the second won't be tested. Just waiting on tests to pass.

Issue #70, opened 2020-04-01
Add _swy suffix to watershed results vector name


phargogh:
The other routed models (NDR, SDR) have a `watershed_results_ndr.shp` output vector, but SWY does not ... it's just ~~`watershed_results.shp`~~ `aggregated_results.shp`.  It would be nice to update this filename for the sake of consistency with the other models.

PR #248, opened 2020-07-29
Make SWY output filename consistent with NDR, SDR

emlys:
Changed aggregate_results.shp to aggregate_results_swy.shp for consistency with output file naming in NDR and SDR models.
Addresses natcap/invest#70

In the test file, `aggregated_results.shp` was mentioned 4 times, so I replaced those mentions with a new variable set to the new filename. Let me know if this is overkill/doesn't deserve a global variable.

### DIFF
--- a/tests/test_seasonal_water_yield_regression.py
+++ b/tests/test_seasonal_water_yield_regression.py
@@ -739,7 +739,7 @@ class SeasonalWaterYieldRegressionTests(unittest.TestCase):
         make_agg_results_csv(agg_results_csv_path, climate_zones=True)
 
         SeasonalWaterYieldRegressionTests._assert_regression_results_equal(
-            os.path.join(args['workspace_dir'], 'aggregated_results_cz.shp'),
+            os.path.join(args['workspace_dir'], 'aggregated_results_swy_cz.shp'),
             agg_results_csv_path)
 
     def test_user_recharge(self):
--- a/tests/test_seasonal_water_yield_regression.py
+++ b/tests/test_seasonal_water_yield_regression.py
@@ -13,7 +13,6 @@ import pygeoprocessing.testing
 REGRESSION_DATA = os.path.join(
     os.path.dirname(__file__), '..', 'data', 'invest-test-data',
     'seasonal_water_yield')
-AGGREGATED_RESULTS_FILENAME = 'aggregated_results_swy.shp'
 
 
 def make_simple_shp(base_shp_path, origin):
@@ -452,7 +451,7 @@ class SeasonalWaterYieldUnusualDataTests(unittest.TestCase):
         l_path = os.path.join(self.workspace_dir, 'L.tif')
         make_recharge_raster(l_path)
         aggregate_vector_path = os.path.join(self.workspace_dir,
-                                             AGGREGATED_RESULTS_FILENAME)
+                                             'aggregated_results_swy.shp')
         make_simple_shp(aggregate_vector_path, (1180000.0, 690000.0))
         seasonal_water_yield._aggregate_recharge(aoi_path, l_path, l_path,
                                                  aggregate_vector_path)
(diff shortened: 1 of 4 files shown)


tests/test_seasonal_water_yield_regression.py
```diff
@@ -451,7 +452,7 @@ def test_aggregate_vector_preexists(self):
         l_path = os.path.join(self.workspace_dir, 'L.tif')
         make_recharge_raster(l_path)
         aggregate_vector_path = os.path.join(self.workspace_dir,
-                                             'aggregated_results.shp')
+                                             AGGREGATED_RESULTS_FILENAME)
```
phargogh, line 455: Very interesting, and very DRY!  In this case, I feel like the benefit of having the filename written in-place (even if it is a bit redundant) might outweigh the benefits of having the filename only defined in one place.  There are definitely some cases where being a little more explicit will end up making things easier for us to maintain in the long run and I wonder if this indirection is an example of that ... but I'd love to hear what you think!

And by the way, one of the things that I do like about this approach is that even though there is a bit of indirection, we almost don't lose the meaning here because of the variable name you've chosen ... very nice and explicit and human-readable.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -738,7 +739,7 @@ def test_climate_zones_regression(self):
         make_agg_results_csv(agg_results_csv_path, climate_zones=True)
 
         SeasonalWaterYieldRegressionTests._assert_regression_results_equal(
-            os.path.join(args['workspace_dir'], 'aggregated_results_cz.shp'),
+            os.path.join(args['workspace_dir'], 'aggregated_results_swy_cz.shp'),
```
phargogh, line 742: Ah, this is a really interesting example that might help us decide on an approach for the filename!

If we wanted to go with the DRY approach, we would _also_ want this string here to be updated as well ... but doing so would also hide the meaning a bit behind some string manipulations:

```python
'_cz'.join(os.path.splitext(AGGREGATED_RESULTS_FILENAME))
```
Is a lot harder to read (as a human, anyways :) ) than `aggregated_results_swy_cz.shp`

So for the sake of explicitness, I might suggest we use the static strings rather than a module-level variable.  But again, what do you think about this reasoning?

tests/test_seasonal_water_yield_regression.py
```diff
@@ -782,7 +783,7 @@ def _assert_regression_results_equal(
             result_vector_path (string): path to the summary shapefile
                 produced by the SWY model.
             agg_results_path (string): path to a csv file that has the
-                expected aggregated_results.shp table in the form of
+                expected aggregated_results_swy.shp table in the form of
```
phargogh, line 786: Come to think of it, this, too, would ideally use the module-level variable somehow if it needed to refer to the filename.

tests/test_seasonal_water_yield_regression.py
emlys: Thanks for the feedback, I wasn't sure which way to go on this. I suppose it's really unlikely that this filename will be changing much in the future, so the benefit of having it defined in one place doesn't matter.

tests/test_seasonal_water_yield_regression.py
emlys: Sounds good, I'll revert it to static strings. Agree that ideally all usages of the string would be replaced with the variable, which seems impractical here.

phargogh:
Thanks for updating the UG as well, @emlys !  I see that commit went through on the [automated build of the page](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/seasonal_water_yield.html), so that all looks good.  Could you update the user's guide revision SHA in `Makefile` to point to that revision as well?

The `Makefile` update isn't _strictly_ necessary (the minor release process would grab the SHA of the latest change on the UG master branch), but I do think it'd be good practice.  After that, we should be good to go!

PR #252, opened 2020-07-31
Pr resolution

emlys:
[https://github.com/natcap/invest/pull/251](url) Merge conflict resolution for this automatically generated PR


tests/test_hra.py
```diff
@@ -1146,3 +1146,56 @@ def test_validate_negative_resolution(self):
...
+                                     (a_feature_fid, b_feature_fid))
+            a_feat = None
+            b_feat = None
+            a_feat = a_layer.GetNextFeature()
+            b_feat = b_layer.GetNextFeature()
+
+        a_layer = None
+        b_layer = None
+        a_vector = None
+        b_vector = None
```
phargogh, line 1201: Hey @emlys I think this function was removed in `release/3.9` but then maybe `git` incorrectly re-added it here?

In `release/3.9` we refactored some of our custom assertions into `natcap.invest.utils`, including this function.

tests/test_recreation.py
```diff
@@ -1188,6 +1189,44 @@ def test_bad_predictor_table_header(self):
...
+                    assert(ev is None)
+            feature = None
+            actual_feature = None
+    finally:
+        actual_layer = None
+        actual_vector = None
+        expected_layer = None
+        expected_vector = None
+
+
```
phargogh, line 1229: I think this function was removed (or at least moved) in `release/3.9` as well.  Maybe git reintroduced the function in the merge?

tests/test_sdr.py
```diff
@@ -27,11 +27,22 @@ def assert_expected_results_in_vector(expected_results, vector_path):
     watershed_results_layer = None
     watershed_results_feature = None
     for key in expected_results:
-        numpy.testing.assert_approx_equal(
-            actual_results[key], expected_results[key], significant=6,
-            err_msg=(
-                f'actual_results: {actual_results}, '
-                f'expected_results: {expected_results}'))
+        # numpy.testing.assert_almost_equal(
+        #     expected_results[key], actual_results[key], decimal=6)
```
phargogh, line 31: Is there a benefit to leaving these lines commented out?  If there isn't, could we remove them?

tests/test_sdr.py
```diff
@@ -27,11 +27,22 @@ def assert_expected_results_in_vector(expected_results, vector_path):
...
-                f'expected_results: {expected_results}'))
+        # numpy.testing.assert_almost_equal(
+        #     expected_results[key], actual_results[key], decimal=6)
+
+        # In order to pass with GDAL<2.3 and GDAL>2.3:
+        # asserting equality to 5 significant figures instead of 6 decimal
+        # places. GDAL 2.3 introduced new warping behavior yielding different
+        # pixel values when also using a smoothing interpolation method.
+        # Surprisingly, these differences are not washed out by an
+        # aggregation such as zonal statistics.
```
phargogh, line 38: Since we're merging this into `release/3.9`, which requires `GDAL>3`, I think we can safely remove this comment.

And on a related note, since we no longer need to accommodate two versions of GDAL that have different warping behaviors (leading to small but notable numerical differences), is it reasonable to bump down the tolerance back to 6 significant digits?  Or maybe you're already doing this in the `assert_allclose` call, just below this?

tests/test_wind_energy.py
```diff
@@ -732,6 +732,83 @@ def test_missing_valuation_params(self):
...
+                        (a_field, a_feature_fid, a_value, b_feature_fid, b_value))
+
+            a_feat = None
+            b_feat = None
+            a_feat = a_layer.GetNextFeature()
+            b_feat = b_layer.GetNextFeature()
+
+        a_shape = None
+        b_shape = None
+
```
phargogh, line 811: The wholesale inclusion of this function (like with in `test_hra.py` and `test_recreation.py` makes me wonder if `git` included the function by mistake in the merge.  Is this function used in the file?  Is it needed?

emlys: Oh, thanks for catching these! Yes it must have been a result of the merge.

tests/test_sdr.py
emlys: Good point! Decreased rtol to 0.000001

Issue #3, opened 2020-01-29
Use pytest instead of nosetests for testing


phargogh:
PyTest has become probably the most-used testing framework for python applications, while `nosetests` (https://pypi.org/project/nose/#history) hasn't been updated since 2015.

Although nosetests works, I don't know of a reason to stick with something that isn't actively maintained and hasn't been for some time.  I think we should switch to `pytest`.

PR #253, opened 2020-08-03
#3 switched make test from nose to pytest

emlys:
Addresses natcap/invest#3 

Some notes about the switch to pytest:

1. Test timers: There doesn't seem to be an exact equivalent of nose-timer for pytest. There is a package pytest-timer [https://github.com/skudriashev/pytest-timer](url) which I tried out, but it threw an error. There is also the `--durations=n` option that comes with pytest. This displays the durations of the n-slowest tests in a separate section at the bottom; setting `--durations=0` displays durations for all tests. Neither this nor pytest-timer seems to have an option to display the times inline, like nose-timer does.

2. Coverage: nose seems to automatically(?) ignore anything in the invest/ui/ directory when calculating coverage. Pytest was including invest/ui/ which made the overall coverage much lower. I added the config file `.coveragerc` with a setting to omit invest/ui/.

3. Making use of pytest features. Pytest has some additional features such as not requiring tests to be contained in a class, and encouraging the use of `assert` statements rather than `self.assert*()`. I think nose has some similar features which we don't appear to be using. Since pytest also accepts tests in the standard unittest format like we have, I think it's best to leave things as is rather than conform to the pytest style. That way, if we want to stop using pytest in the future, the test files will likely remain compatible with other frameworks.

4. At the end of the tests, pytest displays a rather large section of deprecation warnings. I'm guessing these weren't caught by nose because it hasn't been updated in a long time. Probably worth addressing these in a separate ticket.

5. There's no `-P` (don't make changes to sys.path) option in pytest. The closest thing is `--import-mode=importlib`: "new in pytest-6.0, this mode uses importlib to import test modules. This gives full control over the import process, and doesn’t require changing sys.path or sys.modules at all." For some reason this causes 3 tests in `test_validation.py` to fail with an import error. I can look into this more if it's important that sys.path isn't changed?


Makefile
```diff
@@ -154,7 +153,10 @@ $(BUILD_DIR) $(DATA_DIR) $(DIST_DIR) $(DIST_DATA_DIR):
 	$(MKDIR) $@
 
 test: $(GIT_TEST_DATA_REPO_PATH)
-	$(TESTRUNNER) tests
+	@echo "[run]" > .coveragerc
+	@echo "omit = */invest/ui/*" >> .coveragerc
```
emlys, line 157: I originally did this in one line with an `\n` using `echo -e`, but the windows tests were failing because the `-e` flag wasn't recognized.

phargogh: Interesting!  Although it looks like the Windows tests are still failing with an exception ([example line](https://github.com/natcap/invest/runs/942585849?check_suite_focus=true#step:12:260)), perhaps related to the encoding of the file?

Makefile
```diff
@@ -154,7 +153,10 @@ $(BUILD_DIR) $(DATA_DIR) $(DIST_DIR) $(DIST_DATA_DIR):
...
+	-$(TESTRUNNER) tests/test_validation.py
```
phargogh, line 158: Looks like this still just running the tests for validation and not the full suite of model tests.

phargogh:
@emlys I'll respond to each of your notes in the PR description here.

1. The `--durations=n` option sounds just fine to me.  We had `nose-timer` in there left over from an older project where we were trying to reduce the runtime of the test suite.  The pytest-provided `--durations` option sounds like it'll serve the same purpose when we're looking to reduce the runtime further.
2. That sounds like a nice solution about the `.coveragerc`!  Nose would only show coverage for the files that were imported by any tests that ran, so this sounds like a notable difference in how the two test runners behave.
3. Your assessment about test harnesses in python sounds spot on, and I'll add two more justifications: 
    1. the python standard library is slow to change, and so anything within the stdlib should continue to be a supported format for some time into the future.  Therefore, if we want to maximize compatibility across future test runners, we should go with unittest, which is the official test harness.
    2. The `assert` keyword is interesting because they are [disabled by calling `python -O` (turning on optimizations)](https://docs.python.org/3.8/reference/simple_stmts.html#the-assert-statement).  So for how conveniently brief and expressive they are, `assert` statements are a bit unreliable if we're running python within a context where we might not have control over how it's invoked.  Not that we have that issue at present, but it's something to be aware of when writing software to be run on other people's computers.
4. I agree that any warnings we have control over should be addressed.  Could you create some new bugfix issues for the warnings that a5e within this project?
4. The `-P` flag was originally there because `nosetests` would reach into the `src` directory and attempt to import the files it found there.  This is problematic because our cython-compiled modules aren't located in `src/**`, they're compiled and installed into our activated conda/virtualenv environment.  Adding the `-P` flag would force `nosetests` to ignore `src/`.  So, I'd defer to you on what you think would be the best approach here, with the goal of having all of our tests be able to import what they need to import and test what they need to test.

Makefile
emlys: Oh! I had added `-` to the line running the tests so it would continue to the next line and remove `.coveragerc` whether the tests pass or fail. Just realized that is a really bad idea since the tests will always appear to pass!

src/natcap/invest/validation.py
```diff
@@ -833,8 +833,19 @@ def _wrapped_validate_func(args, limit_to=None):
...
+        # Pytest in importlib mode makes it impossible for test modules to 
+        # import one another. This causes a problem in test_validation.py,
+        # which gets imported into itself here and fails.
+        # Since this decorator might not be needed in the future,
+        # just ignore failed imports; assume they have no ARGS_SPEC.
+        try:
+            model_module = importlib.import_module(validate_func.__module__)
+        except:
+            LOGGER.warning('Unable to import module %s: assuming no ARGS_SPEC.'
+                            % validate_func.__module__)
```
phargogh, line 845: Ah, sorry, this is a pretty trivial request, but I suspect a linter will suggest that we pass `validate_func.__module__` as a parameter rather than through string formatting.  I used to think that the parameter-passing approach here was the recommended way, but now I'm not sure where I saw that and suspect it came from pylint or the like.  In any case, the parameter-passing approach is how most non-`f`-string log messages are handled throughout InVEST, so I think this could be worth changing for the sake of consistency.

And as a side note, I had previously thought that the thing I'm suggesting was (for security purposes) the recommended approach for logging, but that doesn't seem to be the case.  All that `logging` does behind the scenes (by default, anyways) is [exactly what you have here, just a `%`-based parameter substitution](https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L363).  The one hypothetical advantage to passing args by parameter is that it does give us the _option_ down the road to sanitize the parameters somehow (like if we're dealing with user input in a webapp) before sending it to a remote logging server or something.

emlys: Interesting about what `logging` is doing! I put it in a parameter instead of string formatting

Issue #51, opened 2020-03-05
Use the args key in datastack archive rasters and vectors


phargogh:
When creating a datastack archive, InVEST creates a randomly-generated directory that will definitely not conflict in name with other directories in the same datastack.  To do this, a randomish directory name is generated.  A downside to this is that you have no idea what that raster is supposed to represent unless you're also looking at the `parameters.json` file.

Can't we at least include the args key names in the directory name?  It would be very nice to be able to just look at the uncompressed datastack on disk and know which parameter I'm looking at.

Issue #209, opened 2020-07-07
Rebuild Coastal Blue Carbon


phargogh:
The InVEST Coastal Blue Carbon model has a few critical issues, doesn't conform to our current development standards and is just plain hard to debug when an issue comes up.  I'd like to rebuild this from the ground up.

I'd also like to make a couple of changes while I'm in there.  Specifically:

- [x] Replace the UI's `MultiInput`s with a CSV (and delete `MultiInput` from `natcap.invest.ui`)
- [x] Update the User's Guide to clarify what should happen before the first transition (and make sure the model doesn't do anything before the first transition)
- [x] Improve error handling when duplicate landcover class names are provided
- [x] ~~Fix the longstanding, undocumented issue that carbon values _should_ be calculated per pixel, but are actually calculated per hectare.  This is in contrast with what's stated in the User's Guide: "We estimate accumulation by multiplying habitat specific rates of carbon accumulation by the given cell area".~~ We decided not to do this, as many carbon scientists don't actually need it and this is counterproductive when doing global carbon modelling (where pixel sizes differ across the map).
- [x] Include some validation for half-life values to make sure that `halflife > 0` ([related forums issue](https://community.naturalcapitalproject.org/t/coastal-blue-carbon-empty-outputs/1181/2?u=jdouglass)).
- [x] Properly describe the model math like what we have in other user's guide chapters.  Not having this be well specified leads to occasional questions on the forum about what, exactly, is happening.  ([recent example](https://community.naturalcapitalproject.org/t/calculation-principle-of-carbon-storage-in-coastal-blue-carbon-model/1179/4?u=jdouglass))

This rebuild would also address an issue that came up in [this forums thread](https://community.naturalcapitalproject.org/t/coastal-blue-carbon-output-questions/976/15) where we discovered that the model was producing emissions outputs before the first transition (which shouldn't be happening).  As clarified by Stacie:

> So my two cents: I also would assume that no emissions happen until the first transition, that seems perfectly reasonable, since the transition is what's causing the emissions. If we were to decide otherwise, what would it be based on, given the required input data? The one thing that I *would* assume happens between baseline and first transition is carbon sequestration, since we provide accumulation information for baseline habitat types.

So between baseline and 1st transition, we should still be accumulating carbon, but not emitting until after the first transition.

This would ultimately subsume #17 and #135 and hopefully make our lives a whole lot easier when issues come up on the forums.

phargogh:
[...]

PR #254, opened 2020-08-06
#51 use filename as prefix for datastack folder name

emlys:
Use the filenames as a prefix onto the datastack data folder names, so you can tell what's inside them.

E.g. using the SDR sample data, the datastack `sdr_natcap.invest/data/` would contain a folder named `raster_r9brgocp`, which contains `DEM_gura.tif` and the associated files. Now that folder will be named `DEM_gura_raster_r9brgocp`.

The folder name prefix is derived from the parameter value (a filepath) not the parameter key, so if the user inputs confusingly-named files, the folder names will be similarly unhelpful. But hopefully they will know what it means. Using the parameter key instead would require a more extensive refactor.


src/natcap/invest/datastack.py
```diff
@@ -85,10 +85,17 @@ def _collect_spatial_files(filepath, data_dir):
...
+    filename = os.path.basename(os.path.splitext(filepath)[0])
+
     with utils.capture_gdal_logging():
         raster = gdal.OpenEx(filepath, gdal.OF_RASTER)
         if raster is not None:
-            new_path = tempfile.mkdtemp(prefix='raster_', dir=data_dir)
+            # give the folder a descriptive name with a unique
+            # random suffix to avoid name conflicts
+            new_path = tempfile.mkdtemp(
+                prefix='%s_raster_' % filename,
```
dcdenu4, line 97: This might be nitpicky but I feel like we've slowly been adopting f-strings for formatting. I think it reads a little bit nicer. Not a big deal.

src/natcap/invest/datastack.py
```diff
@@ -123,7 +130,9 @@ def _collect_spatial_files(filepath, data_dir):
                 vector = None
                 return None
 
-            new_path = tempfile.mkdtemp(prefix='vector_', dir=data_dir)
+            new_path = tempfile.mkdtemp(
+                prefix='%s_vector_' % filename,
```
dcdenu4, line 134: Same note about maybe using f-strings instead?

src/natcap/invest/datastack.py
emlys: I saw some other old-style formatted strings in there, but I agree it's probably better to switch to `.format()`. Just fixed these

emlys:
> Hey @emlys I only have a few comments. One is the mention of maybe using f-string for string formatting. The other is to update the docstring for the `_collect_spatial_files` function. I think in that docstring there is mention of how the prefixing works.
> 
> I also wonder if it's worth talking about not being able to grab the Args Keys for the prefix? You mention it in the PR, it would also be good to document your thoughts on that decision in the issue itself, in case we ever re-open that issue.

Thanks for pointing out the docstring! 

> Hey @emlys I only have a few comments. One is the mention of maybe using f-string for string formatting. The other is to update the docstring for the `_collect_spatial_files` function. I think in that docstring there is mention of how the prefixing works.
> 
> I also wonder if it's worth talking about not being able to grab the Args Keys for the prefix? You mention it in the PR, it would also be good to document your thoughts on that decision in the issue itself, in case we ever re-open that issue.

After thinking about it more, I like it better using the args keys than the user-supplied file names. I updated it to generate the folder name that way instead; let me know what you think

src/natcap/invest/datastack.py
```diff
@@ -341,8 +355,12 @@ def _recurse(args_param, handler, nested_key=None):
                                  possible_path, filepath)
                     return filepath
                 except KeyError:
+                    # turn the nested key into a nice name for a folder
+                    # e.g. args['category']['data_path'] --> category_data_path
+                    folder_prefix = nested_key[6 : -2].replace('\'][\'', '_')
```
emlys, line 360: I'm using the nested_key here because, if there were to be a nested params dictionary, it would uniquely identify each param, making for a helpful folder name.
However, it's not ideal if there is a list... see above

src/natcap/invest/datastack.py
```diff
@@ -325,7 +339,7 @@ def _recurse(args_param, handler, nested_key=None):
                 handler.removeFilter(args_key_filter)
             return new_dict
         elif isinstance(args_param, list):
-            return [_recurse(list_item, handler) for list_item in args_param]
+            return [_recurse(list_item, handler, nested_key=nested_key) for list_item in args_param]
```
emlys, line 342: Without `nested_key=nested_key`, `nested_key` would get reset to `None` anytime there's a list. That's a problem for the folder-naming, so I added this. It will affect the log filtering as well, but I don't think that's a problem... any idea why we would want the log filter to be reset when there's a list?

The other issue here is that, if there is a list, there is no longer a unique prefix for each folder. Each path in the list would go into a new folder with the same prefix. The only way to identify them separately would be to incorporate something from the path, such as the filename, into the folder name. Since I was previously using only the filename as the prefix, and not the nested_key, if this is needed I would be in favor of just naming folders according to filenames.

dcdenu4: Hey @emlys I think I'm finally wrapping my head around what Datastack is trying to do when archiving. 

For the list issue, is it possible to pass that `nested_key` like you have, but tack on the index of the value being passed to the `nested_key`? Something like:
```python
return [_recurse(list_item, handler, nested_key=f"{nested_key}{idx}") for idx, list_item in enumerate(args_param)]
```
I'm trying to think of how that could fail...?

src/natcap/invest/datastack.py
dcdenu4: See above my comments on lists. I think this replacement looks good!

emlys:
> Hey @emlys, I added a comment on a possible way to handle the list issue? It's highly possible I am missing something and my proposed solution won't work. But maybe it'll spark a better idea. Let me know!

Thanks for the suggestion, that would definitely be the better way to deal with lists! However when #209 is complete, it seems that list inputs will be obsolete, so @phargogh said it's ok to just ignore this case and take out the whole list condition. That would simplify the folder naming as well

emlys:
Here's what the datastack looks like now:

Screen Shot 2020-08-07 at 2 55 19 PM

Issue #214, opened 2020-07-09
Strip leading and trailing spaces from CSV columns


phargogh:
We commonly run into issues on the forums where leading and trailing spaces in a CSV column title causes indexing issues later on.  If it's reasonable to strip leading and trailing spaces from tabular inputs, we might be able to save ourselves some time debugging down the road.

Example forums post from recently: https://community.naturalcapitalproject.org/t/habitat-quality-vauleerror-encounter/1137/2

PR #259, opened 2020-08-07
Standardize reading CSVs to dataframes

emlys:
#214 
`utils.build_lookup_from_csv()` is used throughout the models for reading CSVs into dictionaries in a standard way. There are a lot of places though where the CSV is read into a dataframe instead with `pandas.read_csv()`. I propose adding a new utils method, `read_csv_to_dataframe()`, to handle these cases and make it easier to standardize column names.


src/natcap/invest/utils.py
```diff
@@ -538,6 +541,30 @@ def build_lookup_from_csv(
...
+
+    Args:
+        path (string): path to a CSV file
+        to_lower (bool): if True, convert all column names to lowercase
+        strip (bool): if True, strip leading/trailing whitespace from all
+            column names
+        **kwargs: any kwargs that are valid for ``pandas.read_csv`` or
+            ``pandas.read_excel``
+    """
+    dataframe = pandas.read_csv(path, **kwargs)
```
phargogh, line 559: One of the critical things that `build_lookup_from_csv` does is checking for the presence of a BOM, which then changes the encoding that we assume.  I'd suggest that we include that checking here as well BUT, the presence of `**kwargs` makes that a little more interesting because a user should also be able to provide the `encoding` keyword argument and have that take precedent.

src/natcap/invest/wave_energy.py
```diff
@@ -619,9 +618,9 @@ def execute(args):
         output_dir, 'GridPts_prj%s.shp' % file_suffix)
 
     grid_data = grid_land_data.loc[
-        grid_land_data['TYPE'].str.lower() == 'grid']
+        grid_land_data['type'].str.lower() == 'grid']
     land_data = grid_land_data.loc[
-        grid_land_data['TYPE'].str.lower() == 'land']
+        grid_land_data['type'].str.lower() == 'land']
```
phargogh, line 623: Would the `.str.lower()`s here be subsumed by the `to_lower` parameter?

src/natcap/invest/wind_energy.py
```diff
@@ -929,22 +928,18 @@ def execute(args):
...
-        ]
+        grid_land_df = utils.read_csv_to_dataframe(args['grid_points_path'])
 
         # Make separate dataframes based on 'TYPE'
         grid_df = grid_land_df.loc[(
-            grid_land_df['TYPE'].str.upper() == 'GRID')]
+            grid_land_df['type'].str.upper() == 'GRID')]
         land_df = grid_land_df.loc[(
-            grid_land_df['TYPE'].str.upper() == 'LAND')]
+            grid_land_df['type'].str.upper() == 'LAND')]
```
phargogh, line 937: Looks like this could be simplified as well with the `to_lower` kwarg.

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -1458,8 +1458,8 @@ def _validate_same_projection(base_vector_path, table_path):
     """
     # This will load the table as a list paths which we can iterate through 
     # without bothering the rest of the table structure
-    data_paths = pandas.read_csv(
-        table_path, squeeze=True, usecols=['path']).tolist()
+    data_paths = utils.read_csv_to_dataframe(
+        table_path, squeeze=True)[['path']].tolist()
```
phargogh, line 1462: Sorry, this is in part due to my own naivete with `pandas`: what are the tradeoffs to consider between the `usecols` approach and the indexing (`[['path']]`) here?

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2891,7 +2889,7 @@ def _validate_habitat_table_paths(habitat_table_path):
     Raises:
         ValueError if any vector in the ``path`` column cannot be opened.
     """
-    habitat_dataframe = pandas.read_csv(habitat_table_path, header=0)
+    habitat_dataframe = utils.read_csv_to_dataframe(habitat_table_path)
```
phargogh, line 2892: Do we lose anything here by removing the `header=0` kwarg?

src/natcap/invest/utils.py
```diff
@@ -538,6 +541,30 @@ def build_lookup_from_csv(
...
+    This helps avoid common errors caused by user-supplied CSV files with
+    column names that don't exactly match the specification.
+
+    Args:
+        path (string): path to a CSV file
+        to_lower (bool): if True, convert all column names to lowercase
+        strip (bool): if True, strip leading/trailing whitespace from all
+            column names
+        **kwargs: any kwargs that are valid for ``pandas.read_csv`` or
+            ``pandas.read_excel``
```
phargogh, line 557: This mentions `read_excel` as well ... was the intent here to also add excel support?

src/natcap/invest/hra.py
```diff
@@ -2457,10 +2457,11 @@ def _get_info_dataframe(base_info_table_path, file_preprocessing_dir,
...
                          base_info_table_path)
 
-    # Convert column names to upper case, and encode it to strings if it's not
-    # a string (e.g. unicode)
+    # Convert column names to upper case, strip whitespace,
+    # and encode it to strings if it's not a string (e.g. unicode)
     info_df.columns = [
-        col.upper() if isinstance(col, str) else col.encode('utf-8').upper()
+        col.strip().upper() if isinstance(col, str) 
+        else col.encode('utf-8').upper()
```
phargogh, line 2464: This `.encode` stuff reminds me of how we used to have to handle encodings for python2/python3 cross-compatibility.  Now that `natcap.invest` is python3-only, perhaps the `.encode` stuff here can be removed now that all `str` objects are unicode?

src/natcap/invest/hra.py
```diff
@@ -2606,7 +2607,7 @@ def _get_criteria_dataframe(base_criteria_table_path):
 
     # Validate the column header, which should have 'criteria type'
     criteria_df.columns = [
-        x if isinstance(x, str) else None for x in
+        x.strip() if isinstance(x, str) else None for x in
```
phargogh, line 2610: Wouldn't this also be handled by the `strip` kwarg?

src/natcap/invest/wave_energy.py
emlys: This raises another point, which is should all string data in the CSV be converted to lowercase, or just the column headers? Right now `read_csv_to_dataframe()` is only modifying the column headers, so the `.lower()` here would still be needed, but I think it could make sense to apply it to the entire dataframe

src/natcap/invest/recreation/recmodel_client.py
emlys: `usecols` will only read in the given columns into the dataframe, so if the column name had extra whitespace e.g. 'path ', I don't think it would work. So rather than selecting the column when the CSV is read, I'm first reading in all the columns, removing extra whitespace, and then indexing by 'path'.

src/natcap/invest/coastal_vulnerability.py
emlys: I'm not sure why it was there, I don't believe it's doing anything since no column names are being passed. It should be equivalent to the default behavior. see pandas docs: [https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html](url)

src/natcap/invest/utils.py
emlys: Oh yeah... after seeing that excel was an option for the HRA model, I tried adding it, but I think it's better to handle the HRA model inputs separately unless we decide that all models should allow excel files

src/natcap/invest/hra.py
emlys: `criteria_df` and `info_df` (above) are still created from `pandas.read_csv()` here, not `utils.read_csv_to_dataframe()`. This is partly because of the need to handle possible excel files in this one model, and partly because they are being converted to uppercase instead of lowercase. The data is indexed throughout the model using all-uppercase titles, which I'd be in favor of replacing with lowercase, unless there's a reason for it to be uppercase..

src/natcap/invest/wind_energy.py
```diff
@@ -1769,7 +1763,7 @@ def _read_csv_wind_data(wind_data_path, hub_height):
             to dictionaries that hold wind data at that location.
 
     """
-    wind_point_df = pandas.read_csv(wind_data_path)
+    wind_point_df = utils.read_csv_to_dataframe(wind_data_path, to_lower=False)
```
emlys, line 1766: Here's another place where it would be nice to have everything in lowercase, but all-uppercase columns are used throughout this model. I'd like to standardize it (to either upper- or lower-case) for all the models, but it seems like there might be a preference for uppercase here?

src/natcap/invest/wave_energy.py
phargogh: That _is_ an interesting question.  Usually we like to use a rule of 3 for creating general solutions, to avoid premature optimization, and to make sure we're developing abstractions that will definitely be used.  In this case, do we have 3 cases in InVEST where we benefit from case-insensitive string comparisons in InVEST?

src/natcap/invest/recreation/recmodel_client.py
phargogh: Ah,  I see ... very interesting.

src/natcap/invest/coastal_vulnerability.py
phargogh: Interesting!  This does look a bit redundant to have here, then.

src/natcap/invest/utils.py
phargogh: 👍

src/natcap/invest/hra.py
phargogh: Ah, that's right ... the excel/CSV compatibility will make things a little more interesting.

As for whether the CSV headers are all-uppercase or all-lowercase, I don't personally have a preference and it sounds like an implementation decision made by the engineer who wrote this version of the model (who has since moved on).  Consistency is nice to have, so I'd support converting these to be lowercase in HRA.

src/natcap/invest/wind_energy.py
phargogh: You know, the all-uppercase thing might just be an artifact of age.  Both HRA and Wind Energy were created back when our standard table input format was `.dbf`, and I believe that ArcGIS would create column names that were in all-caps.

I personally don't have a preference about all-uppercase vs. all-lowercase, so long as we can support case-insensitivity.  In this particular case @dcdenu4 implemented the model.  Doug, do you remember if we made a conscious choice in Wind Energy about using all-uppercase or all-lowercase column indexing?

src/natcap/invest/wave_energy.py
emlys: It looks like `utils.build_lookup_from_csv()` does make the entire dataframe lowercase by default, which must be used in several places. And there are definitely at least 3 places in this PR where the case of the column names matter. The case of the values matters here, in wind_energy.py below, and in hra.py: [https://github.com/natcap/invest/pull/259/files#diff-ef8666f31ebc9043012bf1e7a9abadd5R2480](url), and likely several other places.

src/natcap/invest/utils.py
emlys: Do you mean user, as in where `read_csv_to_dataframe` is called, or as in the actual user supplying the files? I don't think there's currently the option for them to supply a specific encoding

src/natcap/invest/wave_energy.py
phargogh: I don't doubt that a third one exists, but I'd also suggest that we confirm it before abstracting the functionality into the proposed helper function.

You do also make a good point that `build_lookup_from_csv` also provides this, and we could offer it in this new function in the interest of consistency.

src/natcap/invest/utils.py
phargogh: In this case, "user" is the caller of the function.  You're right that there isn't currently an option to provide a specific encoding to `utils.read_csv_to_dataframe` (and that's probably OK for our purposes until we know otherwise).  But the thing is that this proposed function is accepting _all possible keyword arguments_ that are accepted to the `pandas.read_csv`, and so if the caller provides an `encoding` kwarg (which is a valid input to `pandas.read_csv()`), we should honor that.

The catch is that if the caller _doesn't_ provide an `encoding` kwarg, then we should take advantage of our experience dealing with the CSV encodings we usually deal with and assume UTF-8, either with or without a BOM.  If we don't support an optional BOM we risk needing to spend more time debugging this on the forums down the road and needing to patch it later on.

So, the main question I have for you comes back to the use cases for this function: what are the features that we demonstrably need (given our 3+ examples of where this function will be used)?  What are the problems that this helper function will solve?  And, more to the point, is the `**kwargs` needed for our use cases?

src/natcap/invest/hra.py
emlys: After taking a poll on slack, it looks like lowercase is preferred by all, so I'll go ahead with that.

src/natcap/invest/wave_energy.py
emlys: Here's a summary:

Table values should be made lowercase:
- Crop Production (so crop names match those in accepted list)
- Wave energy (grid connection points TYPE value)
- Wind energy (grid connection points TYPE value)

Table values should not be made lowercase:
- HRA Habitat and Criteria CSVs (contains filepaths)
- Forest Carbon Edge Effect (keep case in description field)
- Recreation client (contains filepaths)

Doesn't matter:
- Carbon
- Coastal blue carbon
- Pollination
- SWY
- NDR
- SDR
- Urban cooling
- Urban flood risk
- Habitat quality
- Wave energy (other tables)
- Wind energy (other tables)

Scenic Quality: No CSV input.

Finfish, Fisheries, Recreation server: Reads CSVs in a specialized way line by line. It seems best not to mess with this.

emlys: My conclusions:
- Making the column names lowercase is always good
- Making all the values lowercase is never necessary, it is only needed for a couple of columns in a couple of inputs.
- Therefore, the `to_lower` arg to `build_lookup_from_csv` should be changed so it only affects the column names, not the entire dataframe. `.lower()`ing of the data should be done in the model only where necessary.
- We could do away with the `to_lower` arg for `build_lookup_from_csv` and `read_csv_to_dataframe` entirely, and have them always make the column names lowercase.

phargogh: > Making the column names lowercase is always good

It's not about a value statement, it's about the problem we're trying to solve.  And in this case, the problem we're trying to solve here is that we want to handle column names in a case-insensitive way and (ideally) consistently across models.  To that end, using lowercase column names sounds like a good convention.

> Making all the values lowercase is never necessary, it is only needed for a couple of columns in a couple of inputs.

You're right, I don't think we can safely lowercase _all_ string table values, and we've probably just gotten lucky with the ones that have been lowercased!  Case will often matter (depending on the OS) for filepaths provided via a table, for example.

> Therefore, the `to_lower` arg to `build_lookup_from_csv` should be changed so it only affects the column names, not the entire dataframe. `.lower()`ing of the data should be done in the model only where necessary.

This sounds right to me.  I don't think we should be manipulating user input without good reason, and that would best be done within the model on a case-by-case basis.

> We could do away with the to_lower arg for build_lookup_from_csv and read_csv_to_dataframe entirely, and have them always make the column names lowercase.

So long as we're committing to this convention of table indexes using lowercase strings, then yes, it would follow that we could remove the `to_lower` parameter from the function.

> If we're standardizing all column names to lowercase, this should also apply to the names of columns in CSVs that the model outputs.

We should take a closer look at this!  It is a related problem, but it affects different audiences.

Standardizing internal table indexing is a strictly _internal_ convention, so members of the software team are the only ones likely to be affected by the change.  Changing fieldnames in outputs, however, is an _external_-facing change, and so we need to also consider who this change will affect and to what degree it will affect them.

Because of these considerations, I suggest that we defer this for the time being.  Could you create an issue for it so we can take a look at it later on?

emlys: Thanks! Linked issue here: https://github.com/natcap/invest/issues/262

src/natcap/invest/utils.py
```diff
@@ -482,9 +476,13 @@ def build_lookup_from_csv(
     if col_list and key_field not in col_list:
         col_list.append(key_field)
    
+   # utf-8-sig will work whether or not there is a BOM
     table = pandas.read_csv(
         table_path, sep=None, index_col=False, engine='python', 
-        encoding=encoding)
+        encoding='utf-8-sig')
```
emlys, line 482: In reading about Unicode I learned that utf-8-sig works for decoding with or without a BOM, so it shouldn't be necessary to check for it here. (https://docs.python.org/3/howto/unicode.html#reading-and-writing-unicode-data)

src/natcap/invest/hra.py
emlys: I would think so, all the column names should be `str`

src/natcap/invest/hra.py
emlys: Is there any reason that a column name wouldn't be a `str`?

src/natcap/invest/utils.py
emlys: By including a kwarg `encoding='utf-8-sig'` in the method signature, I believe it should default to UTF-8 with/without a BOM, and still allow the caller to override it, see comment below about 'utf-8-sig'. 

There aren't any cases where stripping the whitespace is undesirable, so I don't think there needs to be a kwarg for that. Seems safe to assume it should always be stripped.

There are at least 3 cases where we don't want to make column names lowercase, at least until we decide if that should be a convention. So I'm making that a kwarg `to_lower=False`.

There are 5 cases where the `**kwargs` is used: see changes in recmodel_client.py, globio.py, validation.py, wave_energy.py, wind_energy.py. So that makes sense to me to include as a feature of the helper function.

src/natcap/invest/ui/model.py
```diff
@@ -1069,15 +1069,15 @@ def post_warnings(self, validation_warnings):
...
+            # html.escape handles escaping of characters <, >, &, " for HTML.
             self.title_label.setText(
                 '<h2>Validation warnings found</h2>'
                 '<h4>To ensure the model works as expected, please fix these '
                 'erorrs:</h4>')
             self.label.setText(
                 '<ul>%s</ul>' % ''.join(
                     ['<li><b>%s</b>: %s</li>' % (
-                        ', '.join(labels), cgi.escape(warning_, quote=True))
+                        ', '.join(labels), html.escape(warning_))
```
emlys, line 1080: Ran into this issue a little while ago, fixing it in this PR. `cgi.escape` has been deprecated since 3.2 and doesn't exist in 3.8 (https://docs.python.org/3.7/library/cgi.html#cgi.escape). `html.escape` now defaults to `quote=True`.

src/natcap/invest/wave_energy.py
davemfish: > > Making all the values lowercase is never necessary, it is only needed for a couple of columns in a couple of inputs.
> 
> You're right, I don't think we can safely lowercase _all_ string table values, and we've probably just gotten lucky with the ones that have been lowercased! Case will often matter (depending on the OS) for filepaths provided via a table, for example.

You're both on point. I just got tripped up by this in the Rec model on linux, where filepaths are being lowercased by `utils.build_csv...`.

src/natcap/invest/wind_energy.py
dcdenu4: That was indeed a long time ago. Off the top of my head I have no recollection of an all-uppercase vs. all-lowercase decision.  I'd have to check it out.

src/natcap/invest/hra.py
phargogh: I can't think of one, but I suppose a user might provide a numeric column name or something, which would probably be a good case to test against in our test suite.

src/natcap/invest/utils.py
phargogh: Great!  That whole BOM encoding thing was mighty frustrating for a long time, and if `utf-8-sig` handles the BOM if it's there, then that'll help simplify quite a few things!  I added an issue for this at https://github.com/natcap/invest/issues/264

src/natcap/invest/utils.py
```diff
@@ -538,6 +541,38 @@ def build_lookup_from_csv(
...
+            if first_line.startswith(codecs.BOM_UTF8):
+                encoding = 'utf-8-sig'
+    dataframe = pandas.read_csv(path, encoding=encoding, **kwargs)
+    # this won't work on non-string types
+    try:
+        dataframe.columns = dataframe.columns.str.strip()
+        if to_lower:
+            dataframe.columns = dataframe.columns.str.lower()
+    except:
+        pass
```
phargogh, line 572: Could I ask about this?  Is there an error that we're expecting to handle here?  Or is this a 'just in case' sort of exception handler?  When could we expect this handler to be triggered?

The reason I'm asking is that if there's a specific exception that we're expecting will be raised, it would be better to catch that specific exception and handle it appropriately.  If this is the case (and depending on the error), it might also be worth describing the circumstances under which the exception might be raised.

But the other big thing is that errors should not pass silently ... that'll come back to bit us _hard_ in the future when we can't find out exactly what happened from the log or immediately from the input files.  I think it'd be more maintainable for us in the long run for this to catch the specific exception we expect, or if this is really a clause that's likely to be triggered, I'd suggest that we add a `LOGGER.exception` here.

src/natcap/invest/wind_energy.py
dcdenu4: Okay, so I think the Wind model is probably okay for moving to lowercase, though some constants in the model are hard coded to uppercase. But there is a CSV input that is constructed such that the first column is a list of variable names and the second column are the values. 

Wave Energy also has some non standard CSV inputs and usage of shapefiles for column access. I think you're right James, I think the dbf files upper case the column names. Oh, Wave Energy also has a CSV input of just numbers. It's a wave machine performance input. 

So there are certainly some exceptions out there! But they might not play a role in this PR.

src/natcap/invest/wind_energy.py
```diff
@@ -1654,8 +1651,9 @@ def _read_csv_wind_parameters(csv_path, parameter_list):
 
     """
     # use the parameters in the first column as indeces for the dataframe
+    # this doesn't benefit from `utils.read_csv_to_dataframe` because there
+    # is no header to strip whitespace
```
phargogh, line 1655: This is very helpful context!  👍

src/natcap/invest/utils.py
```diff
@@ -538,6 +541,38 @@ def build_lookup_from_csv(
...
+        **kwargs: any kwargs that are valid for ``pandas.read_csv``
+    """
+    # Check if the file encoding is UTF-8 BOM first
+    # allow encoding kwarg to override this if it's provided
+    if not encoding:
+        with open(path, 'rb') as file_obj:
+            first_line = file_obj.readline()
+            if first_line.startswith(codecs.BOM_UTF8):
+                encoding = 'utf-8-sig'
+    dataframe = pandas.read_csv(path, encoding=encoding, **kwargs)
```
phargogh, line 565: Hey @emlys sorry, I thought I had read a comment somewhere that this logic could be entirely removed if we just defaulted to using `encoding='utf-8-sig'`, which is also what the python docs make it sound like is the case.  Is that not actually the case?

src/natcap/invest/utils.py
phargogh: And if this is related to the note above the block about not working on string types, can we replicate this in the tests?

src/natcap/invest/utils.py
emlys: Update: I'm no longer sure about this. `utf-8-sig` seems to work for UTF-8 with or without a BOM, but not for any other encodings. So `test_csv_not_utf_8` was failing. Weirdly this test passes with `encoding=None`. But `encoding=None` does not work when there is a BOM. I spent some time reading about this and still not sure what's going on. I don't think pandas guesses what the encoding is, so maybe the 'iso8859_5' Cyrillic encoding used in the test just happens to be interpretable as UTF-8?

phargogh: Uffda, encodings (and localization more generally) are quite the rabbit hole!  Thanks for the background and for looking into this.  The behavior you're describing does sound a bit bizarre ... I'm sorry I don't have any magic solutions here, and guessing encodings can be problematic at best (I highly doubt `pandas` is doing any guessing of encodings).  It's quite possible that the strings in the cyrrilic test are valid UTF-8 (if gibberish), and if that's true then there wouldn't be an error when decoding as UTF-8.

Fortunately, most of the issues we've encountered are merely with the presence or absence of the BOM, so we might be able to defer the true solution to the encodings issue until later.  There are other localization issues at hand, too, (example: `196,96 00` --> `196.96`) so we'll want to take a look at how we can offer more reliable interpretation of tabular inputs some time down the road and probably in the workbench.

Side note: If you're ever looking for some background information on encodings, the first couple chapters of https://www.oreilly.com/library/view/fonts-encodings/9780596102425/ are _awesome_ background and history on the subject.

src/natcap/invest/utils.py
emlys: I originally did this for the case where `header=None`, which cause the `dataframe.columns` to be an Index with dtype int64, so the string operations didn't work. Later decided there's no point in using this method at all if `header=None`, but kept the `try` block to handle the case where the CSV does have non-strings for some reason. I will make the exception more specific though.

tests/test_utils.py
```diff
@@ -765,3 +765,166 @@ def test_csv_latin_1_encoding(self):
...
+            file_obj.write(textwrap.dedent(
+                """
+                h1;h2;h3
+                a;b;c
+                d;e;f
+                """
+            ).strip())
+        # using sep=None with the default engine='python',
+        # it should infer what the separator is
+        df = utils.read_csv_to_dataframe(csv_file, sep=None)
```
phargogh, line 903: Ooo thanks for reminding me of this!  I'm sorry I had forgotten about it, but I would like to make a case for assuming `sep=None`, at least as a default.  Many users of InVEST (especially when the computer's language is not English) will export Excel tables to CSV but then their separator will be semicolons rather than commas.  Passing `sep=None` fixes that, and this is exactly what `build_lookup_from_csv`does.  What do you think?

tests/test_utils.py
```diff
@@ -765,3 +765,166 @@ def test_csv_latin_1_encoding(self):
...
+    def setUp(self):
+        """Make temporary directory for workspace."""
+        self.workspace_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        """Delete workspace."""
+        shutil.rmtree(self.workspace_dir)
+
+    def test_read_csv_to_dataframe(self):
+        """Test the default behavior"""
```
phargogh, line 782: Could I suggest that the docstring here be prefixed with `Utils: `, and the same applied to all of the `test_*` methods in this fixture?  It makes for easier skimming in the test logs.

tests/test_utils.py
emlys: That sounds like a good default to have!

phargogh:
Thanks for these updates, @emlys ! Functionally, this looks really good to me.  Since we now more uniformly ensuring that the models are case-insensitive across the board (a user-facing change), would this be worth noting in HISTORY?

emlys:
> Thanks for these updates, @emlys ! Functionally, this looks really good to me. Since we now more uniformly ensuring that the models are case-insensitive almost across the board (a user-facing change), would this be worth noting in HISTORY?

Based on the discussion from last week, I'd thought we were setting aside the case-insensitivity issue for now (https://github.com/natcap/invest/pull/259#discussion_r468892671). Did you just mean to set aside the part about changing the output table headers? There are some places where it would be rather awkward to keep the output table headers as they are while having an internal convention of all-lowercase... they're sort of intertwined. The current state of the PR is mostly leaving capitalization as it was, just introducing the `to_lower` option, which is used in some places and hopefully expanded to everywhere when case-insensitivity is more standardized. Let me know what you think about that.

Stripping whitespace is also user-facing, though shouldn't really affect how people use the program, hopefully no one is leaving in extra whitespace on purpose. I'll add a HISTORY note for it though!

phargogh:
> Based on the discussion from last week, I'd thought we were setting aside the case-insensitivity issue for now (#259 (comment)). Did you just mean to set aside the part about changing the output table headers?

Yep!  I only meant the part about the table headers.  We're absolutely still setting the issue of case within the table fields aside.  Like the stripping of spaces in column names, the case-insensitivity of column names is a user-facing change and should also be mentioned in `HISTORY`.

PR #174, opened 2020-06-17
Bugfix/171 sq too many open files

phargogh:
This PR addresses an issue of having "too many files open" experienced by a user of our mac binaries who was running Scenic Quality with over 7000 viewpoints and the model was naively passing all 7000+ rasters to a single `raster_calculator` call.

So, this PR updates Scenic Quality to make better, more responsible use of memory and disk I/O when given large numbers of viewpoints.  Updates include:

* A refactor to the main `execute` function.  Previously, the model would iterate through every point in the viewpoints vector on every run through the model.  This looping (which also determines which points are valid viewpoints) has been refactored into a separate task (`_determine_valid_viewpoints`) in the graph to avoid that (potentially expensive) loop.
* Timed progress logging has been added to a few loops that may take some time when operating on many points.
* The task `_count_and_weight_visible_functions` has been refactored to load each block from each visibility raster in sequence rather than passing them all in to a single call to `pygeoprocessing.raster_calculator`.

There's also a bonus change to SDR that removes a bit of the `execute` docstring documenting an args key that was not actually used by the model.

Fixes #171

PR #260, opened 2020-08-10
Reclassification Draft Proposal

dcdenu4:
This commit is a test sample for how to handle reclassification messages for users. Once we decide on this format other models that use `reclassify_rasters` can adopt it. Here `sdr` uses `pygeoprocessing.ReclassificationMissingValuesError` to provide a more useful error message.

image


dcdenu4:
@phargogh I just made some more edits that deviate from the original plan. Specifically if you look at the SWY changes you'll see I've started to take the approach of checking LULC values against input tables up front in the models. As I was diving into abstracting out `pgp.reclassify_raster` tasks into other functions, it struck me that maybe it would be better to handle this issue before ever getting to this point, especially in the cases where the Users' Guide asserts LULC values must appear in the table.

Let me know what you think. I ultimately felt like I was obfuscating the problem by trying to wrap `reclassify_raster` calls to add a `try/except` around them.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -400,6 +400,25 @@ def _execute(args):
...
+    # fail early if soil group raster has values other than 1-4
+    missing_soil_groups = utils.raster_values_missing_from_dict_map(
+            (args['soil_group_path'], 1), {1: '', 2: '', 3: '', 4: ''})
+    if missing_soil_groups:
+        raise ValueError("Values in the Soil Group raster were found that are"
+                         " not {1, 2, 3, 4} which correspond to soil groups"
+                         " A, B, C, and D, respectively. The other values"
+                         " found in the soil group raster were:"
+                         f" {missing_soil_groups}.")
+
```
phargogh, line 421: My biggest comment about checking these two up-front is that if we do, we should be careful that these are nodes in the model's task graph:

1. Having them be tasks will allow us to skip the checking if it's already been checked.  I'm especially worried about this one .... if a user provides a gigabytes-large global LULC and a corresponding biophysical table, we should only check it the first time we encounter the raster, not each time `execute` is called (which is what's happening in this example)
2. Having these operations be tasks will allow the checking to be parallelized when `n_workers>0`.  For very large rasters, this could be _very_ useful.

src/natcap/invest/utils.py
```diff
@@ -746,3 +746,32 @@ def _assert_vectors_equal(
...
+    Returns:
+        A set of raster values not found as keys in ``dict_map``
+    """
+
+    raster_path = raster_path_band[0]
+    # declare a set to store unique values from raster
+    raster_unique_values = set()
+
+    for _, raster_block in pygeoprocessing.iterblocks(raster_path_band):
+        raster_unique_values.update(numpy.unique(raster_block))
```
phargogh, line 768: For very large rasters, the disk I/O could end up being substantial and we wouldn't want there to be a perception of a hanging process.  What about adding some timed logging to this, like once every 5s or so, like what we have in most of our raster processing loops throughout InVEST?

dcdenu4:
Hey James, thanks for your feedback per usual. I definitely thought it would be too costly to have this type of checking in any kind of validation, but I didn't think through a large raster making it seem like the model was hanging. Putting in a task and logging would be the right thing to do for this approach. However, I also had the feeling that programming by contract was desirable and the exception should ultimately be handled on failure. I went back and forth with these two options and decided on the checking up front as opposed to breaking out all the `reclassify_raster` calls into helper functions for `task`ing. However, now it is apparent the operation should be `task`ed anyway, I'm convinced the better option is to break the `reclassify_raster` calls into helper functions.

Thanks for the help as always!

> Hey Doug, this is a neat idea!
> 
> The use case that I'm most worried about is when we have very, very large incoming rasters. Is it worth doubling the number of read operations for a reclassification in order to fail sooner in the model's execution? On very very large raster operations (as we're starting to see more of), this overhead could be significant.
> 
> Part of the benefit of a taskgraph approach is that the up-front validation is suddenly less necessary (especially for harder-to-compute things, like this reclassification issue), since the model will just pick up where it last failed at the beginning of a reclassification. This really does lend itself to a programming-by-contract approach to software design, which would be shorter and easier for us to maintain, too.
> 
> Anyways, I'd love to hear what you think about this, and happy to talk about the tradeoffs!

phargogh:
Sounds good to me!  And very happy to talk more if anything else comes up!  These are interesting topics, for sure.

Yes, I think you're right on the money about it being too costly about this being in validation! _(Side note: we _might_ be able to check a RAT vs a biophysical table when a RAT already exists ... that could be fast enough for validation)_

On the topic of doing lots of stuff in `execute`, I got burned in Scenic Quality pretty recently (https://github.com/natcap/invest/pull/174), where I had assumed that the number of viewpoints being passed to the model would be pretty small ... and it turned out that when a user provided like 2000 viewpoints, it ran pretty slow and caused some other issues, too.  Anyways, just trying to save some pain down the road!

dcdenu4:
Putting on hold until a `pgp.convolve_2d` bug has been worked through.

dcdenu4:
Closing this draft PR since the above mentioned proposal has been slightly altered. Will open a new PR that references this one but more clearly states the current methods.

Issue #228, opened 2020-07-15
Undefined nodata values cause nodata checking to fail in numpy.isclose


phargogh:
We have encountered a number of issues on the forums ([here's a recent example](https://community.naturalcapitalproject.org/t/seasonal-water-yield-model-keeps-giving-ufunc-isfinite-error/1150/5)) where the model is crashing during the comparison of a floating-point array with an undefined nodata value.  In the model, it looks something like this:

```python
valid_pixels = numpy.isclose(array, nodata_value)
```

But if `nodata_value=None`:
```python
>>> import numpy
>>> array = numpy.array([1]))
>>> valid_pixels = numpy.isclose(array, None)
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Note that if we have an integer array, comparing with `None` is no problem at all:

```python
>>> import numpy
>>> numpy.ones((1,1), dtype=numpy.uint8) == None
array([[ False]])
```

As documented in #226, having an unset nodata value is a valid configuration of a raster and should be supported throughout InVEST.

- [x] Locate all cases where we're using `numpy.isclose` with a possibly-undefined nodata value
- [x] Determine and implement a fix
- [ ] Ensure that we're covering this case in our tests

phargogh:
Just saw another instance of this in https://community.naturalcapitalproject.org/t/habitat-quality-typeerror-ufunc-isfinite/1168/2

richpsharp:
Hi @emlys! I'm going through old email and just saw this issue and draft pull request. I know this isn't the biggest deal, and I'm a little worried I don't entirely understand the issue, but if I do can I suggest a change to the proposed function name? Is this is a fix to pass through to `numpy.is_close` but guard against a `None` or some other invalid value? If so, could I propose that this function be called `utils.is_close` since that is the behavior it's mimicking? That way code will read the same and we won't have to guess what "valid" means or what the arguments are, etc? This would also save having to logically reverse any `~` operator in the code now or in a future maintenance issue where for some reason there are a lot of `numpy.is_close`s sitting around (`numpy.is_close` is easier to change to `utils.is_close` rather than `~utils.is_valid`?) 

I hope that's helpful, and in case I'm not understanding this issue at all, please feel free to disregard.

emlys:
[...]

PR #269, opened 2020-08-17
Handle comparison to undefined nodata values

emlys:
#228 
* Add a new function `utils.is_valid(array, nodata)` that handles the condition where the nodata value is undefined. This replaces dozens of instances of `~numpy.isclose(array, nodata)`.
* Add unit tests for the new function in test_utils.py.
* Add more complex test to test_seasonal_water_yield_regression.py, testing actual usage of the function


emlys:
see convo here https://github.com/natcap/invest/issues/228#issuecomment-676564189

src/natcap/invest/utils.py
```diff
@@ -746,3 +746,27 @@ def _assert_vectors_equal(
...
+    Returns:
+        boolean numpy.ndarray with the same shape as the input array. 
+        Each element is False if its value is close to the nodata value,
+        True otherwise.
+
+    """
+    if nodata:
+        return ~numpy.isclose(array, nodata)
+    else:
+        return array != nodata
```
phargogh, line 772: There are a couple of things that will probably turn out to be problematic here:

1. Python's truthiness rules allow for values of `0`, `False`, empty iterables and a few other things to all be interpreted as `False`.  How will this function behave if we have a nodata value of `0`, which is a valid input?
2. similar to @richpsharp 's comment at https://github.com/natcap/invest/issues/228#issuecomment-675160937, the function name `is_valid` makes it sound like this function is for checking whether a pixel is valid ... but all it's actually checking for is whether the value is close to `nodata`.  Function names seem like a fairly trivial thing, but it's so so important for us to be clear about the intent of a function and what it's really trying to do.  So if the intent is to determine whether a value is close to `nodata`, then we should really be expressing that in the title.  What, exactly, is a valid pixel, is highly dependent on the use case.
3. If the intent here is actually to create a single function that checks to see if a pixel value is close to `nodata`, then we need to accommodate the differences in floating-point and integer matrices.  It turns out that if we know in advance that we have an integer matrix, then we can take advantage of order-of-magnitude speedups in per-element comparisons: `integer_array != nodata` is an order of magnitude faster than `numpy.isclose(integer_array, nodata)`.   On very large datasets, this can make a noticeable difference.  Here's the timing details:

```python
>>> timeit.timeit('numpy.isclose(ones, 0)', setup='import numpy; ones=numpy.ones((50, 50), dtype=numpy.float32)', number=100000)
5.014724600000022
>>> timeit.timeit('numpy.isclose(ones, 0)', setup='import numpy; ones=numpy.ones((50, 50), dtype=numpy.int16)', number=100000)
4.701722299999915
>>> timeit.timeit('ones != 0', setup='import numpy; ones=numpy.ones((50, 50), dtype=numpy.int16)', number=100000)
0.315173499999986
```
4. By making it necessary to have an extra function callthrough, we're not only obfuscating the intent of the function, there's also a (relatively minor) computational overhead.  The main point here is that the value gained from having a function should outweigh the cognitive and computational overhead of needing to look elsewhere at what the function is doing.  This function is so simple that I'm not sure this threshold is met.

emlys:
> @emlys You're very right that the common pattern you mention (initialize the output array, create a mask of valid pixels, limit the calculations to valid pixels) is indeed to limit computation to only those places where it is needed. In general, the internals of InVEST should do what's best and most efficient for the user in the general case, even if it's a little trickier for us to program and to maintain (without gutting maintainability, though!), and also even if it underperforms in certain specific edge cases. All that within the context of the problem, of course!
> 
> In the simplest of cases, where a `valid_mask` is either present or not present when operating on a single raster (like in the [globio example you linked to above](https://github.com/natcap/invest/pull/269/files#diff-9c24b665c47f29dc86d5c6394d33d1dbR578)), might I suggest the following slight tweak to the conditional you had proposed?
> 
> ```python
> if globio_nodata is None:
>     valid_mask = slice(None)  # the ``slice`` equivalent of ``:``
> else:
>     valid_mask = ~numpy.isclose(msa_f, globio_nodata)
> result[valid_mask] = msa_f[valid_mask] * msa_lu[valid_mask] * msa_i[valid_mask]
> ```
> 
> Or, more abbreviated:
> 
> ```python
> valid_mask = slice(None)  # the ``slice`` equivalent of ``:``
> if globio_nodata is not None:
>     valid_mask = ~numpy.isclose(msa_f, globio_nodata)
> result[valid_mask] = msa_f[valid_mask] * msa_lu[valid_mask] * msa_i[valid_mask]
> ```
> 
> With this slight tweak, the `slice(None)` object creation happens in constant time regardless of the size of the input data, the indexing is an order of magnitude less than boolean array indexing, and we can easily (and cheaply, thanks to python's memory management) overwrite the `slice` if the nodata value _is_ defined. Interestingly, this doesn't completely avoid indexing, and so in some edge cases this _will_ underperform ... but I think this might be the fastest general-case, DRY solution given what we know about numpy's array indexing.
> 
> I found the timings for this to be rather surprising, but `slice(None)` appears to be the equivalent of using `:` in array indexing notation (and the [python docs support this](https://docs.python.org/dev/glossary.html#term-slice)). Timings are included below if you'd like to try it out!
> Timings
> 
> Anyways, what do you think about this RE: tradeoffs between efficiency, readability, maintainability and general-case solutions?

I like the `slice(None)`! Don't think I've ever seen `slice()` used before. I just replaced all my uses of `numpy.full(True)` with it.
Agree that with that notation the decrease in readability is minimal, and worth it to avoid slowing down the computation.

emlys:
> I did have one other question, though, which was that it looks like this PR handles the use of `numpy.isclose` for comparing nodata values. Were you also able to take a look at the places where nodata checking is done by `!=`?

I did not, but it seems that where `!=` is used, it shouldn't matter if the nodata value is defined or not.
```
>>> a = numpy.array([1,2,3])
>>> a != None
array([ True,  True,  True])
```
Is this what you're referring to?

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -1282,7 +1282,8 @@ def extract_bathymetry_along_ray(
                 'bathymetry input fully cover the fetch ray area?'
                 % (value, {'xoff': ix, 'yoff': iy,
                            'win_xsize': win_size, 'win_ysize': win_size}))
-        if ~numpy.isclose(value, bathy_nodata):
+
+        if bathy_nodata is None or ~numpy.isclose(value, bathy_nodata):
```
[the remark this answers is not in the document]
emlys: cool! Is there some resource you have for finding faster alternatives like this?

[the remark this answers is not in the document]
emlys: update: `value` is actually an array, not a single value, so `math.isclose` doesn't work :( I changed the variable name to `values` to be less confusing.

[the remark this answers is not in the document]
phargogh: Ha, sorry to disappoint!  I unfortunately don't know of such a resource.  In this case, I just saw that we were using numpy but then only checking an array with a single value in it.  Numpy's great for operating on arrays that might be very large, but for a single value, it's often faster to just directly compare the values.  Plus, numpy does a [whole lot of extra work](https://github.com/numpy/numpy/blob/e2428594b77aef952610f930bacf4085ef1fcdf9/numpy/core/numeric.py#L2201) to ensure that it can actually compare the values, compared with the python implementation of `isclose`, which is [about as short as you'd expect](https://github.com/python/cpython/blob/82e79480d6e61940d7007d9026fbff0b1a11ad9a/Modules/mathmodule.c#L2932) for this.

Also I was a bit wrong about the timings I had included above.  Since this is a 1-element numpy array, here are the corrected timings:

```python
>>> timeit.timeit('~numpy.isclose(array, 1)', setup='import numpy; array=numpy.array([[1.1]], dtype=numpy.float32)')
42.59300510000003
>>> timeit.timeit('not math.isclose(array[0][0], 1)', setup='import numpy; import math; array=numpy.array([[1.1]], dtype=numpy.float32)')
0.5673721999999941
```

Not _quite_ as fast thanks to the need to do that indexing, but still substantially faster.

src/natcap/invest/globio.py
```diff
@@ -456,10 +456,12 @@ def _primary_veg_mask_op(lulc_array, globio_nodata, primary_veg_mask_nodata):
     """Masking out natural areas."""
     # lulc_array and nodata could conceivably be a float here,
     # if it's the user-provided globio dataset
-    valid_mask = ~numpy.isclose(lulc_array, globio_nodata)
     # landcover type 1 in the GLOBIO schema represents primary vegetation
     result = numpy.empty_like(lulc_array, dtype=numpy.int16)
-    result[:] = primary_veg_mask_nodata
+    result = lulc_array == 1
```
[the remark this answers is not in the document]
emlys: oops, you're totally right!

phargogh:
Ah, good point about the `!=` comparison!  So yes, we should be functionally ok on that front.  I suppose it could be a bit faster to avoid the comparison entirely, but that's outside the scope of this issue and it won't break anything as-is.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -1273,17 +1273,18 @@ def extract_bathymetry_along_ray(
...
                 'got a %s when trying to read bathymetry at %s. Does the '
                 'bathymetry input fully cover the fetch ray area?'
-                % (value, {'xoff': ix, 'yoff': iy,
+                % (values, {'xoff': ix, 'yoff': iy,
                            'win_xsize': win_size, 'win_ysize': win_size}))
-        if ~numpy.isclose(value, bathy_nodata):
-            bathy_values.append(value)
+
+        if bathy_nodata is None or ~numpy.isclose(values, bathy_nodata):
+            bathy_values.append(values)
```
phargogh, line 1287: Hey I'm sorry about the somewhat confusing chain of timings and comments over in the other thread.  You're right that we can't use `math.isclose` directly on the array, but since we know ahead of time that this will be a single value, we can just index into that one item, `values[0][0]` or the like: `not math.isclose(values[0][0])`

src/natcap/invest/globio.py
```diff
@@ -943,22 +950,21 @@ def _collapse_infrastructure_op(*infrastructure_array_list):
...
-            infrastructure_mask = (
-                infrastructure_mask |
-                ((infrastructure_array_list[index] > 0) & ~current_nodata))
-
-            nodata_mask = (nodata_mask & current_nodata)
+        
+        nodata_mask = numpy.full(infrastructure_array_list[0].shape, True)
+        infrastructure_mask = numpy.full(infrastructure_array_list[0].shape, False)
+
+        for index in range(0, len(infrastructure_array_list)):
```
phargogh, line 957: This is a pretty trivial change, but since we're starting from index 0 (the default), this can be abbreviated to `range(len(infrastructure_array_list))` if you like to achieve the same effect.

src/natcap/invest/coastal_vulnerability.py
emlys: Oh okay, that makes sense. The nodata value was also (incorrectly) a list so I fixed that in `calculate_wind_exposure`

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -621,6 +621,14 @@ cpdef calculate_local_recharge(
...
                         kc_nodata = kc_m_nodata_list[m_index]
 
+                        # make sure that user input nodata values are defined
+                        # precipitation and evapotranspiration data should 
+                        # always be non-negative
+                        if precip_nodata is None:
+                            precip_nodata = -1
+                        if et0_nodata is None:
+                            et0_nodata = -1
+
```
phargogh, line 631: I was thinking about this over the weekend.  Checking for and assigning values for  `precip_nodata` and `et0_nodata` here will make these extra lines run many, many times (once for each month, but also for each pixel in the `work_queue`, which ends up being for each valid pixel).  Of course it's quite possible that the compiler will optimize this condition away so that it doesn't need to be run so many times, but since we know in advance that this can just be defined elsewhere, would you be ok with moving it outside of the main processing loop?  Offhand, it looks like these could be checked for once, up near the top of this function (around line 448 for `et0`, 456 for `precip`).

emlys: That's a good point! Just moved it outside of the loop

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -441,19 +441,23 @@ cpdef calculate_local_recharge(
...
+    # set to -1 if not defined
+    # precipitation and evapotranspiration data should 
+    # always be non-negative
     et0_m_raster_list = []
     et0_m_nodata_list = []
     for et0_path in et0_path_list:
         et0_m_raster_list.append(_ManagedRaster(et0_path, 1, 0))
         et0_m_nodata_list.append(
-            pygeoprocessing.get_raster_info(et0_path)['nodata'][0])
+            pygeoprocessing.get_raster_info(et0_path)['nodata'][0] or -1)
```
phargogh, line 453: Neat!  this is a bit of python syntax that I was not expecting to work, but it absolutely does.

phargogh (approved):
Thanks for all your work on this @emlys !

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
emlys: Oh heck! This is going to have the same problem with truthiness of 0... so `0 or -1` will be `-1`. I can fix this in my next PR...

phargogh: Whoops, so it does!  I'm sorry I didn't catch that.  Yes, let's do this in the next PR.

Issue #172, opened 2020-06-15
Fisheries: Batch processing CSV directory input fails validation even when UI input disabled


phargogh:
To reproduce:

1. Open the InVEST Fisheries UI (this was observed by Jess on InvEST 3.8.0)
2. Load up a sample datastack
3. Check the 'Batch Processing' checkbox to enable the CSV directory input
4. Manipulate the CSV directory input so that validation fails
5. Uncheck the 'Batch Processing' checkbox in order to disable the CSV directory input
6. Observe that even though the input is disabled, the 'Run' Button displays a red X and prevents the model from executing.

PR #273, opened 2020-08-21
Fix checkbox validation bug #172

emlys:
The issue here was that the user input values were being validated before determining if conditionally-required values are actually required. 
There needs to be a distinction between two types of non-required values:
* _optional_ values, such as the output suffix, which have `required: False` and should be validated
* _conditionally excluded_ values, such as fields that are greyed out when a checkbox is unchecked. These have  `required: 'condition that evaluates to False'` and should not be validated.

I changed the order of the validation steps so that conditionally excluded keys are identified first, and then validation is done on `set(sufficient keys) - set(conditionally excluded keys)`.

It seems that generally, fields that are conditionally-required are checkboxes, and for that case this change makes sense. Relying on the assumption that a conditionally-required field will never be optional (meaning that if the condition is `False`, the field value won't be used for anything, and so shouldn't be validated). Are there any situations where this assumption doesn't hold? Such as a field which is optional, unless a certain checkbox is checked, in which case it's required?

### DIFF
--- a/src/natcap/invest/ui/model.py
+++ b/src/natcap/invest/ui/model.py
@@ -9,7 +9,7 @@ import collections
 import json
 import requests
 import textwrap
-import cgi
+import html
 import tarfile
 import contextlib
 import functools
@@ -1069,7 +1069,7 @@ class WholeModelValidationErrorDialog(QtWidgets.QDialog):
         self.warnings = validation_warnings
 
         if validation_warnings:
-            # cgi.escape handles escaping of characters <, >, &, " for HTML.
+            # html.escape handles escaping of characters <, >, &, " for HTML.
             self.title_label.setText(
                 '<h2>Validation warnings found</h2>'
                 '<h4>To ensure the model works as expected, please fix these '
@@ -1077,7 +1077,7 @@ class WholeModelValidationErrorDialog(QtWidgets.QDialog):
             self.label.setText(
                 '<ul>%s</ul>' % ''.join(
                     ['<li><b>%s</b>: %s</li>' % (
-                        ', '.join(labels), cgi.escape(warning_, quote=True))
+                        ', '.join(labels), html.escape(warning_))
                      for labels, warning_ in validation_warnings]))
             self.label.repaint()
             self.label.setVisible(True)
--- a/src/natcap/invest/ui/model.py
+++ b/src/natcap/invest/ui/model.py
@@ -1143,8 +1143,7 @@ class InVESTModel(QtWidgets.QMainWindow):
     If any of these attributes are not overridden, a warning will be raised.
     """
 
-    def __init__(self, label, target, validator, localdoc,
-                 suffix_args_key='results_suffix'):
+    def __init__(self, label, target, validator, localdoc):
         """Initialize the Model.
 
         Parameters:
@@ -1155,8 +1154,6 @@ class InVESTModel(QtWidgets.QMainWindow):
                 function.
             localdoc (string): The filename of the user's guide chapter for
                 this model.
-            suffix_args_key='results_suffix' (string): The args key to use for
-                suffix input.  Defaults to ``results_suffix``.
         """
         QtWidgets.QMainWindow.__init__(self)
         self.label = label
@@ -1268,7 +1265,7 @@ class InVESTModel(QtWidgets.QMainWindow):
                 model=modelname)))
 
         self.suffix = inputs.Text(
-            args_key=suffix_args_key,
+            args_key='results_suffix',
             helptext=(
                 'A string that will be added to the end of the output file '
                 'paths.'),
--- a/src/natcap/invest/ui/wind_energy.py
+++ b/src/natcap/invest/ui/wind_energy.py
@@ -11,8 +11,7 @@ class WindEnergy(model.InVESTModel):
             label='Wind Energy',
             target=wind_energy.execute,
             validator=wind_energy.validate,
-            localdoc='wind_energy.html',
-            suffix_args_key='suffix'
+            localdoc='wind_energy.html'
         )
 
         self.wind_data = inputs.File(
--- a/tests/test_recreation.py
+++ b/tests/test_recreation.py
@@ -1182,7 +1182,7 @@ class RecreationValidationTests(unittest.TestCase):
             'aoi_path': os.path.join(SAMPLE_DATA, 'andros_aoi.shp')})
 
         self.assertEqual(validation_warnings, [(['predictor_table_path'], 
-            "Fields are missing from this table: ['ID', 'PATH', 'TYPE']")])
+                                                 expected_message)])
 
         validation_warnings = recmodel_client.validate({
             'compute_regression': True,
@@ -1194,9 +1194,9 @@ class RecreationValidationTests(unittest.TestCase):
             'aoi_path': os.path.join(SAMPLE_DATA, 'andros_aoi.shp')})
         
         self.assertEqual(validation_warnings, [(['predictor_table_path'], 
-            "Fields are missing from this table: ['ID', 'PATH', 'TYPE']"), 
-            (['scenario_predictor_table_path'], 
-              "Fields are missing from this table: ['ID', 'PATH', 'TYPE']")])
+                                                 expected_message), 
+                                               (['scenario_predictor_table_path'], 
+                                                 expected_message)])
 
 
 def _assert_vector_attributes_eq(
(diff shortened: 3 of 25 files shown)


emlys:
For the tests to pass, I need to update the test data (one args JSON file mistakenly has the key `"suffix"` instead of `"results_suffix"`). Because in this PR I changed it to iterate over a set of the args keys, rather than the spec keys, this now causes an error when there's a key in the provided args that doesn't exist in the spec. Alternatively we could just ignore keys that don't exist in the spec, but I'd think that shouldn't normally happen.

dcdenu4:
Hey @emlys , I haven't really looked into your changes here but was just reading through. I saw you asked:
>Are there any situations where this assumption doesn't hold? Such as a field which is optional, unless a certain checkbox is checked, in which case it's required?

I believe a few models do this, Wind Energy for sure. In Wind Energy if you select the `Valuation` checkbox then there are some required fields in there that were _hidden_. If this isn't the case you're thinking of then please ignore!

phargogh:
>>Are there any situations where this assumption doesn't hold? Such as a field which is optional, unless a certain checkbox is checked, in which case it's required?

> I believe a few models do this, Wind Energy for sure. In Wind Energy if you select the Valuation checkbox then there are some required fields in there that were hidden. If this isn't the case you're thinking of then please ignore!

@dcdenu4 you're right that Wind Energy's valuation checkbox allows some additional inputs to be revealed, and also that this is a common pattern across InVEST.  If I'm understanding your meaning, I think these inputs (which are _conditionally required_ ... they're required when the 'valuation' checkbox is checked) fit into the model that @emlys is describing.  Is that the case?

@emlys I think the conditional requirement pattern we most frequently use (having a checkbox that enables or disables an input) actually represents the sort of conditional-enabling that @dcdenu4 is describing.  If this checkbox is checked, then this input (or set of inputs) is required.  I can't think of a concrete example of where we deviate from that pattern, so I think what we have here could make sense.  But I could be wrong about this!  I tried clicking around a few of the UIs using a build from this PR and I didn't see any obvious issues ... @dcdenu4 @emlys have you tried spot-checking a few of the UIs to see that they work as expected?

emlys:
> 
> @dcdenu4 @emlys what do you think?

I did look through some models and it seems that in general, conditionally-required fields are disabled and greyed out in the UI when the condition evaluates to False. In the UI model files I've looked at, enabled status is connected to the sufficiency for conditionally-required fields (https://github.com/natcap/invest/blob/master/src/natcap/invest/ui/carbon.py#L116-L128).

dcdenu4:
> ... @dcdenu4 @emlys have you tried spot-checking a few of the UIs to see that they work as expected?

I just tested out Wind Energy and the behavior is different. To replicate the issue:

1) Open Wind Energy model under `bug/172` branch
2) Select `Valuation` checkbox
3) Select `Use Price Table` checkbox

Inputs under both of the checkboxes do not highlight red showing that a valid input needs to be entered.

image

emlys:
> I just tested out Wind Energy and the behavior is different. To replicate the issue:

Oh interesting! Thanks for catching this. It seems this was happening because of a bug in the wind energy UI. There was a `KeyError` happening in `validate`, which was causing the function to exit before returning the list of validation warnings, hence no red X's. If you clicked Run, it would fail and show the error. 

The `KeyError` was happening because the wind energy UI gave the results suffix arg the name `suffix` instead of `results_suffix`, and `suffix` isn't in the spec. There's a kwarg to `model.__init__` that allows overriding the default name `results_suffix`, but it was only used in this one place, so I removed it. I'm guessing that was left over from a time when this wasn't standardized, but it seems better to make it consistent now. Thoughts?

src/natcap/invest/ui/model.py
```diff
@@ -1268,7 +1265,7 @@ def _settings_saved_message():
                 model=modelname)))
 
         self.suffix = inputs.Text(
-            args_key=suffix_args_key,
+            args_key='results_suffix',
```
dcdenu4, line 1268: Right, since `results_suffix` is standardized across InVEST there shouldn't be the option to name it something else. (just commenting to confirm my own understanding)

src/natcap/invest/ui/wind_energy.py
```diff
@@ -11,8 +11,7 @@ def __init__(self):
             label='Wind Energy',
             target=wind_energy.execute,
             validator=wind_energy.validate,
-            localdoc='wind_energy.html',
-            suffix_args_key='suffix'
```
dcdenu4, line 15: Ah, I see. Yea, this looks fine to remove to stick with the standardized format of InVEST Spec. I wonder if any other models have this issue.

dcdenu4:
Hey @emlys thanks for the dive into the Wind Energy failing. I think everything you mentioned in your comment makes sense. It looks like it was just left behind when standardizing to `results_suffix`. 

I don't think I was the main reviewer on this and didn't go through all the details, but I did notice that you handled some `store_results` for a future TaskGraph release. I don't think we've bumped InVEST requirements for TaskGraph yet and maybe that should be done for the 3.9 release. So it might be best to leave those changes to a separate PR? Let me know what you think.

dcdenu4:
> I don't think I was the main reviewer on this and didn't go through all the details, but I did notice that you handled some `store_results` for a future TaskGraph release. I don't think we've bumped InVEST requirements for TaskGraph yet and maybe that should be done for the 3.9 release. So it might be best to leave those changes to a separate PR? Let me know what you think.

Just realized in another PR that `requirements.txt` is `gt` 0.9.1 for TaskGraph. Over in that PR I had suggested maybe we should pin `requirements.txt` to 0.9.1 instead of making these TaskGraph changes, as committing those changes to `master` will put us in a situation where building from `master` will be broken because not all the models have been updated for 0.10.0. Lets chat about this!

src/natcap/invest/ui/wind_energy.py
emlys: I briefly looked through all the other model UI files, and none of them were using `suffix_args_key` in the model `__init__()`, so I think it's okay to remove the option

src/natcap/invest/ui/model.py
phargogh: Yep, there shouldn't be that option at all, and this is something that should have been addressed way back when we were building this version of the UI.  It was only fairly recently that we actually did formally establish a standard suffix.

phargogh:
It looks like this might still be waiting on @dcdenu4 , so I'll hold off on merging for now.  What do you think @dcdenu4 ?

tests/test_recreation.py
```diff
@@ -1173,19 +1173,30 @@ def test_bad_predictor_table_header(self):
...
+        validation_warnings = recmodel_client.validate({
+            'compute_regression': True,
+            'predictor_table_path': table_path,
+            'start_year': '2012',
+            'end_year': '2016',
+            'workspace_dir': self.workspace_dir,
+            'aoi_path': os.path.join(SAMPLE_DATA, 'andros_aoi.shp')})
+
+        self.assertEqual(validation_warnings, [(['predictor_table_path'], 
+            "Fields are missing from this table: ['ID', 'PATH', 'TYPE']")])
```
dcdenu4, line 1185: It looks like there is a line above with `expected_message`. Should that go here instead of re-writing the same text?

tests/test_recreation.py
```diff
@@ -1173,19 +1173,30 @@ def test_bad_predictor_table_header(self):
...
+            'scenario_predictor_table_path': table_path,
+            'start_year': '2012',
+            'end_year': '2016',
+            'workspace_dir': self.workspace_dir,
+            'aoi_path': os.path.join(SAMPLE_DATA, 'andros_aoi.shp')})
+        
+        self.assertEqual(validation_warnings, [(['predictor_table_path'], 
+            "Fields are missing from this table: ['ID', 'PATH', 'TYPE']"), 
+            (['scenario_predictor_table_path'], 
+              "Fields are missing from this table: ['ID', 'PATH', 'TYPE']")])
```
dcdenu4, line 1199: Should `expected_message` replace this text as well?

tests/test_ucm.py
```diff
@@ -282,8 +282,6 @@ def test_bad_args(self):
             'workspace_dir': self.workspace_dir,
             'results_suffix': 'test_suffix',
             't_ref': 35.0,
-            't_obs_raster_path': os.path.join(
-                REGRESSION_DATA, "Tair_Sept.tif"),
```
dcdenu4, line 286: Why is this key pair being removed from the test?

emlys: `t_obs_raster_path` doesn't exist in the spec for UCM, and with the changes I made to validation, it now causes a key error when there are extra keys. I'm guessing this was just outdated

tests/test_recreation.py
emlys: good point, thanks for catching this!

tests/test_ucm.py
dcdenu4: oh interesting! Thanks.

Issue #263, opened 2020-08-13
Rec model should catch invalid predictor table "type"


davemfish:
The model expects to find very specific strings in the `type` column of the predictor table, but if those strings don't match exactly, that predictor will be silently ignored from the regression. See an example reported here: https://community.naturalcapitalproject.org/t/recreation-coefficients/1222/7

It would be nice to strip any extra whitespace from the values before using them in comparisons, and it might be even nicer to validate those `type` values early on in the runtime or during validation.

Issue #275, opened 2020-08-24
TaskGraph add_task has new optional param


dcdenu4:
TaskGraph will soon have an additional argument in `add_task` of `store_result=False`.
>Adds a store_result flag to add_task to conditionally store a result rather than do it by default. In turn, get will now raise an exception if invoked on a Task created where the store_result flag is False. This PR adds that functionality, tests to cover it, and an update to history.

https://github.com/natcap/taskgraph/pull/34

We should make sure that InVEST is up to date with this change when upgrading to a new TaskGraph release.

dcdenu4:
From `grep` it looks like the following models will need to add this new argument:

- [ ] habitat_quality
- [ ] scenic_quality
- [ ] urban_flood_risk_mitigation

PR #276, opened 2020-08-25
Validate recreation predictor types

emlys:
Addresses #263 

As suggested by @davemfish, validating the type values early in `execute()`. Leading/trailing whitespace is ignored. Following the pattern of more complex validation being done in a separate function outside of `validate()` e.g. `_validate_same_id_lengths` and `_validate_same_projection`.

### DIFF
--- a/src/natcap/invest/recreation/recmodel_client.py
+++ b/src/natcap/invest/recreation/recmodel_client.py
@@ -252,6 +252,8 @@ def execute(args):
                       with the response polygon in projected units of AOI
                     * 'polygon_area': area of the polygon contained within
                       response polygon in projected units of AOI
+                    * 'polygon_percent_coverage': percent (0-100) of area of 
+                      overlap between the predictor and each AOI grid cell.
 
         args['scenario_predictor_table_path'] (string): (optional) if
             present runs the scenario mode of the recreation model with the
@@ -1520,16 +1522,16 @@ def _validate_predictor_types(table_path):
         type, ignoring leading/trailing whitespace.
     """
     df = utils.read_csv_to_dataframe(table_path, to_lower=True)
-    type_list = df['type'].tolist()
-    valid_types = ['raster_mean', 'raster_sum', 'point_count', 
+    # ignore leading/trailing whitespace because it will be removed
+    # when the type values are used
+    type_list = set([type.strip() for type in df['type']])
+    valid_types = set({'raster_mean', 'raster_sum', 'point_count', 
                    'point_nearest_distance', 'line_intersect_length',
-                   'polygon_area_coverage', 'polygon_percent_coverage']
-    for type_value in type_list:
-        # ignore leading/trailing whitespace because it will be removed
-        # when the type values are used
-        if type_value.strip() not in valid_types:
-            raise ValueError(f'The table contains an invalid type value: \
-                {type_value}. The allowed types are: {", ".join(valid_types)}')
+                   'polygon_area_coverage', 'polygon_percent_coverage'})
+    difference = type_list.difference(valid_types)
+    if difference:
+        raise ValueError('The table contains invalid type value(s): '
+            f'{difference}. The allowed types are: {valid_types}')
 
 
 def delay_op(last_time, time_delay, func):
--- a/src/natcap/invest/recreation/recmodel_client.py
+++ b/src/natcap/invest/recreation/recmodel_client.py
@@ -1502,6 +1502,31 @@ def _validate_same_projection(base_vector_path, table_path):
             "projection of the base vector")
 
 
+def _validate_predictor_types(table_path):
+    """Validate the type values in a predictor table.
+
+    Args:
+        table_path (string): path to a csv table that has at least
+            the field 'type' 
+
+    Returns:
+        None
+
+    Raises:
+        ValueError if any value in the ``type`` column does not match a valid
+        type, ignoring leading/trailing whitespace.
+    """
+    df = utils.read_csv_to_dataframe(table_path, to_lower=True)
+    type_list = df['type'].tolist()
+    valid_types = ['raster_mean', 'raster_sum', 'point_count', 
+                      'point_nearest_distance', 'line_intersect_length',
+                      'polygon_area_coverage', 'polygon_percent_coverage']
+    for type_value in type_list:
+        if type_value.strip() not in valid_types:
+            raise ValueError(f'The table contains an invalid type value: \
+                {type_value}. The allowed types are: {", ".join(valid_types)}')
+
+
 def delay_op(last_time, time_delay, func):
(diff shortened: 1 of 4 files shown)


src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -1502,6 +1505,33 @@ def error_handler(err_level, err_no, err_msg):
...
+    Returns:
+        None
+
+    Raises:
+        ValueError if any value in the ``type`` column does not match a valid
+        type, ignoring leading/trailing whitespace.
+    """
+    df = utils.read_csv_to_dataframe(table_path, to_lower=True)
+    type_list = df['type'].tolist()
+    valid_types = ['raster_mean', 'raster_sum', 'point_count',
```
dcdenu4, line 1524: I noticed that under the `execute` docstring the `predictor_table_path` doesn't include the `polygon_percent_coverage`, but I do see it in the User's Guide and that it is used in the model. Could you add it to the `execute` docstring?

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -1502,6 +1505,33 @@ def error_handler(err_level, err_no, err_msg):
...
+    df = utils.read_csv_to_dataframe(table_path, to_lower=True)
+    type_list = df['type'].tolist()
+    valid_types = ['raster_mean', 'raster_sum', 'point_count', 
+                   'point_nearest_distance', 'line_intersect_length',
+                   'polygon_area_coverage', 'polygon_percent_coverage']
+    for type_value in type_list:
+        # ignore leading/trailing whitespace because it will be removed
+        # when the type values are used
+        if type_value.strip() not in valid_types:
+            raise ValueError(f'The table contains an invalid type value: \
```
dcdenu4, line 1531: The backlash `\` for strings over multiple lines works here, but inside parens you can also extend strings like:
```python
raise ValueError(
    'The table contains an invalid type value:'
    f' {type_value}. The allowed types are: {", ".join(valid_types)}')
```
I end up doing this a lot and feel that it is a little more clean than the backlash...

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -1502,6 +1505,33 @@ def error_handler(err_level, err_no, err_msg):
...
+
+    Returns:
+        None
+
+    Raises:
+        ValueError if any value in the ``type`` column does not match a valid
+        type, ignoring leading/trailing whitespace.
+    """
+    df = utils.read_csv_to_dataframe(table_path, to_lower=True)
+    type_list = df['type'].tolist()
```
dcdenu4, line 1523: I think what you have here is just fine, but I wonder if using `sets` would be a bit less code and cleaner?
```python
type_set = set(df['type'].str.strip().tolist())
valid_types = set([ ... ])
invalid_types = type_set.difference(valid_types)
if invalid_types:
    raise ValueError(...)
```

Sorry if this feels like over designing, I've just been loving `sets` for list comparisons recently.

tests/test_recreation.py
```diff
@@ -1187,6 +1187,51 @@ def test_bad_predictor_table_header(self):
...
+            'end_year': '2014',
+            'grid_aoi': False,
+            'predictor_table_path': table_path,
+            'workspace_dir': self.workspace_dir,
+        }
+
+        # there should be no error when the type has trailing whitespace
+        recmodel_client.execute(args)
+        output_path = os.path.join(self.workspace_dir, 'regression_coefficients.txt')
+        print('regression_coefficients.txt:')
```
dcdenu4, line 1222: Should this be a logging statement? Is it needed here?

tests/test_recreation.py
```diff
@@ -1187,6 +1187,51 @@ def test_bad_predictor_table_header(self):
...
+        # predictor id, meaning it wasn't dropped from the regression
+        with open(output_path, 'r') as output_file:
+            self.assertTrue(predictor_id in ''.join(output_file.readlines()))
+
+        # try it again with the table that has a misspelled type
+        # it should raise an error
+        args['predictor_table_path'] = bad_table_path
+        with self.assertRaises(ValueError):
+            recmodel_client.execute(args)
+
```
dcdenu4, line 1234: Could I suggest breaking this into two tests? One that tests for predictor id being used with whitespace and one that fails from typo? I think it's best to have each test in charge of one thing.

Also for the `assertRaises` test, could test against the error message? I feel like it's prudent to make sure we are not just catching an arbitrary `ValueError` that could have come from somewhere else.
```python
with self.assertRaises(ValueError) as cm:
    recmodel_client.execute(args)
self.assertTrue("Expected error message chunk here" in str(cm.exception))
```

Thanks!

src/natcap/invest/recreation/recmodel_client.py
emlys: Oh, good catch!

tests/test_recreation.py
emlys: oops! Left that in by mistake

src/natcap/invest/recreation/recmodel_client.py
emlys: good idea! Sets are very cool.

phargogh:
Offhand, it looks like these failures might be related to the new parameter documented in #275 , which by default does _not_ store the output of functions in the task database.

emlys:
> Offhand, it looks like these failures might be related to the new parameter documented in #275 , which by default does _not_ store the output of functions in the task database.

Just set `store_result=True` where needed, tests should be passing now

dcdenu4:
Ah, I see. I had thought that `requirements.txt` pinned to a certain version of TaskGraph but I see it's just `gt` 0.9.1. Hmmm, obviously this is an issue with how we handle dependency versions (which we've talked about), but I don't think it's ideal making changes that are not related to the current work of the PR. In this case it's a larger issue, because as soon as we commit to `master` then we are dependent on the latest TaskGraph, which means we have to get all the other models up to speed before being able to release a bugfix. I would almost suggest we remove the `store_result` changes and instead pin `requirements.txt` to 0.9.1... Thoughts?

emlys:
> Ah, I see. I had thought that `requirements.txt` pinned to a certain version of TaskGraph but I see it's just `gt` 0.9.1. Hmmm, obviously this is an issue with how we handle dependency versions (which we've talked about), but I don't think it's ideal making changes that are not related to the current work of the PR. In this case it's a larger issue, because as soon as we commit to `master` then we are dependent on the latest TaskGraph, which means we have to get all the other models up to speed before being able to release a bugfix. I would almost suggest we remove the `store_result` changes and instead pin `requirements.txt` to 0.9.1... Thoughts?

That makes more sense that what I did! I guess I assumed it was using the most current version for some reason, and just worked around it. But it would probably be better to be consistent with the versions of everything

Issue #278, opened 2020-08-27
Installation failing because requirements-all.yml doesn't list conda-forge


jsta:
I'm attempting installation and running the following command:

`make ENV=invest env`

but I'm getting the following error:

```
conda env update -p invest --file requirements-all.yml
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - taskgraph[version='>=0.9.1']
```

I solved the error by manually adding conda-forge as a channel to the `yml` file and re-running:

`conda env update -p invest --file requirements-all.yml`

phargogh:
Thanks for letting us know about this!  The script does indeed check conda-forge for the presence of packages, so it would follow that the environment file it produces should include the conda-forge channel.  I'll get a patch in for this.

dcdenu4:
Hey @phargogh , just confirming that this is good to go now.

jsta:
Works for me.

phargogh:
Thanks for letting us know, @jsta !

PR #279, opened 2020-08-27
Adding channels to the environment YML file. RE:#278

phargogh:
This PR addresses a minor issue with our conda environment file generation script where the script checks for whether packages were present on conda-forge, but then it wasn't adding the conda-forge channel to the output requirements file.  Since a user noted this in #278 , I've added a note about this to `HISTORY.rst` since it's a user-facing change.


emlys:
Looks like this is going to need the taskgraph version pinned as well :(

phargogh:
Yep, absolutely right!  This'll have to wait until the taskgraph version pinning goes through.

PR #281, opened 2020-08-27
Pr resolution

emlys:
Merge conflict resolution for https://github.com/natcap/invest/pull/280


dcdenu4:
@emlys I think the failing tests are due to pygeoprocessing 2.1. Could you pin pygeoprocessing to 2.0 in this pr-resolution. If That fixes the tests like I think it will then we should be good to go here. I have a PR coming soon going straight into release/3.9 that will actually bump the pygeoprocessing version to the latest revision until 2.1.1 is released.

Issue #261, opened 2020-08-11
Validation should handle files stored in on-demand file streaming services


phargogh:
When a file is saved to an on-demand cloud file streaming service such as Google Drive (and served with Google File Stream), the file must be downloaded before the file can be validated.  This delay leads to an unresponsive Qt UI (because validation completes before the UI window is shown), degraded model runtimes (because network resources are used before a file is loaded), and slower validation.

Looking forward to the new UI, validation should handle the case where files are stored in a Google File Stream-like system.  When this happens, it would be nice if validation were to fail more quickly and report back a helpful message of some sort.  

Maybe we can set a timeout on validation function calls?

PR #289, opened 2020-08-31
Timeout functions that read files in validation

emlys:
Addresses #261 

Using `threading` to set a timeout on functions that open a file. These could potentially be slowed down by Google File Stream-like services if the file has to be downloaded.

Note: I had a really hard time trying to replicate the issue with Google File Stream. Using a 2GB file on my google drive, it seemed to stay available locally no matter what I did. I tried limiting the download speed using the File Stream setting to be sure it would take a noticeable amount of time to download. Even after uninstalling, clearing the cache, and reinstalling, the file would show up almost instantly.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -27,6 +27,10 @@ Unreleased Changes
 * Recreation
     * Validate values in the type column of predictor tables early in execution. Raise
       a ValueError if a type value isn't valid (leading/trailing whitespace is okay).
+* Validation
+    * Set a timeout on validation functions that open a file. This will raise an error to
+      prevent the UI from hanging if a file has to be downloaded from Google Drive File
+      Stream or a similar service.
 
 3.8.7 (2020-07-17)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -36,9 +36,8 @@ Unreleased Changes
     * Validate values in the type column of predictor tables early in execution. Raise
       a ValueError if a type value isn't valid (leading/trailing whitespace is okay).
 * Validation
-    * Set a timeout on validation functions that open a file. This will raise an error to
-      prevent the UI from hanging if a file has to be downloaded from Google Drive File
-      Stream or a similar service.
+    * Set a 5-second timeout on validation functions that access a file. This will raise 
+      a warning and prevent validation from slowing down the UI too much.
 
 3.8.7 (2020-07-17)
 ------------------
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -292,6 +292,7 @@ def check_raster(filepath, projected=False, projection_units=None):
         A string error message if an error was found.  ``None`` otherwise.
 
     """
+    print('starting check_raster...')
     file_warning = check_file(filepath, permissions='r')
     if file_warning:
         return file_warning
@@ -312,6 +313,7 @@ def check_raster(filepath, projected=False, projection_units=None):
         return projection_warning
 
     gdal_dataset = None
+    print('check_raster finished.')
     return None
 
 
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -667,7 +667,6 @@ def validate(args, spec, spatial_overlap_opts=None):
         found, an empty list is returned.
 
     """
-    print('in validation.validate')
     validation_warnings = []
 
     # step 1: check absolute requirement
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -628,7 +628,10 @@ def timeout(func, *args, timeout=5, **kwargs):
     thread.join(timeout=timeout)
 
     if thread.is_alive():
-        raise RuntimeError('File checking thread timed out.')
+        return ('Validation of this file timed out. If this file is stored in '
+               'a file-streaming service, it may be taking too long to '
+               'download. Try storing it locally.')
+
     else:
         LOGGER.info('File checking thread completed.')
         # get any warning messages returned from the thread
@@ -804,14 +807,8 @@ def validate(args, spec, spatial_overlap_opts=None):
             LOGGER.exception(
                 'Error when validating key %s with value %s',
                 key, args[key])
-            if str(ex) == 'File checking thread timed out.':
-                warning_msg = ('Validation of this file timed out. If this '
-                    'file is stored in a file-streaming service, it may be '
-                    'taking too long to download. Try storing it locally.')
-            else:
-                warning_msg = 'An unexpected error occurred in validation'
             validation_warnings.append(
-                ([key], warning_msg))
+                ([key], 'An unexpected error occurred in validation'))
     # step 5: check spatial overlap if applicable
     if spatial_overlap_opts:
         spatial_keys = set(spatial_overlap_opts['spatial_keys'])
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -9,6 +9,7 @@ import re
 import threading
 import functools
 import importlib
+import queue
 
 import pygeoprocessing
 import pandas
@@ -612,7 +613,11 @@ def timeout(func, *args, timeout=5, **kwargs):
 
     Raises:
     """
-    thread = threading.Thread(target=func, args=args, kwargs=kwargs)
+    # use a queue to share the return value from the file checking thread
+    # the target function puts the return value from `func` into shared memory
+    message_queue = queue.Queue()
+    thread = threading.Thread(target=message_queue.put(func(*args, **kwargs)))
+
     LOGGER.info(f'Starting file checking thread with timeout={timeout}')
     thread.start()
     thread.join(timeout=float(timeout))
@@ -621,6 +626,8 @@ def timeout(func, *args, timeout=5, **kwargs):
         raise RuntimeError('Thread timed out.')
     else:
         LOGGER.info('File checking thread completed.')
+        # get any warning messages returned from the thread
+        return message_queue.get()
 
 
 # checks that open a file could potentially take a long time
@@ -635,8 +642,8 @@ _VALIDATION_FUNCS = {
     'freestyle_string': check_freestyle_string,
     'number': check_number,
     'option_string': check_option_string,
-    'raster': functools.partial(timeout, check_raster),
-    'vector': functools.partial(timeout, check_vector),
+    'raster': check_raster, #functools.partial(timeout, check_raster),
+    'vector': check_vector, #functools.partial(timeout, check_vector),
     'other': None,  # Up to the user to define their validate()
 }
 
@@ -777,6 +784,7 @@ def validate(args, spec, spatial_overlap_opts=None):
         except KeyError:
             validation_options = {}
         type_validation_func = _VALIDATION_FUNCS[parameter_spec['type']]
+        print('type validation func:', type_validation_func)
         if type_validation_func is None:
             # Validation for 'other' type must be performed by the user.
             continue
@@ -784,6 +792,7 @@ def validate(args, spec, spatial_overlap_opts=None):
         try:
             warning_msg = type_validation_func(
                 args[key], **validation_options)
+            print('warning message:', warning_msg)
 
             if warning_msg:
                 validation_warnings.append(([key], warning_msg))
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -11,6 +11,7 @@ import functools
 import importlib
 import queue
 import signal
+import warnings
 
 import pygeoprocessing
 import pandas
@@ -610,9 +611,11 @@ def timeout(func, *args, timeout=5, **kwargs):
             Defaults to 5.
 
     Returns:
-        None
+        A string warning message if the thread completed in time and returned
+        warnings, ``None`` otherwise.
 
     Raises:
+        ``RuntimeWarning`` if the thread does not complete in time.
     """
     # use a queue to share the return value from the file checking thread
     # the target function puts the return value from `func` into shared memory
@@ -626,9 +629,11 @@ def timeout(func, *args, timeout=5, **kwargs):
     thread.join(timeout=timeout)
 
     if thread.is_alive():
-        return ('Validation of this file timed out. If this file is stored in '
-               'a file-streaming service, it may be taking too long to '
-               'download. Try storing it locally.')
+        # first arg to `check_csv`, `check_raster`, `check_vector` is the path
+        warnings.warn(f'Validation of file {args[0]} timed out. If this file '
+            'is stored in a file streaming service, it may be taking a long '
+            'time to download. Try storing it locally instead.')
+        return None
 
     else:
         LOGGER.info('File checking thread completed.')
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -6,6 +6,8 @@ import logging
 import pprint
 import os
 import re
+import threading
+import functools
 import importlib
 
 import pygeoprocessing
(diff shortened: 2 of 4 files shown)


phargogh:
FYI @emlys I just changed the target branch to `main`, which looks to have been pretty seamless!

src/natcap/invest/validation.py
```diff
@@ -596,17 +600,58 @@ def check_spatial_overlap(spatial_filepaths_list,
...
+    def wrapper_func():
+        message_queue.put(func(*args, **kwargs))
+
+    thread = threading.Thread(target=wrapper_func)
+    LOGGER.info(f'Starting file checking thread with timeout={timeout}')
+    thread.start()
+    thread.join(timeout=timeout)
+
+    if thread.is_alive():
+        return ('Validation of this file timed out. If this file is stored in '
```
emlys, line 631: @phargogh and I discussed that it might be better to raise an error on timeout and exit validation entirely. I first tried it out this way, just returning the timeout message like any other validation warning message, because it took the least modification to things outside of `timeout()`. But this could maybe be a good use for a custom exception?

src/natcap/invest/validation.py
```diff
@@ -522,7 +526,7 @@ def check_csv(filepath, required_fields=None, excel_ok=False):
         # use sep=None, engine='python' to infer what the separator is
         dataframe = pandas.read_csv(
             filepath, sep=None, engine='python', encoding=encoding)
-    except Exception:
+    except Exception as e:
```
dcdenu4, line 529: Was this a linting change?

src/natcap/invest/validation.py
```diff
@@ -596,17 +600,58 @@ def check_spatial_overlap(spatial_filepaths_list,
...
+        args: arguments to pass to the function
+        timeout (number): how many seconds to allow the function to run.
+            Defaults to 5.
+
+    Returns:
+        None
+
+    Raises:
+    """
+
```
dcdenu4, line 617: extra line here?

src/natcap/invest/validation.py
dcdenu4: I also thing there's a `TimeoutError` that could be appropriate as opposed to a custom exception.

dcdenu4: I do like the idea of not holding the user hostage if this timeouts. Meaning, on a given timeout I as the user should still be able to click run. Is that what happens in this case?

src/natcap/invest/validation.py
```diff
@@ -596,17 +600,58 @@ def check_spatial_overlap(spatial_filepaths_list,
...
+               'a file-streaming service, it may be taking too long to '
+               'download. Try storing it locally.')
+
+    else:
+        LOGGER.info('File checking thread completed.')
+        # get any warning messages returned from the thread
+        return message_queue.get()
+
+
+# checks that open a file could potentially take a long time
```
dcdenu4, line 641: Should this comment be: 
>opening a file could potentially take a long time

src/natcap/invest/validation.py
```diff
@@ -758,7 +803,7 @@ def validate(args, spec, spatial_overlap_opts=None):
             if warning_msg:
                 validation_warnings.append(([key], warning_msg))
                 invalid_keys.add(key)
-        except Exception:
+        except Exception as ex:
```
dcdenu4, line 806: Was this a linting suggestion as well? Just curious.

tests/test_validation.py
```diff
@@ -665,6 +690,47 @@ def test_wrong_filetype(self):
...
+
+        # make a mock function to replace pandas.read_csv
+        # this will sleep for 10 seconds, then return a dataframe
+        mock_read_csv = unittest.mock.Mock(
+            side_effect=delay, 
+            return_value=pandas.DataFrame([['1','2','3']]))
+
+        # replace the copy of pandas.read_csv that's in the validation module
+        # with the mock function, and try to validate
+        with unittest.mock.patch('natcap.invest.validation.pandas.read_csv', mock_read_csv):
```
dcdenu4, line 728: Neat. I haven' done much Mocking. Would that line behave the same as:
```python
with unittest.mock.patch('natcap.invest.validation.check_csv', mock_read_csv)
```
?

src/natcap/invest/validation.py
emlys: It would still block you from running the model, that input would show up with a red X. That is interesting though, should it be possible to run the model anyway?

I kind of like this approach because it ensures that the UI will show up. Apparently if there are saved inputs from a previous run, validation happens before the model window opens. So if one of the inputs needs to be downloaded it could appear to hang while that happens. I like that this way the error message shows up in the UI. If it raised and exception instead it would have to show up in the terminal window instead and the user might not notice it.

src/natcap/invest/validation.py
emlys: no I did that when I was going to use an exception instead of warning message. Forgot to change it back 🙃

src/natcap/invest/validation.py
emlys: oops! no this was just left over from a previous change

src/natcap/invest/validation.py
emlys: Thanks for catching that! Rephrased it to make more sense.

tests/test_validation.py
emlys: I think that would work also... to be consistent I'd need to change the `Mock` return value to be a list, not a dataframe. But yes that should also trigger the timeout. Do you think it would be better that way? I guess it decreases the amount of irrelevant code that's run in the test

emlys:
Adding @phargogh because we had talked about this issue

tests/test_validation.py
dcdenu4: Hmm, I'm not sure about "better" but I'll leave it up to you. Maybe your last comment of decreasing the amount of code that's run makes sense? I guess `pandas.read_csv` is the actual function that would take awhile to load, but to me `check_csv` is a more clear function to mock... I guess 6 in one hand, half dozen in the other.

dcdenu4 (changes requested):
Hey @emlys thanks for these comment and changes. I just had one more comment about raising a `RuntimeError`.

src/natcap/invest/validation.py
```diff
@@ -596,17 +601,60 @@ def check_spatial_overlap(spatial_filepaths_list,
...
+        args: arguments to pass to the function
+        timeout (number): how many seconds to allow the function to run.
+            Defaults to 5.
+
+    Returns:
+        A string warning message if the thread completed in time and returned
+        warnings, ``None`` otherwise.
+
+    Raises:
+        ``RuntimeWarning`` if the thread does not complete in time.
```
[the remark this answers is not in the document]
dcdenu4: Oh, oops. I thought this said `RuntimeError`. I see. Please disregard.

dcdenu4 (approved):

src/natcap/invest/validation.py
phargogh: For future us, we were chatting about this on our internal slack channel and decided that an inability to run validation shouldn't prevent an informed user from running a model.  Rather, we'll want to display some sort of warning about the inability to validate and let the user make an informed decision.

In the longer-run, this will involve reworking the validation API to support status levels.

HISTORY.rst
```diff
@@ -35,6 +35,10 @@ Unreleased Changes
 * Recreation
     * Validate values in the type column of predictor tables early in execution. Raise
       a ValueError if a type value isn't valid (leading/trailing whitespace is okay).
+* Validation
+    * Set a timeout on validation functions that open a file. This will raise an error to
+      prevent the UI from hanging if a file has to be downloaded from Google Drive File
+      Stream or a similar service.
```
phargogh, line 41: The original issue #261 has to do with file streaming services, but the approach we're taking here is more general and would apply to any file-based validation that takes too long.

And would it be worth mentioning that it'll raise a `warning` rather than an error?

src/natcap/invest/validation.py
```diff
@@ -596,17 +601,60 @@ def check_spatial_overlap(spatial_filepaths_list,
...
     'folder': check_directory,
     'directory': check_directory,
     'freestyle_string': check_freestyle_string,
     'number': check_number,
     'option_string': check_option_string,
-    'raster': check_raster,
-    'vector': check_vector,
+    'raster': functools.partial(timeout, check_raster),
+    'vector': functools.partial(timeout, check_vector),
     'other': None,  # Up to the user to define their validate()
```
phargogh, line 658: Neat use of `functools.partial`!

Do you think it'd be useful to also wrap `file`, `folder` and `directory` with the timeout function as well?  Since the idea of the new `timeout` function here is to have a timeout on file-based operations and these three also make OS calls to check if a file exists, I'm wondering if we're likely to get into a case where even checking to see if a file or folder exists on disk could take some time, like if someone is checking an NFS share or something like that.

As a separate note about the use of `functools.partial`, wrapping the timeout functions in this way unfortunately does not allow for the timeout to differ per use case.  Here are a few that come to mind that might benefit from different timeouts:

1. The Qt UI - here I think the default 5s timeout could work just fine ... we're just trying to prevent a feeling of unresponsiveness.
2. The workbench - we've previously talked about timeouts in the context of that it would be great to be able to basically guarantee fast validation ... so in that case, maybe we'd want validation to return within a much shorter timeframe.  In any case, I suspect it would be useful to be able to configure the timeout so that the workbench and/or server can set their own limits as needed.
3. Model execution - there are a number of model that call `validate` within the model's execute.  In this case it could be reasonable to permit a longer-running validation.

All that being said, perhaps this isn't the right place to address these timeout use cases, and the solution as stated _does_ address the issue at hand so maybe it'd be worth addressing this along with the other design issues in the validation API that came up on the slack thread yesterday.  What do you think @dcdenu4 @emlys ?

tests/test_validation.py
```diff
@@ -199,6 +201,34 @@ def validate(args, limit_to=None):
...
+            time.sleep(10)
+
+        # this will return a warning if the timeout is exceeded
+        # timeout defaults to 5 seconds so this should fail
+        with warnings.catch_warnings(record=True) as ws:
+            # cause all warnings to always be triggered
+            warnings.simplefilter("always")
+            validation.timeout(func, 'arg')
+            self.assertTrue(len(ws) == 1)
+            self.assertTrue('timed out' in str(ws[0].message))
```
phargogh, line 230: Do you think it would be reasonable to trim down the sleep time here?  10s isn't a terribly long time to wait in the scope of the rest of the test suite, but it also doesn't seem like that long of a wait is required to test the timeout function.

tests/test_validation.py
```diff
@@ -665,6 +695,49 @@ def test_wrong_filetype(self):
...
+        # make an actual file so that `check_file` will pass
+        path = os.path.join(self.workspace_dir, 'slow.csv')
+        with open(path, 'w') as file:
+            file.write('1,2,3')
+
+        # define a side effect for the mock that will sleep
+        # for longer than the allowed timeout
+        def delay(*args, **kwargs):
+            time.sleep(10)
+
```
phargogh, line 711: Same here ... between the `time.sleep` and the timeout we can set, would it be reasonable to turn down the expected runtime of these tests?  Just hoping to have our tests run as quickly as is reasonable, is all!

tests/test_validation.py
```diff
@@ -665,6 +695,49 @@ def test_wrong_filetype(self):
...
+        mock_read_csv = unittest.mock.Mock(
+            side_effect=delay, 
+            return_value=pandas.DataFrame([['1','2','3']]))
+
+        # replace the copy of pandas.read_csv that's in the validation module
+        # with the mock function, and try to validate
+        with unittest.mock.patch('natcap.invest.validation.pandas.read_csv', mock_read_csv):
+            with warnings.catch_warnings(record=True) as ws:
+                # cause all warnings to always be triggered
+                warnings.simplefilter("always")
```
phargogh, line 736: This is pretty minor since we don't use warnings all that much throughout InVEST, but this looks to me like installing this filter is altering a global `warnings` state ... is that the case?  If so, would it be possible to uninstall the filter once the test completes, on both test success and test failure so we can avoid the side effect?  Something like

```python
try:
    warnings.simplefilter("always")
    validation.validate(args, spec)
    self.assertTrue(len(ws) == 1)
    self.assertTrue('timed out' in str(ws[0].message))
finally:
    <uninstall the warnings filter>
```

src/natcap/invest/validation.py
```diff
@@ -6,7 +6,12 @@
 import pprint
 import os
 import re
+import threading
+import functools
 import importlib
+import queue
+import signal
```
phargogh, line 13: Is signal unused?  Maybe this was left over from a prototype?

phargogh (changes requested):
Hey @emlys this looks like a nice solution to the file streaming issue!

I had a couple of pretty minor requests, and then one larger question about the nature of a timeout in the scope of a couple of different use cases we have that I think would be worth talking about at least a little and absolutely might be outside of the scope of this PR.,  But I wanted to bring it up and I'd love to hear what you think!

tests/test_validation.py
emlys: Happy to change it!

tests/test_validation.py
emlys: Yes definitely, 6 seconds should be fine

src/natcap/invest/validation.py
emlys: oops, forgot to take that out!

tests/test_validation.py
emlys: Looks like it should be restored to the original state after exiting the context manager: https://docs.python.org/3/library/warnings.html#testing-warnings

src/natcap/invest/validation.py
emlys: It could definitely be useful to have a timeout for `file`, `folder`, and `directory` as well! I didn't include them at first because I'm pretty sure that doing `os.path.exists` on a file stream filepath wouldn't trigger it to download, but there are probably other situations where it could take a long time, so I'll add the timeouts there.

It should work to pass in a different time to `timeout`...
`functools.partial(timeout, check_csv, timeout=10)` --> `timeout(check_csv, *args, timeout=10, **kwargs)`, and then the thread target becomes `check_csv(*args, **kwargs)`. Changing the timeout this way does seem to work in the tests, so I'm pretty confident, but it's definitely not very intuitive how the args get passed around

phargogh: Ok great!  Yes, probably safest to add timeouts to all of the filesystem operations in case those take some time.

For the timeouts, it makes sense to pass the timeout argument around, and after some reflection, I think I might suggest that we wait to address that until we have a more concrete plan ironed out for a round of updates to validation.  What do you think?

tests/test_validation.py
emlys: Turns out that it's not so simple to replace `'natcap.invest.validation.pandas.read_csv'` with `'natcap.invest.validation.check_csv'` because `validate` references `check_csv` via `functools.partial` in the `_VALIDATE_FUNCS` dictionary. Ended up mocking the dictionary instead and replacing `functools.partial(timeout, check_csv)` with `functools.partial(timeout, delay)` :)

src/natcap/invest/validation.py
emlys: Agree that we can wait on adding that option until we plan to use it for those different cases 👍

HISTORY.rst
emlys: good point!

tests/test_validation.py
phargogh: Huh, so it does!  Thanks for looking into that.

dcdenu4:
@phargogh @emlys I'm in agreement that it's fine to hold off on some of the `timeout` implementation and consider it's use in a larger redesign of the validation module.

phargogh (approved):
Great, thanks @emlys !

Issue #93, opened 2020-04-23
CV - validate SLR input is a point geometry


davemfish:
Providing other geometry types yields a whole lot of ` utils._log_gdal_errors(67) ERROR [errno 6] Incompatible geometry for operation` during `execute` and probably some bogus values in the results.

Either validate for points in `validate` or at the very least raise an exception on other geom types during `execute`.

PR #295, opened 2020-09-02
Validate coastal vulnerability SLR geometry type

emlys:
Addresses #93 
Added a step to `coastal_vulnerability.validate` that checks that the SLR vector is a point or multipoint geometry, and a corresponding test case.
(Should single point geometries be allowed? User's guide isn't clear on this)

### DIFF
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -2941,6 +2941,19 @@ def validate(args, limit_to=None):
                 (['shelf_contour_vector_path'],
                  'Must be a polyline vector'))
 
+    if ('slr_vector_path' in sufficient_keys and 
+       'slr_vector_path' not in invalid_keys):
+        vector = gdal.OpenEx(args['slr_vector_path'], gdal.OF_VECTOR)
+        layer = vector.GetLayer()
+        # all OGR point/multipoint codes
+        if layer.GetGeomType() not in [
+            ogr.wkbPoint, ogr.wkbMultiPoint,
+            ogr.wkbPointM, ogr.wkbMultiPointM,
+            ogr.wkbPointZM, ogr.wkbMultiPointZM,
+            ogr.wkbPoint25D, ogr.wkbMultiPoint25D]:
+            validation_warnings.append((['slr_vector_path'], 
+                f'Must be a point or multipoint geometry.'))
+
     if ('slr_vector_path' not in invalid_keys and
             'slr_field' not in invalid_keys and
             'slr_vector_path' in sufficient_keys and
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -2936,7 +2936,10 @@ def validate(args, limit_to=None):
         layer = vector.GetLayer()
         # linestring codes: https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogr_core.h
         if layer.GetGeomType() not in [
-                2, 5, 2002, 2005, 3002, 3005, -2147483646, -2147483643]:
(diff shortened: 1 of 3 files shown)


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2941,6 +2941,19 @@ def validate(args, limit_to=None):
...
+    if ('slr_vector_path' in sufficient_keys and 
+       'slr_vector_path' not in invalid_keys):
+        vector = gdal.OpenEx(args['slr_vector_path'], gdal.OF_VECTOR)
+        layer = vector.GetLayer()
+        # all OGR point/multipoint codes
+        if layer.GetGeomType() not in [
+            ogr.wkbPoint, ogr.wkbMultiPoint,
+            ogr.wkbPointM, ogr.wkbMultiPointM,
+            ogr.wkbPointZM, ogr.wkbMultiPointZM,
+            ogr.wkbPoint25D, ogr.wkbMultiPoint25D]:
```
davemfish, line 2953: Ahh there's always a surprising number of somewhat obscure geom types to handle. Using the constants like this is much nicer than hardcoding the integers like I did in the `shelf_contour_vector_path` test. 

What do you think about replacing those integers with the constants for that test as well? That would be a clear improvement for readability and make us more consistent across tests. I'll add an in-line comment in the test I'm referring to.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2941,6 +2941,19 @@ def validate(args, limit_to=None):
                 (['shelf_contour_vector_path'],
```
davemfish, line 2941: It's not related to the SLR issue, but it would be a clear improvement to replace these integer codes with the `ogr` constants like you have them in the SLR test. Do you think it's a good idea to do that while we're here?

tests/test_coastal_vulnerability.py
```diff
@@ -1428,6 +1428,26 @@ def test_missing_sealevelrise_field(self):
...
+        srs.ImportFromEPSG(4326)
+        polygon_vector.CreateLayer('layer', srs, ogr.wkbPolygon)
+        polygon_vector = None
+
+        args = {'slr_vector_path': polygon_path}
+        warning_messages = coastal_vulnerability.validate(args)
+
+        with self.assertRaises(ValueError):
+            for keys, message in warning_messages:
+                print(keys, message)
```
davemfish, line 1447: Probably don't need to keep this `print`, right?

davemfish (changes requested):
Thanks for taking this one @emlys . The check and the test look great, so great that I commented that another very similar test for geometry types could benefit from a quick update to look more like this one. It's not critical, but please see what you think there.

src/natcap/invest/coastal_vulnerability.py
davemfish: This comment belongs on the previous line. For some reason github won't let me add a comment to that line.

src/natcap/invest/coastal_vulnerability.py
emlys: Sure I think that would be an improvement!

tests/test_coastal_vulnerability.py
emlys: ahh thanks for catching that!

davemfish (approved):
@emlys , thanks for the additional update here. I realized I didn't directly answer your question about single Point geometries being okay. I think they are. It doesn't necessarily make a lot of sense to use the SLR input with only one point, as it's meant to capture spatial variation in SLR within the AOI, but results would still be valid and predictable. So similar to our slack thread about validation warnings vs errors, I wouldn't want to stop a user from doing this.

It looks like we're done with this one, but I'll approve and let you click merge just in case I'm jumping the gun and you weren't actually done here.

PR #306, opened 2020-09-10
Allow extra args keys in validation

### LINKED ISSUE #305: Check all models for mismatched args spec
https://community.naturalcapitalproject.org/t/water-yield-output-error/1315
This looks to be an error introduced/uncovered by this PR: https://github.com/natcap/invest/pull/273
The hydropower UI file includes the arg `do_scarcity_and_valuation`, which doesn't exist in the model `ARGS_SPEC`, and because of the changes I made in this validation PR, args that don't have a match cause a `KeyError`. I guess this slipped through the tests. We should check the other models for mismatched args. This won't be a problem in the workbench because the UI will reference the `ARGS_SPEC` directly.

emlys:
Fixes #305 
Put the key lookup in a try/except to be consistent with other error handling

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -1218,7 +1218,7 @@ class TestValidationFromSpec(unittest.TestCase):
                 'required': True
             }
         }
-        message = 'Provided key b does not exist in ARGS_SPEC'
+        message = 'DEBUG:natcap.invest.validation:Provided key b does not exist in ARGS_SPEC'
         
         with self.assertLogs('natcap.invest.validation', level='DEBUG') as cm:
             validation.validate(args, spec)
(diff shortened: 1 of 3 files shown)


tests/test_validation.py
```diff
@@ -1204,3 +1204,23 @@ def test_spatial_overlap_error_optional_args(self):
...
+            'a': {
+                'type': 'freestyle_string',
+                'name': 'a',
+                'about': 'a freestyle string',
+                'required': True
+            }
+        }
+        message = 'DEBUG:natcap.invest.validation:Provided key b does not exist in ARGS_SPEC'
+        
+        with self.assertLogs('natcap.invest.validation', level='DEBUG') as cm:
```
phargogh, line 1223: I did not know this assertion existed ... this is great!

phargogh (approved):
Thanks, @emlys !

PR #316, opened 2020-09-22
Rebuild Coastal Blue Carbon

### LINKED ISSUE #17: UI tests on Multi inputs crash
This has happened fairly consistently for a while, but is not reproducible outside of the appveyor VM. Since only one InVEST model (coastal blue carbon) uses a Multi input, we’ll skip this class of tests for now, and consider refactoring the CBC UI to avoid this troublesome input.

For now we're skipping the test. Eventually we might factor out the instances of the Multi objects altogether.

Originally reported by @davemfish at https://bitbucket.org/natcap/invest/issues/3936/ui-tests-on-multi-inputs-crash-on-appveyor

### LINKED ISSUE #135: CBC model doesn't use its half-life inputs
The CBC model doesn't use its half-life parameters at all, despite being a documented (and required) input to the model.  This is missing from the Emissions calculations and is dramatically affecting the numerical outputs experienced by a user on the forums (by 8-10 orders of magnitude).

### LINKED ISSUE #209: Rebuild Coastal Blue Carbon

phargogh:
This is a complete rebuild of the InVEST Coastal Blue Carbon model that addresses a number of longstanding issues with the model including

* **Correct outputs**
  The former implementation failed to use certain inputs (#135), and our overall confidence in the correctness of the model's outputs was very low.  While implementing, other equations were implemented in fundamentally incorrect ways, leading to incorrect outputs in all cases involving emissions and valuation.
* **Improved model debugging**
  The prior implementation was incredibly difficult to debug due to the complex, 3-dimensional way the model tracked timeseries data.  This new implementation breaks the model's operations into tasks and raster calculator operations, which simplifies debugging at the cost of most of the model code being task orchestration and a lot of files being written to disk per timestep.
* **Memory efficiency**
  The prior implementation was vulnerable to maxing out the available memory in cases where the user was running the model over a timeseries with a large number of timesteps.  This vulnerability is eliminated by breaking the model's operations into a series of tasks that each do their own spatial operations.
* **An 'advanced' mode for users who want to provide spatially-explicit accumulation or emissions layers**
  Colleagues across partner institutions have occasionally desired to provide spatially-explicit rasters of rates of carbon accumulation and emissions, which the 'official' model inputs make very difficult to do.  This refactor now includes a second public-facing entrypoint that can grant a scientist lower-level access to the interesting computations in the emissions timeseries (which are more complicated than the baseline period leading up to the first transition), and support these sorts of analyses.
* **Consolidated input files**
  Several of the required inputs to the CBC model and its preprocessor have been combined, leading to a simpler UI with clearer input names, and clearer-to-comprehend input files themselves.

Additionally, this PR also removes the `Multi` element from the UI and its tests (#17 ), as CBC was the last model to use this input type.

I'm submitting this PR for review now, but I have yet to revisit the user's guide chapter.  So look forward to an update of the UG hash.

Fixes #209 
Fixes #135 
Fixes #17 

@emlys I've added you as a reviewer, but anyone, feel free to chime in if you like!  This is kind of a big one.


richpsharp:
Yay James, you did it!

phargogh:
Ha, thanks @richpsharp !  I'm just glad it's done (minus a few cleanup things like the tests failing), and I'm glad that it's probably correct ... or at least more correct than the old version.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+        else:
+            inflation_rate = float(args['inflation_rate']) * 0.01
+            annual_price = float(args['price'])
 
-    Args:
-        snapshot_years (list): list of snapshot years.
-        snapshot_idx (int): index of snapshot
+            try:
+                max_year = max(transition_years.union(set([analysis_year])))
+            except ValueError:
```
emlys, line 297: Curious why the try/except here rather than `if transition_years... else...`? I've heard that using try/except is more pythonic. Is that true and why would you choose one over the other for a case like this?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+                      stock_rasters[year][pool]),
+                dependent_task_list=current_stock_dependent_tasks,
+                target_path_list=[stock_rasters[year][pool]],
+                task_name=f'Calculating {pool} carbon stock for {year}')
+
+            # Calculate disturbance volume if we're at a transition year.
+            if year in transition_years:
+                # We should only switch around the transition years the first
+                # time we encounter this, not for each pool.
+                if current_transition_year != year:
```
emlys, line 774: I'm a little confused about what the comment is saying here vs. what it's doing. It looks like `current_transition_year` will get updated on each iteration, because `year` will increment each time. And it makes sense that you're keeping track of the prior and current transition years on lines 775 and 776. But I'm not sure why line 774 is necessary (shouldn't it always be true?) and the comment seems to contradict that. Please correct me if I'm missing something :)

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+                    current_transition_year = year
+
+                disturbance_vol_rasters[year][pool] = os.path.join(
+                    intermediate_dir,
+                    DISTURBANCE_VOL_RASTER_PATTERN.format(
+                        pool=pool, year=year, suffix=suffix))
+                current_disturbance_vol_tasks[pool] = task_graph.add_task(
+                    func=pygeoprocessing.raster_calculator,
+                    args=([(disturbance_magnitude_rasters[year][pool], 1),
+                           # TODO: why is this using prior year's stocks?
```
emlys, line 785: Is it okay to leave this TODO for a future issue?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+                   stock_rasters[year][POOL_BIOMASS],
+                   yearly_accum_rasters[current_transition_year][POOL_LITTER]],
+                  total_carbon_rasters[year]),
+            dependent_task_list=[
+                current_stock_tasks[POOL_SOIL],
+                current_stock_tasks[POOL_BIOMASS]],
+            target_path_list=[total_carbon_rasters[year]],
+            task_name=f'Calculating total carbon stocks in {year}')
+
+        # If in the last year before a transition of the last year before the
```
emlys, line 885: Should this say 'or' instead of 'of'?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+    valuation_tasks = {}
+
+    first_transition_year = min(transition_years)
+    final_year = int(args['analysis_year'])
+
+    summary_net_sequestration_tasks = []
+    summary_net_sequestration_raster_paths = {
+        first_transition_year: args['sequestration_since_baseline_raster']}
+
+    for year in range(first_transition_year, final_year+1):
```
emlys, line 735: Curious why this loop is over every year in the range, rather than just the transition years? Maybe could avoid some extra intermediate rasters if this could be done only on transition years

src/natcap/invest/ui/cbc.py
```diff
@@ -15,28 +14,39 @@ def __init__(self):
...
-            label='LULC Lookup Table (CSV)',
-            validator=self.validator)
-        self.add_input(self.lulc_lookup_uri)
-        self.lulc_snapshot_list = inputs.Multi(
-            args_key='lulc_snapshot_list',
-            callable_=functools.partial(inputs.File, label="Input"),
-            label='Land Use/Land Cover Rasters (GDAL-supported)',
-            link_text='Add Another')
-        self.add_input(self.lulc_snapshot_list)
+        def _ui_keys(args_key):
```
emlys, line 17: Since this function is repeated in both `CoastalBlueCarbon` and `CoastalBlueCarbonPreprocessor`, could it be defined just one time outside of the classes?

emlys (commented):
This is really cool! Since we talked about it at a higher level yesterday, I'm just leaving some minor comments and questions. Overall it looks like a big improvement and a lot more consistent with other models as far as the structure and quality.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
                 "The discount rate on future valuations of "
-                "sequestered carbon, compounded yearly."),
-        }
+                "sequestered carbon, compounded yearly.  A "
+                "value of 5 would represent a 5% discount."),
+        },
     }
 }
 
-_INTERMEDIATE = {
```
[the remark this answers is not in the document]
phargogh: Nice balance between the two approaches!  I've done this in 3c76f41b which also includes another raster path that I hadn't made into a module-level constant.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+
+    snapshots = _extract_snapshots_from_table(args['landcover_snapshot_csv'])
+
+    # Phase 1: alignment and preparation of inputs
+    baseline_lulc_year = min(snapshots.keys())
+    baseline_lulc_path = snapshots[baseline_lulc_year]
+    baseline_lulc_info = pygeoprocessing.get_raster_info(
+        baseline_lulc_path)
+    min_pixel_size = numpy.min(numpy.abs(baseline_lulc_info['pixel_size']))
+    target_pixel_size = (min_pixel_size, -min_pixel_size)
```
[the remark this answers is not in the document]
phargogh: I believe it will be ok, and you do make a really interesting point here.  I've made a note to include this in the updated UG, just to clarify the behavior for anyone who's wondering.

Interestingly, the model's former implementation actually did two things: the preprocessor used the pixel minimum, and the model resampled to the baseline.  So in a way, using the minimum resolution would be adhering to (part of) the original behavior.  In this case, because we're working with discrete/classified values, I suspect that the underlying results will be OK given that there's already a degree of error stemming from discretization.

In practice, all of the CBC model runs I've ever seen have all had very similar or identical cell sizes.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: It's true!  It is accepted and even recommended to use reasonable and expected exceptions for control flow, which generally follows the form of:

```
try:
    the thing you would like to have work as expected
except possible_error:
    the backup action instead if it doesn't work
```

So in the case of the above, the assumption is `len(transition_years) > 0`.  In this case, that is an easy enough thing to specify, and so in this case I actually think that your suggested approach is far more direct.  The `ValueError` I had here does not make it immediately clear what the exception was and why it happened, whereas `if transition_years` is much clearer.

Exceptions are especially useful because they allow us to assume that the thing we want to work will work as expected, and yet we are able to express backup or cleanup solutions as needed in a fairly compact syntax.  Exceptions can also be useful for breaking out of multiple nested loops or function calls, ensuring something happens regardless of success or error state, or just jumping into a possibly-expensive operation without worrying ahead of time whether it might fail.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+
+    transition_years = set()
+    analysis_year = int(args['analysis_year'])
+
+    base_paths = [baseline_lulc_path]
+    aligned_lulc_paths = {}
+    aligned_paths = [os.path.join(
+        intermediate_dir,
+        f'aligned_lulc_baseline_{baseline_lulc_year}{suffix}.tif')]
```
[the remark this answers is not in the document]
phargogh: Yes indeed!  Patched in 3c76f41b.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+    aligned_lulc_paths[baseline_lulc_year] = aligned_paths[0]
+    for snapshot_year in snapshots:
+        if snapshot_year == baseline_lulc_year:
+            continue
+        base_paths.append(snapshots[snapshot_year])
+        transition_years.add(snapshot_year)
+        aligned_paths.append(
+            os.path.join(
+                intermediate_dir,
+                f'aligned_lulc_snapshot_{snapshot_year}{suffix}.tif'))
```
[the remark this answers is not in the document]
phargogh: Also in 3c76f41b

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+        aligned_lulc_paths[snapshot_year] = aligned_paths[-1]
```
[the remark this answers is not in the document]
phargogh: I agree, those structures a re redundant when `aligned_lulc_paths` can itself be used more directly, and sorted when needed.  Patched in 33077eae.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+
+    prices = None
+    if ('do_economic_analysis' in args and
+            args['do_economic_analysis'] not in ('', None)):
+        if args.get('use_price_table', False):
```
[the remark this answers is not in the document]
phargogh: Yes indeed!  `.get` could be used for such a case.  There are a few other cases where similar checks are performed in CBC, which have been included in 71a67073.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+        transition_years.add(snapshot_year)
+        aligned_paths.append(
+            os.path.join(
+                intermediate_dir,
+                f'aligned_lulc_snapshot_{snapshot_year}{suffix}.tif'))
+        aligned_lulc_paths[snapshot_year] = aligned_paths[-1]
+
+    prices = None
+    if ('do_economic_analysis' in args and
+            args['do_economic_analysis'] not in ('', None)):
```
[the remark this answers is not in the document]
phargogh: The UI will always include the value of `do_economic_analysis`, but the same can't be said about someone using the model through a script.

You're very right about this condition being incorrect!  I've patched it and updated it to use the `.get()` notation above (and also further simplifying because we can rely on python's truthiness here) in 26f8b326

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+        target_path_list=aligned_paths,
+        task_name='Align input landcover rasters.')
+
+    # We're assuming that the LULC initial variables and the carbon pool
+    # transient table are combined into a single lookup table.
+    biophysical_parameters = utils.build_lookup_from_csv(
+        args['biophysical_table_path'], 'code')
+
+    (biomass_disturb_matrix, soil_disturb_matrix,
+        biomass_accum_matrix, soil_accum_matrix) = _read_transition_matrix(
```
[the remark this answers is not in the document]
phargogh: Sure would, especially since this is the only place where it's used.  Patched in e3ef41c7

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+    accumulation_matrices = {
+        'soil': soil_accum_matrix,
+        'biomass': biomass_accum_matrix,
+    }
+
+    # Baseline stocks are simply reclassified.
+    # Baseline accumulation are simply reclassified
+    # There are no emissions, so net sequestration is only from accumulation.
+    # Value can still be calculated from the net sequestration.
+    end_of_baseline_period = baseline_lulc_year
```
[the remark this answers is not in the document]
phargogh: How strange ... must have been left over from a refactored condition.  Patched in 30d406e4

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+            total_net_sequestration_for_baseline_period),
+        'stocks_at_first_transition': {
+            POOL_SOIL: stock_rasters[end_of_baseline_period-1][POOL_SOIL],
+            POOL_BIOMASS: stock_rasters[
+                end_of_baseline_period-1][POOL_BIOMASS],
+            POOL_LITTER: stock_rasters[end_of_baseline_period-1][POOL_LITTER],
+        }
+    }
+
+    if ('do_economic_analysis' in args and args['do_economic_analysis']):
```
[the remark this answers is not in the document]
phargogh: Sure can!  Patched in 027b91bf

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +187,1485 @@
...
+    taskgraph_cache_dir = os.path.join(
+        args['workspace_dir'], TASKGRAPH_CACHE_DIR_NAME)
+    task_graph = taskgraph.TaskGraph(
+        taskgraph_cache_dir, n_workers)
+
+    suffix = args.get('suffix', '')
+    intermediate_dir = os.path.join(
+        args['workspace_dir'], INTERMEDIATE_DIR_NAME)
+    output_dir = os.path.join(
+        args['workspace_dir'], OUTPUT_DIR_NAME)
```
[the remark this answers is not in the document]
phargogh: Yeah, I was on the fence about having the setup be in a single function, since it's often clearer to have the setup clearly stated where it's needed rather than hidden away in a function.  In this case, the benefit clearly outweighs the cost of having this in a helper function, and the setup is more consistent as a result.  Patched in 6517b782

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: Ha, this check was the result of a neat bug in bookkeeping! 

An abbreviated form of the timeseries loop up to this conditional would be:
```python
for year in range(first_transition_year, final_year+1):
    # ...
    for pool in (POOL_SOIL, POOL_BIOMASS):
        # ...
        if year in transition_years:
            # ...

            # We should only switch around the transition years the first
            # time we encounter this, not for each pool.
            if current_transition_year != year:
                prior_transition_year = current_transition_year
                current_transition_year = year
```

If we don't have the `if current_transition_year != year`, the lines
```python
prior_transition_year = current_transition_year
current_transition_year = year
```
would be executed twice in the matching year, once each for `POOL_SOIL` and `POOL_BIOMASS`, resulting in both `prior_transition_year` and `current_transition_year` pointing to the same year, which we don't want.  The condition here ensures that we only update the prior/current transition years the first time we encounter it.

So if we're visiting `if current_transition_year != year` the first time, you're right that the two variables, `current_transition_year` and `year` will not have the same value.  So then we execute the next lines, setting `prior_transition_year`, and then `current_transition_year` is now set to `year.  So then for the next carbon pool, the next time we reach line 774, ` current_transition_year` will have already been set to `year`, and we don't change our understanding of the prior/current year.

I do think it would be clearer to have this logic elsewhere in the timeseries loop so that we could avoid this possible duplication, but I'm not convinced that this could be done without duplicating conditionals elsewhere as well.  I'm happy to revisit this if you like ... let me know what you think!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: I've convinced myself that the emissions calculations as stated are using the correct carbon stocks raster, so c5451cb6 changes this from a `TODO` to a block comment describing my rationale for basing disturbance volume on last year's stocks.

phargogh: Actually, I take that back (and reverted c5451c b6).  effcb738 now includes the actual fix, which is consistent with the way that the model handles stocks elsewhere in the model.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: 040c24e1

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: This was absolutely a consideration, and, in fact, the implementation that this replaces does almost exactly that.  Certainly the prior model only wrote rasters for each of the transitions, containing as much as possible in numpy arrays in memory.

There were 3 major issues that arose from this approach:

1. **memory constraints**
  This model has no upper bound on how many years the timeseries can operate over.  So even if we were to have each task operating on only those years between transitions, even that number could be very large and exhaust available memory.  We had many issues with memory usage in the prior model, and iterating over lots of files is the safest general-case solution for a timeseries model with an unbounded number of years.
2. **in practice, the intermediate rasters are useful for analyses**
  The prior implementation wrote as few rasters out as it could, which ultimately ended up making it harder for some of our partner organizations to to the science they needed to do.  By providing all these intermediate rasters, we allow our users to do their own analytics and summaries on how carbon stocks change over time, which would have been game-changing to our colleagues in analyses past.
3. **debugging is easier with rasters than with multidimensional arrays**
  The prior implementation used 3D arrays (`m` x `n` x `n_years`) for its analysis, which was more concise to implement but so incredibly difficult to debug when there were issues.  If we were to only output a few summary rasters per transition year, we would still have these 3D arrays, and debugging would be much trickier.  Now, there's probably a way to make all this much easier to debug than the prior implementation, but even if we were to do so, we'd still have the unbounded-array-size problem, mentioned in point 1.

I'm also not particularly happy with the sheer number of rasters produced (and the resulting hit to runtime), but I do honestly think that the model will be more useful for our users and easier to debug as a result, and I'm also glad to say that this should never ever run out of memory on consumer hardware.

src/natcap/invest/ui/cbc.py
phargogh: 424a623a reworks this into a shared function, using `functools.partial` for the instance's reference to the appropriate validator function.

phargogh:
@emlys I believe I've addressed all of your comments, so feel free to take another look if you like!  I've re-requested your review on this.

One thing that is definitely not complete yet, though, is the User's Guide, so please don't merge this just yet!

src/natcap/invest/ui/cbc.py
emlys: Neat!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
emlys: Cool, thanks for the explanation!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
emlys: Ohh, that makes sense! I had missed the `for pool in (POOL_SOIL, POOL_BIOMASS)` loop. Thanks for explaining!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
emlys: Those sound like really good reasons to do it this way! This could be another decision that's worth recording for posterity, either in a big comment or in the user's guide (could pretty much copy-and-paste what you wrote above)

emlys (approved):

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: Added some notes about a couple of things that have been discussed about the model relating to its implementation in ca89ccb9, much of which came from the discussion in this thread.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+            dtype=numpy.float32)
+        sum_array[:] = 0.0
+
+        # Assume everything is valid until proven otherwise
+        valid_pixels = numpy.ones(sum_array.shape, dtype=numpy.bool)
+        pixels_touched = numpy.zeros(sum_array.shape, dtype=numpy.bool)
+        for raster_path in raster_path_list:
+            raster = gdal.OpenEx(raster_path, gdal.OF_RASTER)
+            band = raster.GetRasterBand(1)
+            band_nodata = band.GetNoDataValue()
```
richpsharp, line 1519: I imagine you've thought this through but I'm a little surprised to see you'd want to open and close file on every block iteration. Is there a possibility there might be too many files open at once? If so I think this is fine rather than some kind of dance to try to keep track of the pixels_touched array. Otherwise would you consider a list of raster/band/nodata tuples to avoid opening and closing N files for M iterblocks?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+    # lucode we're indexing with plus 1 (to account for 1-based counting).
+    n_rows = max_lucode + 1
+    soil_disturbance_matrix = scipy.sparse.dok_matrix(
+        (n_rows, n_rows), dtype=numpy.float32)
+    biomass_disturbance_matrix = scipy.sparse.dok_matrix(
+        (n_rows, n_rows), dtype=numpy.float32)
+    soil_accumulation_matrix = scipy.sparse.dok_matrix(
+        (n_rows, n_rows), dtype=numpy.float32)
+    biomass_accumulation_matrix = scipy.sparse.dok_matrix(
+        (n_rows, n_rows), dtype=numpy.float32)
```
richpsharp, line 1589: I think I see what you're doing, and I can tell you've thought it through since below you mention the "dict-based approach" from before. But it looks like you're using these just like a dict? To me it's confusing as to why you'd use sparse matrices instead of a dictionary that you index with a tuple. i.e. I wouldn't find `biomass_accumulation_dict[(from_lucode, to_lucode)]` so annoying that I'd use `scipy` matrices. Was there something more complicated before, or is there a larger issue I don't see?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+
+    # I don't actually know if this uses less memory than the dict-based
+    # approach we had before since that, too, was basically sparse.
+    # At least this approach allows numpy-style indexing for easier 2D
+    # reclassifications.
+    # If we really wanted to save memory, we wouldn't duplicate the float32
+    # values here and instead use the transitions to index into the various
+    # biophysical values when reclassifying. That way we rely on python's
+    # assumption that ints<2000 or so are singletons and thus use less memory.
```
richpsharp, line 1598: I'm also missing the memory issue here. I'd normally expect that a user would have orders of magnitude fewer landcover classes than pixels in the raster. Is the concern that there'd be a unique lulc per pixel?

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
richpsharp: Ahh I think I see. Is this so you can use the numpy index indexing on line 1605 below?

src/natcap/invest/utils.py
```diff
@@ -484,10 +484,8 @@ def build_lookup_from_csv(
     """
     # Check if the file encoding is UTF-8 BOM first
     encoding = None
-    with open(table_path, 'rb') as file_obj:
-        first_line = file_obj.readline()
-        if first_line.startswith(codecs.BOM_UTF8):
-            encoding = 'utf-8-sig'
+    if has_utf8_bom(table_path):
+        encoding = 'utf-8-sig'
```
richpsharp, line 488: Just looking at the `has_utf8_bom` here and below for the dataframe. I thought Pandas had a way where you could export a dataframe as a dictionary? I'm wondering if there's duplicate code here such that `build_lookup_from_csv` could be implemented from a call to `read_csv_to_dataframe` which handles all the weird CSV formatting an encodings?

phargogh:
OK! I just updated the user's guide revisions to reflect the latest changes to the model (contained on `release/3.9` in the UG repo).

And I was chatting with @richpsharp about the model, and he asked to take a look at it.  So just waiting on your review here, Rich!

src/natcap/invest/coastal_blue_carbon/preprocessor.py
```diff
@@ -76,292 +67,175 @@ def execute(args):
...
+        dependent_task_list=[alignment_task],
+        task_name='Determine transitions and write transition table')
+
+    # Creating this table should be cheap, so it's not in a task.
+    # This is only likely to be expensive if the user provided a landcover
+    # lookup with many many rows. If that happens, this tool will have other
+    # problems (like exploding memory in the transition table creation).
+    target_biophysical_table_path = os.path.join(
+        output_dir, BIOPHYSICAL_TABLE.format(suffix=suffix))
+    _create_biophysical_table(landcover_table, target_biophysical_table_path)
```
richpsharp, line 149: Just to note (I don't think this is the case here) but an advantage of creating a file in TaskGraph even if it's fast is if that file were used by a task later. If you recreate it every time then the task down the line will also reexecute every time.

src/natcap/invest/coastal_blue_carbon/preprocessor.py
```diff
@@ -76,292 +67,175 @@ def execute(args):
...
+    def _read_block(raster_path, offset_dict):
+        try:
+            raster = gdal.OpenEx(raster_path, gdal.OF_RASTER)
+            band = raster.GetRasterBand(1)
+            array = band.ReadAsArray(**offset_dict)
+            nodata = band.GetNoDataValue()
+        finally:
+            band = None
+            raster = None
+        return array, nodata
```
richpsharp, line 167: This is a fun function that I feel like you could use in the main module where I left a comment about opening a raster file per iterblock. If that is something you want to do, this could be a nice function to simplify that code in the other file.

richpsharp (commented):
Hi @phargogh, thanks for letting me review this! It looks like a lot of work and done well. I have some rando comments about variable names and sparse matrices vs. dicts and such that don't seem that important so I'll leave this as "comment" rather than "request changes"

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+        # Initial accumulation values are a simple reclassification
+        # rather than a mapping by the transition.
+        yearly_accum_rasters[pool] = os.path.join(
+            intermediate_dir, ACCUMULATION_RASTER_PATTERN.format(
+                pool=pool, year=baseline_lulc_year, suffix=suffix))
+        baseline_accum_tasks[pool] = task_graph.add_task(
+            func=pygeoprocessing.reclassify_raster,
+            args=(
+                (aligned_lulc_paths[baseline_lulc_year], 1),
+                {lucode: values[f'{pool}-yearly-accumulation']
```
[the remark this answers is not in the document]
phargogh: In this case, I feel like `f'{pool}-yearly-accumulation'` directly stated might be clearer than having a templated string up at the top of the model.  If we were to use a templated version, 

```python
values[f'{pool}-yearly-accumulation']
```
would turn into something like this
```python
values[ACCUM_FIELD_NAME.format(pool=pool)]
```
which requires holding the `ACCUM_TABLE_KEY` pattern in mind when we can just explicitly state the keyname and pool via f-strings.  It isn't that much longer, but it's less clear what's being accessed.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+    prices = None
+    if args.get('carbon_prices_per_year', None):
+        prices = {int(year): float(price)
+                  for (year, price) in args['carbon_prices_per_year'].items()}
+    discount_rate = float(args['discount_rate'])
+    baseline_lulc_year = int(args['baseline_lulc_year'])
+
+    # ASSUMPTIONS
+    #
+    # This rebuild assumes that this timeseries analysis is ONLY taking place
```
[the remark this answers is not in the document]
phargogh: Whoops!  Forgot to remove that. 5412879e

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+
+        # If in the last year before a transition or the last year before the
+        # final year of the analysis (which might not be a transition):
+        #  * sum emissions since last transition
+        #  * sum accumulation since last transition
+        #  * sum net sequestration since last transition
+        if (year + 1) in transition_years or (year + 1) == final_year:
+            emissions_rasters_since_transition = []
+            net_seq_rasters_since_transition = []
+            for _year in range(current_transition_year, year + 1):
```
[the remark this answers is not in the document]
phargogh: Yep, I agree! 8afd8b61

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+                                          dtype=numpy.float32)
+    disturbed_carbon_volume[:] = NODATA_FLOAT32
+    valid_pixels = (~numpy.isclose(disturbance_magnitude_matrix,
+                                   NODATA_FLOAT32) &
+                    ~numpy.isclose(stock_matrix, NODATA_FLOAT32))
+    disturbed_carbon_volume[valid_pixels] = (
+        disturbance_magnitude_matrix[valid_pixels] *
+        stock_matrix[valid_pixels])
+    return disturbed_carbon_volume
+
```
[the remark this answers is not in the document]
phargogh: Since this is a [top-level function in the module, wouldn't there be two lines](https://www.python.org/dev/peps/pep-0008/#blank-lines)?  FWIW, `pycodestyle` isn't complaining, but I did some light linting in 1fc4864f

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
     Returns:
-        num_blocks (int): number of blocks in raster
+        ``None``.
+
     """
-    ds = gdal.OpenEx(raster_uri)
-    n_rows = ds.RasterYSize
-    n_cols = ds.RasterXSize
+    for target_raster_year, target_raster_path in sorted(
+            target_raster_years_and_paths.items(), key=lambda x: x[0]):
```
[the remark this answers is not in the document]
phargogh: Yep!  Not strictly needed. 05d17f42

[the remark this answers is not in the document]
phargogh: I also removed a few more lambdas in a3efd314, which also ended up clarifying that I was accidentally sorting on the wrong attribute!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
 
-    If a nodata value is not provided, the function will return an array with
-    NaN values in its place to mark cells that could not be reclassed.​
+    Stocks for the given pool at the end of the baseline period are a function
+    of the initial baseline stocks, the yearly accumulation rate, and the
+    number of years in the baseline period.
 
     Args:
-        array (numpy.array): input data
-        d (dict): reclassification map
```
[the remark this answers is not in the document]
phargogh: As one who has had to do it before, I can attest that it is inconvenient.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
 
     Args:
-        raster_uri (str): filepath to raster
+        net_sequestration_rasters (dict): A dict mapping int years to string
+            paths to sequestration rasters.  The year keys correspond to the
+            years marking the end of a transition period.  All rasters must
+            share the same projected CRS and have identical dimensions.
+        prices_by_year (dict): A dict mapping int years between the baseline
+            year and the final year.
+        discount_rate (float): The discount rate (between 0 and 1).
```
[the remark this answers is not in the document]
phargogh: I've clarified this in the docstring and in `ARGS_SPEC` in 71bfbf09, and I confirm that the model makes no restrictions around whether a discount rate is positive or negative, or even that it be constrained to the range between 0 and 1.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+                              dtype=numpy.float32)
+            npv[:] = NODATA_FLOAT32
 
-    band = ds.GetRasterBand(1)
-    cols_per_block, rows_per_block = band.GetBlockSize()
+            matrix_sum = numpy.zeros(npv.shape, dtype=numpy.float32)
+            valid_pixels = numpy.ones(npv.shape, dtype=numpy.bool)
+            for matrix in sequestration_matrices:
+                valid_pixels &= ~numpy.isclose(matrix, NODATA_FLOAT32)
+                matrix_sum += matrix
```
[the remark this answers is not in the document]
phargogh: Well that's pretty bizarre ... I can't imagine why I set it up like this only to not use the `valid_pixels` mask.  Patched in 45bd61e8

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
 
-def reclass_transition(a_prev, a_next, trans_dict, out_dtype=None,
-                       nodata_mask=None):
-    """Reclass arrays based on element-wise combinations between two arrays.
+def _calculate_accumulation_over_time(
+        annual_biomass_matrix, annual_soil_matrix,
+        annual_litter_matrix, n_years):
+    """Calculate the total accumulation over a period of years.
+
+    This is a shortcut for adding up 3 rasters per year over n years.
```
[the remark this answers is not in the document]
phargogh: Struck: 8c82e339

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+        accumulation_raster_path, emissions_raster_path, target_raster_path):
+    """Calculate net sequestration for a given timestep and pool.
+
+    Sequestration is the per-pixel tallying of carbon accumulated or
+    emitted in this timestep and for this pool.  This model assume that a given
+    pixel may be either in a state of accumulation or a state of emissions, but
+    not both.  It is possible to have a state where a pixel might have both,
+    since a user is able to provide rasters mapping the spatial distribution of
+    carbon accumulated for a transition year.  If this happens, we assume that
+    emissions takes precedence.
```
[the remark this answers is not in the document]
phargogh: Yeah, there are cases like this that are pretty ambiguous in the model's definition.  In reality, I suspect that a single cell could be _both_ emitting _and_ accumulating carbon, with the net effect being whatever the effect is.  But that is definitely not what's in the old implementation and the strict form of transitions described in the UG implies that carbon can only be emitting or accumulating, therefore carbon emissions are halted upon entering an accumulation state.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: Possibly?  The hope with opening/closing each raster was to hopefully avoid issues of too many open files.  I've implemented the update you suggested in 82d7cb61, which also includes some bonus timed logging, since this function can take a while to run on enough sufficiently-large rasters.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: Yes!  The numpy indexing is extremely convenient, and I also wouldn't be surprised if there's some numpy broadcasting magic happening behind the scenes due to the upfront definition of the sparse matrices' data type.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: Yes!  In fact, I ran across this very issue once, and it was due to my own (foolish) suggestion about how to work around the model's prior inability to handle spatial variation in biophysical parameters.  That's now addressed with having a second entrypoint for 'advanced' usage.

src/natcap/invest/utils.py
phargogh: Yes indeed!  Updated in bb8dc54c.

src/natcap/invest/coastal_blue_carbon/preprocessor.py
phargogh: Sounds good to me!  In this case, there aren't any downstream tasks, but I don't have an otherwise good reason to keep it out of a task.  Wrapped in e5749c40

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -188,941 +194,1521 @@
...
+                pool=pool, year=baseline_lulc_year, suffix=suffix))
+        pool_stock_task = task_graph.add_task(
+            func=pygeoprocessing.reclassify_raster,
+            args=(
+                (aligned_lulc_paths[baseline_lulc_year], 1),
+                {lucode: values[f'{pool}-initial'] for (lucode, values)
+                    in biophysical_parameters.items()},
+                stock_rasters[baseline_lulc_year][pool],
+                gdal.GDT_Float32,
+                NODATA_FLOAT32),
```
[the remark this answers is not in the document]
phargogh: Sounds good to me!  Renamed to `NODATA_FLOAT32_MIN` and `NODATA_UINT16_MAX` in 3334ebd8

src/natcap/invest/coastal_blue_carbon/preprocessor.py
phargogh: I've been profiling the reimplementation, and found that out of a total profiled runtime of 1338.949s, only about 10.5s (about 0.8% of total runtime) was spent across almost 6500 calls to `gdal.OpenEx`:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     6776   10.498    0.002   10.498    0.002 {built-in method _gdal.OpenEx}
```

So I feel comfortable with just opening the raster when needed, reading the block and closing the raster, and since there may well be a [practical limit to the number of open files allowed](https://docs.microsoft.com/en-us/cpp/c-runtime-library/file-handling?redirectedfrom=MSDN&view=vs-2019), I think this would be safest.  I'll patch this shortly.

phargogh: Patched in f118a2b2 so that both the model and the preprocessor use a shared function that reads a block on demand.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: There's more discussion related to this in https://github.com/natcap/invest/pull/316#discussion_r501258814 (comment below), and I've re-patched this in f118a2b2 to read in a block at a time since opening GDAL rasters seems to have a negligible effect on the overall runtime of the program.

phargogh:
I've added a few extra commits that have come up while addressing various concerns, and a few more beyond those addressing some interesting issues arising from the forums discussion at https://community.naturalcapitalproject.org/t/coastal-blue-carbon-zero-net-c-sequestration-in-disturbed-blue-carbon-habitats/1398/2?u=jdouglass.  Since KWarnell's data is proving to be an interesting stress test for the model, I'm using this as a chance to profile some of the functionality and also make sure everything looks good on her inputs before saying this is ready for a re-review.  More soon!

src/natcap/invest/coastal_blue_carbon/preprocessor.py
phargogh: @richpsharp just wanted to let you know that I've settled on keeping the raster objects in a list.  After some further profiling, it became clear that I was reading my initial profile results incorrectly and that replacing these repeated `gdal.OpenEx`s with using the original raster values, I could trim the raster summing function's runtime in the main model by about 40%, which has been done (in both the model and preprocessor) in 84a2a20c.

phargogh:
OK!  @emlys @richpsharp , I believe I've addressed all of your comments, and I've also addressed a few extra things that came up when running the model on https://community.naturalcapitalproject.org/t/coastal-blue-carbon-zero-net-c-sequestration-in-disturbed-blue-carbon-habitats/1398/4.  The only things that have been added since  your reviews have been 1) faster iteration over raster datasets in `_sum_n_rasters`, 2) improved handling of missing or incorrect fieldnames in the transition table and 3) some additional output files have been created for the cumulative carbon accumulation across all pools between transitions.  Very happy to talk in greater depth about these changes if you like, but otherwise, back to you!

phargogh:
Following up on our conversation this morning, [here's a diff of everything that's changed since the first review](https://github.com/natcap/invest/compare/ca89ccb..190df09) in case that's helpful.

richpsharp (approved):

phargogh:
OK!  I just resolved the merge conflicts on `HISTORY.rst` and the `Makefile`, and at this point all known issues with the model have been resolved.  Back to you!

phargogh:
@emlys and I chatted about this today and decided it's good to go, so I'll proceed with the merge!

PR #328, opened 2020-10-08
Fix Finfish Suffix Handling

### LINKED ISSUE #307: Finfish should use suffix
The finfish model UI, like all the others, includes a suffix field, but the model `ARGS_SPEC` is missing the `results_suffix` key, so inputting a suffix in the UI has no effect. 
- Add the standard `results_suffix` arg to the spec
- Apply the suffix to all output files

dcdenu4:
This PR fixes the issue of Finfish not using the input suffix. The suffix is added to the output shapefile and the output html file. Added a test for checking file path names.

Closes #307


emlys (approved):
Looks great, thanks!

Issue #334, opened 2020-10-12
Proposal: Allow warnings in InVEST, but elevate warnings to errors in tests


phargogh:
In a [recent forums thread](https://community.naturalcapitalproject.org/t/seasonal-water-yield-error-warn-method/1405) a user reported an issue with the model where the model is elevating all `warnings` to Exceptions, and a `DeprecationWarning` related to a `LOGGER.warn` call is causing the model to crash.

Historically, we've used [warnings such as through `numpy.seterr`](https://github.com/natcap/invest/wiki/spec_invest#handling-warnings-in-numpy-operations) as a way to find numerical errors and other errors during model development, and, admittedly, raising `warnings` as Exceptions can be helpful in that context.

But it's now also leading to a case where a user is being prevented from running the model.  Obviously we should fix this deprecation issue, but we shouldn't be preventing the user from running the model in a reasonable way.  Also, explicitly raising warnings to Exceptions in a binary distribution also opens us up to issues where warnings that are completely outside of our control prevent the model from executing.

# Use Cases

## End-user of our binaries

When running InVEST through our binaries, it is reasonable to expected to be able to run the model without crashing if their inputs are correct.  A `DeprecationWarning` being treated as an Exception (despite the method still working fine) may result in extra work for us on the forums that could have been avoided.

## Developing InVEST

When we're developing InVEST, any warnings raised during the execution of our models (such as `DeprecationWarnings` from calls we make to deprecated python functions, or `numpy` numerical warnings) should be raised to Exceptions and dealt with through our normal development processes.

# Proposal

I propose that we change our development workflow and standard way of handling warnings to make it easier for us to find and fix them when in development, and make `warnings` less fatal for users of our binaries by making the following changes:

1. We remove all instances of `warnings` elevation across InVEST
2. We agree that we won't generally include `warnings` elevation in future model development that makes it out to the user (except where it makes sense to do so)
3. We add a warnings filter to our testing suite so that warnings emitted by InVEST within our tests _are_ elevated to Exceptions, which allows us a chance to fix them proactively. (See `pytest.ini` example below for a possible approach)

<details>
<summary>Example pytest.ini configuration</summary>

This text, contained within `pytest.ini` in the InVEST project root, would allow any warnings raised throughout InVEST to be reported as Exceptions:

```python
[pytest]

# Convert warnings emitted anywhere in InVEST to exceptions.
filterwarnings =
        error:::natcap.invest[.*]
```

Example output from pytest:

```shell
[...]

PR #336, opened 2020-10-12
Replacing LOGGER.warn with LOGGER.warning across InVEST. RE:#333

### LINKED ISSUE #333: Find and replace LOGGER.warn --> LOGGER.warning
We've known for some time that `LOGGER.warn` is deprecated and will be removed eventually.  We should fix this throughout InVEST.

This recently came up on the forums (https://community.naturalcapitalproject.org/t/seasonal-water-yield-error-warn-method/1405), where SWY is raising all warnings to exceptions and a deprecated call to `LOGGER.warn` is causing the model to crash.

phargogh:
This PR replaces `LOGGER.warn` with `LOGGER.warning` throughout InVEST in response to [this forums post](https://community.naturalcapitalproject.org/t/seasonal-water-yield-error-warn-method/1405).

There's still the outstanding issue of what we should do as 'normal' InVEST behavior, which is proposed as #334 .

Fixes #333

### DIFF
--- a/src/natcap/invest/carbon.py
+++ b/src/natcap/invest/carbon.py
@@ -403,7 +403,7 @@ def execute(args):
             if os.path.exists(tmp_filename):
                 os.remove(tmp_filename)
         except OSError as os_error:
-            LOGGER.warn(
+            LOGGER.warning(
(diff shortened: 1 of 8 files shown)


dcdenu4 (approved):
Looks good.

PR #340, opened 2020-10-14
Annual water yield conditional bug

### LINKED ISSUE #332: Hydropower rsuppy not calculated if demand is 0
In `hydropower_water_yield.compute_rsupply_volume`, the `rsupply` field is only set `if wyield_mn and consump_mn`. If either of these are 0, the condition doesn't pass. Seems like 0 should be a valid value, though. This should probably be 
`if wyield_mn is not None and consump_mn is not None`.

Came up in this issue: https://community.naturalcapitalproject.org/t/error-hydropower-evaluation-annual-water-yield/1385/8

dcdenu4:
This PR fixes a bug related to a conditional check that should have been guarding against `None` but instead was guarding against any non-true returning value. This came up in a User Forum post which is linked in the issue: #332 . Ultimately, this fix now allows for a demand table where all demand values are 0.

Closes #332

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -34,6 +34,9 @@
 .. :changelog:
 Unreleased Changes (3.9)
 ------------------------
+* Annual Water Yield:
+    * Fixing bug that limited ``rsupply`` result when ``wyield_mn`` or
+      ``consump_mn`` was 0.
 * General:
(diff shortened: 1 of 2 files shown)


emlys (approved):
Looks great, thanks!

PR #345, opened 2020-10-16
Remove duplicate validation calls

emlys:
While experimenting with early logging from the UI before the model run logfile is created, I noticed `model.validate` was being called a lot of times. The last run args are loaded after `connect`ing args changes to `validate`. So validation was being called once for each arg when it gets autofilled. 
Getting rid of this duplication made the `invest run` command noticeably faster for me.

### DIFF
--- a/src/natcap/invest/ui/model.py
+++ b/src/natcap/invest/ui/model.py
@@ -1165,6 +1165,7 @@ class InVESTModel(QtWidgets.QMainWindow):
             localdoc (string): The filename of the user's guide chapter for
                 this model.
         """
+
         QtWidgets.QMainWindow.__init__(self)
         self.label = label
         self.target = target
@@ -1851,16 +1852,7 @@ class InVESTModel(QtWidgets.QMainWindow):
         def _validate(new_value):
             # We want to validate the whole form; discard the individual value
             self.validate(block=False)
-
-        self.validate(block=False)
-        for input_obj in self.inputs:
-            input_obj.value_changed.connect(_validate)
-            try:
-                input_obj.validity_changed.connect(_validate)
-            except AttributeError:
-                # Not all inputs can have validity (e.g. Container, dropdown)
-                pass
-
+        
         # Set up quickrun options if we're doing a quickrun
         if quickrun:
             @QtCore.Slot()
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Makes sense to me!

src/natcap/invest/ui/model.py
```diff
@@ -1908,6 +1900,15 @@ def _quickrun_close_model():
         self.raise_()  # raise window to top of stack.
         self.validate(block=False)  # initial validation for the model
 
+        for input_obj in self.inputs:
+            input_obj.value_changed.connect(_validate)
+            try:
+                input_obj.validity_changed.connect(_validate)
+            except AttributeError:
+                # Not all inputs can have validity (e.g. Container, dropdown)
+                pass
```
phargogh, line 1909: Ah, I see the issue here that this change corrects.  So under the previous setup, calling `self.load_lastrun()` would trigger the validation as each input's value was being updated, and once again when `self.validate(block=False)` is called.

Under this proposed change, is validation still being called on this set of inputs?

Offhand, it clearly makes sense to populate the inputs first and _then_ validate everything.  But it also looks like the `self.validate(block=False)` should come after this `for input_obj in self.inputs:` loop.  Does that sound right?

phargogh (commented):
I just had a minor comment about the sequence of events, since I bet you're deeper into this at the moment than I am @emlys !  But overall, I agree with bumping the signal/slot connection setup step until after the inputs have been populated, and my only concern is that we're careful that the validation state presented to the user properly reflects the validity of the input set.

src/natcap/invest/ui/model.py
emlys: `self.load_lastrun()` is called before `self.validate(block=False)`, so it is validating the autofilled inputs. I'm not sure if it matters whether the `for input_obj in self.inputs:` happens before or after `self.validate(block=False)`, so long as it's after `self.load_lastrun()`?

phargogh: > I'm not sure if it matters whether the `for input_obj in self.inputs:` block happens before or after `self.validate(block=False)`, so long as it's after `self.load_lastrun()`?

Ah, I see what's happening here, and you're right: it doesn't matter whether the `self.validate()` call is before or after the block where we connect all of the inputs' `value_changed` signals to the model windows' validation method for this use case.

My main concern was that whole-model validation was being called on the new set of inputs, and it is!  We're calling it here in this method, and the UI calls it again when the user clicks the 'Run' button in the UI.  So we should be all set here.

@emlys since the speedup on model UI launch is a user-facing change, should there be a note in `HISTORY`?

emlys: Good point! Just added the note.

phargogh (approved):
Great, thanks!

phargogh:
For future us, the mac binary build is failing, but that's due to the homebrew issue we've previously mentioned, and a workaround for this has been introduced in #358 .

PR #364, opened 2020-10-29
update delineateit in cli.py

emlys:
Update delineateit in cli.py, fixing bug that kept `invest getspec delineateit` and the workbench server from accessing the model. 
I introduced this bug into the release branch so I don't think a history note is needed.


dcdenu4 (approved):
Easy PR!

PR #369, opened 2020-11-04
split valuation option into productivity and energy

### LINKED ISSUE #367: Separate Urban Cooling valuation options
In the Urban Cooling model, there are two valuation types: energy consumption and work productivity. They are mutually independent, but are grouped together under one 'do valuation' option. Because of this you need to input data for both even if you just want to use one. A user just posted on the forum asking about this (https://community.naturalcapitalproject.org/t/how-to-run-work-loss-option-without-energy-consumption-option/1472) and while using dummy data is a workaround, it shouldn't be necessary. We should split these out into two separate options.

emlys:
Split work productivity and energy savings valuation into separate options. Fixes #367


dcdenu4 (commented):
Hey @emlys this looks good to me for the most part. It does look like the updated invest-sample-data revision is causing a data validation error? 

Also, should we add a few tests that use the new functionality? `True, False`, `False, True`, for the energy savings and work productivity respectfully?

emlys:
> Hey @emlys this looks good to me for the most part. It does look like the updated invest-sample-data revision is causing a data validation error?
Ah, I did something weird and accidentally merged in CBC changes to the task/367 branch of invest-sample-data. Should be fixed now!

> Also, should we add a few tests that use the new functionality? `True, False`, `False, True`, for the energy savings and work productivity respectfully?
Sure! I wasn't sure if new tests were really needed here, but I should err on the side of more tests :)

dcdenu4 (commented):
Hey @emlys I'm re-running the tests again because it still looked like validating sample data was failing. I was also thinking that we should update History as well, was that here before or did I miss it in my first review? Also, are there any UG update needs related to this? There might not be if the UG never stated dependence on how options are run.

Just putting `comment` to see what the tests return.

dcdenu4:
@emlys Now it looks like a Windows UI test is failing! I re-ran them again, because it seems odd that the 3.7 but not 3.8 version is failing...

emlys:
Oof, CBC related sample data issues. Looks like it's all good now!

dcdenu4 (approved):
Hurray!

PR #371, opened 2020-11-06
Bugfix/125 source dist includes too much

### LINKED ISSUE #125: Python source distribution includes more than it should
The python distribution (`python setup.py sdist`) for InVEST 3.8.2 includes files that are not strictly part of the python build itself (like the `installer` and `test` directories).  This isn't necessarily a bug, but it could be a problem down the road if we're inadvertently distributing a very large distribution for something that really should be pretty small (< 1 MB).

phargogh:
This PR fixes #125 by trimming a few directories from the python source distribution.


dcdenu4 (approved):
Looks good to me!

PR #373, opened 2020-11-09
Codesign the macOS binaries

### LINKED ISSUE #229: Sign the InVEST mac binaries
We've been signing our windows binaries with the Stanford-issued certificate for some time, and that's been pretty great.  It would be nice to also sign the mac binaries, since our users currently have to jump through a couple of hoops to get around the 'unidentified developer' warnings.

emlys:
# Code-signing the macOS distribution
Fixes #229. Whew! 

## Switch back to DMG distribution for macOS
As part of this PR, I propose switching our mac distribution from a zip file of an app bundle, to a zip file of a DMG which holds the app bundle. *(Does it still need to be zipped?)*
### pros:
- Code signing is a lot easier since there's only one item to sign
- IMO, it looks more "professional". Most software I've downloaded outside of the app store is in a dmg format.
- We can customize the dmg background
### cons:
- @phargogh told me that in the past, some users would not copy InVEST to their Applications directory, causing issues. I've tried to clarify this step by updating the background.
- Have to do the "right-click > open > open anyway" twice: when opening the DMG, and when opening InVEST the first time once it's copied to Applications/

@phargogh and I talked about improving the app bundle structure as well. I'm splitting that out into a separate issue, since now that we're using a DMG, it no longer affects the code signing.

## New Github secret
The new `make codesign_mac` command relies on a new Github secret, `KEYCHAIN_PASS`. As far as I know we can't get the login password of the actions user, so we can't use the default login keychain. Instead it makes a new keychain and sets the password to this known value.

## Leave HISTORY.rst and HTML docs out of the distribution
I propose leaving these out of the DMG distribution because:
1. They are readily available online, and at least for me, it's easier to access them there.
2. They can't go in the top level of the InVEST app bundle because it breaks the bundle structure. Putting them in a lower level would make them hard to find. Putting them in the DMG alongside the app bundle would work, but could clutter up the "drag and drop" background and confuse people.

As a result, the script `build_dmg.sh` is no longer complex enough to deserve being a script. The `dmgbuild` command now lives in the `Makefile`.
If we still want to provide the option to access docs locally without an internet connection, I propose that they be downloaded separately.

## Code signing certificate
I'm using the same certificate and password that we use for Windows code-signing. We can swap this out for the Apple Developer cert when that becomes available. 
Apple does not make it easy to build apps that meet their standards without using Xcode. There are little to no official docs that go into detail on this process. After much trouble with the build hanging due to `codesign` prompting for the keychain password, I came across the mysterious `security set-key-partition-list` option. Still not fully sure what it does, but it works, and I made a [post](https://developer.apple.com/forums/thread/666107) on the Apple Developer forums to get confirmation that this is correct and secure. 
Maybe once we have an Apple Developer account, we can open a support ticket and get some help to be sure that we are doing this in the best way.

@phargogh since you have a mac, would you mind testing that the DMG works on your computer as well? This would just be: download the DMG artifact, verify that the background looks as expected, verify that the app will run, and that it's signed: `codesign --verbose --verify <path to DMG>`


emlys:
Here's how the DMG looks on my screen. I added the text and made it larger and more colorful than the past version, so hopefully there won't be any more trouble with users skipping this step.
Screen Shot 2020-11-10 at 2 50 46 PM

.github/workflows/binary-applications.yml
```diff
@@ -218,7 +217,8 @@ jobs:
               shell: bash -l {0}
               run: |
                   make install
-                  make mac_zipfile
+                  # make mac_zipfile
```
dcdenu4, line 220: If the zip isn't need anymore, can probably do away with this commented line.

HISTORY.rst
```diff
@@ -61,6 +61,7 @@ Unreleased Changes (3.9)
       a small but noticeable amount of time in launching a model.
     * The number of files included in the python source distribution has been
       reduced to just those needed to install the python package and run tests.
+    * Code-sign the macOS distribution, and switch to a DMG distribution format.
```
dcdenu4, line 64: Might want to mention local doc changes if we decide they should not be bundled with dmg.

Makefile
```diff
@@ -143,7 +143,7 @@ help:
 	@echo "  userguide         to build HTML version of the users guide"
 	@echo "  python_packages   to build natcap.invest wheel and source distributions"
 	@echo "  windows_installer to build an NSIS installer for distribution"
-	@echo "  mac_installer     to build a disk image for distribution"
+	@echo "  mac_dmg     to build a disk image for distribution"
```
dcdenu4, line 146: Sorry about my OCD here, but could that line be indented to match the others?

dcdenu4 (commented):
Thanks for all this work and really nicely formatted PR @emlys ! I'm certain I'll have the least to say here since I'm a bit Mac illiterate. I did make some general comments for you to take a look at.

From my _brief_ google searching of Mac app packaging and sandboxing everything you lay out here seems to fall in line with what I saw. I think that given the outstanding issue of Mac sandboxing and local html docs, maybe this provides a window for addressing that in a better way for Mac. Maybe it makes sense to throw html docs down a level in `docs` or like you mention, maybe local docs should be optional and treated like invest sample data. 

I certainly think that your updated graphic for installing InVEST as an APP gets the point across. Again, I'm not a Mac user, but if this is for the most part normal for Mac apps, I think we can expect users to be "smart" to some extent.

Looking forward to Dave and James's feedback!

davemfish (approved):
Thanks for figuring all this out @emlys ! All the decisions you described up top sound very reasonable to me. I was going to ask what happens if someone clicks the "Model Documentation" link in the UI? But then I tried it and see we already offer a link to the online version if the local docs are unavailable!

davemfish:
> Here's how the DMG looks on my screen. I added the text and made it larger and more colorful than the past version, so hopefully there won't be any more trouble with users skipping this step.
> Screen Shot 2020-11-10 at 2 50 46 PM

Looks great! Not sure if it's just the screenshot or not, but that second invest logo in the lower right is pretty low-res and since it's also kinda redundant, maybe we should ditch it?

Makefile
```diff
@@ -332,23 +332,46 @@ $(WINDOWS_INSTALLER_FILE): $(INVEST_BINARIES_DIR) $(USERGUIDE_ZIP_FILE) build/vc
...
+	# add the certificate to the keychain
+	# -T option says that the codesign executable can access the keychain
+	# for some reason this alone is not enough, also need the following step
+	security import $(BUILD_DIR)/$(P12_FILE) -k '$(KEYCHAIN_NAME)' -P '$(CERT_KEY_PASS)' -T /usr/bin/codesign
+	# this is essential to avoid the UI password prompt
+	security set-key-partition-list -S apple-tool:,apple: -s -k '$(KEYCHAIN_PASS)' '$(KEYCHAIN_NAME)'
+	# sign the dmg using certificate that's looked up by unique identifier 'Stanford Univeristy'
+	codesign --verbose --sign 'Stanford University' $(MAC_DISK_IMAGE_FILE)
+	# relock the keychain (not sure if this is important?)
+	security lock-keychain '$(KEYCHAIN_NAME)'
```
phargogh, line 373: This all makes a lot of sense!  I don't know much about the mac keychains, but this all sounds consistent with the way they handle the keychain more generally.

Makefile
```diff
@@ -332,23 +332,46 @@ $(WINDOWS_INSTALLER_FILE): $(INVEST_BINARIES_DIR) $(USERGUIDE_ZIP_FILE) build/vc
...
+	# create a new keychain (so that we can know what the password is)
+	security create-keychain -p '$(KEYCHAIN_PASS)' '$(KEYCHAIN_NAME)'
+	# add the keychain to the search list so it can be found
+	security list-keychains -s '$(KEYCHAIN_NAME)'
+	# unlock the keychain so we can import to it (stays unlocked 5 minutes by default)
+	security unlock-keychain -p '$(KEYCHAIN_PASS)' '$(KEYCHAIN_NAME)'
+	# add the certificate to the keychain
+	# -T option says that the codesign executable can access the keychain
+	# for some reason this alone is not enough, also need the following step
+	security import $(BUILD_DIR)/$(P12_FILE) -k '$(KEYCHAIN_NAME)' -P '$(CERT_KEY_PASS)' -T /usr/bin/codesign
```
phargogh, line 367: Following up on our previous conversations about how secrets should be shared between processes (such as through environment variables), make will by default print out each step in the recipe:

Screenshot 2020-11-12 130750

GH Actions thankfully redacts the secrets, but should we be silencing these recipe steps as a precaution?

emlys:
> Looks great! Not sure if it's just the screenshot or not, but that second invest logo in the lower right is pretty low-res and since it's also kinda redundant, maybe we should ditch it?

Good point! I had just lazily increased the size of the old image, but you're right that it's redundant. Removed that and replaced it with a clear image of the natcap logo:
background

Makefile
emlys: Good catch!

phargogh (approved):
This looks great, @emlys, and congratulations on getting this all to work under github actions!

I had a couple of very minor comments sprinkled throughout the PR, and I'll respond to a few other things here

> Does the binary still need to be zipped?

Probably not ... although, is there a case where someone wouldn't be able to mount a DMG?

> Looks great! Not sure if it's just the screenshot or not, but that second invest logo in the lower right is pretty low-res and since it's also kinda redundant, maybe we should ditch it?

I'm seeing the same blurriness as well.  The clarity (visual and conceptual) of the text and arrow in this new version is a good upgrade and about as clear as we possibly can be.

> Have to do the "right-click > open > open anyway" twice: when opening the DMG, and when opening InVEST the first time once it's copied to Applications/

Yes, I'm seeing this behavior as well.  So once we have a developer ID from the university, we should be able to just swap it in here and have the verification work as expected?

### Verifying the DMG

```shell
(base) ➜  ~ codesign --verbose --verify /Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/
/Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/: valid on disk
/Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/: satisfies its Designated Requirement
```

Is that what you were expecting?

### Leave HISTORY.rst and HTML docs out of the distribution

I don't remember if there was a specific problem we were trying to solve by including the docs and changelog in with the mac binaries.  My best guess is that it was solely for greater consistency with the windows installer.  If we do want to include the documentation and changelog, we could put all of these artifacts into a single folder to then be dragged into `/Applications`.  Doesn't address the issue of being able to access the docs from the running application, of course.

All of that aside, I'd be just fine with not including these documents with the InVEST mac application since these resources can all be easily found online.

And looking forward to the Workbench, I think we've talked about including these resources (or at least the UG) within the application itself.   A common pattern I've seen in electron applications is to make the changelog and the application's docs available as a dialog or pane within the application, and that seems reasonable.

emlys:
> Probably not ... although, is there a case where someone wouldn't be able to mount a DMG?

I was more wondering about zipping the DMG itself? It seems kinda unnecessary to me. As far as I know the disk image format is standard and we can expect that people can use it.

> > Have to do the "right-click > open > open anyway" twice: when opening the DMG, and when opening InVEST the first time once it's copied to Applications/
> 
> Yes, I'm seeing this behavior as well. So once we have a developer ID from the university, we should be able to just swap it in here and have the verification work as expected?

Yep! * fingers crossed *

> ```shell
> (base) ➜  ~ codesign --verbose --verify /Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/
> /Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/: valid on disk
> /Users/jdouglass/Downloads/InVEST_3.8.7.post1153+g0fac9558.dmg/: satisfies its Designated Requirement
> ```
> 
> Is that what you were expecting?

Yes, that looks correct! Thanks for helping me double-check it.

> All of that aside, I'd be just fine with not including these documents with the InVEST mac application since these resources can all be easily found online.
> 
> And looking forward to the Workbench, I think we've talked about including these resources (or at least the UG) within the application itself. A common pattern I've seen in electron applications is to make the changelog and the application's docs available as a dialog or pane within the application, and that seems reasonable.

I agree that including them in the electron app would be ideal, and it seems okay to leave them out for the 1 (?) release that will happen before the workbench is ready.

Makefile
dcdenu4: I think throwing an `@` in front of the commands in the Makefile would suppress the echoing of the calls. Likewise from the yml file, `make -s` should run everything in silent mode. But yea, I don't see why we shouldn't hide this output.

dcdenu4 (commented):
Hey @emlys , I saw some tests failed so I'm giving them a re-run now. Other than that I just left a comment on something James mentioned about silencing the echo from the make calls. I see both James and Dave approved but I don't think that comment was addressed one way or the other.

Makefile
```diff
@@ -332,23 +332,46 @@ $(WINDOWS_INSTALLER_FILE): $(INVEST_BINARIES_DIR) $(USERGUIDE_ZIP_FILE) build/vc
 	-$(RM) $(WINDOWS_INSTALLER_FILE)
 	makensis /DVERSION=$(VERSION) /DBINDIR=$(INVEST_BINARIES_DIR) /DARCHITECTURE=$(PYTHON_ARCH) /DFORKNAME=$(INSTALLER_NAME_FORKUSER) /DDATA_LOCATION=$(DATA_BASE_URL) installer\windows\invest_installer.nsi
 
+DMG_CONFIG_FILE := installer/darwin/dmgconf.py
+mac_dmg: $(MAC_DISK_IMAGE_FILE) 
+$(MAC_DISK_IMAGE_FILE): $(DIST_DIR) $(MAC_APPLICATION_BUNDLE) $(USERGUIDE_HTML_DIR)
+	dmgbuild -Dinvestdir=$(MAC_APPLICATION_BUNDLE) -s $(DMG_CONFIG_FILE) "InVEST $(VERSION)" $(MAC_DISK_IMAGE_FILE)
```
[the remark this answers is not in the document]
emlys: Good idea!

Makefile
emlys: Yeah, something could always go wrong with the redacting. I like the `make -s` option!

Makefile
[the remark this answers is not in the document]
emlys: Sadly this doesn't seem to work: ` /usr/local/miniconda/envs/macbin-env/bin/python3: No module named dmgbuild.__main__; 'dmgbuild' is a package and cannot be directly executed`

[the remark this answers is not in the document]
phargogh: Looks like they haven't (yet) implemented `__main__`.  Oh well, thanks for considering!

Makefile
emlys: It does still print out some info about the certificate, but I don't think that's confidential? https://github.com/emlys/invest/runs/1397377958?check_suite_focus=true#step:10:17

dcdenu4 (approved):
Thanks @emlys , looks good to me!

.github/workflows/binary-applications.yml
```diff
@@ -122,7 +122,7 @@ jobs:
                         GSUTIL="gsutil" \
                         BIN_TO_SIGN="$INSTALLER_BINARY" \
                         SIGNTOOL="$SIGNTOOL_PATH" \
-                        signcode_windows
+                        -s signcode_windows
```
phargogh, line 125: 👍

phargogh (approved):
Looks great!

phargogh:
I just resolved a trivial merge in `HISTORY`, and everything else looks good to go so I'll merge this without waiting for the builds/tests to finish.

emlys:
Ok! I was just going to update the user's guide revision, but no worries - I can put it in a new PR along with the `make env` fix :)

PR #378, opened 2020-11-10
Create pull_request_template.md

emlys:
Adding a PR template with reminders to include issue number, update HISTORY.rst and user's guide.
I have seen some repos use extensive templates: https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/
That's probably overkill for us, but let me know if there's anything else that would be helpful for you to have here!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+# Description
+Fixes #
+
+# Checklist
+- [ ] Updated HISTORY.rst (if needed)
+
+- [ ] Updated the user's guide (if needed)


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+# Description
+Fixes #
+
+# Checklist
+- [ ] Updated HISTORY.rst (if needed)
```
phargogh, line 5: Since we have a convention of when to update the `HISTORY` (when we have a change that affects the user's experience of the application), I wonder if it'd be useful to note that here.   You could imagine the PR submitted yesterday wondering what 'if needed' is, and different projects may have different approaches.

phargogh (commented):
I'd be interested to hear what everyone thinks about this template, but this seems like a good starting place for us and we can always iterate on it as needs evolve.  My one comment is just about what 'if needed' means, should someone without experience in this project take a look at the checklist, but it's more of a question than a request for a change.

davemfish (approved):
I agree with James that the History note might as well say "if these changes are user-facing" or something like that.

I think this a great idea!

dcdenu4 (approved):
This seems great to me. The HISTORY.rst update James suggests sounds fine too. But this definitely should be geared to help us as opposed to 3rd parties I think, since it's so rare for someone outside our group to do PRs. Yes I know I made the comment in Slack about helping 3rd parties with PRs haha.

emlys:
Agree that clarifying the HISTORY note would be good! I accidentally branched this off of natcap/invest rather than emlys/invest, and it won't let me add more commits to the branch... sigh

PR #382, opened 2020-11-13
Update gha setup miniconda version

### LINKED ISSUE #346: Github actions warnings about add-path
In looking through a few recent GH actions runs associated with a PR ([example actions run](https://github.com/natcap/invest/actions/runs/298225931)), there are a pile of annotations/warnings about `add-path` being deprecated and replaced by environment files.  We should look into this and update accordingly.

Screenshot 2020-10-16 114952

dcdenu4:
# Description
Update github action versions for setup-miniconda to patch the warning / security issue for `add-path` being deprecated. Also bumped the github actions cache.

Fixes #346

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -42,7 +42,7 @@ jobs:
               run: git fetch origin +refs/tags/*:refs/tags/*
 
             - name: Restore pip cache
-              uses: actions/cache@v1
+              uses: actions/cache@v2
               with:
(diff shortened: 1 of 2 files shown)


emlys (approved):

emlys:
Just realized that maybe this should have gone into `main`, not `release/3.9`... without these changes, the workflows on `main` will probably break starting on Monday. But if we're going to do the release next week anyway, it's not a big deal

PR #389, opened 2020-11-17
Patching set union/iteration in CBC accumulation. RE:#209

phargogh:
# Description
This PR addresses an issue that came up in https://community.naturalcapitalproject.org/t/coastal-blue-carbon-negative-carbon-stocks/780/12 where, under specific conditions, the accumulation, sequestration and valuation rasters would be empty.  Because the model hasn't been released, I just piggybacked on #209 .  No updates to HISTORY, UG needed for this.

### DIFF
--- a/src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
+++ b/src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
@@ -663,7 +663,7 @@ def execute(args):
     total_accumulation_tasks = {}
     prior_transition_year = baseline_lulc_year
     for current_transition_year in sorted(
-            transition_years.union(set([analysis_year]))):
+            transition_years.union([analysis_year])):
         total_accumulation_rasters[current_transition_year] = os.path.join(
             output_dir, ACCUMULATION_SINCE_TRANSITION_RASTER_PATTERN.format(
                 start_year=prior_transition_year,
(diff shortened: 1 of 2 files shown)


src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -662,7 +662,8 @@ def execute(args):
     total_accumulation_rasters = {}  # Across all 3 pools
     total_accumulation_tasks = {}
     prior_transition_year = baseline_lulc_year
-    for current_transition_year in sorted(transition_years) + [analysis_year]:
+    for current_transition_year in sorted(
+            transition_years.union(set([analysis_year]))):
```
emlys, line 666: `transition_years.add(analysis_year)` would be slightly more succinct, but this is also just fine!

emlys (approved):
Looks good!

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
phargogh: In terms of making it more succinct, I hadn't realized that the extra `set()` isn't needed in order to take the union of two sets, which I've patched in 3be2bb4c.  It should at least be a bit more readable for the creation of a one-off set for use just in this one case.

I wish I could just use `add` here!  That would be lovely ... but as it stands, `set.add` modifies the set but doesn't return the new set.

Let me know what you think @emlys !

emlys: Oh, you're right about `set.add`! This looks great.

Issue #18, opened 2020-02-13
Fisheries-HST FloatingPointError: invalid value encountered in true_divide 


phargogh:
This sometimes causes our tests to fail (https://ci.appveyor.com/project/davemfish/invest/builds/28437818/job/y1psai9ejk8h2htg)

It's clear what the problem is, but unclear what the correct scientific solution is.

fisheries_hst.py line 144 divides by array n_a, which is populated by a user-input matrix of 1s and 0s (see lines 109, 110 for n_a, and trace it back to the user input in fisheries_hst_io.py)

fisheries_hst.pyline 110 is especially suspicious: n_a[n_a == 0] = 0

Originally reported by @davemfish at https://bitbucket.org/natcap/invest/issues/3925/fisheries-hst-floatingpointerror-invalid

PR #391, opened 2020-11-17
Address deprecation warnings when running tests

### LINKED ISSUE #255: Resolve deprecation warnings in tests
When running `make test` or `make test_ui`, pytest displays a bunch of deprecation warnings and some other warnings near the end of the test output. Nosetests probably wasn't catching these because it hasn't been updated in a long time. We should go through and resolve each issue causing a warning to prevent future problems. 

Most of them are pretty straightforward e.g. replacing `warn` with `warning`.

emlys:
# Description
Fixes #255 
* Replace one instance of `LOGGER.warn` with `LOGGER.warning`
* Remove deprecated `U` option of `open` function. From the [python docs](https://docs.python.org/3/library/functions.html#open): 
    > There is an additional mode character permitted, 'U', which no longer has any effect, and is considered deprecated. It previously enabled universal newlines in text mode, which became the default behaviour in Python 3.0. "
* Replace `imp` with `importlib`, see [this post](https://stackoverflow.com/questions/19009932/import-arbitrary-python-source-file-python-3-3) for explanation
* Replace `locale.format` with `locale.format_string`
* Replace unittest `assertEquals` with `assertEqual`
* Explicitly set `rcond=-1` for `numpy.linalg.lstsq`. From the [docs](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html):
    > If not set, a FutureWarning is given. The previous default of -1 will use the machine precision as rcond parameter, the new default will use the machine precision times max(M, N). To silence the warning and use the new default, use rcond=None, to keep using the old behavior, use rcond=-1.
* Replace numpy matrix with array in fisheries
* Resolve weird numpy dtype metadata warning. Discussed this with @richpsharp . See [here](https://gist.github.com/emlys/49b2102203743581370d365bf3e832cd) for a minimal example and [here](https://gist.github.com/richpsharp/04dae344eb7f165534f51cafba6f6608) for the solution he gave me. These changes are copied from that with modification to allow the `'STOP'` strings as well as numpy arrays.

When I run `make test`, the only outstanding warning relates to #18 .
There are still 2 warnings when I run `make test_ui` that come from Qt. I'm ignoring those since we have the workbench coming soon.

No changes to HISTORY or the user's guide are needed.

### DIFF
--- a/src/natcap/invest/recreation/recmodel_server.py
+++ b/src/natcap/invest/recreation/recmodel_server.py
@@ -59,18 +59,14 @@ def _numpy_dumps(numpy_array):
 
 
 def _numpy_loads(queue_string):
-    """Safely unpickle numpy array to string.
-    Items in our queue may either be a pickled numpy array string, or the
-    string 'STOP'. If 'STOP', don't try to load as a numpy array.
+    """Safely unpickle string to numpy array.
+    
     Args:
         queue_string (str): binary string representing a pickled
             numpy array.
     Returns:
         A numpy representation of ``binary_numpy_string``.
     """
-    if queue_string == 'STOP':
-        return queue_string
-
     with BytesIO(queue_string) as file_stream:
         return numpy.load(file_stream)
 
@@ -622,14 +618,15 @@ def construct_userday_quadtree(
         n_points = 0
 
         while True:
-            # if the queue item is 'STOP', this returns 'STOP', not an array.
-            point_array = _numpy_loads(numpy_array_queue.get())
-            if (isinstance(point_array, str) and
-                    point_array == 'STOP'):  # count 'n cpu' STOPs
+            payload = numpy_array_queue.get()
+            # if the item is a 'STOP' sentinel, don't load as an array
+            if payload == 'STOP':
                 n_parse_processes -= 1
                 if n_parse_processes == 0:
                     break
                 continue
+            else:
+                point_array = _numpy_loads(payload)
 
             n_points += len(point_array)
             ooc_qt.add_points(point_array, 0, point_array.size)
--- a/src/natcap/invest/recreation/recmodel_server.py
+++ b/src/natcap/invest/recreation/recmodel_server.py
@@ -13,7 +13,7 @@ import threading
 import collections
 import logging
 import queue
-from io import StringIO
+from io import BytesIO, StringIO
 
 import Pyro4
 import numpy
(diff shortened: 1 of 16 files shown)


src/natcap/invest/recreation/recmodel_server.py
```diff
@@ -589,7 +622,8 @@ def _populate_offset_queue(block_offset_size_queue):
         n_points = 0
 
         while True:
-            point_array = numpy_array_queue.get()
+            # if the queue item is 'STOP', this returns 'STOP', not an array.
+            point_array = _numpy_loads(numpy_array_queue.get())
```
richpsharp, line 626: If you don't mind, I prefer a little more explicit check when it comes to having sentinels in a queue before trying to process the result. Something like this:

```
payload = numpy_array_queue.get()
if payload == 'STOP':
    n_parse_processes -= 1
    if n_parse_processes == 0:
        break
    continue
else:
    point_array = _numpy_loads(payload)
    ....
```

Then you won't have to have `_numpy_loads` doing double duty by also having to know the context in which the data come in, you can just have it load blindly.

src/natcap/invest/recreation/recmodel_server.py
```diff
@@ -45,6 +45,36 @@
...
+    Items in our queue may either be a pickled numpy array string, or the
+    string 'STOP'. If 'STOP', don't try to load as a numpy array.
+    Args:
+        queue_string (str): binary string representing a pickled
+            numpy array.
+    Returns:
+        A numpy representation of ``binary_numpy_string``.
+    """
+    if queue_string == 'STOP':
+        return queue_string
```
richpsharp, line 72: I mention below that I think it's a cleaner solution to explicitly check the sentinel before trying to decode it, if you agree, you can simplify this part.

richpsharp (changes requested):
Hi @emlys, this looks good. I have one comment that I'd prefer the sentinel handling in the context in which it occurs rather than a passthrough function, see what you think?

src/natcap/invest/recreation/recmodel_server.py
emlys: Agree it is cleaner to keep this out of the `_numpy_loads` function!

src/natcap/invest/recreation/recmodel_server.py
emlys: You're absolutely right! not sure why I did it that way

emlys:
FYI we just added the readthedocs build to the PR checks and it's expected that it's going to fail for now

richpsharp (approved):

richpsharp:
Looks great!

PR #392, opened 2020-11-20
Make InVEST run on Big Sur

### LINKED ISSUE #384: InVEST 3.9.0 development build hangs on Mac OS X Big Sur
I'm using a recent development build of InVEST 3.9.0 (revision 7bbbec2dc120f8662361a34efb7ea58942ecf94d, on the mac codesigning branch), and launching InVEST through the mac binaries causes the application to freeze and use up a ton of CPU power while doing so.  The application hangs before any window is presented or any logging is created.sy

So far, I've tried running these binaries in these ways with the same results:

1. By double-clicking on the application (and bypassing the 'unregognized developer' warnings)
2. By running the application from a terminal:
    a. `./InVEST.app/Contents/MacOS/InVEST`
    b. `./InVEST.app/Contents/MacOS/invest_dist/invest --debug launch`

I took a sample of the process through activity monitor and attached it here.  Best guess at the moment is that it's stuck in a PyQt event loop.  [Sample of invest.txt](https://github.com/natcap/invest/files/5538932/Sample.of.invest.txt)

phargogh:
# Description

The fix to this issue came down to an issue in Qt that's addressed by the new environment variable `QT_MAC_WANTS_LAYER` that's defined for the binaries on mac only.  I modified a test by extending the timeout on one of the functions to ensure that the test would behave as expected (it was failing on Windows/python3.8).

I've tested the fix locally on my computer.

Fixes #384

# Checklist


phargogh:
https://github.com/natcap/invest/pull/392/commits/c084f83609fd05f6557932bb3e9d67ac170f259e allows the _build_ to pass, but the .app won't actually launch on Big Sur.  Still looking into this.

phargogh:
Alright, I've downgraded the build environment to 10.15 again and I'm hoping the `hdiutil` issue is just a fluke that has been resolved.  I've confirmed that the binaries work on Big Sur.  @emlys do you still have Catalina, and, if so, could you try running the binaries on your computer and see if they work as expected?

I also added the InVEST license to the dmgbuild configuration for good measure.

tests/test_validation.py
```diff
@@ -732,16 +732,16 @@ def test_slow_to_open(self):
         # define a side effect for the mock that will sleep
         # for longer than the allowed timeout
         def delay(*args, **kwargs):
-            time.sleep(6)
+            time.sleep(7)
             return []
 
         # make a copy of the real _VALIDATION_FUNCS and override the CSV function
-        mock_validation_funcs = {key: val for key, val in validation._VALIDATION_FUNCS.items()}
+        mock_validation_funcs = validation._VALIDATION_FUNCS.copy()
```
emlys, line 739: Thanks for cleaning this up!

installer/darwin/dmgconf.py
```diff
@@ -31,4 +31,5 @@ def get_size(start_path = '.'):
 #background = 'builtin-arrow'
 default_view = 'icon-view'
 
-
+format = 'UDZO'
+license = {'licenses': {'en_US': 'LICENSE.txt'}}
```
emlys, line 35: Clearly it's not a serious issue, but the [dmgbuild docs](https://dmgbuild.readthedocs.io/en/latest/settings.html#license-settings) say that a `'default-language'` key is also required in the `license` dictionary.

Also, is it okay that `LICENSE.txt` is relative to the top-level and not installer/darwin/dmgconf.py ?

emlys (commented):
Looks good! The PR doesn't trigger building the dmg, and it looks like the most recent actions run on your branch failed with that flaky dmgbuild issue. Could you trigger it again or send me the dmg to test?

installer/darwin/dmgconf.py
phargogh: > Clearly it's not a serious issue, but the dmgbuild docs say that a 'default-language' key is also required in the license dictionary.

Ah, good catch!  Since it is required, patched this in e96ed0c0.

> Also, is it okay that LICENSE.txt is relative to the top-level and not installer/darwin/dmgconf.py ?

Based on what's in the rest of `dmgconf.py`, I'd be inclined to think so ... but just in case, I've modified the configuration program to use the file's path for the DMG badge icon and the license file in d849a071.

phargogh: Just following up on this, turns out that `__file__` is not available within the `dmgconf.py` so I've updated the filepath variable to clarify intent and added a few comments to clarify that this is all assumed to be executed from the project root.  Patched in da8296a1.

As for a DMG, the latest binary run is still testing, but here's the actions build from e96ed0c that should allow you to download and use the dmg artifact directly from github actions: https://github.com/natcap/invest/runs/1476979402  That'll be the best thing to test!

phargogh:
Thanks again for taking a look at this @emlys !  Is there anything we're waiting on for this PR?

emlys (approved):

PR #398, opened 2020-11-30
Change UFRM field names from using dot to underscore

### LINKED ISSUE #397: Attribute field from UFRM vector output sometimes doesn't load in ArcGIS
A user was having trouble opening up the attribute table for the vector output `flood_risk_service.shp` in ArcGIS. It would load fine in QGIS (how I viewed it), but in ArcGIS would give a loading table data error message. It could have to do with the `.` in the field names.

Rich also commented:
>it's weird i would have put that precision at 36. i don't remember explicitly doing that. if you're willing to try, could you bump it down to 24 and see if that work for them?
>could also try converting it to a GPKG and seeing if they can load that? I remember I originally had GPKGs but Perrine was worried folks might not be able to load it, but as you know, esri shapefiles are pretty nasty in their cross implmeentations.

Test data for this issue is here: 
https://drive.google.com/drive/folders/11pkojxRqRHBsbSdHGWhXcX3hTrHs4KW5?usp=sharing

Forum reference here:
https://community.naturalcapitalproject.org/t/urban-flood-risk-mitigation-error/1480/10

dcdenu4:
# Description
Fixes #397 

This is a simple PR that fixes an issue where dots in the field names of the output vector would cause ArcGIS to fail when looking at the attribute table.

# Checklist
- [x] Linted corresponding files

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -141,6 +141,10 @@ Unreleased Changes (3.9)
 * Urban Cooling
     * Split energy savings valuation and work productivity valuation into 
       separate UI options.
+* Urban Flood Risk
+    * Changed output field names ``aff.bld`` and ``serv.blt`` to ``aff_bld``
+      and ``serv_blt`` respectively to fix an issue where ArcGIS would not 
+      display properly.
 
 ..
 ..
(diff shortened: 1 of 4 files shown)


emlys (approved):
Looks good! Weird that this hasn't come up before now

PR #399, opened 2020-11-30
Tracking release/workbench-alpha branch to support the workbench

### LINKED ISSUE #341: Proposal: Include the python side of the workbench in this repo
The Workbench is being developed in github.com/natcap/invest-workbench. It has some python dependencies including a small Flask app and `natcap.invest`. 

The workbench build process duplicates a lot of the invest build process, as it needs to package up frozen binaries for those python packages & dependencies. Those build processes (PyInstaller scripts & Actions workflows) are kind of high-maintenance in general - dealing with different OS, python & numpy versions, etc - so it's important to not duplicate them. 

### Proposal:
**Add a `release/workbench-alpha` branch to this repo and make the following changes:**
* Add the Workbench's `server.py` flask app
* Add tests for `server.py` - now they are in the workbench at `tests/flaskapp.test.js` but if they are to live in natcap/invest the tests should be refactored to python.
* Add a CLI endpoint like `invest serve` to launch the flask app
* ~~Add a section to the existing PyInstaller `invest.spec` to include `server.py` as a second entry-point~~
* Add `Flask` to `requirements-gui.txt` ~~(and remove the Qt reqs)~~
* rename `server.py` since it's not a server, it's a flask app that runs on a server
* Add or amend an Actions workflow to upload the PyInstaller output to a bucket where it can be retrieved by the Workbench's build process

The benefits are 1) less duplication and 2) much quicker workbench builds. 

The downside is that changes to the flask app during workbench development will have to be made in a different project from the rest of the workbench. Though it is possible to test local flask app changes in the workbench without running PyInstaller. Also, the flask app doesn't change too much. Even while the workbench is still very unstable in it's pre-alpha state, the `server.py` module rarely changes.

davemfish:
I just created the new branch `release/workbench-alpha` - branched from `release/3.9`. This PR includes the changes needed to support the python side of the invest-workbench. Those changes are listed in #341 and amount to a small flask app that now lives under ~~`natcap.invest.web_api`~~ `natcap.invest.ui_server` (thoughts on this name?) and a new `invest serve` CLI entry-point to start the flask dev server and run the app.

There is a test here to cover the new `invest serve` subcommand. It imports the new module `natcap.invest.ui_server`, but so far there are no tests here to cover the flask app's endpoints. We do have tests for those endpoints in the workbench's test suite. And right now the workbench is the only use-case for the `natcap.invest.ui_server` module. I'm interested to hear opinions about whether those tests should live here in this repo instead.

Fixes #341 

# Checklist
- [x] Updated HISTORY.rst  -  Technically this is a user-facing change, with the new CLI entry-point. But we don't exactly expect anyone to use that except us for the workbench. Thoughts?


src/natcap/invest/cli.py
```diff
@@ -288,7 +288,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 
 
 def main(user_args=None):
-    """CLI entry point for launching InVEST runs.
+    """CLI entry point for launching InVEST runs and other useful utilities.
```
dcdenu4, line 291: Nice

src/natcap/invest/web_api.py
```diff
@@ -0,0 +1,211 @@
+"""A Flask app with HTTP endpoints used by the InVEST Workbench."""
```
dcdenu4, line 1: I'm wondering if the name `web_api` is misleading... What about `ui_server.py`?

dcdenu4 (commented):
Hey @davemfish just to echo my thoughts from the workbench PR, I think this makes sense overall. Here's a doozy of a question that I'm sure you and James have thought and talked about in the past, but is there any case or argument for having the `workbench` alongside `invest` in `invest/src/natcap`? I'm just thinking about this back and forth and wondering if it's a sign that they shouldn't be separated? I haven't spent too much time thinking deeply about this, but thought I'd throw it out there!

I do think that `web_api` might be better as `ui_server.py` or something more representing what it is and is doing...?

I think it's fine to leave the tests over in the `workbench` repo. I feel like the Flask App is fairly straightforward and has no impact on OG `invest`. So I don't think running tests on this side of the puzzle makes a lot of sense if the only user isn't in this repo as well... Those are some fast and loose thoughts.

Looking forward to talking about this more!

src/natcap/invest/cli.py
davemfish: Some of finest docs work, right there. Haha, I didn't feel like trying to duplicate all the `parser` helptext that follows and explains all the other entry-points.

src/natcap/invest/web_api.py
davemfish: I dunno? I thought `server` was kind of inaccurate since this module defines the *app* that runs on the *server*, right? I like having `ui` in the name though. Hopefully we get some more opinions here!

phargogh (commented):
Makes sense to me!

> Updated HISTORY.rst - Technically this is a user-facing change, with the new CLI entry-point. But we don't exactly expect anyone to use that except us for the workbench. Thoughts?

What if we added this to `HISTORY`, but noted that it's likely to change?

> I think it's fine to leave the tests over in the workbench repo. I feel like the Flask App is fairly straightforward and has no impact on OG invest. So I don't think running tests on this side of the puzzle makes a lot of sense if the only user isn't in this repo as well... Those are some fast and loose thoughts.

I agree with Doug, seems completely fine to keep the tests for the interface on the workbench side of things until we have a need to use the flask interface from somewhere within this repo.

> Here's a doozy of a question that I'm sure you and James have thought and talked about in the past, but is there any case or argument for having the workbench alongside invest in invest/src/natcap? I'm just thinking about this back and forth and wondering if it's a sign that they shouldn't be separated? 

It's a great question!  @davemfish please correct me here if I'm missing anything or misremembering ... my recollection was that it made sense to develop these as separate repositories because they were pretty functionally distinct (though there is overlap), and it didn't seem to make sense to impose all of the same development rules that we have here in the InVEST repository on a technology stack we were still figuring out.  Of course InVEST is a part of the workbench (and the workbench will need to run InVEST), but it's also quite possible that the set of features of the workbench will eventually include more than just running InVEST models.

@dcdenu4  you bring up a great point about the back-and-forth here as a possible sign that they shouldn't be separated, and if the two do indeed end up being tightly coupled, then maybe they should be in the same repo!  Another possible way that we could ensure compatibility between the two layers is careful versioning of the HTTP interface.  In any case, there's always bound to be a little more iteration as we're learning more about how to build an application like this, but we should absolutely discuss this further and revisit this question as needed.

src/natcap/invest/web_api.py
dcdenu4: Naming things is always the hardest part! I guess I shied away from `web` because we're only running this locally and the `web` is not really involved. `http_api` seems fine to me. I don't care too deeply, so I'll be fine with whatever you choose!

davemfish:
> > Here's a doozy of a question that I'm sure you and James have thought and talked about in the past, but is there any case or argument for having the workbench alongside invest in invest/src/natcap? I'm just thinking about this back and forth and wondering if it's a sign that they shouldn't be separated?
> 
> It's a great question! @davemfish please correct me here if I'm missing anything or misremembering ... my recollection was that it made sense to develop these as separate repositories because they were pretty functionally distinct (though there is overlap), and it didn't seem to make sense to impose all of the same development rules that we have here in the InVEST repository on a technology stack we were still figuring out. Of course InVEST is a part of the workbench (and the workbench will need to run InVEST), but it's also quite possible that the set of features of the workbench will eventually include more than just running InVEST models.
> 
> @dcdenu4 you bring up a great point about the back-and-forth here as a possible sign that they shouldn't be separated, and if the two do indeed end up being tightly coupled, then maybe they should be in the same repo! Another possible way that we could ensure compatibility between the two layers is careful versioning of the HTTP interface. In any case, there's always bound to be a little more iteration as we're learning more about how to build an application like this, but we should absolutely discuss this further and revisit this question as needed.

I don't think we really ever considered a single repo, but I think it's worth considering it now and going forward.  At the start, certainly it made sense to do the prototyping & early iterations in a separate repo. And James makes good points about dev rules & technology stack being quite different. 

I do hope and expect that the development of each project will happen mostly in isolation with respect to the other, and it will be relatively rare for a developer to need to make changes to both repos at the same time in order to advance a single feature. Of course, that scenario _is_ playing out right now in PRs (this one & also the one Doug is working on for port configuration) that change the interface between invest & the workbench. But I'm hopeful that once we get that interface right, it will be rare to need to coordinate PRs across repos. 

But if I'm wrong and that continues to be a source of friction in the future, then I think we could consider combining projects.

src/natcap/invest/web_api.py
emlys: I like `http` over `web`. Maybe `ui_api` since the relevant info is that it's entrypoints for the UI, not how it's implemented (http)? But `http_api` seems fine too

emlys:
> Here's a doozy of a question that I'm sure you and James have thought and talked about in the past, but is there any case or argument for having the workbench alongside invest in invest/src/natcap? I'm just thinking about this back and forth and wondering if it's a sign that they shouldn't be separated?

I like the separate repos because it reinforces for me thinking about them as completely separate components that only interact through the API. If we can specify an exact invest version with a commit hash, like we do for the user's guide and data, I think that should be enough. Once we have a first release of the workbench I'd expect the workflow to be similar to how we update the user's guide and data: we only need to update the commit hashes when there's a relevant feature changed. To me it's worth the extra bit of upkeep in order to maintain invest as a totally separate component.

src/natcap/invest/web_api.py
davemfish: Thanks for all the input, I renamed it to `ui_server`

emlys (approved):
Looks good to me! I tried out the new `invest serve` command. Love how you can exit it with ctrl+C 😀 My only comment is about the output of `serve`:
```
$ invest serve
 * Serving Flask app "natcap.invest.ui_server" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:56789/ (Press CTRL+C to quit)
 ```
 What is the 'Environment' setting and if this is a development server, should we change it to say 'Development'?

**Comment by davemfish** [comment]
> Looks good to me! I tried out the new `invest serve` command. Love how you can exit it with ctrl+C 😀 My only comment is about the output of `serve`:
> 
> ```
> $ invest serve
>  * Serving Flask app "natcap.invest.ui_server" (lazy loading)
>  * Environment: production
>    WARNING: This is a development server. Do not use it in a production deployment.
>    Use a production WSGI server instead.
>  * Debug mode: off
>  * Running on http://127.0.0.1:56789/ (Press CTRL+C to quit)
> ```
> 
> What is the 'Environment' setting and if this is a development server, should we change it to say 'Development'?

Hey @emlys , thanks for bringing this up. By default, flask starts up in production mode. I haven't read about all the implications of that, but one implication is the subsequent WARNING about using the dev server in production. So the options seem to be:

1. leave as-is on the grounds that it's appropriate to warn the user that ran `invest serve`
2. use a production WSGI server. If this were a "real" webserver that needed to handle concurrent client requests, we would for sure need to do this. And maybe it's a good idea anyway? 
3. leave as-is, but on the workbench side we can launch `invest serve` after setting an env variable to set "Environment: development" as you suggest. That should suppress the WARNING.

File: src/natcap/invest/cli.py [hunk 3]
```diff
@@ -391,6 +391,12 @@ def main(user_args=None):
         help=('The model for which the spec should be fetched.  Use "invest '
               'list" to list the available models.'))
 
+    serve_subparser = subparsers.add_parser(
+        'serve', help=('Start the flask app on the localhost.'))
+    serve_subparser.add_argument(
+        '--port', type=int, default=56789,
+        help='Port number for the Flask server')
```
dcdenu4, line 398: Nice!

dcdenu4 (commented):
Hey @davemfish this looks good to me. I'd be fine with punting the production vs development issue Emily brought up into a separate issue. Off the top of my head I'd say number 1 or 3 would be the way to go.

Also, did you add a comment to History? I know James had mentioned maybe adding something.

emlys:
I agree it's fine to go in a separate issue

davemfish:
Okay we have passing builds now! Well, except for Mac. 

Since you all last looked at this, Doug & I added the optional subargument to specify a port when starting the flask app. e.g. `invest serve --port 54321`. And there's a new test for that.

Also brought this branch up-to-date with `main`.

requirements.txt
```diff
@@ -14,13 +14,13 @@
 
 GDAL>=3.1.2
 Pyro4==4.77  # pip-only
-pandas>=1.0
+pandas>=1.0,<1.2.0
 numpy>=1.11.0,!=1.16.0
-Rtree>=0.8.2,!=0.9.1
+Rtree>=0.8.2,!=0.9.1,<=0.9.4  # rtree/pyinstaller compatibility
```
dcdenu4, line 19: Thanks for the comment!

dcdenu4 (approved):
Thanks Dave!

phargogh (approved):
Thanks, Dave!

I agree with the other comments about discussing the use of a webserver in a separate thread as well.

phargogh:
Looks like I'm the last one to submit an approval and all of the builds are passing (except RTD, which is currently an expected failure), so I'll merge this.

Issue #402, opened 2020-12-07
InVEST 3.9 Release Issues


dcdenu4:
I did a sweep of the InVEST 3.9 pre release and came across the following issues when running each model with the sample data and examining the output.

- [x] **Forest Carbon EE** has an `OverflowWarning` at some point in the log output. **Emily**
- [x] **Forest Carbon EE** raster output looks bugged where `nodata` should be?
- [x] **DeliniateIt** Stacie has some comments in a Slack post that might be worth addressing.
- [x] **DeliniateIt** has output watershed that might be overlapping or multipolygon? Is this right? **Emily**
- [x] **Forest Carbon EE** local doc link is broken. 
- [x] **SDR** Stacie says she got a local docs error from this model, that I did not get.
- [x] **Validation** `validation.timeout` messages are more appropriate as `debug` than `info` level 
- [x] ~~**HQ/CBC** should these models have a note in Data Needs about the updated inputs? Should they indicate 3.9+?~~ I think we agree that this isn't necessary since the UG is versioned along with invest (dave)
- [x] **InVEST** is there a way to have the Windows items be alphabetical and CBC grouped? ( Dave? )
- [x] **HQ** add clarification to UG around Threats and Sensitivity table ( Doug )
- [x] **Wave Energy** remove 3.0 Beta comment in UG.
- [x] **DeliniateIt** has has number indentation issue in UG. **Emily**
- [x] **Crop Production Reg** `soybean reg prod` raster looks bugged with values `-1.79e308` to `1.79e308`
- [x] **Coastal Vulnerability** `population` column / field has values `-1.3106e+39`, is that right?
- [x] **SDR** has output raster with small `sed_dep` value... might be worth checking
- [x] **SWY** has output raster with `-1.05e-05` value... might be worth checking
- [x] **CV, DelineateIt, HRA, Urban Flood** all seem to suffer from `VirtualXpath` dumps during validation in console output. **Emily**
- [x] **GLOBIO** has outputs that look like artifacts or streaking that could be caused by a bug
- [x] **Scenic Quality** output looks not quite right, but I don't know what it should look like?  **James**

If someone would like to look into any of these issues, feel free to edit this with your name at the end of the line.

emlys:
[...]

PR #404, opened 2020-12-07
Pin GDAL to 3.1.2

### LINKED ISSUE #390: pytest output has strange XPath lines
Observed on macOS, pytest version 6.1.2 (latest), python 3.7.9.
Pytest output includes a bunch of repetitive lines:
```
tests/test_hra.py::HraUnitTests::test_missing_index_from_criteria_csv PASSED
tests/test_hra.py::HraUnitTests::test_simplify_geometry 	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
PASSED
tests/test_hra.py::HraUnitTests::test_simplify_geometry_lines 	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
	- 'VirtualXPath'	[XML Path Language - XPath]
PASSED
```
I'm pretty sure these tests used to print out more useful information. This could be a pytest issue, though I don't see it mentioned on their repo.

emlys:
# Description
 #390 , #402 

The `VirtualXPath` lines only happen with GDAL>=3.1.3. For the release, we can just pin the version. I don't see any related issue on the GDAL github so I'll look into it further and open an issue there if appropriate

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-GDAL>=3.1.2
+GDAL==3.1.2
 Pyro4==4.77  # pip-only
 pandas>=1.0
 numpy>=1.11.0,!=1.16.0


dcdenu4 (commented):
I'm just re-running the checks since windows binary failed for some reason.

dcdenu4 (commented):
Hey @emlys . I was just thinking about this some more. Since this issue is already fixed in a forthcoming 3rd party library, ultimately is not a bug in InVEST,  doesn't effect the models other than scrolling output in a background console, and doesn't spam the logfile, should we maybe just not do anything?

Sorry, that was a lot of ANDs haha.

emlys:
@dcdenu4 I agree it's not critical and not really our issue. It seemed like it was annoying people but I'm good either way!

emlys:
@phargogh can we get your opinion on this? Talking about it just now, we're all kind of ambivalent about whether or not to pin the version.
Pros:
* Avoid annoying output for users and ourselves
* It's technically a bug
Cons:
* It's not our bug
* The annoying output is really insignificant to users (goes in the background console)
* Small chance that someone can only use GDAL==3.1.3 and thus can't use InVEST

davemfish:
It occurred to me that another option I don't think we discussed is to pin the version for the release, but then immediately unpin it when we merge the release back into `main`. That way we avoid distributing a release with the annoying console noise, but developers & python users never really have to feel restricted (assuming they're more likely to install from the head of `main` than the release branch/tag).

davemfish:
And we do occasionally ask users to copy/paste their console output when they're having some other issue with the UI that we're trying to troubleshoot.

dcdenu4 (approved):
Hey @emlys @davemfish . I've come around to thinking it is fine to pin this to 3.1.2 for now. Our main community of users is through the UI application and I'd hate to see them be confused with this background noise and post questions about it to the forums. So, I like keeping this bug out of the 3.9.0 release of InVEST. Thanks for entertaining my earlier counter arguments!

PR #405, opened 2020-12-07
masking nodata values needs to use a numpy.isclose equality check

davemfish:
A couple array masking operations were not properly masking nodata values because they weren't using `numpy.isclose` to compare floats.

This addresses one of the items in #402 

# Checklist
- [ x] Updated HISTORY.rst (if these changes are user-facing) - I'm pretty sure we never released a version with this bug.

- [ x] Updated the user's guide (if needed)


dcdenu4 (approved):
Looks good! Thanks @davemfish!

PR #408, opened 2020-12-08
Resolve Forest Carbon overflow error

### LINKED ISSUE #413: Forest Carbon overflow error
When running Forest Carbon with the sample data, we get an overflow error in the logs:
`forest_carbon_edge_effect.py:832: RuntimeWarning: overflow encountered in exp
  thetas[:, :, 0] - thetas[:, :, 1] * numpy.exp(`

This comes down to all 3 regression models being calculated on all points even though only one is used. Large fake values for theta3 resulted in overflow. Luckily the resulting overflow values weren't used, so this bug wasn't affecting the output.

emlys:
# Description
#402 
Fixes #413 
This resolves the overflow errors seen in the logs when running with the sample data. All 3 regression functions were being calculated for all points, even though only 1 per point is used. Some points had large fake values for theta 3, leading to an overflow error.

I simplified the array manipulation a bit, but I'm not super confident about it so a thorough review is appreciated!

# Checklist

### DIFF
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -831,15 +831,13 @@ def _calculate_tropical_forest_edge_carbon_map(
 
         # For each forest pixel x, for each of its k nearest neighbors, the 
         # chosen regression method (1, 2, or 3). model_index shape: (x, k)
-        model_index = numpy.empty(indexes.shape, dtype=numpy.int8) 
-        model_index[:] = 0
+        model_index = numpy.zeros(indexes.shape, dtype=numpy.int8) 
         model_index[valid_index_mask] = ( 
             method_model_parameter[indexes[valid_index_mask]]) 
 
         # biomass shape: (x, k)
-        biomass = numpy.empty((indexes.shape[0], indexes.shape[1]), 
-            dtype=numpy.float32) 
-        biomass[:] = 0
+        biomass = numpy.zeros((indexes.shape[0], indexes.shape[1]), 
+            dtype=numpy.float32)
 
         # mask shapes: (x, k)
         mask_1 = model_index == 1
@@ -883,9 +881,8 @@ def _calculate_tropical_forest_edge_carbon_map(
                       biomass[valid_denom], axis=1) / denom[valid_denom])
 
         # Ensure the result has nodata everywhere the distance was invalid
-        result = numpy.empty(
-            edge_distance_block.shape, dtype=numpy.float32)
-        result[:] = CARBON_MAP_NODATA
+        result = numpy.full(edge_distance_block.shape, CARBON_MAP_NODATA, 
+            dtype=numpy.float32)
         # convert biomass to carbon in this stage
         result[valid_edge_distance_mask] = (
             average_biomass * biomass_to_carbon_conversion_factor)
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -734,7 +734,7 @@ def _calculate_tropical_forest_edge_carbon_map(
         None
 
     """
-    # load spatial indeces from pickle file
+    # load spatial indices from pickle file
     # let d = number of precalculated model cells (2217 for sample data)
     #   kd_tree.data.shape: (d, 2)
     #   theta_model_parameters.shape: (d, 3)
@@ -831,7 +831,7 @@ def _calculate_tropical_forest_edge_carbon_map(
         # For each forest pixel x, for each of its k nearest neighbors, the 
         # chosen regression method (1, 2, or 3). model_index shape: (x, k)
         model_index = numpy.empty(indexes.shape, dtype=numpy.int8) 
-        model_index[:] = -1
+        model_index[:] = 0
         model_index[valid_index_mask] = ( 
             method_model_parameter[indexes[valid_index_mask]]) 
 
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -735,6 +735,10 @@ def _calculate_tropical_forest_edge_carbon_map(
 
     """
     # load spatial indeces from pickle file
+    # let d = number of precalculated model cells (2217 for sample data)
+    #   kd_tree.data.shape: (d, 2)
+    #   theta_model_parameters.shape: (d, 3)
+    #   method_model_parameter.shape: (d,)
     kd_tree, theta_model_parameters, method_model_parameter = pickle.load(
         open(spatial_index_pickle_path, 'rb'))
 
@@ -804,7 +808,8 @@ def _calculate_tropical_forest_edge_carbon_map(
             row_coords[valid_edge_distance_mask].ravel(),
             col_coords[valid_edge_distance_mask].ravel()))
         # note, the 'n_jobs' parameter was introduced in SciPy 0.16.0
-        distances, indexes = kd_tree.query(
+        # for each forest point x, for each of its k nearest neighbors
+        distances, indexes = kd_tree.query( # distances: (x, k) indexes: (x, k)
             coord_points, k=n_nearest_model_points,
             distance_upper_bound=DISTANCE_UPPER_BOUND, n_jobs=-1)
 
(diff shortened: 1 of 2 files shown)


src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
             row_coords[valid_edge_distance_mask].ravel(),
             col_coords[valid_edge_distance_mask].ravel()))
         # note, the 'n_jobs' parameter was introduced in SciPy 0.16.0
-        distances, indexes = kd_tree.query(
+        # for each forest point x, for each of its k nearest neighbors
+        distances, indexes = kd_tree.query( # distances: (x, k) indexes: (x, k)
```
richpsharp, line 812: Ugh, sorry about this because it's so nitpicky, but inline comments need two spaces before the # don't they? If this is my only comment then no need to change this.

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
...
         # reshape to an N,nearest_points so we can multiply by thetas
         valid_edge_distances_km = numpy.repeat(
             edge_distance_block[valid_edge_distance_mask] * cell_size_km,
             n_nearest_model_points).reshape(-1, n_nearest_model_points)
 
-        # asymptotic model
+        # For each forest pixel x, for each of its k nearest neighbors, the 
+        # chosen regression method (1, 2, or 3). model_index shape: (x, k)
+        model_index = numpy.empty(indexes.shape, dtype=numpy.int8) 
+        model_index[:] = 0
```
richpsharp, line 834: If you're making an empty array and then setting them all to zeros, you can do this in one step with a `model_index = numpy.zeros(indexes.shape, dtype=numpy.int8)`

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
...
+        model_index[valid_index_mask] = ( 
+            method_model_parameter[indexes[valid_index_mask]]) 
+
+        # biomass shape: (x, k)
+        biomass = numpy.empty((indexes.shape[0], indexes.shape[1])) 
+        biomass[:] = 0
```
richpsharp, line 840: Same, here, one step, but I think it's also a good idea to name the type if you know what it's going to be. I bet it'll be a float 32 so `biomass = numpy.zeros((indexes.shape[0], indexes.shape[1], dtype=numpy.float32)) `

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
...
+        mask_3 = model_index == 3
+
+        # exponential model
         # biomass_1 = t1 - t2 * exp(-t3 * edge_dist_km)
-        biomass_model[:, :, 0] = (
-            thetas[:, :, 0] - thetas[:, :, 1] * numpy.exp(
-                -thetas[:, :, 2] * valid_edge_distances_km)
+        biomass[mask_1] = (
+            thetas[mask_1][:,0] - thetas[mask_1][:,1] * numpy.exp(
+                -thetas[mask_1][:,2] * valid_edge_distances_km[mask_1])
```
richpsharp, line 851: Yep, great.

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
...
             ) * cell_area_ha
 
         # logarithmic model
         # biomass_2 = t1 + t2 * numpy.log(edge_dist_km)
-        biomass_model[:, :, 1] = (
-            thetas[:, :, 0] + thetas[:, :, 1] * numpy.log(
-                valid_edge_distances_km)) * cell_area_ha
+        biomass[mask_2] = (
+            thetas[mask_2][:,0] + thetas[mask_2][:,1] * numpy.log(
+                valid_edge_distances_km[mask_2])) * cell_area_ha
```
richpsharp, line 858: Great

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -804,59 +808,60 @@ def _calculate_tropical_forest_edge_carbon_map(
...
-        # reduce the axis=1 dimensionality of the model by selecting the
-        # appropriate value via the model_index array. Got this trick from
-        # http://stackoverflow.com/questions/18702746/reduce-a-dimension-of-numpy-array-by-selecting
-        biomass_y, biomass_x = numpy.meshgrid(
-            numpy.arange(biomass_model.shape[1]),
-            numpy.arange(biomass_model.shape[0]))
-        biomass = biomass_model[biomass_x, biomass_y, model_index]
+        biomass[mask_3] = (
+            thetas[mask_3][:,0] + thetas[mask_3][:,1] *
+            valid_edge_distances_km[mask_3]) * cell_area_ha
```
richpsharp, line 864: Great, very good.

richpsharp (changes requested):
Hi Emily, excellent. I have the smallest of stylistic requests on this PR. I hope they're not annoying.

src/natcap/invest/forest_carbon_edge_effect.py
emlys: Good catch! I was bending the rules to make it fit in 80 characters haha. I'll move it to its own line.

src/natcap/invest/forest_carbon_edge_effect.py
emlys: I've been using `numpy.empty` because I thought it was faster. I just tested it:
```
>>> timeit.timeit('a = numpy.zeros(100000)', setup='import numpy')
18.696336253000027
>>> timeit.timeit('a = numpy.full(100000, 0)', setup='import numpy')
15.37874967600004
>>> timeit.timeit('a = numpy.empty(100000); a[:] = 0', setup='import numpy')
14.114311920000034
```
And I think we use `numpy.empty` rather than `numpy.full` in a lot of places. However, [this person](https://stackoverflow.com/a/26213344/14451410) got the opposite results. And it doesn't really make sense that `numpy.zeros` would be less optimized. Maybe you have some insight into this?

src/natcap/invest/forest_carbon_edge_effect.py
emlys: Thanks!

src/natcap/invest/forest_carbon_edge_effect.py
richpsharp: Ha this is straight out of code review I did with James years ago where he was surprised that the empty.. [:] paradigm was faster when i was doing a empty... [:] = nodata set. That's a great answer you linked and and I agree with that person that I'd "use it for the name" and "it's harder for numpy folks to optimize two statements instead of one". Also there's a comment on there about someone guessing that numpy is doing an uninitialized `mmap` for very large arrays, so timing it like this isn't accurate since the memory isn't really set yet. You can see the difference there if you do something that requires reading all the elements of the arrays. Something like this:

```
>>> timeit.timeit('a = numpy.zeros(100000); x=numpy.sum(a)', setup='import numpy')
72.45531159999996
>>> timeit.timeit('a = numpy.empty(100000); a[:] = 0.0; x=numpy.sum(a)', setup='import numpy')
107.3589187
```
(my computer is much slower than yours it seems) 

But even still, in the last statement you're talking about a difference of 0.072 *micro*seconds vs 0.107 microseconds per raster block. Not noticeable on a 4k X 4k raster and taking about 3ish seconds longer on a full global 128k x 64k raster. So even if the results were refersed and empty->[:] were were 3 seconds slower, I think I'd prefer the single statement instead of the two.

But given this I hadn't realized there was a `numpy.full` function! I'll start using that in the future instead of the two step `empty` `[:]=init_val` paradigm that I might have poisoned us with casually back in the day.

Thanks for making the effort on this on what I thought was just a casual style comment.

richpsharp (changes requested):
Hi Emily, please see my too-long comment about timings and uninitialized c `mmap`s!

src/natcap/invest/forest_carbon_edge_effect.py
emlys: That's wacky! Thanks for the explanation. Weirdly I still get the opposite result:
```
>>> timeit.timeit('a = numpy.zeros(100000); x=numpy.sum(a)', setup='import numpy')
46.84990651900034
>>> timeit.timeit('a = numpy.empty(100000); a[:] = 0.0; x=numpy.sum(a)', setup='import numpy')
43.15799723200007
```
Maybe an operating system difference? But I agree it's worth using `zeros` just for the convenience and that the time difference is insignificant! Fixed in 4175906

richpsharp: It might be, at a deep level it's up to the OS about how, or when, or if ever, to allocate memory. I'll merge this PR but I'm curious if you'd be willing to try another experiment? I'll slack you.

richpsharp (approved):

richpsharp:
Hi @emlys, I approved the merge and manually fixed the merge conflict on `HISTORY.rst`. It's running tests now, so feel free to merge if you get to it before I do! Thanks for doing a few iterations on this with me.

richpsharp:
Hey @emlys, the `sdist` test is failing. Is that because I screwed up HISTORY.rst? I should have let you do it... I'll pass it back to you okay? Sorry...

emlys:
No worries, I think it was just a flaky test. Looked like a conda http error. Everything's passing now so I'm going to merge!

PR #415, opened 2020-12-09
Update HQ default half sat for sample data

dcdenu4:
# Description
Handles an issue from #402 brought up by Stacie about our default value not demonstrating a good output for quality.

Update value from 0.5 to 0.05. Also update the revisions for sample data and users guide that now reflect this change.

I mentioned not being able to see differences between runs when changing that half saturation value. I no longer see that as the case and confirmed that the reexecution hashes were different from TG. Seems to be okay now.

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -133,7 +133,7 @@ Unreleased Changes (3.9)
       value. Now, the decay does not ignore those nodata edges causing values
       on the edges to decay more quickly. The area of study should have
       adequate boundaries to account for these edge effects.
-    * Update default half saturation value for sample data to 0.05 from 0.5.
+    * Update default half saturation value for sample data to 0.05 from 0.1.
 * Seasonal Water Yield
     * Fixed a bug where precip or eto rasters of ``GDT_Float64`` with values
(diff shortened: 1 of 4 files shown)


HISTORY.rst
```diff
@@ -133,6 +133,7 @@ Unreleased Changes (3.9)
       value. Now, the decay does not ignore those nodata edges causing values
       on the edges to decay more quickly. The area of study should have
       adequate boundaries to account for these edge effects.
+    * Update default half saturation value for sample data to 0.05 from 0.5.
```
davemfish, line 136: It's trivial, but the former value was 0.1 https://bitbucket.org/natcap/invest-sample-data/commits/b7a51f189315e08484b5ba997a5c1de88ab7f06d

davemfish (changes requested):
@dcdenu4 I thought I had remembered changing this a while back (though I clearly forgot to change the UI text about it!) to 0.1. I'm happy if we change it again to 0.05 if that is better than 0.1. Do you think it's worth looking at the 0.1 results? Or maybe you and Stacie already had been.

davemfish (approved):

PR #417, opened 2020-12-10
Mask nodata areas of forest carbon raster output

### LINKED ISSUE #414: Forest Carbon output in nodata areas
Running Forest Carbon with the sample data, there are some weird shapes showing up in the output carbon map raster. The glowing dots in the bottom left and light colored shapes in the bottom right are outside of the input data areas.
Screen Shot 2020-12-09 at 11 50 25 AM

emlys:
# Description
#402
Fixes #414 

The shapes in what should be nodata areas of the output carbon map were happening because the function `pygeoprocessing.distance_transform_edt` does not preserve nodata areas. Its docstring says,

> It is computationally convenient to calculate the distance transform on the entire raster irrespective of nodata placement and thus produces a raster that will have distance transform values even in pixels that are nodata in the base.

Masking out the areas of the `edge_distance` raster that are nodata in the `lulc` raster isn't super convenient because you have to read in both rasters blockwise simultaneously and then overwrite the `edge_distance` raster.

I see a few options and I'd like to get another opinion before going further with this and updating the test data:
1. Mask out the nodata areas in `pygeoprocessing.distance_transform_edt`. I like this option because I feel like the current behavior is going against a pattern of preserving nodata across operations. 
2. (implemented in this PR) Mask out the nodata areas in `forest_carbon_edge_effect._map_distance_from_tropical_forest_edge`. You could argue that the model has the responsibility to modify the output from `pygeoprocessing` if it needs it to be masked.
3. Ignore it? I'm guessing we ignored this issue on purpose when the model was first made. Maybe it's fine to have those extra shapes and users will know to ignore them.

# Checklist

### DIFF
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -24,8 +24,8 @@ LOGGER = logging.getLogger(__name__)
 # grid cells are 100km. Becky says 500km is a good upper bound to search
 DISTANCE_UPPER_BOUND = 500e3
 
-# helpful to have a global nodata defined for all the carbon map rasters
-CARBON_MAP_NODATA = -9999
+# helpful to have a global nodata defined for the whole model
+NODATA_VALUE = -1
 
 ARGS_SPEC = {
     "model_name": "Forest Carbon Edge Effect Model",
@@ -378,7 +378,7 @@ def execute(args):
         func=pygeoprocessing.raster_calculator,
         args=(carbon_maps_band_list, combine_carbon_maps,
               output_file_registry['carbon_map'], gdal.GDT_Float32,
-              CARBON_MAP_NODATA),
+              NODATA_VALUE),
         target_path_list=[output_file_registry['carbon_map']],
         task_name='combine_carbon_maps')
 
@@ -415,10 +415,10 @@ def combine_carbon_maps(*carbon_maps):
     nodata_mask = numpy.empty(carbon_maps[0].shape, dtype=numpy.bool)
     nodata_mask[:] = True
     for carbon_map in carbon_maps:
-        valid_mask = carbon_map != CARBON_MAP_NODATA
+        valid_mask = carbon_map != NODATA_VALUE
         nodata_mask &= ~valid_mask
         result[valid_mask] += carbon_map[valid_mask]
-    result[nodata_mask] = CARBON_MAP_NODATA
+    result[nodata_mask] = NODATA_VALUE
     return result
 
 
@@ -550,7 +550,7 @@ def _calculate_lulc_carbon_map(
             is_tropical_forest = 0
         if ignore_tropical_type and is_tropical_forest == 1:
             # if tropical forest above ground, lookup table is nodata
-            lucode_to_per_cell_carbon[int(lucode)] = CARBON_MAP_NODATA
+            lucode_to_per_cell_carbon[int(lucode)] = NODATA_VALUE
         else:
             try:
                 lucode_to_per_cell_carbon[int(lucode)] = float(
@@ -570,7 +570,7 @@ def _calculate_lulc_carbon_map(
 
     utils.reclassify_raster(
         (lulc_raster_path, 1), lucode_to_per_cell_carbon,
-        carbon_map_path, gdal.GDT_Float32, CARBON_MAP_NODATA,
+        carbon_map_path, gdal.GDT_Float32, NODATA_VALUE,
         reclass_error_details)
 
 
@@ -607,7 +607,6 @@ def _map_distance_from_tropical_forest_edge(
         if int(ludata['is_tropical_forest']) == 1]
 
     # Make a raster where 1 is non-forest landcover types and 0 is forest
-    forest_mask_nodata = 255
     lulc_nodata = pygeoprocessing.get_raster_info(
         base_lulc_raster_path)['nodata']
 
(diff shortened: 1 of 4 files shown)


src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -612,20 +612,48 @@ def _map_distance_from_tropical_forest_edge(
...
+    edge_distance_band = edge_distance_raster.GetRasterBand(1)
+    for (lulc_offset, lulc_block), (_, distance_block) in zip(
+            pygeoprocessing.iterblocks(
+                (base_lulc_raster_path, 1), largest_block=2**12),
+            pygeoprocessing.iterblocks(
+                (edge_distance_path, 1), largest_block=2**12)):
+        # where LULC has nodata, overwrite edge distance with nodata value
+        masked_distance_block = numpy.where(
+            lulc_block == lulc_nodata, -1, distance_block)
+        edge_distance_band.WriteArray(
```
emlys, line 652: Is there anything wrong with alternately reading and writing to a raster like this? Because it's reading one block, then overwriting that block, then reading the next block, I think it shouldn't conflict.

dcdenu4: That's a good question. I think it'd be fine. BUT, if you know the rasters are aligned you could do:
```python

base_lulc_raster = gdal.OpenEx(base_lulc_raster_path)      
base_lulc_band = base_lulc_raster.GetRasterBand(1)         
                                                  
edge_distance_raster = gdal.OpenEx(edge_distance_raster_path)
edge_distance_band = edge_distance_raster.GetRasterBand(1)   

for offset_dict in pygeoprocessing.iterblocks(
    (base_lulc_raster_path, 1), offset_only=True):

     lulc_block = lulc_band.ReadAsArray(**offset_dict)
     edge_distance_block = edge_distance_band.ReadAsArray(**offset_dict)
```

BUT, it might also just be better to call another `raster_calculator` call too, with a simple op that masks the nodata. This would require a temporary file of sorts for that intermediate step.

dcdenu4 (changes requested):
Hey @emlys I commented on the `iterblocks` question you had. In general I feel like `pygeoprocessing` shouldn't necessarily have to be responsible for masking the nodata. I think if the docstring is clearly defined then the programmer should be responsible for handling the result. It certainly would be a convenience. In this instance I think it'd make sense to handle the issue in the model.

src/natcap/invest/forest_carbon_edge_effect.py
emlys: I think I like your code snippet the best, it seems the most simple.

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -618,17 +617,17 @@ def mask_non_forest_op(lulc_array):
...
-        # where LULC has nodata, set value to forest_mask_nodata value (255)
+        # where LULC has nodata, set value to nodata value (-1)
         # where LULC has data, set to 0 if LULC is a forest type, 1 if it's not
-        return numpy.where(nodata_mask, forest_mask_nodata, non_forest_mask)
+        return numpy.where(nodata_mask, NODATA_VALUE, non_forest_mask)
 
     pygeoprocessing.raster_calculator(
         [(base_lulc_raster_path, 1)], mask_non_forest_op,
-        target_non_forest_mask_path, gdal.GDT_Byte, forest_mask_nodata)
+        target_non_forest_mask_path, gdal.GDT_Byte, NODATA_VALUE)
```
dcdenu4, line 630: Hmm. Will this be a problem that we're returning a `gdal.GDT_Byte` but using a `-1` nodata value? Since we're not setting metadata for `IMAGE_STRUCTURE` of the raster, I'm thinking GDAL will assume Unsigned Byte. Thoughts?

dcdenu4 (changes requested):
Hey @emlys looking good to me! Just had one question about a possible Byte, -1 nodata issue. Thanks for taking a look.

dcdenu4:
Oh, it looks like there's conflicts now as too. Probably since the other FCEE PR merged?

src/natcap/invest/forest_carbon_edge_effect.py
emlys: That's a good point! I just wanted to use a consistent nodata value throughout the model. I'll use type int8 instead

dcdenu4:
Bah, I keep reviewing before you've actually requested / re-requested me too. Sorry about that, it probably comes off as super pushy!

emlys:
Thanks for the input @dcdenu4! I added the carbon raster output to the test data since before we were only testing the aggregated shapefile output. And resolved the conflict.
> Sorry about that, it probably comes off as super pushy!
Not at all! Happy to go back and forth quickly on this one

davemfish:
@emlys I didn't review this one, but may I make a small request to just give it a descriptive title before it's merged. It's occasionally useful to search for something in old PRs later on, so titles help. Thank you!

src/natcap/invest/forest_carbon_edge_effect.py
emlys: Just going to leave it as is with the 255 value and byte type

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -607,25 +607,54 @@ def _map_distance_from_tropical_forest_edge(
...
-        """Converts forest lulc codes to 1."""
-        non_forest_mask = ~numpy.in1d(
-            lulc_array.flatten(), forest_codes).reshape(lulc_array.shape)
+        """Convert forest lulc codes to 0.
+        Args:
+            lulc_array (numpy.ndarray): array representing a LULC raster where
+                each forest LULC code is in `forest_codes`.
+        Returns:
+            numpy.ndarray with the same shape as lulc_array. All pixels are
+                0 (forest), 1 (non-forest), or -1 (nodata).
```
dcdenu4, line 621: Ah! Sorry to be picky but I think this -1 should be back to a 255 now?

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -607,25 +607,54 @@ def _map_distance_from_tropical_forest_edge(
...
+        """
+        non_forest_mask = ~numpy.isin(lulc_array, forest_codes)
         nodata_mask = lulc_array == lulc_nodata
+        # where LULC has nodata, set value to nodata value (-1)
```
dcdenu4, line 625: Here too for nodata value?

dcdenu4 (changes requested):
Thanks for all this @emlys and tackling the test_data and tests. I have one last wicked picky comment about some comments still having the -1 nodata where I think it should now be 255...

src/natcap/invest/forest_carbon_edge_effect.py
emlys: Ahh thanks for catching that!

dcdenu4 (approved):
Thanks @emlys ! Looks good to me.

PR #430, opened 2021-01-05
Add logger name to logging format

### LINKED ISSUE #15: pygeoprocessing.routing not showing up in InVEST model logs
When running an InVEST model, specifically an under development version of SDR that uses PyGeoprocessing 1.6.x, logging does not appear in the log for pygeoprocessing.routing, though it does seem to show up as a log to Task._call. James and I did some debugging and we find that outside of InVEST the the routing module does log and so we suspect something in the way InVEST is handling logging. Here's a short snippet that uses routing and taskgraph that shows log messages from the routing module:

```python
import logging

import taskgraph
import pygeoprocessing.routing

logging.basicConfig(level=logging.DEBUG)


def main():
    task_graph = taskgraph.TaskGraph('taskgraph_cache', 8)

    target_filled_dem_raster_path = 'filled.tif'
    dem_raster_path_band = (r"C:\Users\rpsharp\Documents\bitbucket_repos\invest\data\invest-data\Base_Data\Freshwater\dem", 1)
    task_graph.add_task(
        func=pygeoprocessing.routing.fill_pits,
        args=(
            dem_raster_path_band, target_filled_dem_raster_path),
        target_path_list=[target_filled_dem_raster_path])

    task_graph.close()
    task_graph.join()


if __name__ == '__main__':
    main()
```

Originally reported by @richpsharp at https://bitbucket.org/natcap/invest/issues/3859/pygeoprocessingrouting-not-showing-up-in

emlys:
# Description
Fixes #15 (sort of)
See my comment [here](https://github.com/natcap/invest/issues/15#issuecomment-754910712) about why this isn't a perfect solution. Having the extra info should be useful for all cython modules, in both invest and pygeoprocessing. 
I'm attaching an example of how the logs look now (search 'routing', 'geoprocessing_core', 'sdr_core' to see examples). Is it annoying having the redundant `name` and `module` info for non-cython logging? I could try adding it as a custom handler for each cython module. Personally I don't mind the redundancy.
[InVEST-Sediment-Delivery-Ratio-Model-(SDR)-log-2021-01-05--14_37_29.txt](https://github.com/natcap/invest/files/5772748/InVEST-Sediment-Delivery-Ratio-Model-.SDR.-log-2021-01-05--14_37_29.txt)

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -32,6 +32,12 @@
 
 
 .. :changelog:
+3.9.1
+-----
+* General:
+    * Include logger name in the logging format. This is helpful for the cython
+      modules, which can't log module, function, or line number info.
+
 3.9.0 (2020-12-11)
 ------------------
 * General:
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -20,6 +20,7 @@ import pygeoprocessing
 LOGGER = logging.getLogger(__name__)
 LOG_FMT = (
     "%(asctime)s "
+    "(%(name)s) "
     "%(module)s.%(funcName)s(%(lineno)d) "
     "%(levelname)s %(message)s")
 


davemfish (approved):
I thought about suggesting we remove the `module`, now that we have `name`, to get lines like these without the redundancy:

```
2021-01-05 14:37:29,971 (natcap.invest.utils) prepare_workspace(124) INFO
2021-01-05 14:37:29,984 (taskgraph.Task) __init__(250) DEBUG 
2021-01-05 14:37:30,903 (pygeoprocessing.routing.routing) _call(1234) INFO 100.0% complete
```
But I'm not sure that's a good idea.  `_call` is the function on `Task`, which is no longer referenced at all. And if I searched for `utils.prepare_workspace` I wouldn't find anything. We could drop the parens and join `name` to `function` with a `.`, but then we'd end up with `pygeoprocessing.routing.routing._call`, which is definitely not helpful.

So, I'm good with the solution you have here!

emlys:
Thanks!

PR #432, opened 2021-01-06
Fisheries HST divide by zero warning

### LINKED ISSUE #18: Fisheries-HST FloatingPointError: invalid value encountered in true_divide

emlys:
# Description
Fixes #18
IMO it's clear enough that the intent here was to set `H_xa_weighted` to 0 where `n_a == 0`.  
It looks like this model and the main fisheries model come from a different era. They're kind of a mess by our current standards, and I wouldn't be surprised if there are other silent bugs. Maybe we should discuss the future of these models and whether it's worth refactoring them or discontinuing them.

# Checklist


phargogh (approved):
@emlys you're exactly right that this model is from a different era and will undoubtedly be worth an update at some point.  FWIW, the PSC is reviewing our list of supported models, our ability to maintain them, and community interest, and deprecation of Fisheries is for sure up for discussion!

So given that there's a lot about the Fisheries model that's up in the air, this fix is great: minimal and sufficient, consistent with the existing implementation, and documented.

PR #435, opened 2021-01-07
remove fix for old actions bug

emlys:
# Description
Actions bug has apparently been fixed, so this workaround isn't needed any longer


davemfish (approved):
Much appreciated!

PR #436, opened 2021-01-08
pandas 1.2.0 compatibility (and rtree/pysinstaller compatibility)

### LINKED ISSUE #428: pandas 1.2.0 compatibility
It looks there are some breaking changes in this pandas release with test failures/errors observed in `utils`, `validation` and `wind_energy`.

https://pandas.pydata.org/docs/whatsnew/v1.2.0.html

Here's an example failing test suite:
https://github.com/natcap/invest/pull/399/checks?check_run_id=1645871551

### LINKED ISSUE #429: rtree & pyinstaller incompatibility
This seems to be a recurring problem. The current iteration of the problem seems to be rtree's logic for finding dll's in a pyinstaller frozen distribution. Our pyinstaller script seems to be doing the job of finding the spatialindex dlls correctly - adding them to the `binaries` object and including them in the frozen dist - but rtree still can't find them on import.

Here's an example build failure
https://github.com/natcap/invest/runs/1646423120?check_suite_focus=true

### LINKED ISSUE #439: rtree & pyinstaller incompatibility
This seems to be a recurring problem. The current iteration of the problem seems to be rtree's logic for finding dll's in a pyinstaller frozen distribution. Our pyinstaller script seems to be doing the job of finding the spatialindex dlls correctly - adding them to the `binaries` object and including them in the frozen dist - but rtree still can't find them on import.

Here's an example build failure
https://github.com/natcap/invest/runs/1646423120?check_suite_focus=true

davemfish:
# Description
Fixes #428 and fixes #439 

This PR makes invest compatible with pandas 1.2.0 and maintains backwards-compatibility. The following changes needed to be accommodated.
* The default behavior of the `read_csv` python engine changed to allow a `UnicodeDecodeError` to bubble up instead of replacing problematic characters.
* `openpyxl` is now preferred over `xlrd` for reading `.xlsx` files and so new types of exceptions are raised when reading fails.
* Changes to default float precision in `read_csv`

This PR also restricts `rtree` to a version that is compatible with our pyinstaller frozen environments, for the time-being.

# Checklist
- [ ] Updated HISTORY.rst - no user-facing changes

- [ ] Updated the user's guide - no changes

### DIFF
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -587,7 +587,8 @@ def read_csv_to_dataframe(
         dataframe = pandas.read_csv(
             path, engine=engine, encoding=encoding, sep=sep, **kwargs)
     except UnicodeDecodeError as error:
-        LOGGER.error(f'{path} must be encoded as utf-8')
+        LOGGER.error(
+            f'{path} must be encoded as utf-8 or avoid non-ASCII characters')
         raise error
     # this won't work on integer types, which happens if you set header=None
     # however, there's little reason to use this function if there's no header
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -582,8 +582,12 @@ def read_csv_to_dataframe(
     # allow encoding kwarg to override this if it's provided
     if not encoding and has_utf8_bom(path):
         encoding = 'utf-8-sig'
-    dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
-                                sep=sep, **kwargs)
+    try:
+        dataframe = pandas.read_csv(
+            path, engine=engine, encoding=encoding, sep=sep, **kwargs)
+    except UnicodeDecodeError as error:
+        LOGGER.error(f'{path} must be encoded as utf-8')
+        raise error
     # this won't work on integer types, which happens if you set header=None
     # however, there's little reason to use this function if there's no header
     dataframe.columns = dataframe.columns.str.strip()
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -886,7 +886,10 @@ class ReadCSVToDataframeTests(unittest.TestCase):
                 """
             ).strip())
         with self.assertRaises(UnicodeDecodeError) as cm:
-            utils.read_csv_to_dataframe(csv_file)
+            # In pandas 1.2.0, there's a bug where sep=None combined
+            # with this UnicodeDecodeError leaves the file handle open,
+            # and our tearDown function errors. So setting sep=',' for now.
+            utils.read_csv_to_dataframe(csv_file, sep=',')
         self.assertTrue("decode byte" in str(cm.exception))
 
     def test_override_default_encoding(self):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -871,7 +871,7 @@ class ReadCSVToDataframeTests(unittest.TestCase):
         self.assertEqual(df['HEADER2'][1], 5)
 
     def test_non_utf8_encoding(self):
-        """utils: test that non-ASCII characters raise an error"""
+        """utils: test non-ASCII chars with non-UTF8 encoding raises error"""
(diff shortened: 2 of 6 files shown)


src/natcap/invest/utils.py
```diff
@@ -583,8 +583,12 @@ def read_csv_to_dataframe(
     # allow encoding kwarg to override this if it's provided
     if not encoding and has_utf8_bom(path):
         encoding = 'utf-8-sig'
-    dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
-                                sep=sep, **kwargs)
+    try:
+        dataframe = pandas.read_csv(
+            path, engine=engine, encoding=encoding, sep=sep, **kwargs)
+    except UnicodeDecodeError as error:
+        LOGGER.error(f'{path} must be encoded as utf-8')
```
emlys, line 590: I don't know if it's worth mentioning that other encodings are fine as long as they only use the ASCII-compatible subset of characters? Maybe that wouldn't be any help to most users

src/natcap/invest/utils.py
```diff
@@ -583,8 +583,12 @@ def read_csv_to_dataframe(
     # allow encoding kwarg to override this if it's provided
     if not encoding and has_utf8_bom(path):
         encoding = 'utf-8-sig'
```
emlys, line 585: If `encoding` isn't specified, should we default to using `encoding = 'utf-8'`? It seems like that would be the `pandas` default, but it isn't clear from the [docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) what happens when `encoding=None`.

tests/test_utils.py
```diff
@@ -871,7 +871,7 @@ def test_utf8_bom_encoding(self):
         self.assertEqual(df['HEADER2'][1], 5)
 
     def test_non_utf8_encoding(self):
-        """utils: test that non-UTF8 encoding doesn't raise an error"""
+        """utils: test that non-ASCII characters raise an error"""
```
emlys, line 874: to be pedantic, should this be "non-UTF8, non-ASCII characters"?

emlys (commented):
Just a couple small comments!

src/natcap/invest/utils.py
davemfish: If `encoding` is not specified then we do let pandas do it's own thing (offhand I'm not sure what that is either). But in the rare case where the file has a BOM, I guess pandas does not handle the BOM correctly when encoding=None, hence the need to identify the BOM first and specify encoding as `utf-8-sig`.

davemfish:
Thanks, @emlys , back to you!

src/natcap/invest/utils.py
emlys: Maybe we should be explicit about assuming UTF8 when there isn't a BOM too? something like 
```
if not encoding:
    if has_utf8_bom(path):
        encoding = 'utf-8-sig'
    else:
        encoding = 'utf-8'
```

davemfish:
Okay, @emlys , this is now actually ready for review again. Originally I thought we could keep backwards compatibility with pandas<1.2.0, but then after accidentally trying that I saw that some tests failed with pandas==1.1.5. It's really only the tests, not the source code, that is not backwards-compatible, but I'm not sure there's a compelling reason to go out of our way to support the older versions.

The tests in question are 1. the wind energy test, where a floating point value in our regression data is read in slightly differently with pandas versions. And 2. an HRA test that writes an excel file as part of setup for a test. With older pandas, it wants to use `xlwt`, but issues a FutureWarning. With new pandas, we can use `openpyxl` and eliminate `xlrd, xlwt` dependencies.

emlys:
@davemfish I'm happy with bumping the pandas requirement to 1.2.0. It seems like an improvement since it's reading in that float more correctly (without the extra 0.000000005), and `ValueError` is nicer than `xlrd.biffh.XLRDError`.

emlys (approved):

PR #447, opened 2021-01-19
Changes to go with new UI spec format

emlys:
Goes with https://github.com/natcap/invest-workbench/pull/84.
* Adding an endpoint that accepts a vector filepath, opens the vector, and returns a list of its column names. A couple of models use this to fill in a dropdown menu input's list of options.
* Add an endpoint that calls `delineateit._vector_may_contain_points`. This is used in the DelineateIt UI to determine if the 'snap points' option should be enabled.
* Fix a couple of args spec properties

I just realized that using a filepath could be problematic in the future. If we tried to `invest serve` on a remote server or in a docker container it wouldn't have access to the file. Maybe we should look into doing small geospatial functions like these on the javascript side.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -38,16 +38,18 @@ workbench-alpha
 * General:
     * Added ``invest serve`` entry-point to the CLI. This launches a Flask app
       and server on the localhost, to support the workbench.
-    * Add ``/colnames`` POST endpoint to the server API. This endpoint accepts a
-      vector filepath and returns a list of the vector's column names (used to
-      determine options for a dropdown menu input).
+    * Add ``/colnames`` POST endpoint to the server API. This endpoint accepts
+      a vector filepath and returns a list of the vector's column names (used 
+      to determine options for a dropdown menu input).
     * Add ``/vector_has_points`` POST endpoint to the server API. This endpoint
       accepts a vector filepath and returns a boolean indicating whether the
-      vector might have point geometries (can't be certain because the type could
-      be ``ogr.wkbUnknown``). This is used by the DelineateIt UI.
-    * Rename finfish ARGS_SPEC ``module_name`` to ``model_name`` to be consistent
-      with all other models.
+      vector might have point geometries (can't be certain because the type 
+      could be ``ogr.wkbUnknown``). This is used by the DelineateIt UI.
+    * Rename finfish ARGS_SPEC ``module_name`` to ``model_name`` to be 
+      consistent with all other models.
     * Minor rephrasing of some ``ARGS_SPEC.args.arg.name`` values.
+    * Dynamically populated dropdown inputs' arg specs now have a ``type`` of
+      ``option_string`` and ``validation_options['options']`` is an empty list.
 
 Unreleased Changes (3.9.1)
 --------------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -45,6 +45,9 @@ workbench-alpha
       accepts a vector filepath and returns a boolean indicating whether the
       vector might have point geometries (can't be certain because the type could
       be ``ogr.wkbUnknown``). This is used by the DelineateIt UI.
+    * Rename finfish ARGS_SPEC ``module_name`` to ``model_name`` to be consistent
+      with all other models.
+    * Minor rephrasing of some ``ARGS_SPEC.args.arg.name`` values.
 
 Unreleased Changes (3.9.1)
 --------------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -38,9 +38,13 @@ workbench-alpha
 * General:
     * Added ``invest serve`` entry-point to the CLI. This launches a Flask app
       and server on the localhost, to support the workbench.
-    * Add ``/colnames`` endpoint to the server API. This endpoint accepts a
+    * Add ``/colnames`` POST endpoint to the server API. This endpoint accepts a
       vector filepath and returns a list of the vector's column names (used to
       determine options for a dropdown menu input).
+    * Add ``/vector_has_points`` POST endpoint to the server API. This endpoint
+      accepts a vector filepath and returns a boolean indicating whether the
+      vector might have point geometries (can't be certain because the type could
+      be ``ogr.wkbUnknown``). This is used by the DelineateIt UI.
 
 Unreleased Changes (3.9.1)
 --------------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -41,7 +41,7 @@ workbench-alpha
     * Add ``/colnames`` POST endpoint to the server API. This endpoint accepts
       a vector filepath and returns a list of the vector's column names (used 
       to determine options for a dropdown menu input).
-    * Add ``/vector_has_points`` POST endpoint to the server API. This endpoint
+    * Add ``/vector_may_have_points`` POST endpoint to the server API. This endpoint
       accepts a vector filepath and returns a boolean indicating whether the
       vector might have point geometries (can't be certain because the type 
       could be ``ogr.wkbUnknown``). This is used by the DelineateIt UI.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -38,6 +38,9 @@ workbench-alpha
 * General:
     * Added ``invest serve`` entry-point to the CLI. This launches a Flask app
       and server on the localhost, to support the workbench.
+    * Add ``/colnames`` endpoint to the server API. This endpoint accepts a
+      vector filepath and returns a list of the vector's column names (used to
+      determine options for a dropdown menu input).
 
 Unreleased Changes (3.9.1)
 --------------------------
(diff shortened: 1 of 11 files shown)


src/natcap/invest/finfish_aquaculture/finfish_aquaculture.py
```diff
@@ -37,8 +37,11 @@
                 "tables). Additionally, the numbers underneath this "
                 "farm identifier name must be unique integers for all "
                 "the inputs."),
-            "type": "freestyle_string",
+            "type": "option_string",
```
emlys, line 40: dynamically populated dropdowns have had the `freestyle_string` type which is kinda incorrect but worked. It doesn't work any more because the `type` property determines what kind of input field is rendered. I don't really like having the set of options here in `validation_options.options`, but it doesn't really make sense to go in the UI spec either

src/natcap/invest/ui_server.py
```diff
@@ -102,6 +104,60 @@ def get_invest_validate():
...
+    LOGGER.debug(payload)
+    vector_path = payload['vector_path']
+    colnames = []  
+    # a lot of times the path will be empty so don't even try to open it
+    if vector_path:  
+        try:
+            vector = gdal.OpenEx(vector_path, gdal.OF_VECTOR)
+            colnames = [defn.GetName() for defn in vector.GetLayer().schema]
+        except:
+            LOGGER.error(f'Could not read column names from {vector_path}')
```
davemfish, line 128: I don't mind catching the broad exception here because I think flask would catch it anyway, but is it worth logging the actual exception here as well?

src/natcap/invest/ui_server.py
```diff
@@ -102,6 +104,60 @@ def get_invest_validate():
...
+    vector_path = payload['vector_path']
+    # default True because it may have point geometries unless proven otherwise
+    has_points = True
+    # a lot of times the path will be empty so don't even try to open it
+    if vector_path:
+        try:
+            has_points = delineateit._vector_may_contain_points(vector_path)
+        except:
+            LOGGER.error(
+                f'Could not tell if vector {vector_path} contains points.')
```
davemfish, line 156: Same comment as above about exceptions

src/natcap/invest/ui_server.py
```diff
@@ -102,6 +104,60 @@ def get_invest_validate():
...
+    LOGGER.debug({'has_points': has_points})
+    return json.dumps({'has_points': has_points})
```
davemfish, line 158: I was wondering if the plain boolean value could be returned as valid json without the object syntax, and I learned the answer is no!

```
>>> json.dumps(True)
'true'
>>> json.dumps({'has_points': True})
'{"has_points": true}'
```

davemfish (changes requested):
Looks good to me, @emlys ! I mentioned one more ARGS_SPEC (CV - slr field) that needs updated to `option_string`. And a small request for a bit more logging.

davemfish:
> I just realized that using a filepath could be problematic in the future. If we tried to `invest serve` on a remote server or in a docker container it wouldn't have access to the file. Maybe we should look into doing small geospatial functions like these on the javascript side.

Yeah that's a good point. And it also applies to the validation endpoint, which sends lots of filepaths to `validation.py`. I don't think it's problem we should solve yet, but good to think about it.

src/natcap/invest/finfish_aquaculture/finfish_aquaculture.py
emlys: oof glad you caught that!

src/natcap/invest/ui_server.py
emlys: i wish it didn't have to have an arbitrary key name but no way around it :/

emlys:
Thanks @davemfish! That coastal vulnerability arg is definitely the only other dynamic dropdown. I added the exception to the logging. I also decided it's worth renaming the `has_points` endpoint and its variables to `may_have_points`, since it can't tell for sure if the file can't be opened or if there are geometries with unknown type.

HISTORY.rst
```diff
@@ -38,6 +38,18 @@ workbench-alpha
 * General:
     * Added ``invest serve`` entry-point to the CLI. This launches a Flask app
       and server on the localhost, to support the workbench.
+    * Add ``/colnames`` POST endpoint to the server API. This endpoint accepts
+      a vector filepath and returns a list of the vector's column names (used 
+      to determine options for a dropdown menu input).
+    * Add ``/vector_has_points`` POST endpoint to the server API. This endpoint
```
davemfish, line 44: Here's another to place to update the endpoint name.

davemfish (approved):
@emlys I found a spot in history that could use an update. Other than that this looks good and I'll approve and pass it on to @phargogh

tests/test_ui_server.py
```diff
@@ -0,0 +1,65 @@
...
+import json
+import unittest
+
+TEST_DATA_PATH = os.path.join(
+    os.path.dirname(__file__), '..', 'data', 'invest-test-data')
+
+
+class EndpointFunctionTests(unittest.TestCase):
+
+    def test_get_vector_colnames(self):
```
phargogh, line 11: Missing docstring?

tests/test_ui_server.py
```diff
@@ -0,0 +1,65 @@
...
+        path = os.path.join(
+            TEST_DATA_PATH, 'aquaculture', 'Input', 'Finfish_Netpens.shp')
+        response = test_client.post('/colnames', json={'vector_path': path})
+        colnames = json.loads(response.get_data(as_text=True))
+        self.assertEqual(colnames, ['FarmID'])
+        # a non-vector file shouldn't raise an error
+        path = os.path.join(TEST_DATA_PATH, 'ndr', 'input', 'dem.tif')
+        response = test_client.post('/colnames', json={'vector_path': path})
+        colnames = json.loads(response.get_data(as_text=True))
+        self.assertEqual(colnames, [])
```
phargogh, line 28: It makes sense to me that a non-vector file won't have any column names, but isn't that an error state?  Should that return an error HTTP code as well?

src/natcap/invest/ui_server.py
phargogh: Yep, that's unfortunately the case ... conformant JSON objects must have key-value pairs.

src/natcap/invest/ui_server.py
```diff
@@ -102,6 +104,63 @@ def get_invest_validate():
...
+    vector_path = payload['vector_path']
+    colnames = []  
+    # a lot of times the path will be empty so don't even try to open it
+    if vector_path:  
+        try:
+            vector = gdal.OpenEx(vector_path, gdal.OF_VECTOR)
+            colnames = [defn.GetName() for defn in vector.GetLayer().schema]
+        except Exception as e:
+            LOGGER.error(f'Could not read column names from {vector_path}. '
+                         f'ERROR: {e}')
```
phargogh, line 129: Could I suggest that we change the `LOGGER.error` to `LOGGER.exception`?  As it stands, `f'ERROR: {e}'` will only capture the text of the exception, which with GDAL isn't usually all that helpful without the traceback (which `LOGGER.exception` does capture):

```
>>> try:
...     raster = gdal.Open('foo.tif')
...     band = raster.GetRasterBand(1)
... except Exception as e:
...     print(f'{e}')
...
'NoneType' object has no attribute 'GetRasterBand'
```

src/natcap/invest/ui_server.py
```diff
@@ -102,6 +104,63 @@ def get_invest_validate():
...
+    may_have_points = True
+    # a lot of times the path will be empty so don't even try to open it
+    if vector_path:
+        try:
+            may_have_points = delineateit._vector_may_contain_points(
+                vector_path)
+        except Exception as e:
+            LOGGER.error(
+                f'Could not tell if vector {vector_path} may contain points. '
+                f'ERROR: {e}')
```
phargogh, line 159: Same suggestion here about `logger.Exception`

tests/test_ui_server.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_get_vector_may_have_points(self):
```
phargogh, line 30: Docstring?

phargogh (changes requested):
Thanks, Emily!  I had just a couple of minor comments and a question or two ... let me know what you think!

tests/test_ui_server.py
emlys: I was thinking to make it return the value that's relevant to the UI (an empty list) but you're right that it would make more sense to be an error, and then have the UI handle that error. Probably a `422` status code? (https://stackoverflow.com/questions/7939137/right-http-status-code-to-wrong-input)

emlys:
Thanks @phargogh! I changed the endpoints to return `422` if the path isn't a valid vector file. I also moved the `_vector_may_contain_points` logic into `ui_server.py` and modified so it will also return an error if the path isn't valid.

src/natcap/invest/ui_server.py
emlys: Sounds good!

phargogh (changes requested):
Thanks, @emlys !  This all looks good, but I see that the delineateIt UI is still pointing to the `delineateit._vector_may_contain_points` function.  I know the target branch here is `release/workbench_alpha`, but I'd worry that this will cause headhaches for us down the road.

Other than that, I think these changes look good!

emlys:
@phargogh I put the function back in delineateit, so both UIs will work

phargogh (approved):

PR #448, opened 2021-01-20
Bugfix/384 dev build hangs on big sur

### LINKED ISSUE #384: InVEST 3.9.0 development build hangs on Mac OS X Big Sur

phargogh:
# Description

This PR (which is a follow-up to #392 ) fixes an issue with launching the InVEST application on Mac OS X Big Sur, which is a stopgap issue until the Qt folks can release an update for this (see link in https://github.com/natcap/invest/issues/384#issuecomment-729107894 for further discussion about Qt build requirements).  It turns out that simply defining the environment variable in the pyinstaller runtime hook wasn't enough, and the most effective workaround is to have the variable defined in the parent shell itself.

There are 2 different behaviors for this environment variable addressed here:

1. **If we're running `InVEST.app` binary bundle** - the environment variable is set in the init script that calls the pyinstaller binary itself.
2. **If we're running InVEST through python** - a `RuntimeWarning` is raised noting that this _might_ be an issue, and, if it is, how to define the variable.

I've tested this fix locally on my Big Sur install, seems to work OK.  The same fix has also worked for the ROOT binaries.

@emlys if you get a chance, could you try this out on your mac as well to make sure it works in High Sierra?  Thanks in advance! 

Fixes #384 

# Checklist

- [x] Updated the user's guide (if needed) _(not needed)_

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -40,6 +40,12 @@ Unreleased Changes (3.9.1)
     * Restructured API reference docs and removed outdated and redundant pages.
     * Include logger name in the logging format. This is helpful for the cython
       modules, which can't log module, function, or line number info.
+    * Fixed an issue with the InVEST application launching on Mac OS X 11
+      "Big Sur".  When launching the InVEST ``.app`` bundle, the environment
+      variable ``QT_MAC_WANTS_LAYER`` is defined.  If running InVEST through
+      python, this environment variable may need to be defined by hand like
+      so: ``QT_MAC_WANTS_LAYER=1 python -m natcap.invest``.  A warning will
+      be raised if this environment variable is not present on mac.
 * Fisheries Habitat Scenario Tool
     * Fixed divide-by-zero bug that was causing a RuntimeWarning in the logs.
       This bug did not affect the output.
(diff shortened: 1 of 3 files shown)


emlys (approved):
Thanks @phargogh! It's working for me on Catalina. Interestingly I'm no longer getting the first "do you really want to open this" warning message (when opening the dmg). Still happens when launching the app for the first time.

phargogh:
> It's working for me on Catalina.

Thanks for verifying this!  I'm glad it's still working.

> Interestingly I'm no longer getting the first "do you really want to open this" warning message (when opening the dmg). Still happens when launching the app for the first time.

You know, the same thing happened to me on Big Sur, though not sure why that would be.

PR #449, opened 2021-01-21
Task/GitHub actions dependency transparency

dcdenu4:
# Description
Adds steps to the `run-model-tests` workflows. The added steps are:
1) List conda environment once setup. Export conda env to `./conda-env.txt`
2) Download an archived `conda-env.txt` from the last successful workflow if one exists. Does not fail workflow on error, such as if no archive can be downloaded. Uses `continue-on-error`.
3) Compare the two conda envs, listing the diff between them. Again does not fail workflow on error, uses `continue-on-error`.
4) Archive the current `conda-env.txt`.

See my fork Actions for a working version of what to expect:
https://github.com/dcdenu4/invest/runs/1744349750?check_suite_focus=true#step:11:1

# Checklist
- [ ] Updated HISTORY.rst (if these changes are user-facing) - Not user facing.

- [ ] Updated the user's guide (if needed) - Not needed.


.github/workflows/build-and-test.yml
```diff
@@ -103,17 +103,43 @@ jobs:
...
+              run: |
+                  conda list
+                  conda list --export > conda-env.txt
+
+            - name: Download Conda Env Artifact
+              continue-on-error: true
+              uses: dawidd6/action-download-artifact@v2
+              with:
+                  workflow: build-and-test.yml
+                  workflow_conclusion: success
```
emlys, line 117: I would put a comment here like "this gets the artifact from the last workflow run that succeeded" just because the meaning of the `workflow_conclusion` option wasn't super obvious to me

emlys (commented):
Thanks @dcdenu4 , this will be really helpful when dependency issues come up again! The tests are failing because of the pandas issue, but the steps that were changed in this PR are passing. To be safe should we wait to merge until that's resolved though?

dcdenu4:
> To be safe should we wait to merge until that's resolved though?

Yea! I'll merge in @davemfish fix for pandas.

emlys (approved):

.github/workflows/build-and-test.yml
```diff
@@ -103,17 +103,46 @@ jobs:
                     requirements-gui.txt > requirements-all.yml
                   conda env update --file requirements-all.yml
 
+            - name: Conda Environment Listing
+              shell: bash -l {0}
+              run: |
+                  conda list
+                  conda list --export > conda-env.txt
```
davemfish, line 110: Is this double `conda list` intentional?

davemfish (approved):
It was pretty cool when tests failed during this PR and then your new diff step alerted you to the new pandas bugfix release, and then we fixed that compatibility issue real quick in a separate PR, and now all checks are passing again here :)

phargogh (approved):
This looks _super_ useful, and pretty great that the pandas change was so easily highlighted!  🎉

.github/workflows/build-and-test.yml
dcdenu4: Yes! to "print" for viewing in Github Actions to see the full conda environment being used for the build and tests. And the second one to save to a file to compare against and archive later.

PR #460, opened 2021-01-26
Bugfix/UI server 455 logging from flask

### LINKED ISSUE #455: ui_server is not logging to stdout
`LOGGER` messages issued from `ui_server` do not appear in stdout. We need to fix that because the workbench listens to stdout in order to transport those messages through to the workbench's node logger.

davemfish:
This PR sets up a sys.stdout StreamHandler for logging from the Flask app, roughly following the "Other Libraries" section of these docs https://flask.palletsprojects.com/en/1.1.x/logging/. 

Previously, using `logging.basicConfig` left some mysterious behavior where messages weren't showing up in stdout or stderr, but sometimes would appear there only when the program closed. 

Fixes #455 

# Checklist

### DIFF
--- a/src/natcap/invest/ui_server.py
+++ b/src/natcap/invest/ui_server.py
@@ -6,17 +6,20 @@ import importlib
 import json
 import logging
 import pprint
+import sys
 import textwrap
 
 from flask import Flask
 from flask import request
-from flask.logging import default_handler
 import natcap.invest.cli
 import natcap.invest.datastack
+from natcap.invest.utils import LOG_FMT
 
-logging.basicConfig(level=logging.DEBUG)
 LOGGER = logging.getLogger(__name__)
-LOGGER.addHandler(default_handler)
+handler = logging.StreamHandler(sys.stdout)
+formatter = logging.Formatter(LOG_FMT)
+handler.setFormatter(formatter)
+LOGGER.addHandler(handler)
 
 app = Flask(__name__)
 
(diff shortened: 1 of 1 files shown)


src/natcap/invest/ui_server.py
```diff
@@ -6,15 +6,20 @@
...
 import natcap.invest.cli
 import natcap.invest.datastack
+from natcap.invest.utils import LOG_FMT
 
-logging.basicConfig(level=logging.DEBUG)
 LOGGER = logging.getLogger(__name__)
+handler = logging.StreamHandler(sys.stdout)
+formatter = logging.Formatter(LOG_FMT)
+handler.setFormatter(formatter)
+LOGGER.addHandler(handler)
```
phargogh, line 22: This makes sense to me for the context of the `ui_server` application, but I did want to check about the intent ...  `logging.basicConfig` sets up the root logger (if the root logger doesn't already have a handler), and this change will set up a logger for `natcap.invest.ui_server`, which means that the logger as configured will only capture logging from the application, not from any models, utilities, or anything else like that.  Is that the intent?

phargogh (commented):
Hey @davemfish The change looks good to me, but I wanted to check about the intent of the change and the expected logging printed to stdout.

src/natcap/invest/ui_server.py
davemfish: Thanks, @phargogh . No that is not the intent. The intent is definitely to capture all the messages from all the modules and stream them to stdout. I'll give it another go.

davemfish:
@phargogh I'm closing this because it turns out the issue behind this PR is invalid (#455 )

PR #469, opened 2021-02-03
Bugfix/431 mac launch failure space in filepath

### LINKED ISSUE #431: InVEST on Mac OS won't launch if there's a space in the filepath
To reproduce:

1. Create `/Applications/InVEST 3.9.0`
2. Download the 3.9.0 release
3. Open the DMG and copy the binary into the folder created in the first step.
4. Observe that the application won't launch

To correct this behavior:

1. Rename `/Applications/InVEST 3.9.0` to `/Applications/InVEST-3.9.0`
2. Launch application, everything's OK

Experienced on Big Sur, but I suspect this will also be an issue on other OS versions.

phargogh:
# Description

This PR updates the method used to determine the directory of the InVEST application bundle in a way that is more tolerant to spaces.

I've tested that this works with a space in the filename (`InVEST 3.app`) on my mac.

Fixes #431 

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -47,6 +47,8 @@ Unreleased Changes (3.9.1)
       python, this environment variable may need to be defined by hand like
       so: ``QT_MAC_WANTS_LAYER=1 python -m natcap.invest``.  A warning will
       be raised if this environment variable is not present on mac.
+    * Fixing an issue on Mac OS X where saving the InVEST application to a
+      filepath containing spaces would prevent the application from launching.
 * Fisheries Habitat Scenario Tool
     * Fixed divide-by-zero bug that was causing a RuntimeWarning in the logs.
       This bug did not affect the output.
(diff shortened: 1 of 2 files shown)


emlys (approved):
I don't really understand why this solves the space-in-filepath issue, but it does! I was looking at [this article](https://mywiki.wooledge.org/BashFAQ/028) linked to in a response to that stackoverflow post, and it suggests there's no perfect solution for getting the script location. Would it be any better to pass in the invest binary location as another argument to `build_app_bundle.sh`? Or as an env variable in the Makefile? Maybe it doesn't matter for this.

phargogh:
@emlys I hadn't read all the way through that article, and yes!  There's no perfect solution here, but I think the `BASH_SOURCE` approach (which is used in this modified version) is the right one for this case, as we know in advance that we're working with a bash script on disk and we're trying to get a path to a file relative to the location of the script.  The `$0` approach is clearly problematic, which is what we had before.

> First and foremost, try to avoid the problem entirely by not depending on the location of your script! 

As the article points out, this is the preferable solution!  But doing this will also require some reworking of the pyinstaller binaries.  It'd probably still be worth doing at some point, but it seems like a lower priority.

For your questions:

> Would it be any better to pass in the invest binary location as another argument to `build_app_bundle.sh`? Or as an env variable in the Makefile?

It absolutely could be if we could depend on the location of `InVEST.app`!  But I don't think we can depend on the location.  Even if the user were to always put `InVEST.app` into `/Applications` as the DMG so nicely suggests, OSX's App Sandboxing relocates the application into a different place on the filesystem each time it executes, which would break any statically-defined link.  Was that what you were thinking of @emlys ?

phargogh:
I just merged in the latest changes on `main` to make sure that the RTD build finishes.  After that, back to you @emlys !

emlys:
Ah okay, that makes sense!

PR #475, opened 2021-02-08
Raise when AOI, Wind Data don't intersect

### LINKED ISSUE #13: Wind Energy fails when no grid points provided intersect the AOI
Mentioned on the forums: https://community.naturalcapitalproject.org/t/error-running-offshore-wind-energy-with-custom-aoi-and-custom-grid-land-points/513

Original issue: https://bitbucket.org/natcap/invest/issues/3922/wind-energy-fails-when-no-grid-points

Apparently I was already working on this in my fork on the `bugfix/WINDENERGY-3922-error-when-no-grid-pts-intersect-aoi` branch.  Might want to dig that out.

dcdenu4:
# Description
Wind Energy would crash when the AOI and Wind Data did not intersect. Because of how Wind Energy is structured I decided to raise a `ValueError ` in `clip_vector_by_vector` when the `target_layer` feature count was 0. 

Added a test with Wind Data points outside the AOI to make sure a `ValueError` is raised.

Fixes #13 

# Checklist

### DIFF
--- a/src/natcap/invest/wind_energy.py
+++ b/src/natcap/invest/wind_energy.py
@@ -2405,6 +2405,11 @@ def _clip_vector_by_vector(
         base_layer_defn.GetName(), base_layer.GetSpatialRef(), base_geom_type)
     base_layer.Clip(clip_layer, target_layer)
 
+    empty_clip = False
+    if target_layer.GetFeatureCount() <= 0:
+        empty_clip = True
+
+    # Allow function to clean up resources
     target_layer = None
     target_vector = None
     clip_vector = None
@@ -2414,6 +2419,12 @@ def _clip_vector_by_vector(
 
     if base_sr_wkt != target_sr_wkt:
         shutil.rmtree(temp_dir, ignore_errors=True)
+
+    if empty_clip:
+        raise ValueError(
+            f"Clipping {base_vector_path} by {clip_vector_path} returned 0"
+            " features. If an AOI was provided this could mean the AOI and"
+            " Wind Data do not intersect spatially.")
 
     LOGGER.info('Finished _clip_vector_by_vector')
 
--- a/src/natcap/invest/wind_energy.py
+++ b/src/natcap/invest/wind_energy.py
@@ -2406,6 +2406,9 @@ def _clip_vector_by_vector(
     base_layer.Clip(clip_layer, target_layer)
 
     empty_clip = False
+    # Check if the feature count is less than 1, indicating the two vectors
+    # did not intersect. This will raise a ValueError below. GetFeatureCount
+    # can return -1 if the count is not known or too expensive to compute.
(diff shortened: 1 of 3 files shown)


src/natcap/invest/wind_energy.py
```diff
@@ -2405,6 +2405,11 @@ def _clip_vector_by_vector(
         base_layer_defn.GetName(), base_layer.GetSpatialRef(), base_geom_type)
     base_layer.Clip(clip_layer, target_layer)
 
+    empty_clip = False
+    if target_layer.GetFeatureCount() <= 0:
```
phargogh, line 2409: What would cause `GetFeatureCount()` to return something `< 0`?

phargogh: Oh, [I see](https://gdal.org/python/osgeo.ogr.Layer-class.html#GetFeatureCount) what could cause this.  Could you add a comment about the `< 0` case here for future us?

tests/test_wind_energy.py
```diff
@@ -734,6 +734,32 @@ def test_missing_valuation_params(self):
...
+            open_table.write('-58.5,24.5,7.59,2.6,10\n')
+            open_table.write('-58.95,24.95,7.59,2.6,10\n')
+            open_table.write('-57.95,24.95,7.59,2.6,10\n')
+            open_table.write('-57.95,25.95,7.59,2.6,10\n')
+
+        args['wind_data_path'] = wind_data_csv
+
+        # AOI and wind data should not overlap, leading to a ValueError in 
+        # clip_vector_by_vector
+        self.assertRaises(ValueError, wind_energy.execute, args)
```
phargogh, line 761: Ah, sorry, would it be possible to check that the error string is what we expect it is?  It looks like the model has ~5 places where `ValueError` has been raised, and it'd be great to be extra sure that we're catching the right `ValueError`.

phargogh (changes requested):
Thanks for following up on this @dcdenu4 !  I just had a couple of really minor comments, but let me know what you think?

src/natcap/invest/wind_energy.py
dcdenu4: Added a comment here: https://github.com/natcap/invest/pull/475/commits/a105ac775642caaf175093dc130a9257b86e2901. Honestly, I think I meant to write `== 0`, and just got lucky about it being possible to be `-1` 😄

tests/test_wind_energy.py
dcdenu4: Added a check here: https://github.com/natcap/invest/pull/475/commits/a105ac775642caaf175093dc130a9257b86e2901

dcdenu4:
Back to you @phargogh !

phargogh (approved):
Thanks @dcdenu4 !

Issue #476, opened 2021-02-08
GEOS and PROJ bump causing Wave Energy / Coastal Vulnerability tests to fail


dcdenu4:
I noticed tests failing not related to PR #475 . When looking at the conda env diff, I saw some important updates. 

When updating to Proj=7.2.0 in Conda locally I could reproduce the test failures and I also got the dependent updates:

```
The following packages will be UPDATED:

  cairo                                1.16.0-hd28d34b_1006 --> 1.16.0-hba8bd2f_1007
  expat                                    2.2.9-h33f27b4_2 --> 2.2.10-h39d44d4_0
  gdal                                 3.2.1-py38hb9e1232_0 --> 3.2.1-py38hb9e1232_4
  geos                                     3.8.1-he025d50_0 --> 3.9.0-h39d44d4_0
  geotiff                                  1.6.0-h8884d1a_3 --> 1.6.0-h5d775bf_4
  icu                                       67.1-h33f27b4_0 --> 68.1-h0e60522_0
  libclang                        10.0.1-default_hf44288c_1 --> 11.0.1-default_h5c34c98_1
  libgdal                                  3.2.1-h0d43894_0 --> 3.2.1-hd376e63_4
  libspatialite                            5.0.0-hf693123_0 --> 5.0.1-h6966176_2
  openjpeg                                 2.3.1-h48faf41_3 --> 2.4.0-h48faf41_0
  poppler                                 0.89.0-h5d62644_1 --> 0.89.0-h3772339_5
  proj                                     7.1.1-h7d85306_3 --> 7.2.0-h3e70539_1
  qt                                      5.12.9-hb2cf2c5_0 --> 5.12.9-h5909a2a_3
  shapely                              1.7.1-py38hc96c142_1 --> 1.7.1-py38hc55ee08_2
  tiledb                                   2.1.6-hf84e3da_1 --> 2.2.3-hf84e3da_0
```

dcdenu4:
The CV error makes it looks like a `shapely.ops.linemerge` call is failing, though the Shapely version has remained the same but is compiled against the newer dependencies. I see `geos` and `libspatiallite` are also bumped a version.

dcdenu4:
PROJ release notes: https://proj.org/news.html

I can't find anything helpful for GEOS change logs.

dcdenu4:
[...]

PR #478, opened 2021-02-08
Create upper bound for geos and bump PROJ

dcdenu4:
# Description
This PR attempts to hotfix some critical issues with dependencies that came up here: #476 . They relate to GEOS and PROJ libraries.

This PR  sets an upper bound `geos<3.9.0` to avoid a confirmed bug in GEOS https://github.com/libgeos/geos/issues/408 .

This PR also updates a Wave Energy test for PROJ 7.2.0 and sets a GDAL lower bound to `3.2.1` that will grab PROJ 7.2.0+. I believe this precision difference is due to improvements on PROJ handling coordindate transformations. 

*NOTE*: There are still conda builds of GDAL 3.2.1 that support PROJ < 7.2.0, but I think it'll grab the later version for us consistently. We just have no direct handle on PROJ without using conda.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,11 +12,10 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-GDAL>=3.1.2
+GDAL>=3.2.1
 Pyro4==4.77  # pip-only
 pandas>=1.2.1
 geos<3.9.0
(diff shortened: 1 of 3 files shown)


dcdenu4:
Closing this in preference of skipping tests.

PR #480, opened 2021-02-08
Remove deprecated numpy.float and numpy.bool

### LINKED ISSUE #255: Resolve deprecation warnings in tests

emlys:
# Description
Fixes #255 
> DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.

> DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.

Just replacing these.

### DIFF
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -833,11 +833,10 @@ def _calculate_tropical_forest_edge_carbon_map(
         col_coords, row_coords = numpy.meshgrid(col_range, row_range)
 
         # query nearest points for every point in the grid
-        # n_jobs=-1 means use all available CPUs
+        # workers=-1 means use all available CPUs
         coord_points = list(zip(
             row_coords[valid_edge_distance_mask].ravel(),
             col_coords[valid_edge_distance_mask].ravel()))
-        # note, the 'n_jobs' parameter was introduced in SciPy 0.16.0
         # for each forest point x, for each of its k nearest neighbors
         # shape of distances and indexes: (x, k)
         distances, indexes = kd_tree.query(
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -412,7 +412,7 @@ def combine_carbon_maps(*carbon_maps):
 
     """
     result = numpy.zeros(carbon_maps[0].shape)
-    nodata_mask = numpy.empty(carbon_maps[0].shape, dtype=numpy.bool)
+    nodata_mask = numpy.empty(carbon_maps[0].shape, dtype=bool)
     nodata_mask[:] = True
     for carbon_map in carbon_maps:
         valid_mask = carbon_map != NODATA_VALUE
@@ -842,7 +842,7 @@ def _calculate_tropical_forest_edge_carbon_map(
         # shape of distances and indexes: (x, k)
         distances, indexes = kd_tree.query(
             coord_points, k=n_nearest_model_points,
-            distance_upper_bound=DISTANCE_UPPER_BOUND, n_jobs=-1)
+            distance_upper_bound=DISTANCE_UPPER_BOUND, workers=-1)
 
         if n_nearest_model_points == 1:
             distances = distances.reshape(distances.shape[0], 1)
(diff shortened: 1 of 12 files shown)


src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -842,7 +842,7 @@ def _calculate_tropical_forest_edge_carbon_map(
         # shape of distances and indexes: (x, k)
         distances, indexes = kd_tree.query(
             coord_points, k=n_nearest_model_points,
-            distance_upper_bound=DISTANCE_UPPER_BOUND, n_jobs=-1)
+            distance_upper_bound=DISTANCE_UPPER_BOUND, workers=-1)
```
dcdenu4, line 845: I think there are some comments above this call that reference `n_jobs`, could you update those?

dcdenu4: Also, does this mean we should set a lower bound on Scipy?

dcdenu4 (changes requested):
Hey @emlys thanks for refactoring this. I noticed that a forest edge effect test was failing. It looks related to the `workers` keyword in the `query` op. Could this be a Scipy version issue?

emlys:
Oh yep, it's `n_jobs` for older versions and `workers` for scipy>=1.6.0, the latest version. And it seems they're not phasing in/out the options like python tends to do. We would need to require scipy>=1.6.0 for this to work. Is that OK given that we've been allowing a wide range of scipy versions?
scipy>=1.6.0 doesn't break anything so maybe it's fine

dcdenu4:
>  Is that OK given that we've been allowing a wide range of scipy versions?

Great question... I'd be curious what @phargogh thinks, but if the only reason to bump the lower bound for Scipy is to avoid this warning... I'd probably hold off on changing this for now. I wonder if there would be users who wouldn't necessarily want to be on the latest greatest Scipy version... But, like Rich said, who really are our python library users? Probably just us and our team.

emlys:
It's unfortunate that the versions are mutually exclusive, but if we have to choose one, I'd lean towards the latest version. Have there been issues before with people unable to use the invest package due to requirements conflicts?

dcdenu4:
Are the version mutually exclusive? If we left the `n_jobs` it just raises a deprecation warning no? But I could still use the latest version right?

I'm usually all about the latest and greatest and would agree with just bumping the version. But I feel like I normally miss something... and  @phargogh typically enlightens me haha

emlys:
Ah, the warning says that `The old name "n_jobs" will stop working in SciPy 1.8.0.`. I guess I'd rather not wait until 1.8.0 comes out and the tests unexpectedly break, but we don't have to fix it right now either... I was trying to close #255 but it's really an eternal issue :/

dcdenu4:
Yea, it looks like Scipy 1.6.0 just came out in late December. Hmmm. Well, the only reason I can think of to not go ahead and bump it is that users will not have upgraded to Scipy 1.6.0 because of it's recent release. There seems to be a lot of good changes in 1.6.0. Let's go for it!

dcdenu4:
I think my only change then is the comments referring to `n_jobs` instead of `n_workers`

emlys:
thanks @dcdenu4! Just updated those comments (removed one since it doesn't seem relevant any more)

dcdenu4:
Great! I'll Approve once the tests go through.

dcdenu4 (approved):
Thanks!

phargogh:
Yeah, I'm usually hesitant to upgrade the requirement for larger binary packages, but we shouldn't go further out of our way than really needed to maintain two parallel versions of a package without a clear use case benefiting from it.

Besides, it's gotten _so_ easy in the last few years to procure binary distributions of packages like `scipy` that requiring a later version shouldn't overall be an issue.

PR #484, opened 2021-02-10
Allow non-admin Windows users to download and run InVEST

### LINKED ISSUE #21: InVEST Requires Windows Admin Privileges
See if there is a way to get around having to be an admin to install InVEST

not sure I'll get to this. I did do some research and it seems possible windows bumps admin privileges when finding words "setup", "install", etc..

Originally reported by @dcdenu4 at https://bitbucket.org/natcap/invest/issues/3498/invest-requires-windows-admin-privileges

dcdenu4:
# Description
This PR allows Windows users without admin privileges to download, install, run, and uninstall InVEST.

I created a non admin account on my PC and was able to confirm that the user did not have admin privileges and was able to download and install InVEST on Windows 10.

Fixes #21 

# Checklist

I'm not sure if Users's Guide mentions anything about this. I'll check. UPDATE: grep didn't turn anything up, so not needed.


HISTORY.rst
```diff
@@ -52,6 +52,8 @@ Unreleased Changes (3.9.1)
     * Fixed an issue on Mac OS when certain models would loop indefinitely and
       never complete.  This was addressed by bumping the ``taskgraph``
       requirement version to ``0.10.3``
+    * Allow Windows users without admin priveleges to download and install
```
davemfish, line 55: typo: privileges

installer/windows/invest_installer.nsi
```diff
@@ -298,7 +298,7 @@ FunctionEnd
 
 Section "InVEST Tools" Section_InVEST_Tools
     AddSize 230793  ; This size is based on Windows build of InVEST 3.4.0
-    SetShellVarContext all
+    SetShellVarContext current
```
davemfish, line 301: Mainly out of curiosity, did this change anything about which parts of the filesystem are used? Either the install location or the application's data (e.g. `AppData/Local/Roaming/...` or whatever the heck that path is)?

installer/windows/invest_installer.nsi
```diff
@@ -406,7 +406,7 @@ SectionEnd
 Section "uninstall"
   ; Need to enforce execution level as admin.  See
   ; nsis.sourceforge.net/Shortcuts_removal_fails_on_Windows_Vista
-  SetShellVarContext all
+  SetShellVarContext current
```
davemfish, line 409: Can you confirm that the uninstall worked? On Vista? ;) Kidding, can we remove that comment?

davemfish (changes requested):
Great idea, @dcdenu4 ! I found a typo in History and a couple questions that came to mind.

installer/windows/invest_installer.nsi
dcdenu4: Oh that's a great question Dave! I don't think this should effect that. But `SetShellVarContext` set to `all` meant that the user could not write the startup menu shortcuts in `ProgramData` because they didn't have permission to do this for All Users. 

We should definitely test what happens when running InVEST models and where we normally expect these folders to be.

Do we have a good handle on all the local paths InVEST uses for temp folders, caches, etc... ?

dcdenu4: Oh, here's the resource: https://nsis.sourceforge.io/Reference/SetShellVarContext

installer/windows/invest_installer.nsi
dcdenu4: Hahah, yes WE CAN. This DOES bring up an interesting point... This might only work on Windows 10, not 7... Are we supporting Windows 7 still? It looks like Microsoft doesn't anymore, so I'm inclined not to.
https://support.microsoft.com/en-us/windows/windows-7-support-ended-on-january-14-2020-b75d4580-2cc7-895a-2c9c-1466d9a53962

dcdenu4: Oh, and yes, I was able to uninstall the program. It'd be great if someone else who has access to a non-admin user could try as well I suppose.

installer/windows/invest_installer.nsi
dcdenu4: @davemfish I ran some models and checked cache folders. Folder being used:
>../AppData/Roaming/Natural Capital Project/

This seems to line up with what I'm seeing on my admin side.

For some reason I thought the non-admin user was also having start up shortcuts go to `ProgramData`, but apparently not. Which makes sense since they don't have permission to write there. Instead they are going to:
>.../AppData/Roaming/Microsoft/Windows/Start Menu/Programs/

So, I think everything is okay.

HISTORY.rst
dcdenu4: Fixed: https://github.com/natcap/invest/pull/484/commits/af0ce411b87d56a87272d09839d0bca219908151

installer/windows/invest_installer.nsi
dcdenu4: Removed old comments and added a new one: https://github.com/natcap/invest/pull/484/commits/d1db4d6fc9cbad86d02ada28e9594f8307a2db79

installer/windows/invest_installer.nsi
davemfish: Thanks for checking on that! I think everything is okay, too.

installer/windows/invest_installer.nsi
davemfish: Okay I'll give it a try as well.

davemfish: Well I made a non-admin account and installed the artifact from this PR. I tried the uninstall shortcut that's in the installation directory and Windows kept telling me that it can't uninstall because InVEST is still running. But I also got that when I tried the same from my Admin account with an old version of invest. 

I've probably never tried that shortcut before. I usually go to "Add/Remove Programs" menu on Windows. Checking that menu on either account - it does not even list the newly installed invest (from this PR's build). The installation seems to work though - I could open a model.

dcdenu4: >I tried the uninstall shortcut that's in the installation directory and Windows kept telling me that it can't uninstall because InVEST is still running

Huh, I didn't have any problem uninstalling this way.

>I usually go to "Add/Remove Programs" menu on Windows

I don't think I've tried this for InVEST! I'll see if I get the same behavior.

phargogh: Ugh, I can't seem to get the installer to download from github actions, so I can't test this out right now (I'll try again later).

Hm, this is interesting: the NSIS docs for [`RequestExecutionLevel`](https://nsis.sourceforge.io/Reference/RequestExecutionLevel) suggest that if we need to write to HKLM in the windows registry (the InVEST installer does), then the `admin` level might be required.  So @davemfish, maybe that's why you aren't able to see it in the Add/Remove programs list?  That list is populate from the Windows Registry.

In any case, it sounds like even if the user doesn't have admin privileges, then the only difference is that they won't be able to uninstall InVEST from add/remove programs.

> Are we supporting Windows 7 still? It looks like Microsoft doesn't anymore, so I'm inclined not to.

Yeah, I don't think we should deliberately support Windows 7 given its age.  It might well still work, but no need to go out of our way to verify it does.

dcdenu4: I also don't see InVEST in the `Add/Remove Programs` tool for non-admin. Let me see if I can fix that.

davemfish: I wouldn't be surprised if I just happened to get Windows into a state where it thought invest was still running - that's wayy too common on Windows - as opposed to some property of the new installer actually being responsible.

phargogh: In truth, there are lots of reasons why that could be!  The uninstaller only ever looks for a process named `invest.exe`, which could be from a totally different version, could be a zombie process, or could even be something not related to InVEST at all!  It's kind of a crude check.  I wouldn't be opposed to removing it.

installer/windows/invest_installer.nsi
phargogh: Oooo I think we might have a pickle here: institutional users.  Perrine wrote to me a few months back about installing InVEST in a computer lab for a class.  IT was going to perform the install.  If the Start Menu shortcuts will be written to `C:\Users\<IT User>\AppData\...`, then anyone _other_ than IT will need to launch the application from the CLI.

Some installers out there have an option to "Install for all users" and then "install for this user only" ... maybe NSIS has a standard way to do something similar?

dcdenu4:
Hey @davemfish and @phargogh , thanks for the feedback. I'm going to take some time and try to do a refactor on our NSIS script to allow for `Install for All Users` for admin and `Install for User` for non-admin. ( or admin! ). I see James echoed this sentiment and we discussed it a little today.

dcdenu4:
Update
--------

This PR now allows for two installation modes: All Users and Current User. With admin privileges a user can install for the whole machine or can install locally for that user. A non-admin user can install locally and can uninstall using the uninstaller (uninstalling via Apps and Features is bugged in Windows 10 and elevates privileges. I don't have a good work around for this.)

Installing as admin for All Users:
- InVEST is put in `Program Files`
- Start menu shortcuts are put in `Program Data/Microsoft/Windows/Start Menu/Programs/InVEST_version`

Installing as current user:
- InVEST is put in `AppData/Local/InVEST_version`
- Start menu shortcuts are put in `AppData/Roaming/Microsoft/Windows/Start Menu/Programs/InVEST_version`

dcdenu4:
@davemfish and @phargogh I think this is worth another look. Please see my Update post above for the latest.

phargogh (approved):
Hey @dcdenu4, thanks for digging into this!  I don't have any suggestions about the new NSIS work ...  what's here makes sense and is working well!  It's also neat to see that the command-line flags also worked exactly as expected, and with a few extra flags for admin or non-admin installs, too.

For what it's worth, on my windows laptop I was able to do a non-admin install and non-admin uninstall from the Apps&Features window without issue at all ... it worked great!

dcdenu4:
> For what it's worth, on my windows laptop I was able to do a non-admin install and non-admin uninstall from the Apps&Features window without issue at all ... it worked great!

Woah! Awesome!

davemfish:
Hey @dcdenu4 , I just tried this out in a couple ways. The admin install for all users worked great - that's a really nice feature. The non-admin install showed me this popup right when I launched the installer from explorer.
installer

davemfish (changes requested):
Now that installs have a new default location (Program Files instead of C:) I think that's worth a note in History.

davemfish:
Some misc other things I noticed:

* The uninstaller shortcut triggers the UAC unknown publisher message and wants to run a program called "Un_A.exe". It works great though, and I see that the UAC message is not new, but previously we had a program name that was less scary: Uninstall_3.9.0.post128+g68a3e1c5.exe
* For the admin "just for me" install, the default install location has changed just like you mentioned. Do you think it's worth trying to put the default back to C:/InVEST....? There is some stuff in there that's useful to get at (e.g. sample data) and so I kinda like to the very "findable" location of "C:/" rather than "AppData/Local/Roaming"

dcdenu4:
@davemfish 
>The non-admin install showed me this popup right when I launched the installer from explorer.

That's not good, especially the inconsistency between your attempt, my attempt, and Jame's attempt...

>The uninstaller shortcut triggers the UAC unknown publisher message and wants to run a program called "Un_A.exe"

I know it! I'll try and dig out where this is coming from... I couldn't find a way to change this initially.

> For the admin "just for me" install, the default install location has changed just like you mentioned. Do you think it's worth trying to put the default back to C:/InVEST....?

You're definitely right in that those paths are not as convenient. However, I would rather default to Windows expected behavior and allow the user to change from that default if they prefer. The only thing the admin user couldn't change would be the start menu shortcut locations.

I also cringe when I see InVEST in root `C:`. I think `Program Files` is where it should live, never have I seen a piece of software want to clutter up my root `C:` space. It's like saving things directly in `root` on linux. But that's just me!

@davemfish and @phargogh could you post what version of Windows you're running? I'm not sure how to debug this difference in behavior.

davemfish:
> You're definitely right in that those paths are not as convenient. However, I would rather default to Windows expected behavior and allow the user to change from that default if they prefer. The only thing the admin user couldn't change would be the start menu shortcut locations.
> 
> I also cringe when I see InVEST in root `C:`. I think `Program Files` is where it should live, never have I seen a piece of software want to clutter up my root `C:` space. It's like saving things directly in `root` on linux. But that's just me!
> 

Yeah you're right about expected Windows behavior and cluttering the root drive, I'm sure others share your feelings there. `Program Files` makes sense and it's easy enough to find. It's _very_ lame that Windows puts a space in that path, but I'm sure our code is robust to that. It's mostly the `AppData/Local/InVEST_version` path that I don't much like, but you're right, that's the expected location. 

> @davemfish and @phargogh could you post what version of Windows you're running? I'm not sure how to debug this difference in behavior.

Windows 10 Pro Version	10.0.19041 Build 19041

dcdenu4:
> It's _very_ lame that Windows puts a space in that path

Definitely agree and dislike the Windows path naming convention.

dcdenu4:
I have Windows: `Microsoft Windows 10 Pro Version 10.0.19042 Build 19042`

phargogh:
> @davemfish and @phargogh could you post what version of Windows you're running? I'm not sure how to debug this difference in behavior.

I'm running Windows 10 Pro version 20H2, OS build 19042.804

> >  It's very lame that Windows puts a space in that path
>
> Definitely agree and dislike the Windows path naming convention.

Completely agree here!  InVEST does appear to be resilient to it (spot checking a few models, they work as expected), and access to sample data does seem to be the main reason to have the path be accessible.

FWIW, we did the same thing for OPAL (program and sample data stored in `C:\Program Files\OPAL...`) and the general reaction was relief that the application was now living in the expected place.  We didn't have many issues with people finding their sample data.  That might not be the same with storing the application in `AppData`, since that directory is hidden from the user by default.

> The uninstaller shortcut triggers the UAC unknown publisher message and wants to run a program called "Un_A.exe"

I, too, saw the "Un_A.exe" UAC warning!  So there's at least that consistency.

dcdenu4:
> I, too, saw the "Un_A.exe" UAC warning! So there's at least that consistency.

😆

dcdenu4:
It looks like James and I have the latest and greatest. I found some changes listed here: https://blogs.windows.com/windows-insider/2021/02/09/releasing-windows-10-build-19042-804-20h2-to-beta-and-release-preview-channels/.

>We fixed a possible elevation of privilege vulnerability in the win32k component.

Not sure if that's part of the reason...

dcdenu4:
Looks like for the `Un_A.exe` I found this to be the reason: http://forums.winamp.com/showthread.php?p=3195988
>The uninstaller creates a copy of itself and then tries to run that copy. It does this because a running exe cannot be deleted and you want to be able to delete the uninstaller in $instdir.
The security policy is the problem.

davemfish:
> http://forums.winamp.com/showthread.php?p=3195988

I wasn't expecting those posts to date from 2020 😆

davemfish:
> It looks like James and I have the latest and greatest. I found some changes listed here: https://blogs.windows.com/windows-insider/2021/02/09/releasing-windows-10-build-19042-804-20h2-to-beta-and-release-preview-channels/.
> 
> > We fixed a possible elevation of privilege vulnerability in the win32k component.
> 
> Not sure if that's part of the reason...

Windows decided I should update last night and the installer works as expected now! Hopefully there was a very limited set of windows versions where this wasn't working?

dcdenu4:
> Windows decided I should update last night and the installer works as expected now! Hopefully there was a very limited set of windows versions where this wasn't working?

Ahhh thanks @davemfish ! Ugh that makes me frustrated though about releasing this, if we can't guarantee it'll work across Windows 10 versions. It's also a wicked pain in the ass to test without having access to an environment to easily swap between Windows versions... I wish I could easily roll back my Windows version and decipher what was going wrong.

dcdenu4:
I'm looking into Windows Sandbox tool to see if that allows us to quickly test Windows 10 versions.

dcdenu4:
Update
--------

Windows Sandbox wasn't very helpful since I couldn't run anything in the VM Sandbox as anything but Admin.

I busted out my old laptop running ` Microsoft Windows 10 Home 10.0.18363` and it installed successfully for me under a no-admin user.

dcdenu4:
I also noticed that if I tried installing InVEST again after just installing, I would get the info message that Dave reported. I'm wondering if it's failing because there's already a version of that specific InVEST in the registry...

dcdenu4:
Okay, If I install as admin for all users and then try to install as a non-admin user I see the window Dave reported. @davemfish, do you think this is the order in which you might have tried things? I don't have that specific windows build but I'm seeing this behavior across multiple windows versions.

davemfish:
> Okay, If I install as admin for all users and then try to install as a non-admin user I see the window Dave reported. @davemfish, do you think this is the order in which you might have tried things? I don't have that specific windows build but I'm seeing this behavior across multiple windows versions.

Yes, that sounds like the same order of steps I did. Nice that it's reproducible @dcdenu4

dcdenu4:
Hey @davemfish and @phargogh,

Assuming the tests pass, could you try out this version of the windows installer: https://github.com/dcdenu4/invest/suites/2152699897/artifacts/44095900. 

Okay, tests failed, but it looks like it is a recreation issue.

The reason I think the failures were happening is because an InVEST version was already installed. Now it won't, and will gladly let you overwrite your install if you don't select a different folder.

davemfish:
> The reason I think the failures were happening is because an InVEST version was already installed. Now it won't, and will gladly let you overwrite your install if you don't select a different folder.

Worked great for me!

davemfish (approved):
Thanks for doing all that NSIS programming for us, @dcdenu4 !

phargogh (approved):
🚀 Thanks a ton for this, @dcdenu4 ... this is really, really exciting!  This works great for me as well, both as admin and non-admin and also reinstalling the application over the top of an existing install.  It all works!

The test failures are recmodel-related, which I think has been resolved separately.

PR #488, opened 2021-02-10
Handle validation message for accidental .ovr files

### LINKED ISSUE #50: Offer a more helpful validation message when a user provides an OVR file
This has happened on the forums a couple of times now, but here's the latest example: https://community.naturalcapitalproject.org/t/pollination-model-error-after-updating-to-invest-3-8-0-unable-to-initiate-run/856

When a user provides an overviews file instead of the GeoTiff it's associated with, the overviews file triggers the validation error stating that the file must be projected.  It would be nice if we could check if this file is an overviews file and provide a more specific and helpful error message for this common use case.

Marked as a 3 difficulty because the slightly more interesting thing will be figuring out how to build an overviews file within GDAL as part of the tests.  I think maybe @richpsharp has done that before?

dcdenu4:
# Description
Overview `.ovr` files pose a validation glitch when input instead of a proper `.tif` raster. These files can be opened as a raster but ultimately is invalid and fails when checking projection, because it's empty. 

This PR adds a validation check specifically for `.ovr` files and provides a more useful validation error message.

Fixes #50 

# Checklist

### DIFF
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -306,6 +306,9 @@ def check_raster(filepath, projected=False, projection_units=None):
 
     if gdal_dataset is None:
         return "File could not be opened as a GDAL raster"
+    # Check that an overview .ovr file wasn't opened.
+    if os.path.splitext(filepath)[1] == '.ovr':
+        return "File found to be an overview '.ovr' file."
 
     srs = osr.SpatialReference()
     srs.ImportFromWkt(gdal_dataset.GetProjection())
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -347,6 +347,30 @@ class RasterValidation(unittest.TestCase):
         error_msg = validation.check_raster(filepath)
         self.assertTrue('could not be opened as a GDAL raster' in error_msg)
 
+    def test_invalid_ovr_raster(self):
+        """Validation: test when a .tif.ovr file is input as a raster."""
+        from natcap.invest import validation
+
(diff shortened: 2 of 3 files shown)


src/natcap/invest/validation.py
```diff
@@ -306,6 +306,9 @@ def check_raster(filepath, projected=False, projection_units=None):
 
     if gdal_dataset is None:
         return "File could not be opened as a GDAL raster"
+    # Check that an overview .ovr file wasn't opened.
+    if os.path.splitext(filepath)[1] == '.ovr':
+        return "File found to be an overview '.ovr' file."
```
phargogh, line 311: Ha, I blanked on what `splitext` would do and had `.split` in my mind.

tests/test_validation.py
```diff
@@ -347,6 +347,30 @@ def test_invalid_raster(self):
...
+        meters_srs.ImportFromEPSG(32731)
+        raster.SetProjection(meters_srs.ExportToWkt())
+        raster = None
+        # I could only create overviews when opening the file, not on creation.
+        # Build overviews taken from:
+        # https://gis.stackexchange.com/questions/270498/compress-gtiff-external-overviews-with-gdal-api
+        raster = gdal.OpenEx(filepath)
+        gdal.SetConfigOption("COMPRESS_OVERVIEW", "DEFLATE")
+        raster.BuildOverviews("AVERAGE", [2, 4, 8, 16, 32, 64, 128, 256])
+        raster = None
```
phargogh, line 368: 🚀

phargogh (approved):
Thanks @dcdenu4 !

PR #492, opened 2021-02-12
Fix carbon valuation when rates are 0

emlys:
# Description
Fixes a bug in the carbon model valuation that came up in [this forum post](https://community.naturalcapitalproject.org/t/economic-value-raster-not-assigning-correct-values-to-pixels/1699). If rate change and discount rate are both 0, the calculated valuation constant became a rate per year instead of a total value, giving results that are off by a factor of `fut_year - cur_year`.

# Checklist

### DIFF
--- a/src/natcap/invest/carbon.py
+++ b/src/natcap/invest/carbon.py
@@ -497,15 +497,21 @@ def _calculate_valuation_constant(
         a floating point number that can be used to multiply a delta carbon
         storage value by to calculate NPV.
     """
-    n_years = int(lulc_fut_year) - int(lulc_cur_year) - 1
+    n_years = lulc_fut_year - lulc_cur_year
     ratio = (
-        1.0 / ((1 + float(discount_rate) / 100.0) *
-               (1 + float(rate_change) / 100.0)))
-    valuation_constant = (
-        float(price_per_metric_ton_of_c) /
-        (float(lulc_fut_year) - float(lulc_cur_year)))
-    if ratio != 1.0:
-        valuation_constant *= (1.0 - ratio ** (n_years + 1)) / (1.0 - ratio)
+        1 / ((1 + discount_rate / 100) * 
+             (1 + rate_change / 100)))
+    valuation_constant = (price_per_metric_ton_of_c / n_years)
+    # note: the valuation formula in the user's guide uses sum notation.
+    # here it's been simplified to remove the sum using the general rule
+    # sum(r^k) from k=0 to N  ==  (r^(N+1) - 1) / (r - 1)
+    # where N = n_years-1 and r = ratio
+    if ratio == 1:
+        # if ratio == 1, we would divide by zero in the equation below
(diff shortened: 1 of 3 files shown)


src/natcap/invest/carbon.py
```diff
@@ -497,15 +497,21 @@ def _calculate_valuation_constant(
...
+    n_years = lulc_fut_year - lulc_cur_year
     ratio = (
-        1.0 / ((1 + float(discount_rate) / 100.0) *
-               (1 + float(rate_change) / 100.0)))
-    valuation_constant = (
-        float(price_per_metric_ton_of_c) /
-        (float(lulc_fut_year) - float(lulc_cur_year)))
-    if ratio != 1.0:
-        valuation_constant *= (1.0 - ratio ** (n_years + 1)) / (1.0 - ratio)
+        1 / ((1 + discount_rate / 100) *
```
emlys, line 502: removed these types and `.0`s because we don't need them now in python 3

src/natcap/invest/carbon.py
```diff
@@ -497,15 +497,21 @@ def _calculate_valuation_constant(
...
-        (float(lulc_fut_year) - float(lulc_cur_year)))
-    if ratio != 1.0:
-        valuation_constant *= (1.0 - ratio ** (n_years + 1)) / (1.0 - ratio)
+        1 / ((1 + discount_rate / 100) * 
+             (1 + rate_change / 100)))
+    valuation_constant = (price_per_metric_ton_of_c / n_years)
+    # note: the valuation formula in the user's guide uses sum notation.
+    # here it's been simplified to remove the sum using the general rule
+    # sum(r^k) from k=0 to N  ==  (r^(N+1) - 1) / (r - 1)
+    # where N = n_years-1 and r = ratio
```
emlys, line 508: I was having trouble figuring out how the sum notation is equivalent to the math here

src/natcap/invest/carbon.py
```diff
@@ -497,15 +497,21 @@ def _calculate_valuation_constant(
...
+    if ratio == 1:
+        # if ratio == 1, we would divide by zero in the equation below
+        # so use the limit as ratio goes to 1, which is n_years
+        valuation_constant *= n_years
+    else:
+        valuation_constant *= (1 - ratio ** n_years) / (1 - ratio)
```
emlys, line 514: by some mathemagic, `(ratio ** n_years - 1) / (ratio - 1)` which you get using the general rule above is equivalent to `(1 - ratio ** n_years) / (1 - ratio)` here. This kinda blew my mind but it's been a while since I did math 🙃

src/natcap/invest/carbon.py
richpsharp: Fun fact, the original ArcGIS version of this model used a loop and I found that annoying. Never though someone would notice years later and I'm glad you did because this comment is great.

PR #499, opened 2021-02-24
Fixes Datastack bug when saving Windows paths

### LINKED ISSUE #52: Datastacks should be able to be loaded on any operating system
As a cross-platform developer, I sometimes need to create a datastack archive or parameter set on Windows or Mac OS X and then load the datastack on the other operating system.  When I create a datastack on mac and load it on Windows, it works well.  When I create a datastack on Windows and load it on mac, it doesn't.  Looks like the paths are being saved with `\\` separators and perhaps that's a part of it.

dcdenu4:
# Description
This PR fixes a bug related to saving a Datastack as a parameter set with relative paths under Windows. Windows paths were not being saved using Linux like path separators `/` but were instead saving with the Windows form `\\`. 

Windows paths are now saved with Linux like separators and testing has been updated to check for this change. In particular this PR introduces `@unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")` syntax to the tests. This helps test that Windows path are being saved like we expect, while allowing macOS or `darwin` OS to ignore and run its own version of the test.

Fixes #52

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -52,6 +52,8 @@ Unreleased Changes (3.9.1)
     * Fixed an issue on Mac OS when certain models would loop indefinitely and
       never complete.  This was addressed by bumping the ``taskgraph``
       requirement version to ``0.10.3``
+    * Fixed a bug where saving a datastack parameter set with relative paths
+      would not convert Windows separators to linux style.
 * Fisheries Habitat Scenario Tool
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Thanks @dcdenu4 !  This looks great.  The one test failure looks like a recmodel connection issue, completely unrelated to this PR.

PR #500, opened 2021-02-25
Feature/cli 237 save to python from cli

### LINKED ISSUE #237: Proposal: Move "Save as Python" from UI to datastack or CLI modules
There's a method `save_to_python` on `natcap.invest.ui.model.InVESTModel`. That method is useful, almost as-is, in the `invest-workbench`, except that the workbench avoids depending on anything from `natcap.invest.ui` because it's likely we'll eventually be able to deprecate the whole `ui` module. 

So here's a proposal to move the core functionality of `save_to_python` -- all the string formatting -- into `natcap.invest.datastack` or `natcap.invest.cli`. Then the current UI and the new workbench can both call on the same code for this feature.

An alternative proposal is to refactor this function and it's tests into javascript and include it directly in the workbench.

davemfish:
This PR moves - from the `ui` module to the `cli` module - the routine for saving a python script that calls the model's `execute`. It facilitates the workbench's need for this function while not relying on `natcap.invest.ui`.

Since the new function is under `natcap.invest.cli`, we decided to also make it available on the CLI via `invest export-py carbon -f myscript.py`

The `-f` is optional and `invest export-py carbon` will default to creating `carbon_execute.py` in the current working directory.

The function itself also takes an optional arguments dictionary to populate the values of `args` in the resulting script. Without this argument, `args` includes all the keys from the model's `ARGS_SPEC.args` with all values set to an empty string. This argument is not available to the CLI because I don't think there's a use for it there.

One minor thing I'd like feedback on: In the new tests, the coverage report is trying & failing to report on the python script that we created. The script itself is not created in the CWD, but coverage exits with:

"No source for code: 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpnkuqq5xp\foo.py'."

I can't reproduce this locally. I'm not sure, but I'm guessing coverage is trying to include this file because we import it during the test (in order to conveniently check it's `args` attribute), but remove it during teardown before coverage looks for it. Anyway, `coverage` suggests using the `-i` flag to just ignore errors when reading source files. So that's what I did for now. Any other ideas?

Fixes #237 

# Checklist
- [ x ] Updated HISTORY.rst (if these changes are user-facing)

- [ ] Updated the user's guide (if needed) - we don't document the CLI in the UG.


ui_tests/test_ui_inputs.py
```diff
@@ -2872,6 +2872,7 @@ def test_save_to_python(self):
             try:
                 spec = importlib.util.spec_from_file_location(module_name, python_file)
                 module = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(module)
```
davemfish, line 2875: We've been skipping this whole class of tests for a while, but I learned when I copied this test to `test_cli.py` that this line is now necessary.

dcdenu4 (commented):
Hey @davemfish, everything looks good to me but maybe I can help talk out the testing issue.

Is it possible to use a `.coveragerc` file to selectively `run` or `omit` our created python module? Maybe in this way we can instruct `.coverage` to just ignore it?
- https://coverage.readthedocs.io/en/coverage-5.4/config.html
- Looks like there is a `source_pkgs` setting as of `5.3.` Could we point to `invest` as what we'd like to cover? https://coverage.readthedocs.io/en/latest/changes.html
- The only other thing I can think of is to not import that module at all, but instead read it as text and do the compare that way? But that sounds like more work and not as graceful. 

Sorry if that wasn't too helpful!

davemfish:
> Is it possible to use a `.coveragerc` file to selectively `run` or `omit` our created python module? Maybe in this way we can instruct `.coverage` to just ignore it?
> 
> * https://coverage.readthedocs.io/en/coverage-5.4/config.html
> * Looks like there is a `source_pkgs` setting as of `5.3.` Could we point to `invest` as what we'd like to cover? https://coverage.readthedocs.io/en/latest/changes.html
> * The only other thing I can think of is to not import that module at all, but instead read it as text and do the compare that way? But that sounds like more work and not as graceful.

Hey Doug, these were great suggestions. I tried a variety of things to no avail before I realized we have even bigger issues with coverage reports. see #506 . We can always add it back properly if people find it useful, but until then this PR also fixes #506 by removing coverage reports from `make test`.

dcdenu4 (approved):
I'm fine with dropping the `coverage`. Thanks!

PR #505, opened 2021-02-26
CV tests sensitive to GEOS changes

davemfish:
This PR un-skips some CV tests now that geos has released a patch for a bug that was failing those tests. And our CV regression data is updated to accommodate other geos changes that happened to change the order that CV point features were created. This does not impact the geometry or data, just the order of the features in the layer.

Here's the test data PR: https://bitbucket.org/natcap/invest-test-data/pull-requests/20/updated-regression-data-for-the-complete

Addresses #476 - we should leave this open though, because the geos bugfix did not resolve the failing wave energy tests mentioned in that issue.

# Checklist


phargogh (commented):
Thanks for taking a look at this @dcdenu4 and I'm sorry for the delay @davemfish

PR #516, opened 2021-03-03
Fixes nodata masking for SWY qf.tif output

### LINKED ISSUE #470: SWY QF output has negative pixels on boundary
A user from the forums posted about negative `QF.tif` outputs in the SWY model. After running their data I did confirm that negative values were occurring on the boundaries of the raster. The values were `-11` and appeared to only occur where the underlying DEM was `nodata`. This seems like a masking or alignment issue. Here's a screenshot of `QF.tif` with negative values in red:

Forum post: https://community.naturalcapitalproject.org/t/seasonal-water-yields-negative-results/1633/8

Doug has the data if someone else dives into this.

dcdenu4:
# Description
A user was seeing some negative values (-11) in their `qf.tif` output. On examining it looked like these values were on the edges where the DEM was `nodata`. This PR fixes a nodata masking issue where during `qf` computation `streams` were not being used to mask out nodata values. In this case `streams` was the only input that would have had the DEM nodata values. So, we don't want values where the DEM didn't exist.

Fixes #470 

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -72,6 +72,9 @@ Unreleased Changes (3.9.1)
 * Fisheries Habitat Scenario Tool
     * Fixed divide-by-zero bug that was causing a RuntimeWarning in the logs.
       This bug did not affect the output.
+* Seasonal Water Yield
+    * Fixed a bug where ``qf.tif`` outputs weren't properly masking nodata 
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Great!  Looks like the existing tests cover this case, so we're good to go.

PR #518, opened 2021-03-04
Bugfix/479 delineateit fix optional inputs interactivity

### LINKED ISSUE #479: DelineateIt UI - bug in enabling/disabling optional inputs
As reported by Jesse on slack:  

> I was able to get the UI to accept both point vector and snap points options (v3.9) by populating the fields in a different order (vector before DEM raster). Once the snap points options become greyed out, I can't access them AND the vector input together without quitting and then restarting InVEST.

The goal here is for Jesse to supply his own outlet point vector, while still using the "snap points" option and associated parameters (distance & threshold flow accum).

phargogh:
This PR updates DelineateIt in three ways:

1. The UI container relating to point snapping is now always interactive.  The cheapo layer type checking as a proxy for trying to detect whether the vector might contain points has been removed.  Workbench PR related to this forthcoming.
2. DelineateIt will now snap `MULTIPOINT` geometries so long as they contain only 1 component point.  OGR uses the WKT string `MULTIPOINT` to refer to `wkbMultiPoint`, `wkbMultiPointM`, `wkbMultiPointZM` and `wkbMultiPoint25D`, so all of these cases are covered, and these cases are now asserted in the tests.
3. When a feature is not able to be snapped (it's not a point/multipoint geometry, has >1 component point or has no geometry at all), a log message is recorded and the original geometry is preserved.  If there's no geometry to preserve, the feature is skipped.  These cases are now asserted.

Fixes #479

# Checklist

### DIFF
--- a/src/natcap/invest/delineateit/delineateit.py
+++ b/src/natcap/invest/delineateit/delineateit.py
@@ -564,7 +564,7 @@ def snap_points_to_nearest_stream(points_vector_path, stream_raster_path_band,
 
         # If the geometry is not a primitive point, just create the new feature
         # as it is now in the new vector.
-        if ((geom_name != 'POINT') or
+        if ((geom_name not in ('POINT', 'MULTIPOINT')) or
                 (geom_name == 'MULTIPOINT' and geom_count > 1)):
             LOGGER.debug(f"Feature {point_feature.GetFID()} is of type "
                          "{geom_name} and cannot be snapped to a stream.")
--- a/src/natcap/invest/delineateit/delineateit.py
+++ b/src/natcap/invest/delineateit/delineateit.py
@@ -687,9 +687,6 @@ def _find_raster_pour_points(flow_dir_raster_path_band):
     raster_info = pygeoprocessing.get_raster_info(flow_dir_raster_path)
     # Open the flow direction raster band
     raster = gdal.OpenEx(flow_dir_raster_path, gdal.OF_RASTER)
-    if raster is None:
-        raise ValueError(
-            "Raster at %s could not be opened." % flow_dir_raster_path)
     band = raster.GetRasterBand(band_index)
     width, height = raster_info['raster_size']
 
--- a/src/natcap/invest/delineateit/delineateit.py
+++ b/src/natcap/invest/delineateit/delineateit.py
@@ -566,8 +566,9 @@ def snap_points_to_nearest_stream(points_vector_path, stream_raster_path_band,
         # as it is now in the new vector.
         if ((geom_name not in ('POINT', 'MULTIPOINT')) or
                 (geom_name == 'MULTIPOINT' and geom_count > 1)):
-            LOGGER.debug(f"Feature {point_feature.GetFID()} is of type "
-                         "{geom_name} and cannot be snapped to a stream.")
+            LOGGER.debug(
+                f"FID {point_feature.GetFID()} ({geom_name}, n={geom_count}) "
+                "only primitive points can be snapped to a stream.")
             new_feature = ogr.Feature(snapped_layer.GetLayerDefn())
             new_feature.SetGeometry(source_geometry)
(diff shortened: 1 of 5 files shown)


src/natcap/invest/delineateit/delineateit.py
```diff
@@ -559,10 +533,25 @@ def snap_points_to_nearest_stream(points_vector_path, stream_raster_path_band,
...
-        if source_geometry.GetGeometryName() != 'POINT':
+        # as it is now in the new vector. MULTIPOINT geometries with a single
+        # component point count as primitive points.
+        # OGR's wkbMultiPoint, wkbMultiPointM, wkbMultiPointZM and
+        # wkbMultiPoint25D all use the MULTIPOINT geometry name.
+        if ((geom_name not in ('POINT', 'MULTIPOINT')) or
+                (geom_name == 'MULTIPOINT' and geom_count > 1)):
+            LOGGER.debug(
+                f"FID {point_feature.GetFID()} ({geom_name}, n={geom_count}) "
+                "only primitive points can be snapped to a stream.")
```
davemfish, line 554: Do you think this message is mainly meant for developers or for users? If it's for devs, then I think it's great. If it's for users who chose the snapping option but are now going to get back "unsnapped" points, do you think it makes more sense as a `warning`? If so I might also remove the "primitive points" reference and say something like, "MULTIPOINT geometries cannot be snapped".

src/natcap/invest/delineateit/delineateit.py
```diff
@@ -571,13 +560,16 @@ def snap_points_to_nearest_stream(points_vector_path, stream_raster_path_band,
             continue
 
         point = shapely.wkb.loads(source_geometry.ExportToWkb())
+        if geom_name == 'MULTIPOINT':
+            point = point.geoms[0]  # take the first geometry of the multipoint
```
davemfish, line 564: What about making this comment say something like, "we already checked & know there is only one point"?

tests/test_delineateit.py
```diff
@@ -188,12 +230,13 @@ def test_point_snapping(self):
         snapped_points_layer = snapped_points_vector.GetLayer()
 
         # snapped layer will include 3 valid points and one polygon.
-        self.assertEqual(4, snapped_points_layer.GetFeatureCount())
+        self.assertEqual(5, snapped_points_layer.GetFeatureCount())
```
davemfish, line 233: Need to update this comment?

davemfish (changes requested):
This looks great to me, @phargogh ! Glad to see the UI getting simpler. Thanks for mentioning the related workbench update too.

I had only a few very minor comments. Oh, and it looks like there's a real trivial merge conflict.

src/natcap/invest/delineateit/delineateit.py
phargogh: Yep, `warning` makes more sense for sure!  This is intended to provide a user-facing record of which geometries were not snapped (though there is enough information in the message for us as developers to see why it was not snapped).

I also updated the language to clarify that the geometry could not be snapped.  Your point (ha) about multipoint geometries is a little tricky because a part of this change is all about snapping `MULTIPOINT` geometries, but only if they have exactly one component point.  It was tough to try to capture that nuance in a log message, so I simplified it ... what do you think about the updated text?

Patched in 1eb6b29c

src/natcap/invest/delineateit/delineateit.py
phargogh: Sure! Patched in da09cb3e

tests/test_delineateit.py
phargogh: Yep!  I added in a missing assertion while I was in there as well.  abc0e171

phargogh:
Thanks @davemfish !  I just touched up those couple of things.  Back to you!

davemfish (approved):
Looks good!

PR #537, opened 2021-04-26
sampledata json files no longer part of nsis installer

### LINKED ISSUE #525: BUILD: sampledata archives do not include the json datastack files
Sampledata archives made during the build process exclude the json datastack files. Instead, the jsons are copied over to the installer by the custom NSIS installer script. I guess the idea there is that the json files should be included in all installations, even when the user chooses not to download any of the sample data? I guess there's no harm in that, but those json files are of limited use without the associated data.

The workbench expects all invest resources to be found in the GCS bucket configured in `package.json: invest`, and right now the json files aren't there.

So, we could change the `make sampledata` targets to include the json files in the zip that is deployed to GCS during `make deploy`.  Workbench users won't get the json files unless they also choose to download the sample data. Is that okay?

davemfish:
Fixes #525 
This PR depends on the changes in https://bitbucket.org/natcap/invest-sample-data/pull-requests/19/moved-json-files-into-their-respective

Sampledata json files were moved into their respective folders in that repo. This means they will be included for any user that chooses to download the sampledata as part of the Windows install process. And that means we no longer need to manage them with the NSIS script. 

# Checklist

- [ ] Updated the user's guide (if needed) - I don't think we describe the location of these files in the UG, but not completely certain.

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -76,19 +76,6 @@ jobs:
                   $env:PYTHON = python -c "import sys, os; print(os.path.dirname(sys.executable))"
                   ./ci/windows-ci-binary-install.ps1
 
-            - name: Clone sample data repo (non-lfs)
-              shell: bash -l {0}
-              run: |
-                  GIT_URL=$(make jprint-GIT_SAMPLE_DATA_REPO)
-                  LOCALPATH=$(make jprint-GIT_SAMPLE_DATA_REPO_PATH)
-                  GIT_REV=$(make jprint-GIT_SAMPLE_DATA_REPO_REV)
-
-                  # Only clone the files ... deliberately not cloning LFS data
-                  # Installer needs .invest.json files only, and these are not
-                  # stored in LFS.
-                  GIT_LFS_SKIP_SMUDGE=1 git clone $GIT_URL $LOCALPATH
-                  git -C $LOCALPATH checkout $GIT_REV
-
             - name: Build userguide, binaries, installer
               shell: bash -l {0}
               run: |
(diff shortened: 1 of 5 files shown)


davemfish:
I thought about whether this should go into `main` or `release/workbench-alpha` and because it depends on the sample-data repo, it seemed much easier to make these changes in `main`. Otherwise we'd have to think about maintaining a separate `workbench` branch in the sample-data repo. And I really don't like the sound of that.

And based on my notes in #525 , I don't think there's any reason these changes should not go straight into `main`. How does that sound?

phargogh:
> And based on my notes in #525 , I don't think there's any reason these changes should not go straight into main. How does that sound?

Yep!  This is pretty much just an issue with the distribution of sample data ... there are few to no links in the UI, just in some scripts on the periphery of InVEST itself, not within the application, and maintaining two separate branches for this would be an unnecessary pain.

phargogh (changes requested):
Functionally, this looks great!  It'll be nice to make our build process simpler and the zipfiles more self-contained.  But could you also take a look at `scripts/invest-autovalidate.py` and see what you think about the best way to handle those JSON entries as well?

davemfish:
> Functionally, this looks great! It'll be nice to make our build process simpler and the zipfiles more self-contained. But could you also take a look at `scripts/invest-autovalidate.py` and see what you think about the best way to handle those JSON entries as well?

Thanks, @phargogh ! I assume you meant the `autotest` script here - good catch I had forgotten that one. Updated now with the new paths, and I ran it locally with no problems.

davemfish:
Ugg I think the git-lfs rules are somehow broken again after merging that sample data PR. For some reason the lfs files aren't being checked out with their real content.

phargogh (approved):
Looks great.  Thanks for updating `scripts/invest-autotest.py` and for following up on this @davemfish !

PR #539, opened 2021-04-29
Bugfix/hra handle 0s for rating, dq, and weight in criteria tables

### LINKED ISSUE #538: HRA: bugs in handling 0s in criteria tables
This was first reported here: https://community.naturalcapitalproject.org/t/invest-3-9-0-hra-zerodivisionerror-float-division-by-zero/1868

There might be a couple things going on:

`_validate_rating` is supposed to catch ratings < 1, but doesn't handle numeric datatypes, which seem to come out of xlsx files (csv tables seem to yield strings like "0", which is what the function expects.).

Even if that function does catch the rating = 0, we can still hit a dividebyZero error if DQ or Weight values are 0 for the same criteria. So something might be broken in the logic to skip/ignore criteria scored with a 0, which is the intent.

davemfish:
Fixes #538 

There were a couple problems with the validation functions for these values. The DQ & Weight validator did not guard against 0s, but those values are used as denominators in divides. The ratings validator did not give the correct result for integer or float types, it always expected strings. When these values originate from a CSV, they are strings, but from an xlsx, they are read as floats/ints.

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -83,7 +83,7 @@ Unreleased Changes (3.9.1)
       This bug did not affect the output.
 * HRA
     * Fixed bugs that allowed zeros in DQ & Weight columns of criteria 
-    table to raise DivideByZero errors.
+      table to raise DivideByZero errors.
 * Seasonal Water Yield
     * Fixed a bug where ``qf.tif`` outputs weren't properly masking nodata 
       values and could show negative numbers.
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Hey @davemfish this looks good to me. I'm not familiar at all with this model but your improvements to the nested conditional statements are great.

PR #548, opened 2021-05-10
Always run pollination total abundance step

### LINKED ISSUE #546: Pollination doesn't calculate total abundance without farm vector
Reported [here](https://community.naturalcapitalproject.org/t/missing-total-pollinator-abundance-and-total-pollinator-yield-in-final-results-example-biophysical-table-data/1873) and in emails that followed.

If the optional farm vector is not provided, `execute` finishes without calculating the `total_pollinator_abundance` output, even though it doesn't depend on any farm data. I can't see any reason not to move this step before the `task_graph.close()`, it's just a sum of the individual pollinator abundance outputs, which are created.
The user's guide is not clear about this or that the `total_pollinator_yield` output also depends on the farm vector.

emlys:
# Description
Fixes #546 
Doing this now so I can send a dev build to the user.
Made a `release/3.9.1` branch of the user's guide, with comments clarifying the pollination outputs: https://github.com/natcap/invest.users-guide/tree/release/3.9.1

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -86,10 +86,14 @@ Unreleased Changes (3.9.1)
     * Fixed divide-by-zero bug that was causing a RuntimeWarning in the logs.
       This bug did not affect the output.
 * HRA
-    * Fixed bugs that allowed zeros in DQ & Weight columns of criteria 
+    * Fixed bugs that allowed zeros in DQ & Weight columns of criteria
       table to raise DivideByZero errors.
+* Pollination
+    * Updated so that the ``total_pollinator_abundance_[season].tif`` outputs
+      are always created. Before, they weren't created if a farm vector was
+      not supplied, even though they are independent.
 * Seasonal Water Yield
-    * Fixed a bug where ``qf.tif`` outputs weren't properly masking nodata 
+    * Fixed a bug where ``qf.tif`` outputs weren't properly masking nodata
       values and could show negative numbers.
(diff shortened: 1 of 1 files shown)


emlys:
any idea what's going on here? https://github.com/natcap/invest/blob/32b81de1c748011fc25cc822ff9217ae65aff526/src/natcap/invest/pollination.py#L1406-L1409 🤯 😲

phargogh:
> any idea what's going on here? 

@emlys Pollination was a very early (maybe the first) InVEST model to use `taskgraph`, and this is a super old approach to determining whether the source code had changed.  The code you highlighed reads the source code of the `__call__` method, encodes it as UTF-8 and takes the SHA1 digest.  These days, `taskgraph` handles this all for us behind the scenes, but it does pretty much the same thing.

emlys:
> This digest is then used to determine if the task needs to be recomputed if/when the source code changes. These days, taskgraph handles this for us behind the scenes, but it does pretty much the same thing.

Wacky! So if the same function name is called with the same arguments, but the source code is different, it would know to re-run the task? Very cool

dcdenu4 (approved):
Hey @emlys , this looks good to me. Thanks for the Users Guide note. 

I was wondering if there should be an added test, but in this case it would just be testing that outputs are generated regardless of farm vector... which doesn't seem that helpful.

PR #551, opened 2021-05-19
Better precision in carbon report

phargogh:
This PR updates `numpy.sum` in the carbon model to use a higher-precision aggregate number.  Python's `float` object is float64, so I thought it'd make sense to just stick with that.  A test has been added to assert the extra precision is retained.

# Checklist


davemfish (approved):
Looks great to me @phargogh , always a good idea to specify a dtype!

PR #552, opened 2021-05-20
Make check_spatial_overlap work with undefined projection

### LINKED ISSUE #519: Validation: check spatial overlap on layer with no projection
If a layer's projection wkt is `None`, `check_spatial_overlap` errors like this:

```
  File "site-packages\natcap\invest\ui\model.py", line 1649, in _logged_target
  File "site-packages\natcap\invest\coastal_vulnerability.py", line 286, in execute
  File "site-packages\natcap\invest\validation.py", line 916, in _wrapped_validate_func
  File "site-packages\natcap\invest\coastal_vulnerability.py", line 2952, in validate
  File "site-packages\natcap\invest\validation.py", line 843, in validate
  File "site-packages\natcap\invest\validation.py", line 580, in check_spatial_overlap
  File "site-packages\pygeoprocessing\geoprocessing.py", line 2878, in transform_bounding_box
  File "site-packages\osgeo\osr.py", line 891, in ImportFromWkt
TypeError: not a string
```

Validation does check if the SRS is `None` during the `check_vector` and `check_raster` steps. They append a validation message when SRS is None, but validation continues to `check_spatial_overlap`. And since that func raises an uncaught error, we never get to see the validation message reported from `check_vector`.

If this happens in the UI, I think the model is allowed to run because no validation messages are returned. CV caught this issue because it happens to call `validate` during `execute` (questionable but useful here!).

emlys:
# Description
Fixes #519 
The error is only raised when `different_projections_ok=True`, because that triggers the call to `transform_bounding_box`, which expects a projection. I added a conditional before that step that returns a message if the file has no projection.

Still not explicitly checking that the projections are the same because of the issue described in #558.

# Checklist

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -92,15 +92,28 @@ class SpatialOverlapTest(unittest.TestCase):
         expected = f'Spatial file {filepath_2} has no projection'
         self.assertEqual(error_msg, expected)
 
+    @unittest.skip("skipping due to unresolved projection comparison question")
     def test_different_projections_not_ok(self):
-        """Validation: different projections not allowed by default."""
+        """Validation: different projections not allowed by default.
+
+        This test illustrates a bug we don't yet have a good solution for
+        (natcap/invest#558)
+        When ``different_projections_ok is False``, we don't check that the
+        projections are actually the same, because there isn't a great way to
+        do so. So there's the possibility that some bounding boxes overlap
+        numerically, but have different projections, and thus pass validation
+        when the shouldn't.
+        """
+
         from natcap.invest import validation
 
         driver = gdal.GetDriverByName('GTiff')
         filepath_1 = os.path.join(self.workspace_dir, 'raster_1.tif')
         filepath_2 = os.path.join(self.workspace_dir, 'raster_2.tif')
 
+        # bounding boxes overlap if we don't account for the projections
         for filepath, geotransform, epsg in (
+
                 (filepath_1, [1, 1, 0, 1, 0, 1], 4326),
                 (filepath_2, [2, 1, 0, 2, 0, 1], 2193)):
             raster = driver.Create(filepath, 3, 3, 1, gdal.GDT_Int32)
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -1212,6 +1212,50 @@ class TestValidationFromSpec(unittest.TestCase):
         self.assertTrue('Bounding boxes do not intersect' in
                         validation_warnings[0][1])
 
+    def test_spatial_overlap_error_undefined_projection(self):
+        """Validation: check spatial overlap message when no projection"""
+        from natcap.invest import validation
+
+        spec = {
+            'raster_a': {
+                'type': 'raster',
+                'name': 'raster 1',
+                'about': 'raster 1',
+                'required': True,
+            },
+            'raster_b': {
+                'type': 'raster',
+                'name': 'raster 2',
+                'about': 'raster 2',
+                'required': True,
+            }
+        }
+
+        driver = gdal.GetDriverByName('GTiff')
+        filepath_1 = os.path.join(self.workspace_dir, 'raster_1.tif')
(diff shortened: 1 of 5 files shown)


tests/test_validation.py
```diff
@@ -66,47 +67,59 @@ def test_overlap_no_reference(self):
...
-        new_feature = ogr.Feature(layer.GetLayerDefn())
-        new_feature.SetGeometry(ogr.CreateGeometryFromWkt('POINT 1 1'))
-
-        new_feature = None
-        layer = None
-        vector = None
-
-        error_msg = validation.check_spatial_overlap(
-            [filepath_1, filepath_2, reference_filepath],
-            vector_srs.ExportToWkt())
```
emlys, line 104: This test was testing a feature that no longer exists. Looks like `check_spatial_overlap` used to have an optional "reference" arg it would compare projections to. That's been replaced with `different_projections_ok`.

davemfish (commented):
Hi @emlys , this looks like it does the job! I like that the skipped test is there for us if we ever do deal with the other issue.

One thing to consider here is that before the check spatial overlap step, validation has already done `_check_projection` for spatial inputs and returned a message if any projection is `None`. But that is not stopping it from continuing to this `check_spatial_overlap`. So with this solution, I wonder if the final feedback to the user will contain two redundant messages about a missing projection? And if it makes sense to avoid the overlap check altogether when there are other `_check_projection` warnings present. What do you think?

emlys:
Good point @davemfish, I'm not quite sure the best thing to do here. I think `check_spatial_overlap` should handle the possibility of an undefined projection, in case we ever use it in another context. I agree that ideally we would skip `check_spatial_overlap` if there's already a `_check_projection` warning. But it may be tricky to know if `_check_projection` has been called already.

For instance, `check_vector` can return before `_check_projection` ever gets called (if the file can't be opened with GDAL, or if it's missing any expected fields). So to determine if `_check_projection` has been called already, I think we'd have to verify that neither of those unrelated warning messages are in the messages list either.

IMO this is kinda a problem with the validation structure in general - which warnings are ultimately returned depends on the order of the conditional statements and isn't easily predictable.

davemfish:
> Good point @davemfish, I'm not quite sure the best thing to do here. I think `check_spatial_overlap` should handle the possibility of an undefined projection, in case we ever use it in another context. 

I agree, the main thing is to avoid the uncaught exception.

> I agree that ideally we would skip `check_spatial_overlap` if there's already a `_check_projection` warning. But it may be tricky to know if `_check_projection` has been called already.
 For instance, `check_vector` can return before `_check_projection` ever gets called (if the file can't be opened with GDAL, or if it's missing any expected fields). So to determine if `_check_projection` has been called already, I think we'd have to verify that neither of those unrelated warning messages are in the messages list either.
> 

Maybe the more general thing to know is just whether the vector is valid up to this point? Because I do think `validate` is designed to do `check_spatial_overlap` last, and to only check the inputs that are otherwise valid.

```
        # Only test for spatial overlap once all the sufficient spatial keys
        # are otherwise valid. And then only when there are at least 2.
        valid_spatial_keys = spatial_keys.difference(
            invalid_keys.union(insufficient_keys))

        if len(valid_spatial_keys) >= 2:
```
But now I'm a little confused about how an invalid input (projection is `None`) actually gets passed to `check_spatial_overlap` 🤔  


> IMO this is kinda a problem with the validation structure in general - which warnings are ultimately returned depends on the order of the conditional statements and isn't easily predictable.

Yeah I agree. I think it's generally designed to check the easiest things first though.

emlys:
> But now I'm a little confused about how an invalid input (projection is None) actually gets passed to check_spatial_overlap 🤔

Ohhh you're absolutely right, this shouldn't be happening at all. so the problem is that `_check_projection` checks its `srs` argument against `None`. But if the projection is undefined it won't be `None`, it'll be an empty `osr.SpatialReference` object. I just added a conditional to compare it against an empty SRS. So if it's empty, it won't even get to the point of `check_spatial_overlap`. Thanks for spotting that!

davemfish (approved):
Thanks @emlys , looks great!

PR #554, opened 2021-05-24
Update to gdal 3.3.0

emlys:
## Description
Fixes #
- Wrap calls to `ExportToWkb()` with `bytes()` to convert it back to the expected type `bytes`, since it [now returns a `bytearray`](https://github.com/OSGeo/gdal/blob/eeeffe624996518655f231125712582551222932/gdal/NEWS#L420). This change is backwards-compatible.
- The same change is needed in pygeoprocessing. That's in [this PR](https://github.com/natcap/pygeoprocessing/pull/196) which we'll need to make a 2.2.1 release for.
- Added python 3.9 to the GHA test matrix. No changes were needed to support python 3.9. (Should we take out 3.7 yet?)
- Added gdal versions to the GHA test matrix to make sure we're testing the full range of versions including 3.3.0

## Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -30,12 +30,8 @@ jobs:
         name: "Build windows binaries"
         needs: check-syntax-errors
         runs-on: windows-latest
-        strategy:
-          fail-fast: false
-          matrix:
-              python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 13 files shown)


emlys:
Merging these changes into #555 and closing, because it's getting too confusing maintaining the changes separately.

PR #555, opened 2021-05-24
Update GDAL and pygeoprocessing

### LINKED ISSUE #609: Clarify SDR/NDR behavior for pixels that don't drain to a stream
SDR and NDR are defined in terms of distance to stream. We get this from the pygeoprocessing `distance_to_channel_mfd` function. Before, `distance_to_channel_mfd` produced meaningless values for pixels that don't drain to any stream (distance to an edge, or something like that). In pygeoprocessing 2.3.0 that's fixed, and now those areas are `nodata`.

If we do nothing, those nodata areas will propagate through all the SDR/NDR outputs. SDR/NDR will only be defined for areas that drain to a stream.

Rich proposed making each pour point be a stream pixel, so that the entire raster will drain to a stream and thus have defined SDR/NDR. I talked with Stacie and Rafa about the change, here are excerpts from our convo:

Emily:
> Every pixel in Watershed A drains to a stream on the map. SDR/NDR is clearly defined.
There are no streams on the map in Watershed B. We’re not sure how to calculate SDR/NDR here, because the “distance to stream” component is unknown.
**Past behavior:**
“distance to stream” is really “distance to raster edge” in areas that don’t drain to a stream. SDR/NDR was calculated based on distance to the raster edge. This was an unintentional result of how pygeoprocessing found the distance.
**New behavior (A):**
“distance to stream” is left undefined in areas that don’t drain to a stream.
**Options for SDR/NDR:**
(A): Allow the NoData area to propagate through all the outputs that depend on distance to stream. SDR/NDR won’t be defined for pixels that don’t drain to a stream on the map.
(C): Treat all pour points like streams. Calculate SDR/NDR based on “distance to stream or pour point”. SDR/NDR will be defined for the entire map.
Rich proposed option (C), and that addresses concerns about users getting confused that their outputs are smaller and have different extents depending on the TFA value. On the other hand, we don’t want users to believe their results are valid in areas where it doesn’t make sense to define SDR/NDR.

Stacie:
>  I'm most interested to hear what Rafa says from the point of view of what's best hydrologically. One other idea is, for pixels where "distance to stream" is undefined, could we assign a particular value like -9999, so people would know that these pixels don't reach a stream? That would be easy to see and symbolize and document.
I'm a little wary of option C, since it's not doing what we say the model does, of routing sediment until it reaches a stream, and it's invisible to the user. If C could be an option in the UI, or otherwise made more transparent, that might be better?

Rafa:
> Indeed, I did not consider this case before. Some thoughts: First of all, because sediment delivery ratio is really important and it depends on the downslope distance to the next drainage, calculating the distance to map edges or pour points introduces a structural error. Basically, the model does what the user guide says in some places, and something different in other places (as Stacie pointed out) and the user won't even notice if they're not a real expert.
Basically, I'd think with both (B) and (C) you'd overestimate the values at the raster edges because you count sediment as exported before the entire hillslope had a chance to retain it. However, where that is the case will be not evident for the user. It be then we would provide a mask showing where we get exact results and where results are an approximation. Of both, (C) makes a bit more sense.
I would not mind (A) if the outputs rasters just had no data values at the borders, as long as they maintain the same extent as the input rasters. The no-data at the borders also points users to another thing to keep in mind: It makes much more sense to run these water models on a hydrologically bounded domain than on a square. If the border region is really of interest - well, the model domain needs to be extended to include the next stream, or the TFA needs top be decreased.

emlys:
# Description

- add `bytes()` around `ExportToWkb()` so the result is still a `bytes` object with GDAL>=3.3.0
- update GHA workflows to run our tests with python 3.9
- my fix for OSGeo/gdal#3898 was released in gdal 3.3.1, hooray! We no longer need the workaround to remove nonexistent paths from the `PATH`.

## SDR and NDR changes
Fixes #609 
These two models use pygeoprocessing's `distance_to_channel_mfd` which was changed in pygeo 2.3.0. 
Due to [this pygeo bug](https://github.com/natcap/pygeoprocessing/issues/171), areas in NDR and SDR's "distance downstream" output (d_dn.tif) that didn't actually drain to any stream, were filled with nonsense values. That bug is fixed in pygeo 2.3.0. Now the areas that don't drain to any stream are nodata. 

Discussed with Stacie and Rafa and decided it's best to let those nodata areas propagate through the outputs. Hence the updates to expected test values. I double checked most of these in QGIS.

### Change to `sdr_core.calculate_sediment_deposition`
Two of this function's inputs, `e_prime` and `sdr`, have the same nodata area propagated from `distance_to_channel_mfd`. But in the `sed_deposition` output, that area was filled with 0 (not nodata). I believe I've fixed the algorithm to stop on nodata pixels. I'm not super confident in editing this cython function, but the output is exactly the same as before in the defined area.

# Checklist
- [x] Updated the user's guide (if needed) Corresponding UG PR:natcap/invest.users-guide#37


.github/workflows/build-and-test.yml
```diff
@@ -42,14 +42,19 @@ jobs:
...
                 include:
                     - python-version: 3.7
                       numpy: "numpy=1.15"  # fuzzy assertion in conda is single '='
+                      gdal: "gdal==3.1.2"
                     - python-version: 3.8
                       numpy: "numpy=1.16"
+                      gdal: "gdal==3.2.0"
+                    - python-version: 3.9
+                      numpy: "numpy=1.20"
+                      gdal: "gdal==3.3.0"
```
dcdenu4, line 57: Should this be 3.3.1 given the bug fix of the environment variable issue?

src/natcap/invest/hra.py
```diff
@@ -2584,11 +2584,11 @@ def _get_criteria_dataframe(base_criteria_table_path):
     file_ext = os.path.splitext(base_criteria_table_path)[1].lower()
     if file_ext == '.csv':
         # use sep=None, engine='python' to infer what the separator is
-        criteria_df = pandas.read_csv(base_criteria_table_path, 
-            index_col=0, header=None, sep=None, engine='python')
+        criteria_df = pandas.read_csv(base_criteria_table_path,
+                                      index_col=0, header=None, sep=None, engine='python')
```
dcdenu4, line 2588: Does this bust the 80 column mark? Maybe something like this here:
```python
criteria_df = pandas.read_csv(
    base_criteria_table_path, index_col=0, header=None, sep=None, 
    engine='python')
```

src/natcap/invest/ndr/ndr.py
```diff
@@ -471,7 +477,7 @@ def _validate_inputs(nutrients_to_process, lucode_to_parameters):
         func=pygeoprocessing.calculate_slope,
         args=((f_reg['filled_dem_path'], 1), f_reg['slope_path']),
         target_path_list=[f_reg['slope_path']],
-        dependent_task_list=[stream_extraction_task],
+        dependent_task_list=[fill_pits_task],
```
dcdenu4, line 480: Yea, this seems like the right dependent task here.

src/natcap/invest/ndr/ndr.py
```diff
@@ -1270,7 +1279,7 @@ def _calculate_sub_ndr(
 
     def _sub_ndr_op(dist_to_channel_array):
         """Calculate subsurface NDR."""
-        # nodata value from this ntermediate output should always be 
+        # nodata value from this ntermediate output should always be
```
dcdenu4, line 1282: Just happen to see a type here: `ntermediate`

dcdenu4:
```suggestion
        # nodata value from this intermediate output should always be
```

src/natcap/invest/ndr/ndr.py
```diff
@@ -1305,9 +1314,9 @@ def _calculate_export_op(
         # these intermediate outputs should always have defined nodata
         # values assigned by pygeoprocessing
         valid_mask = ~(numpy.isclose(modified_load_array, load_nodata) |
-            numpy.isclose(ndr_array, ndr_nodata) |
-            numpy.isclose(modified_sub_load_array, subsurface_load_nodata) |
-            numpy.isclose(sub_ndr_array, sub_ndr_nodata))
+                       numpy.isclose(ndr_array, ndr_nodata) |
+                       numpy.isclose(modified_sub_load_array, subsurface_load_nodata) |
+                       numpy.isclose(sub_ndr_array, sub_ndr_nodata))
```
dcdenu4, line 1319: I think line 1318 busts the 80 col mark.

dcdenu4: I highlighted all those lines to insert a suggestion, but apparently I can't here... Maybe because I selected a non changed line...

src/natcap/invest/ndr/ndr.py
```diff
@@ -3,10 +3,14 @@
 import logging
 import os
 import pickle
+import shutil
+import tempfile
```
dcdenu4, line 7: Are these used?

src/natcap/invest/ndr/ndr.py
```diff
@@ -3,10 +3,14 @@
...
 
 import numpy
 import pygeoprocessing
 import pygeoprocessing.routing
+from pygeoprocessing.geoprocessing_core import \
+    DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
```
dcdenu4, line 13: Is this used anywhere?

src/natcap/invest/sdr/sdr.py
```diff
@@ -13,7 +13,11 @@
 from osgeo import gdal
 from osgeo import ogr
 import numpy
+import shutil
+import tempfile
```
dcdenu4, line 17: Are these used?

src/natcap/invest/sdr/sdr.py
```diff
@@ -13,7 +13,11 @@
...
 
+from pygeoprocessing.geoprocessing_core import \
+    DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
```
dcdenu4, line 20: Is this used anywhere?

src/natcap/invest/sdr/sdr.py
```diff
@@ -213,24 +217,24 @@
     'sdr_bare_soil_path': 'sdr_bare_soil.tif',
     'sdr_path': 'sdr_factor.tif',
     'slope_path': 'slope.tif',
+    'stream_path': 'stream.tif',
```
dcdenu4, line 220: Was it desired to move stream path output to the intermediate folder? I feel like I remember prior conversations where it was requested stream path output go in the output folder...

src/natcap/invest/sdr/sdr.py
```diff
@@ -188,18 +192,18 @@
...
+
+INTERMEDIATE_DIR_NAME = 'intermediate_outputs'
 
 _INTERMEDIATE_BASE_FILES = {
     'cp_factor_path': 'cp.tif',
     'd_dn_bare_soil_path': 'd_dn_bare_soil.tif',
     'd_dn_path': 'd_dn.tif',
     'd_up_bare_soil_path': 'd_up_bare_soil.tif',
     'd_up_path': 'd_up.tif',
-    'dem_offset_path': 'dem_offset.tif',
```
dcdenu4, line 202: This was an unused specified path?

src/natcap/invest/sdr/sdr.py
```diff
@@ -213,24 +217,24 @@
     'sdr_bare_soil_path': 'sdr_bare_soil.tif',
     'sdr_path': 'sdr_factor.tif',
     'slope_path': 'slope.tif',
+    'stream_path': 'stream.tif',
     'thresholded_slope_path': 'slope_threshold.tif',
     'thresholded_w_path': 'w_threshold.tif',
     'w_accumulation_path': 'w_accumulation.tif',
     'w_bar_path': 'w_bar.tif',
     'w_path': 'w.tif',
-    'ws_factor_path': 'ws_factor.tif',
```
dcdenu4, line 221: This was an unused specified path?

src/natcap/invest/sdr/sdr.py
dcdenu4: Oh, does the `stream_and_drainage_path` fill that need?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -374,7 +374,7 @@ def calculate_sediment_deposition(
 
     """
     LOGGER.info('Calculate sediment deposition')
-    cdef float sediment_deposition_nodata = -1.0
+    cdef float sediment_deposition_nodata = -1
```
dcdenu4, line 377: I think it's more readable to have the -1.0 since it is a float, was there a reason for trimming it?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -623,11 +621,11 @@ def calculate_average_aspect(
...
-    cdef float* flow_lengths = [
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2
+    cdef float * flow_lengths = [
+        1, < float > SQRT2,
+        1, < float > SQRT2,
+        1, < float > SQRT2,
+        1, < float > SQRT2
```
dcdenu4, line 628: Just curious if this is recommended Python formatting being applied to cython? Does it make sense to format these differently given context?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -479,6 +477,8 @@ def calculate_sediment_deposition(
                             neighbor_flow_val >> (inflow_offsets[j]*4)) & 0xF
                         if neighbor_flow_weight > 0:
                             f_j = f_raster.get(neighbor_col, neighbor_row)
+                            if f_j == sediment_deposition_nodata:
```
dcdenu4, line 480: We talked about this on a Slack call, as well as this file in general. We're going to try and reach out to Rich to see if he'd be willing to help us comprehend this a bit more.

dcdenu4 (changes requested):
Hey @emlys , thanks for tackling this non-trivial change head on. I have some minor comments and suggestions in here. Along with that we were going to look a bit more into the `sdr_core.pyx` file changes and see if Rich would be up for walking us through the sediment deposition function.

.github/workflows/build-and-test.yml
emlys: yes thanks!

src/natcap/invest/hra.py
emlys: fixed! I need to find out how to configure autopep8 to move args down to the next line

src/natcap/invest/ndr/ndr.py
emlys: removed

src/natcap/invest/ndr/ndr.py
emlys: removed

src/natcap/invest/sdr/sdr.py
emlys: removed

src/natcap/invest/sdr/sdr.py
emlys: removed

src/natcap/invest/sdr/sdr.py
emlys: yea, not used anywhere

src/natcap/invest/sdr/sdr.py
emlys: correct

src/natcap/invest/sdr/sdr_core.pyx
emlys: yeah, I was accidentally autopep8-ing cython files. Turned this off but I guess these snuck through.

src/natcap/invest/sdr/sdr_core.pyx
emlys: i believe trailing `.0`s are discouraged since they're usually left over from python 2 when `1.0` didn't behave the same as `1`. do you prefer it for cython specifically?

src/natcap/invest/sdr/sdr_core.pyx
emlys: I am pretty sure now that `f_j` will never be nodata when it's accessed here.

At the bottom of this function, `sediment_deposition_raster.set()` and `f_raster.set()` are called. Both these rasters start out filled with nodata, they're only modified in that one place, and they will never be set to nodata. So if `sediment_deposition_raster.get(i)` is defined, then so is `f_raster.get(i)`. 

A pixel is only pushed onto the stack once `sediment_deposition_raster.get(j)` is defined for each of its upstream neighbors `j`. ([code](https://github.com/emlys/invest/blob/1b71cd469ba68e10f103bdf38e416113cf50856c/src/natcap/invest/sdr/sdr_core.pyx#L594-L605))

Therefore I'm pretty sure that `f_j` is also defined for each upstream neighbor. And we only access the upstream neighbors because of the `if neighbor_flow_weight > 0:` check two lines up.

But if there is any uncertainty, it will be safer to leave this nodata check in here.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +513,76 @@ def calculate_sediment_deposition(
...
                             continue
-                        # if this direction flows out, add to weighted sum
+                        # if it is a downstream neighbor, add to the sum and
+                        # check if it can be pushed onto the stack yet
                         flow_weight = (flow_val >> (j*4)) & 0xF
                         if flow_weight > 0:
                             sdr_j = sdr_raster.get(neighbor_col, neighbor_row)
-                            if sdr_j == 0.0:
+                            if sdr_j == sdr_nodata:
+                                continue
```
emlys, line 563: This is the only place where `continue`ing on nodata could have an impact beyond 

The first part here is summing up the `downstream_sdr_weighted_sum`. The effect is the same because before, `if sdr_j == sdr_nodata` we would add `0`. Now, we `continue` and don't add anything.

The second part is checking if the neighbor `j` can be pushed onto the stack yet. If`sdr_j == sdr_nodata`, then the pixel `j` will never get pushed onto the stack. That's correct because we can't calculate the outputs on pixel `j` if `sdr_j` is undefined. This may be redundant with the [check at the end](https://github.com/emlys/invest/blob/1b71cd469ba68e10f103bdf38e416113cf50856c/src/natcap/invest/sdr/sdr_core.pyx#L616) but it does make the intent clearer.

The potential impact of this is that, if you have a pixel whose upstream neighbors all have undefined SDR, I think it would never get calculated. I believe that's okay but not sure.

emlys:
Hey @dcdenu4, I addressed all your comments above. As for the `calculate_sediment_deposition` algorithm stuff, I tried to figure out what it's doing and added a bunch of comments into the function. I still don't completely understand it but I'm pretty confident this is correct. Still not totally sure about the `sdr_j == nodata` thing.

src/natcap/invest/sdr/sdr_core.pyx
dcdenu4: Yea I was thinking about Cython specifically, since we define it as a float but then set it to a 'perceived' integer. But, I guess I'd defer to the latest Cython style / docs, if there are any.

src/natcap/invest/sdr/sdr_core.pyx
phargogh: > This may be redundant with the check at the end but it does make the intent clearer.

It'll also break out of unnecessary computation earlier, which is a very good thing in these low-level functions

>The potential impact of this is that, if you have a pixel whose upstream neighbors all have undefined SDR, I think it would never get calculated. I believe that's okay but not sure.

Wouldn't we have a contradiction if this were true?  It seems to me that the only way we'd have undefined SDR for upstream pixels _U_ of pixel _p_ would be for upstream pixels _U_ to be in a distinct hydrological region ... but if they're in a distinct hydrological region (because they flow to a nonexistent stream and not the same stream _p_ flows to) they'd have to be downstream of _p_, right?  

I _think_ we might be OK here, but one thing I might suggest is to try to break it!  Can you construct a case (one that the model would produce given reasonable inputs) that exercises what you're concerned about?  If you can, that would be a very good start to a test case to make sure we handle it correctly!

phargogh (changes requested):
Based on what I can tell about this algorithm, the cython implementation as stated looks correct to me.  I did have a few comments that might be of interest, some about `sdr_core.pyx` and some in other files, and happy to talk more about them as would be helpful!

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -623,11 +683,11 @@ def calculate_average_aspect(
 
     # the flow_lengths array is the functional equivalent
     # of calculating |sin(alpha)| + |cos(alpha)|.
-    cdef float* flow_lengths = [
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2,
-        1.0, <float>SQRT2
+    cdef float * flow_lengths = [
```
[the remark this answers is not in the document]
emlys: I agree. I accidentally autopep8-ed this. I've turned off all linting on `.pyx` files for now

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -356,31 +356,56 @@ cdef class _ManagedRaster:
...
-        Returns:
-            None.
+    r_i  =      dr_i  * (sum over j ∈ J of f_j * p(i,j)) + E'_i
+
+    f_i  = (1 - dr_i) * (sum over j ∈ J of f_j * p(i,j)) + E'_i
+
+
+            (sum over k ∈ K of SDR_k * p(i,k)) - SDR_i
+    dr_i = --------------------------------------------
+                          (1 - SDR_i)
```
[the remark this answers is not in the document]
emlys: Done!

.github/workflows/binary-applications.yml
```diff
@@ -79,6 +79,11 @@ jobs:
             - name: Build userguide, binaries, installer
               shell: bash -l {0}
               run: |
+                  # resolves ModuleNotFoundError during pyinstaller build on windows:
+                  # On Windows, with Python >= 3.8, DLLs are no longer imported from the PATH.
+                  # If gdalXXX.dll is in the PATH, then set the USE_PATH_FOR_GDAL_PYTHON=YES environment variable
+                  # to feed the PATH into os.add_dll_directory().
+                  export USE_PATH_FOR_GDAL_PYTHON=YES
```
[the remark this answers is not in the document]
emlys: I'm not sure... at runtime is GDAL still found on the PATH? I thought maybe when it's bundled then it doesn't need to search for it. 
There's probably no harm in setting the variable in `exe/hooks/rthook.py` either way. Would you be interested to try it out and see what happens?

src/natcap/invest/sdr/sdr.py
emlys: Moved it back to the output folder. With the pour points thing it made more sense as an intermediate output, without that it belongs as a final output.

src/natcap/invest/sdr/sdr_core.pyx
emlys: If we assume that this function will only be called on the rasters produced by `sdr.py`, I think you're right. In the model, the SDR raster should have data everywhere that the flow_dir raster has data.

If `sdr_core.calculate_sediment_deposition` were called with an SDR input that _didn't_ come from the model, you can construct a scenario where this happens:
```
pixel names     flow dir        SDR        E'
 a  d  g        -> -> ->     .5 .5 .5    1  1  1
 b  e  h        -> -> ->     .5 -1 .5    1  1  1
 c  f  i        -> -> ->     .5 -1 .5    1  1  1
```
Pixels a, d, and g are seed pixels. Pixels will be pushed onto the stack in the order:  `a, d, g, b, c`. 
`e` and `f` aren't ever pushed to the stack because SDR is nodata there. And so `h` and `i` are never computed either.

I think this would be correct:
- `dR` is defined in terms of SDR
- therefore `dR` should have nodata where SDR has nodata
- `f` is defined in terms of `dR`
- therefore, `f` should have nodata where SDR has nodata
- `f` is also defined recursively in terms of the immediate upstream neighbors
- therefore, if a pixel `i`'s has an upstream neighbor `j`, and `f_j` is nodata, then `f_i` should be nodata
- therefore, if a pixel `i`'s has an upstream neighbor `j`, and `SDR_j` is nodata, then `f_i` should be nodata
- therefore, `f` should be nodata on all pixels downstream of a nodata SDR pixel

You could argue that pixels `b` and `c` should also be undefined in the output, because `r` and `f` are defined in terms of SDR for all _downstream_ neighbors of `i`, which is `nodata` for these pixels.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -535,48 +592,51 @@ def calculate_sediment_deposition(
...
 
                     sdr_i = sdr_raster.get(global_col, global_row)
-                    if sdr_i == sdr_nodata:
-                        sdr_i = 0.0
                     e_prime_i = e_prime_raster.get(global_col, global_row)
-                    if e_prime_i == e_prime_nodata:
-                        e_prime_i = 0.0
+                    # nodata pixels should propagate to the results
+                    if e_prime_i == e_prime_nodata or sdr_i == sdr_nodata:
+                        continue
```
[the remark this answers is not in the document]
emlys: That makes sense. separated these in 51442129e

src/natcap/invest/sdr/sdr_core.pyx
emlys: I made that change in c99bc44 just so you can see what I mean. I don't know if that makes sense to do.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -356,31 +356,59 @@ cdef class _ManagedRaster:
...
-            mfd_flow_direction_path (string): a path to a raster with
-                pygeoprocessing.routing MFD flow direction values.
-            e_prime_path (string): path to a raster that shows sources of
-                sediment that wash off a pixel but do not reach the stream.
-            f_path (string): path to a raster that shows the sediment flux
-                on a pixel for sediment that does not reach the stream.
-            sdr_path (string): path to Sediment Delivery Ratio raster.
-            target_sediment_deposition_path (string): path to created that
-                shows where the E' sources end up across the landscape.
+    This algorithm outputs both sediment deposition (r_i) and flux (f_i)::
```
dcdenu4, line 361: This documentation is great, thank you!

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -391,11 +419,19 @@ def calculate_sediment_deposition(
     cdef _ManagedRaster sediment_deposition_raster = _ManagedRaster(
         target_sediment_deposition_path, 1, True)
 
+    # given the pixel neighbor numbering system
+    #  3 2 1
+    #  4 x 0
+    #  5 6 7
+    # for a pixel's neighbor in position `i`,
+    # that neighbor's neighbor in position `inflow_offsets[i]`
+    # is the original pixel
```
dcdenu4, line 428: Hmmm, I'm just not quite able to understand this. Does this mean that `inflow_offsets[i]` is `x`?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -419,49 +455,59 @@ def calculate_sediment_deposition(
...
 
         for row_index in range(win_ysize):
             seed_row = yoff + row_index
             for col_index in range(win_xsize):
                 seed_col = xoff + col_index
-                # search to see if this is a good seed
-                if mfd_flow_direction_raster.get(seed_col, seed_row) == 0:
+                # check if this is a good seed
+                # a good seed pixel is a local high point
+                if mfd_flow_direction_raster.get(seed_col, seed_row) == mfd_nodata:
```
dcdenu4, line 467: Does a local high point, a good seed pixel, just mean that it is not `nodata` and drains downstream?

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -419,49 +455,59 @@ def calculate_sediment_deposition(
...
                     if neighbor_col < 0 or neighbor_col >= n_cols:
                         continue
+                    # skip if the neighbor's flow direction is undefined
                     neighbor_flow_val = <int>mfd_flow_direction_raster.get(
                         neighbor_col, neighbor_row)
-                    if neighbor_flow_val == 0:
+                    if neighbor_flow_val == mfd_nodata:
                         continue
+                    # if the neighbor flows into it, it's not a local high
+                    # point and so can't be a seed pixel
```
dcdenu4, line 485: Ahh this makes sense and answers my question above.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -419,49 +455,59 @@ def calculate_sediment_deposition(
...
-                    # processed
+                    # processed. this is true for seed pixels because they
+                    # have no upstream neighbors.
                     flat_index = processing_stack.top()
                     processing_stack.pop()
                     global_row = flat_index // n_cols
                     global_col = flat_index % n_cols
 
-                    # calculate the upstream Fj contribution to this pixel
+                    # (sum over j ∈ J of f_j * p(i,j) in the equation for r_i)
```
dcdenu4, line 507: The `element of` epsilon symbol doesn't show up in my `vim` text editor! I guess I need to update some vim plugins.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
                         if neighbor_col < 0 or neighbor_col >= n_cols:
                             continue
 
-                        # see if there's an inflow
+                        # see if there's an inflow from the neighbor to the pixel
```
dcdenu4, line 520: Line too long

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
...
                         neighbor_flow_val = (
                             <int>mfd_flow_direction_raster.get(
                                 neighbor_col, neighbor_row))
                         neighbor_flow_weight = (
                             neighbor_flow_val >> (inflow_offsets[j]*4)) & 0xF
                         if neighbor_flow_weight > 0:
                             f_j = f_raster.get(neighbor_col, neighbor_row)
+                            if f_j == target_nodata:
+                                continue
+                            # sum up the neighbor's flow dir values in each direction
```
dcdenu4, line 530: I think this is also over 80 chars.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
...
+                            # flow dir values are relative to the total
                             neighbor_flow_sum = 0
                             for k in range(8):
                                 neighbor_flow_sum += (
                                     neighbor_flow_val >> (k*4)) & 0xF
+                            # get the proportion of the neighbor's flow that
+                            # flows into the original pixel
                             p_val = neighbor_flow_weight / neighbor_flow_sum
+                            # add the neighbor's flux value, weighted by the flow proportion
```
dcdenu4, line 539: Line too long

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
...
                             f_j_weighted_sum += p_val * f_j
 
-                    # calculate the differential downstream change in sdr
-                    # from this pixel
-                    downstream_sdr_weighted_sum = 0.0
+                    # calculate sum of SDR values of immediate downstream neighbors,
+                    # weighted by proportion of flow into each neighbor
+                    # (sum over k ∈ K of SDR_k * p(i,k) in the equation above)
```
dcdenu4, line 544:
```suggestion
                    # calculate sum of SDR values of immediate downstream 
                    # neighbors, weighted by proportion of flow into each
                    # neighbor
                    # (sum over k ∈ K of SDR_k * p(i,k) in the equation above)
```

src/natcap/invest/sdr/sdr_core.pyx
dcdenu4:
```suggestion
                            # add the neighbor's flux value, weighted by the 
                            # flow proportion
```

src/natcap/invest/sdr/sdr_core.pyx
dcdenu4:
```suggestion
                            # sum up the neighbor's flow dir values in each
                            # direction.
```

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
...
+                            #
                             # if there is a downstream neighbor it
                             # couldn't have been pushed on the processing
                             # stack yet, because the upstream was just
                             # completed
                             upstream_neighbors_processed = 1
+                            # iterate over each neighbor-of-neighbor
                             for k in range(8):
+                                # no need to push the one we're currently calculating
+                                # back onto the stack
```
dcdenu4, line 586:
```suggestion
                                # no need to push the one we're currently
                                # calculating back onto the stack
```

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -471,62 +517,76 @@ def calculate_sediment_deposition(
...
                                 if inflow_offsets[k] == j:
-                                    # we don't need to process the one
-                                    # we're currently calculating
                                     continue
-                                # see if there's an inflow
+                                # skip if neighbor-of-neighbor is outside raster bounds
```
dcdenu4, line 589:
```suggestion
                                # skip if neighbor-of-neighbor is outside
                                # raster bounds
```

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -535,48 +595,53 @@ def calculate_sediment_deposition(
                                     neighbor_col + COL_OFFSETS[k])
                                 if ds_neighbor_col < 0 or ds_neighbor_col >= n_cols:
                                     continue
+                                # if any upstream neighbor of j hasn't been calculated,
+                                # we can't push j onto the stack yet
```
dcdenu4, line 599:
```suggestion
                                # if any upstream neighbor of j hasn't been
                                # calculated, we can't push j onto the stack
                                # yet
```

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -535,48 +595,53 @@ def calculate_sediment_deposition(
...
-                                                sediment_deposition_nodata):
-                                        # can't push it because not
-                                        # processed yet
+                                    if (sediment_deposition_raster.get(
+                                            ds_neighbor_col, ds_neighbor_row) ==
+                                            target_nodata):
                                         upstream_neighbors_processed = 0
                                         break
+                            # if all upstream neighbors of neighbor j are processed,
+                            # we can push j onto the stack.
```
dcdenu4, line 611:
```suggestion
                            # if all upstream neighbors of neighbor j are
                            # processed, we can push j onto the stack.
```

dcdenu4 (commented):
Hey @emlys,

Thanks for spearheading this and adding such great comments. I had a few questions and line too long linting suggestions. Otherwise it looks good to me and I'll let James stamp the approval.

src/natcap/invest/sdr/sdr_core.pyx
emlys: Oh weird! I like being able to use unicode but I have also run into a few more obscure symbols that don't render.

phargogh: @dcdenu4 are you using a unicode-compatible terminal for your vim session?  I think a modern vim/neovim should be able to handle UTF-8, but only if your terminal does too!

.github/workflows/binary-applications.yml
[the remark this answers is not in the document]
phargogh: > Would you be interested to try it out and see what happens?

I would be very interested to try this out!  Do you happen to have a build handy with this change?

> at runtime is GDAL still found on the PATH? I thought maybe when it's bundled then it doesn't need to search for it.

GDAL _does_ still need to search for it, and the python code that this environment variable affects will still execute when InVEST goes to import GDAL.  I do know that pyinstaller fiddles with the module search path to better handle its use case, but unfortunately I don't know if that ends up affecting `PATH`, specifically ... it sure could, though.

src/natcap/invest/sdr/sdr_core.pyx
emlys: Yeah, I rephrased this a little to hopefully clarify

src/natcap/invest/sdr/sdr_core.pyx
emlys: rephrased this comment slightly

.github/workflows/binary-applications.yml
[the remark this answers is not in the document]
emlys: Thanks! Here is a download link for the windows installer without setting USE_PATH_FOR_GDAL_PYTHON=YES: https://github.com/natcap/invest/suites/3523149147/artifacts/84168783
and with: https://github.com/natcap/invest/suites/3523577250/artifacts/84182605

requirements.txt
```diff
@@ -12,14 +12,14 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-GDAL>=3.1.2,<3.3.0
+GDAL>=3.1.2,!=3.3.0  # 3.3.0 had a bug that broke our windows builds: https://github.com/OSGeo/gdal/issues/3898
```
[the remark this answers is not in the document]
emlys: That's cool, I hadn't seen environment specifiers before! It doesn't look like conda supports a similar thing, so unless we have a particular reason to allow `3.3.0` on mac, I'm inclined to leave it as-is and not have to modify the `environment.yml` script

src/natcap/invest/sdr/sdr_core.pyx
dcdenu4: I'm using Cmder. I'll have to check!

.github/workflows/binary-applications.yml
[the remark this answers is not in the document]
phargogh: The binary you linked runs great on Windows 11 on ARM!  UI starts up with no problems at all.

requirements.txt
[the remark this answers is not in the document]
phargogh: I don't have a specific reason not to disallow 3.3.0 on mac other than it seems unnecessary to prevent someone from using a package version that isn't known to be a problem on a platform.

Since there's already a helpful comment explaining the situation, let's not worry about environment specifiers here.

src/natcap/invest/sdr/sdr_core.pyx
phargogh: >If `sdr_core.calculate_sediment_deposition` were called with an SDR input that didn't come from the model, you can construct a scenario where this happens

Thanks for providing this level of detail around a scenario that would exercise this.  And even though it is possible (as you demonstrate) that we could end up with some undefined behavior, I'd like to argue for the simpler solution of assuming that all of this function's inputs will come from the SDR model, clarifying this by renaming `calculate_sediment_deposition` to `_calculate_sediment_deposition` (and doing the same with the other `sdr_core` cython function) because:

1. All other functions used in the SDR task graph are likewise private
2. Private functions in python imply by convention that they are for internal use only and should not be used directly outside of the context of the model's `execute`
3. All of this would help to enforce that the contract we're exposing to the user is at the `execute` level and not for each individual component.

I suppose it's always possible for someone to call a function in a way they aren't supposed to and with data that aren't formatted correctly, but I'm just really hesitant to try to support something that shouldn't happen under normal model operation.

.github/workflows/binary-applications.yml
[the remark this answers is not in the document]
emlys: Great! Do both of them work (with and without the env variable set)? I'd rather not set it if not necessary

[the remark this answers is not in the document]
phargogh: Just to confirm, I re-ran both of these and:

* InVEST 3.9.0 runs great
* InVEST 3.9.0.post503+gcc10489c (which I believe is _without_ setting `USE_PATH_FOR_GDAL_PYTHON=YES`) does not launch.
* InVEST 3.9.0.post506+g12c334bd (which is with the environment variable) works great 

This is on Windows 11 on ARM, but I'd expect the same behavior on windows 10 x86/x64.

dcdenu4:
@emlys That crop production issue is interesting, thanks for putting the clear run down together. 

>We reclassified nodata to 0 before resampling

Wouldn't the old version of `reclassify` handle `nodata` as long as it was defined? Or was the test also passing in a troublesome float nodata that slipped through? I tried to find the crop production test failing but I only see a `sdr` test failing in actions. Sorry if I'm misunderstanding something here.

emlys:
You can see the test failures for `test_crop_production_percentile` and `test_crop_production_regression` in f1e2119 (have to look at the raw logs because it gets truncated).

> Wouldn't the old version of reclassify handle nodata as long as it was defined?

We explicitly told it to reclassify nodata to 0: https://github.com/natcap/invest/blob/4d0fbcb4561c73cda5ec157d159648dbe403ae44/src/natcap/invest/crop_production_percentile.py#L325-L326 
The 2.3.0 version of `reclassify_raster` allowed you to reclassify nodata to another value. In the 2.3.1 version, it ignores nodata pixels when reclassifying. You can still set the `target_nodata` parameter. That's arguably a bug, but I think anywhere we are replacing nodata with an inferred real value is kind of questionable.

emlys:
I've updated this to pygeoprocessing 2.3.2 and it's looking good!

src/natcap/invest/ndr/ndr.py
phargogh: Hmm, looks like they're still in the latest changes ... maybe there's an unpushed commit?

src/natcap/invest/ndr/ndr.py
phargogh: Same, maybe unpushed?

src/natcap/invest/sdr/sdr_core.pyx
phargogh: @emlys any thoughts on the above?

phargogh (changes requested):
This looks good to me!  But there were a couple of comments that look like maybe they were unresolved or referred to some commits that hadn't been pushed for some reason.  Maybe let me know what you think about those and if they're still relevant?  Thanks!

src/natcap/invest/sdr/sdr_core.pyx
emlys: Whoops, didn't see this! This makes sense to me. I reset c99bc44 after committing it so those changes aren't in here now. I did just add to the docstring explaining this assumption.

emlys:
@phargogh, I removed those unused imports and replied to that other comment.

src/natcap/invest/ndr/ndr.py
emlys: Weird, I definitely remember making that change before!

phargogh (approved):
Looks great, no further comments!

dcdenu4 (approved):
I thought I had approved this already!

PR #562, opened 2021-06-02
Fix rec model formatted strings

### LINKED ISSUE #78: Improperly formatted strings in Rec model error messages
In particular, there's one from `_validate_same_projection` that does not do the string substitution correctly: `raise ValueError("%s did not load", path)`

I'll fix this and look around for others as well.

emlys:
# Description
Fixes #78 
There were several incorrect formatted strings throughout this model so I replaced them all with f-strings for good measure.

# Checklist


davemfish (approved):
Thanks @emlys this is less buggy and so much easier to read now!

PR #571, opened 2021-06-16
Fix user's guide build process

### LINKED ISSUE #559: Building newer versions of the user's guide does not work
Originally pointed out by @richpsharp. When the UG commit hash got updated to a more recent version that includes the changes from https://github.com/natcap/invest.users-guide/pull/21, it broke the InVEST binary builds. That's because the UG build running within the InVEST build expected to have its own copy of the sample data, but it doesn't.

Need to handle both cases:
- Independent UG build (like in github actions on natcap/invest.users-guide). It should check out its own copy of the sample data.
- UG build within the InVEST build (in github actions on natcap/invest). It should ideally use the copy of the sample data already existing in the InVEST directory.

There will be changes to both the invest makefile and the user's guide makefile. The corresponding ticket for the user's guide is here: natcap/invest.users-guide#28

emlys:
# Description
Fixes #559 
Modify the `make userguide` command so that newer versions of the user's guide (that depend on invest-sample-data) can be built. Make a symlink from the existing invest-sample-data clone (at `invest/data/invest-sample-data`) to where the UG build expects to find it (`invest/doc/users-guide/invest-sample-data`) so that we don't have to have two copies.

This PR goes with https://github.com/natcap/invest.users-guide/pull/29. That one should get merged first so that I can update the UG variables here to point to natcap/invest.users-guide.

# Checklist


Makefile
```diff
@@ -268,9 +268,16 @@ $(APIDOCS_TARGET_DIR): | $(DIST_DIR)
 $(APIDOCS_ZIP_FILE): $(APIDOCS_TARGET_DIR)
 	$(BASHLIKE_SHELL_COMMAND) "cd $(DIST_DIR) && $(ZIP) -r $(notdir $(APIDOCS_ZIP_FILE)) $(notdir $(APIDOCS_TARGET_DIR))"
 
+# need to get the working directory path because ln doesn't work with relative paths
+WORKING_DIR := $(shell pwd)
+ifeq ($(OS),Windows_NT)
+	# this setting is necessary for ln to work on git bash for windows
+	export MSYS = winsymlinks:nativestrict
```
davemfish, line 275: I'm slightly surprised this is working with the spaces around `=` because I think that treats `MSYS` as a command and `=` as it's first argument. https://unix.stackexchange.com/questions/258727/spaces-in-variable-assignments-in-shell-scripts

davemfish: Is it possible this block is somehow not running and the symlink is still working?

davemfish (commented):
Hi @emlys this looks like it's working, but I did have one comment about the MSYS variable part.

Makefile
emlys: I just tried commenting out that line in ee01dca and it once again fails with "No space left on device": https://github.com/natcap/invest/pull/571/checks?check_run_id=2858930559#step:8:675 In 9991fa8 I un-commented it and called `echo $(MSYS)` and it is the expected value: https://github.com/natcap/invest/runs/2859138132#step:8:666

[This page](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) says _"To pass down, or export, a variable, make adds the variable and its value to the environment for running each line of the recipe."_ shows the `export variable = value` syntax. That's in the context of setting env variables for sub-make commands, but I think it's the same. [Here's](https://stackoverflow.com/a/27282778/14451410) a stackoverflow answer suggesting the same thing for making a variable available to commands in a make recipe.

emlys: I'm guessing it's different from what that unix.stackexchange post says because it's being handled by gnu make, not bash.

davemfish: Ah thanks for the `make` lesson :)

davemfish (approved):
Approved! I'm not sure if you wanted to keep the `echo` you just added, so I'll let you decide and then merge.

emlys:
Thanks @davemfish! I removed the echo and updated the makefile to point to the just-merged changes on natcap/invest.users-guide. Tests are passing now so I'm going to merge this.

PR #579, opened 2021-07-08
Bugfix/428 pandas unicode decode errors

### LINKED ISSUE #428: pandas 1.2.0 compatibility

davemfish:
This PR fixes #428. From my comment there:

>This issue came back with version 1.3.0 where pandas once again does not catch the UnicodeDecodeError and does not replace the characters. I really don't know why this keeps flipping back and forth - it doesn't seem like a deliberate decision as it's never mentioned in the changelogs.

>Assuming this behavior could switch back again with any given pandas release, I'm thinking the resilient solution for us is:
> 1. catch & re-raise a UnicodeDecodeError in utils.read_csv_to_dataframe with a message that tells users to use utf-8 encoding. That's how we solved this originally.
> 2. eliminate the two tests (test_utils & test_validation) that we have been switching back and forth between asserting the replacement characters are present & asserting the UnicodeDecodeError is raised. I think that test really only tests pandas behavior. And since we seem to be okay with either behavior, I don't think we need the test. Thoughts @phargogh ?

# Checklist

### DIFF
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -587,7 +587,7 @@ def read_csv_to_dataframe(
                                     sep=sep, **kwargs)
     except UnicodeDecodeError as error:
         LOGGER.error(
-            f'{path} must be encoded as utf-8 or avoid non-ASCII characters')
+            f'{path} must be encoded as utf-8 or ASCII')
         raise error
 
     # this won't work on integer types, which happens if you set header=None
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -582,8 +582,13 @@ def read_csv_to_dataframe(
     # allow encoding kwarg to override this if it's provided
     if not encoding and has_utf8_bom(path):
         encoding = 'utf-8-sig'
-    dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
-                                sep=sep, **kwargs)
+    try:
+        dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
+                                    sep=sep, **kwargs)
+    except UnicodeDecodeError as error:
+        LOGGER.error(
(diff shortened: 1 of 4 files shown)


src/natcap/invest/utils.py
```diff
@@ -582,8 +582,13 @@ def read_csv_to_dataframe(
...
     if not encoding and has_utf8_bom(path):
         encoding = 'utf-8-sig'
-    dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
-                                sep=sep, **kwargs)
+    try:
+        dataframe = pandas.read_csv(path, engine=engine, encoding=encoding,
+                                    sep=sep, **kwargs)
+    except UnicodeDecodeError as error:
+        LOGGER.error(
+            f'{path} must be encoded as utf-8 or avoid non-ASCII characters')
```
phargogh, line 590: This is super minor, but we could simplify the sentence by avoiding the double-negative: "must be encoded as utf-8 or ASCII"

phargogh (changes requested):
Thanks for addressing this, @davemfish !

Since this logger message is a user-facing change, would it be worth adding a note about it to HISTORY?

src/natcap/invest/utils.py
davemfish: That is better, thanks!

davemfish:
> Thanks for addressing this, @davemfish !
> 
> Since this logger message is a user-facing change, would it be worth adding a note about it to HISTORY?

Thanks, James. Re-worded the message and added the HISTORY note.

phargogh (approved):

PR #591, opened 2021-07-21
Fix UFRM error when given a string infrastructure 'Type' field.

### LINKED ISSUE #590: UFRM build infrastructure vector doesn't handle str(integer) type values
A [user on the forums](https://community.naturalcapitalproject.org/t/urban-flood-risk-mitigation-model-key-error-1/2057/4) appears to be using a `type` field in the built infrastructure vector that the model is interpreting as a _string_ rather than an integer.  If we were reading in the table with `pandas`, the type would be cast automatically during table load, but OGR will return the type indicated by the datatype of the field, uncast.  So in this case, the string `'1'` rather than the int `1`.

Easy fix if this is indeed the issue.  Awaiting user data to verify.

phargogh:
# Description

Like with many PRs of this nature, the test was more work than the fix itself.

Fixes #590

# Checklist

### DIFF
--- a/tests/test_ufrm.py
+++ b/tests/test_ufrm.py
@@ -6,8 +6,11 @@ import shutil
 import os
 
 from osgeo import gdal
+from osgeo import osr
+from osgeo import ogr
 import numpy
 import pygeoprocessing
+import shapely.geometry
 
 
 class UFRMTests(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tests/test_ufrm.py
```diff
@@ -151,6 +154,84 @@ def test_ufrm_value_error_on_bad_soil(self):
...
+        infra_fields = {'Type': ogr.OFTString}  # THIS IS THE THING TESTED
+        infra_attrs = [
+            {'Type': str(index)} for index in range(len(infra_geometries))]
+
+        infrastructure_path = os.path.join(
+            self.workspace_dir, 'infra_vector.shp')
+        pygeoprocessing.shapely_geometry_to_vector(
+            infra_geometries, infrastructure_path, projection_wkt,
+            'ESRI Shapefile', fields=infra_fields, attribute_list=infra_attrs,
+            ogr_geom_type=ogr.wkbPolygon)
```
dcdenu4, line 209: I'm probably just confused, but we're creating buffered infrastructure points along a line (because `yoff` stays at `100`). Then we take those geoms and make a vector type Polygon. Wouldn't that just be a line and the buffered points be unnecessary? Sorry if I'm completely missing something with how the infrastructure vector should be set up.

dcdenu4 (commented):
Hey @phargogh , just had a question about the test!

tests/test_ufrm.py
phargogh: Thanks for taking a look at this!

You're very right that they're along a line, but the circles are at 40-unit intervals along that line (see line 198) so the end result looks like this:

Screen Shot 2021-07-21 at 12 55 46 PM

dcdenu4: Ah right, sorry, my brain failed to make the buffered points to polygon connection.

dcdenu4 (approved):
I don't think the failing Mac test is related, so I'll merge this in.

PR #595, opened 2021-07-23
Add unit formatting function

emlys:
# Description
Related to natcap/invest.users-guide#30
Moving the `pint.Unit` -> `str` functionality into invest so we can use it in the workbench also

After finding this [style guide for SI units](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-6-rules-and-style-conventions-printing-and-using) and talking to @jagoldstein, I'm convinced we should use the standard unit symbols everywhere rather than spelling them out. This has the benefit of being language-independent and also easier to auto-format.

@davemfish, you were right that the pint formatting features would be useful! It's not mentioned in the tutorial, but there's a whole customizable formatting function that we can use here.

## Question
We now have several custom units defined in `spec_utils.py` which we define programmatically using `u.define()`. These add to/overwrite pint's [default units file](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt). There is the option to use a custom units file to replace that, and all our custom units could go in there. Thoughts?

# Checklist

### DIFF
--- a/src/natcap/invest/spec_utils.py
+++ b/src/natcap/invest/spec_utils.py
@@ -1,5 +1,3 @@
-import re
-
 import pint
 
 
@@ -29,7 +27,9 @@ u.define('international_unit = [biologic_amount] = iu = IU')
 # overwrite the default use of the symbol 'h' for henries
 u.define('henry = weber / ampere')
 u.define('hour = 60 * minute = h = hr')
-# overwrite the year definition to use 'yr' rather than 'a' as default symbol
+# overwrite the year definitionto use 'yr' rather than 'a' as default symbol
+# the symbol 'yr' is english-specific and the international symbol 'a' may
+# not be well-known, so we will need to translate this
 u.define('year = 365.25 * day = yr = a = julian_year')
 # Use u.none for unitless measurements
 u.define('none = []')
@@ -182,7 +182,13 @@ def format_unit(unit):
     Returns:
         String describing the unit.
     """
+    # Optionally use a pre-set format for a particular unit
     custom_formats = {
+        # For soil erodibility (t*h*ha/(ha*MJ*mm)), by convention the ha's
+        # are left on top and bottom and don't cancel out
+        # pint always cancels units where it can, so add them back in here
+        # this isn't a perfect solution
+        # see https://github.com/hgrecco/pint/issues/1364
         u.t * u.hr / (u.MJ * u.mm): 't · h · ha / (ha · MJ · mm)'
     }
     if unit in custom_formats:
--- a/src/natcap/invest/spec_utils.py
+++ b/src/natcap/invest/spec_utils.py
@@ -1,3 +1,5 @@
+import re
+
 import pint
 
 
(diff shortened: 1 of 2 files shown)


jagoldstein:
I'm totally onboard with using SI base units of measurement "symbols" as described by NIST and adopted from the [International Committee for Weights and Measures](https://www.bipm.org/en/measurement-units/si-base-units). The language-independence of these standardized abbreviations are more universally interpretable by users than are full unit names spelled out in English, thus streamlining auto-formatting and implementation in versions of the UI and UG translated to other languages.

Thanks for also updating [our internal Style Guide](https://docs.google.com/document/d/1BHwHDu_I-x0s_2GsbUb4rfVmXMkl7kl97sx2suBTLh8/edit?usp=sharing) to reflect the NIST standards @emlys!

davemfish:
> We now have several custom units defined in `spec_utils.py` which we define programmatically using `u.define()`. These add to/overwrite pint's [default units file](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt). There is the option to use a custom units file to replace that, and all our custom units could go in there. Thoughts?

I'm happy either way at this point. It feels natural to have them here in this `spec_utils.py` module for now.

src/natcap/invest/spec_utils.py
```diff
@@ -1,3 +1,5 @@
+import re
```
davemfish, line 1: Is this an unused import? Or I'm blind.

tests/test_utils.py
```diff
@@ -1437,17 +1447,33 @@ def test_exception_raised_with_details(self):
...
+            "Values in the LULC raster were found that are"
+            " not represented under the 'lucode' column"
+            " of the Biophysical table. The missing values found in"
+            " the LULC raster but not the table are: [3].")
         self.assertTrue(
             expected_message in str(context.exception), str(context.exception))
+
+
+class SpecUtilsTests(unittest.TestCase):
+
```
davemfish, line 1467: Could use a docstring

tests/test_utils.py
```diff
@@ -1437,17 +1447,33 @@ def test_exception_raised_with_details(self):
...
+    def test_format_unit(self):
```
davemfish, line 1468: docstring please

src/natcap/invest/spec_utils.py
```diff
@@ -156,3 +164,40 @@
...
+
+    Args:
+        unit (pint.Unit): the unit to format
+
+    Returns:
+        String describing the unit.
+    """
+    custom_formats = {
+        u.t * u.hr / (u.MJ * u.mm): 't · h · ha / (ha · MJ · mm)'
+    }
```
davemfish, line 187: Could we add a comment here for when/why we would want to use a custom format?

src/natcap/invest/spec_utils.py
```diff
@@ -23,11 +25,17 @@
 # May be converted to weight or volume, but conversion factors are specific
 # to the substance. I couldn't find a definition of its dimensionality.
 u.define('international_unit = [biologic_amount] = iu = IU')
+# use 'h' not 'hr' as the symbol for hour, as per SI guidelines
+# overwrite the default use of the symbol 'h' for henries
+u.define('henry = weber / ampere')
+u.define('hour = 60 * minute = h = hr')
+# overwrite the year definition to use 'yr' rather than 'a' as default symbol
+u.define('year = 365.25 * day = yr = a = julian_year')
```
davemfish, line 33: We talked about not using the alias `yr` and always printing `year` for ease of translation later. I'm not quite sure how the strings here will integrate with translation, so for now whatever you think is best is good with me.

davemfish (changes requested):
This looks like it works great, @emlys . It's easy to envision how we might use this function for the Workbench as well. 

I found a few very minor items to address.

src/natcap/invest/spec_utils.py
emlys: you're right! good catch

src/natcap/invest/spec_utils.py
emlys: I think this is ok for now. I believe we'll be able to translate any part of these with `gettext` so maybe the `yr` symbol can get swapped out for an appropriate word/abbreviation in other languages.

emlys:
@davemfish I'm going to leave the programmatic unit definitions for now. A definition file would be nice but I'm not sure if it makes sense to put that in `src/natcap/invest/` or how to make sure it gets bundled by pyinstaller (?) So I'll come back to that in the future.

src/natcap/invest/spec_utils.py
```diff
@@ -156,3 +164,46 @@
...
+    Args:
+        unit (pint.Unit): the unit to format
+
+    Returns:
+        String describing the unit.
+    """
+    # Optionally use a pre-set format for a particular unit
+    custom_formats = {
+        # For soil erodibility (t*h*ha/(ha*MJ*mm)), by convention the ha's
+        # are left on top and bottom and don't cancel out
```
davemfish, line 188: Weird & interesting, thanks for the comment!

davemfish (approved):

PR #598, opened 2021-07-28
CV test: ignore order of rows when comparing output to regression data

### LINKED ISSUE #597: Mysterious CV test failure
One CV test is failing on my branch: https://github.com/natcap/invest/pull/555/checks?check_run_id=3174860834#step:12:2380 The error is not related to the pygeoprocessing/gdal update.
```
E   AssertionError: DataFrame.iloc[:, 2] (column name="R_hab") are different
E   
E   DataFrame.iloc[:, 2] (column name="R_hab") values are different (25.0 %)
E   [index]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
E   [left]:  [4.05, 5.0, 4.05, 5.0, 5.0, 5.0, 5.0, 5.0, 5.0, 5.0, 1.7999999999999998, 5.0, 5.0, 5.0, 5.0, 5.0]
E   [right]: [4.05, 5.0, 5.0, 5.0, 5.0, 5.0, 5.0, 5.0, 1.7999999999999998, 5.0, 5.0, 5.0, 5.0, 5.0, 4.05, 5.0]
```
I can reproduce this locally on my branch and `main`. I traced it back to 69c34e17 where the test data commit hash was updated. Why has the test been passing on main all this time but failing locally and on my branch?

emlys:
# Description
Fixes #597

# Checklist

### DIFF
--- a/tests/test_coastal_vulnerability.py
+++ b/tests/test_coastal_vulnerability.py
@@ -713,22 +713,67 @@ class CoastalVulnerabilityTests(unittest.TestCase):
         args['slr_field'] = 'Trend'
         coastal_vulnerability.execute(args)
 
-        # shore points aren't necessarily created/numbered in the same order,
-        # so we can't compare the fid or shore_id fields at all
-        # and need to sort the rows before comparing
-        actual_values_df = pandas.read_csv(
-            os.path.join(args['workspace_dir'], 'coastal_exposure.csv')
-        ).drop(['fid', 'shore_id'], axis=1)
-        actual_values_df = actual_values_df.sort_values(
-            by=list(actual_values_df.columns)).reset_index(drop=True)
-        expected_values_df = pandas.read_csv(
-            os.path.join(REGRESSION_DATA, 'expected_coastal_exposure.csv')
-        ).drop(['fid', 'shore_id'], axis=1)
-        expected_values_df = expected_values_df.sort_values(
-            by=list(expected_values_df.columns)).reset_index(drop=True)
+        results_vector = gdal.OpenEx(
+            os.path.join(args['workspace_dir'], 'coastal_exposure.gpkg'),
+            gdal.OF_VECTOR)
+        results_layer = results_vector.GetLayer()
 
-        pandas.testing.assert_frame_equal(
-            actual_values_df, expected_values_df, check_dtype=False)
+        # shore points aren't necessarily created/numbered in the same order,
+        # so we can't compare by the fid or shore_id fields
+        # instead check that each point has the expected attributes
+        fields = [
+            'R_hab', 'R_wind', 'R_wave', 'R_surge', 'R_relief', 'R_geomorph',
+            'R_slr', 'population', 'exposure', 'habitat_role',
+            'exposure_no_habitats']
+        # mapping of x, y point coords to their fields in the order above
+        expected_point_results = {
+            (265195.9118871244, 5465342.359978485): [
+                4.05, 5, 5, 1, 5, 1, 5, 0, 3.063308387, 0.0936167899, 3.1569251777],
+            (275802.584962168, 5469283.47691652): [
+                5, 3, 1, 1, 5, 1, 4, 0, 2.2587827631, 0, 2.2587827631],
+            (376479.9819538344, 5383636.197270025): [
+                4.05, 5, 5, 4, 4, 4, 3, numpy.nan, 4.0989337064, 0.125266204816, 4.2241999112],
+            (355807.04065901926, 5394736.771414153): [
+                5, 5, 4, 4, 3, 4, 2, 0, 3.70591872429, 0, 3.70591872429],
+            (344736.25795214524, 5411347.428706937): [
+                5, 2, 4, 3, 3, 4, 1, 0, 2.8261463109, 0, 2.8261463109],
+            (361290.81087879254, 5427975.474574804): [
+                5, 1, 3, 5, 1, 4, 2, 0, 2.493898362454, 0, 2.493898362454],
+            (368269.9048903524, 5449541.99543876): [
+                5, 1, 1, 5, 2, 4, 2, 0.008, 2.3535468936, 0, 2.3535468936],
+            (361341.1463245464, 5433678.995435326): [
+                5, 2, 1, 5, 1, 4, 1, 0, 2.1316631165, 0, 2.1316631165],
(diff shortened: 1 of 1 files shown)


emlys:
Updated this to test the vector output, not the csv summary.

tests/test_coastal_vulnerability.py
```diff
@@ -713,26 +713,67 @@ def test_complete_run(self):
...
-            actual_values_df['shore_id'], intermediate_csv['shore_id'])
-        pandas.testing.assert_series_equal(
-            actual_values_df['shore_id'], habitat_csv['shore_id'])
+        results_vector = gdal.OpenEx(
+            os.path.join(args['workspace_dir'], 'coastal_exposure.gpkg'),
+            gdal.OF_VECTOR)
+        results_layer = results_vector.GetLayer()
+
+        # shore points aren't necessarily created/numbered in the same order,
+        # so we can't compare by the fid or shore_id fields
```
davemfish, line 722: I think we *can* expect the points be created/numbered in the same order each time the same version of the model runs. I'm guessing something upstream (shapely?) changed to make the order change between versions.

tests/test_coastal_vulnerability.py
```diff
@@ -713,26 +713,67 @@ def test_complete_run(self):
...
+                5, 1, 3, 3, 2, 1, 3, 0, 2.225039271, 0, 2.225039271],
+            (290844.10673833836, 5460752.369983306): [
+                5, 4, 2, 2, 2, 1, 4, 0, 2.5169979012, 0, 2.5169979012],
+            (286971.1600469994, 5475571.525391179): [
+                5, 1, 1, 3, 1, 1, 4, 0, 1.7948229213, 0, 1.7948229213],
+            (268404.5296770451, 5479539.256472221): [
+                5, 2, 2, 1, 1, 1, 5, 0, 1.9306977288, 0, 1.9306977288],
+            (252692.2222200262, 5482540.950248547): [
+                5, 3, 2, 1, 4, 1, 5, 0, 2.49389836245, 0, 2.49389836245]
+        }
```
davemfish, line 762: I like having the expected data hardcoded!

tests/test_coastal_vulnerability.py
```diff
@@ -713,26 +713,67 @@ def test_complete_run(self):
...
+            x = geom.GetX()
+            y = geom.GetY()
+            actual_row = []
+            for field in fields:
+                value = feature.GetField(field)
+                if value is None:
+                    actual_row.append(numpy.nan)
+                else:
+                    actual_row.append(value)
+            expected_row = expected_point_results[(x, y)]
```
davemfish, line 775: If the geometry was not as expected, this would raise a `KeyError` right? Do you think it makes sense to have an assertion for the geometry instead so the test can fail rather than error in that case?

tests/test_coastal_vulnerability.py
davemfish: Out of curiosity, did you notice if these coordinates match the previous regression data? Or did the coords change slightly along with the order of the points? It doesn't really matter either way I don't think.

tests/test_coastal_vulnerability.py
```diff
@@ -713,26 +713,67 @@ def test_complete_run(self):
...
-                args['workspace_dir'],
-                'intermediate/intermediate_exposure.csv'))
-        habitat_csv = pandas.read_csv(
-            os.path.join(
-                args['workspace_dir'],
-                'intermediate/habitats/habitat_protection.csv'))
-        pandas.testing.assert_series_equal(
-            actual_values_df['shore_id'], intermediate_csv['shore_id'])
-        pandas.testing.assert_series_equal(
-            actual_values_df['shore_id'], habitat_csv['shore_id'])
```
davemfish, line 735: I think this block is still important to test. We need to assert that each table has the `shore_id` field and that it can be used to correctly join the tables. It looks like these assertions are comparing the tables within the current run - not comparing against regression data - so I think this check is still useful and resilient to a changing order across model versions.

davemfish (changes requested):
Hi @emlys , thanks for sorting out this test for us! I had a couple comments for minor changes if you agree.

tests/test_coastal_vulnerability.py
phargogh: Yep, I concur.  CV's collection of tables form a nice relational database associated by `shore_id`, so while the FID might not be consistent, the `shore_id`s should allow for the same records to be joined across multiple tables.

tests/test_coastal_vulnerability.py
phargogh: Another thing I'd worry about is the floating-point precision of the coordinates.  If we use, as in the data added above, 11 digits of precision, two numerically close floats can still result in an different hash when used as a dict key despite being within the usual 10e-7 relative tolerance:

```python
>>> hash(0.1)
230584300921369408
>>> hash(0.09999999999)
230584300898310976
```

As an alternative, I would suggest sorting by the coordinates, checking the coordinates for numerical closeness and then asserting the closeness of the fields.

phargogh (changes requested):
Thanks for digging into this, @emlys !  I just had one minor suggestion about dictionary indexing in addition to what @davemfish mentioned in his review.

tests/test_coastal_vulnerability.py
emlys: They looked to be identical.

tests/test_coastal_vulnerability.py
emlys: Added this back in

tests/test_coastal_vulnerability.py
emlys: Good point

tests/test_coastal_vulnerability.py
emlys: The shapely version was the same (1.7.1) between passing and failing runs. I expect they would be created in the same order each time, but I don't think shapely guarantees it. They just describe a MultiPolygon as a "collection".

davemfish (approved):
Looks great, thanks for sorting (ha. ha.) that out @emlys !

emlys:
@phargogh, would you like to look at this again or should I merge?

phargogh (approved):
Thanks @emlys !

PR #582, opened 2021-07-09
Graceful exit from CLI even after invest exception

davemfish:
This fixes #580 . Aside from preventing the trailing 'Failed to execute script cli` message, it also changes the traceback slightly as it now doesn't go all the way back to include 
```
  File "natcap\invest\cli.py", line 663, in <module>
  File "natcap\invest\cli.py", line 590, in main
  File "natcap\invest\carbon.py", line 317, in execute
```
That makes sense, but I can't really explain why there's a different number of Traceback blocks after this change. Maybe there's a different/better way to fix #580?

See the full before & after tracebacks:

**Before:**
```
07/09/2021 11:05:32  pygeoprocessing.geoprocessing ERROR    exception encountered in raster_calculator
Traceback (most recent call last):
  File "pygeoprocessing\geoprocessing.py", line 440, in raster_calculator
  File "pygeoprocessing\geoprocessing.py", line 1849, in _map_dataset_to_value_op
pygeoprocessing.geoprocessing.ReclassificationMissingValuesError: ('The following 11 raster values [52 53 54 55 56 57 58 59 60 61 62] from "C:\\Users\\dmf\\projects\\invest\\data\\bad-data\\Carbon\\lulc_current_willamette.tif" do not have corresponding entries in the ``value_map``: {1: 1.35, 2: 0.45, 3: 0.18, 4: 0.0, 5: 0.9, 6: 0.0, 7: 0.0, 8: 0.0, 9: 0.0, 10: 0.0, 11: 0.0, 16: 0.0, 102: 0.0}.', array([52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 62], dtype=int16))
07/09/2021 11:05:32  pygeoprocessing.geoprocessing INFO     Waiting for raster stats worker result.
07/09/2021 11:05:32  pygeoprocessing.geoprocessing_core WARNING  No valid pixels were received, sending None.
07/09/2021 11:05:32  taskgraph.Task     ERROR    Something went wrong when adding task carbon_map_c_above_cur (0), terminating taskgraph.
Traceback (most recent call last):
  File "natcap\invest\utils.py", line 863, in reclassify_raster
  File "pygeoprocessing\geoprocessing.py", line 1856, in reclassify_raster
  File "pygeoprocessing\geoprocessing.py", line 440, in raster_calculator
  File "pygeoprocessing\geoprocessing.py", line 1849, in _map_dataset_to_value_op
[...]

PR #600, opened 2021-08-02
reverting the changes in #582, to address #580.

### LINKED ISSUE #580: CLI - graceful exit after an invest model error
This originally came up as https://github.com/natcap/invest-workbench/issues/131 where anytime a model's `execute` errors, the traceback is sent to stderr, but then the cli also emits `'Failed to execute script cli'` to stderr. That message is associated with pyinstaller-made executables. 

We can avoid that message by always being sure to exit gracefully from `cli.py`. We do that elsewhere in the module, but aren't doing it right now for the `invest run --headless` command. I think we should avoid the message in this case because it's not relevant to the invest model's exception, and `cli.py` did not really fail to execute in this case.

davemfish:
This PR reverts the changes we made in #582. It was not a good idea to catch all invest exceptions just for the purpose of hiding the final bit of stderr emitted by the Pyinstaller-built executeable. 
Fixes #580 

# Checklist
- [ ] Updated HISTORY.rst - we never updated history during #582, so no need to update it here.

### DIFF
--- a/src/natcap/invest/cli.py
+++ b/src/natcap/invest/cli.py
@@ -587,14 +587,7 @@ def main(user_args=None):
 
             # We're deliberately not validating here because the user
             # can just call ``invest validate <datastack>`` to validate.
-            try:
-                getattr(model_module, 'execute')(parsed_datastack.args)
-            # Graceful exit from this program, otherwise the pyinstaller-made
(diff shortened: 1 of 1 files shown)


dcdenu4 (approved):
Looks good.

Issue #567, opened 2021-06-10
Translations: Consolidate text between user's guide and args specs


emlys:
Corresponding issue: natcap/invest.users-guide#31
We should decide on some guidelines for what goes where (being discussed in a [google doc](https://docs.google.com/document/d/1hIkiB7-O3WQRHNhpfOIJ6liRqCLsrvKhDJjzzebZMBs/edit#)). Ideally we would get a scientist familiar with each model to go over the consolidated text and verify that we haven't lost anything.

emlys:
This was done in #604, #633, and #639.

PR #611, opened 2021-08-05
Unify our two release branches: release/workbench-alpha & release/3.10

### LINKED ISSUE #610: validation tests for option_string should test the new ARGS_SPEC options
I think `test_validation.py` needs some minor updates to better test the new spec for args with `type: option_string` where `options` should be a set or a dict. Right now these probably pass by coincidence because they test on lists.

```
    def test_valid_option(self):
        """Validation: test that a string is a valid option."""
        from natcap.invest import validation
        self.assertEqual(None, validation.check_option_string(
            'foo', options=['foo', 'bar', 'Baz']))

    def test_invalid_option(self):
        """Validation: test when a string is not a valid option."""
        from natcap.invest import validation
        error_msg = validation.check_option_string(
            'FOO', options=['foo', 'bar', 'Baz'])
        self.assertTrue('must be one of' in error_msg)
```

davemfish:
We decided it's no longer necessary to keep a separate release branch devoted to changes needed for the Workbench. We're confident we'll want all those changes going forward, so we're merging them into `release/3.10`. A quick test revealed many conflicts in `ARGS_SPECs`, but no conflicts anywhere else. So I did the following:

* Created this `task` branch from `release/workbench-alpha`
* Reverted the commits in #486 on the `workbench` branch. I believe all that was superseded by the work in #526 on the `3.10` branch. And many of those changes would have passed through without conflict, effectively undoing the work from #526.
* `git merge release/3.10 --strategy-option theirs` -- if there were conflicts from this merge, it will automatically take the changes from `release/3.10`
* Manually updated CV & Finfish `ARGS_SPEC`s that actually did have changes in the Workbench branch that we want to keep.
* Updated HISTORY to replace the Workbench section with a 3.10 section - we haven't had 3.10 section up until now.
* Minor changes to `test_validation` to fix #610 

# Checklist


davemfish:
This PR fixes #612 . And we should also remember to delete the `release/workbench-alpha` branch when this is merged.

emlys (approved):
This looks great @davemfish! And thanks for fixing the options validation thing.
I do think we'll want some of those changes from #486 such as lower-casing, but it will be more efficient to do that in the course of #567 than deal with merge conflicts.

PR #623, opened 2021-08-19
UI server test coverage & ARGS_SPEC Workbench compatibility

### LINKED ISSUE #617: ui_server: ARGS_SPEC is not JSON serializable
Since we added `pint.Unit` types to the `ARGS_SPEC`, the dict as a whole is not JSON serializable, but `ui_server.getspec` expects that it will be. We could either update `ui_server`, or we could start enforcing a contract that `ARGS_SPEC` should always be JSON serializable.

This was caught by the Workbench test suite (`flaskapp.test.js`) so it also begs the question that maybe we should move tests of `ui_server.py` over to `natcap.invest`. That's something we considered when we first moved `ui_server` from `invest-workbench` to `natcap.invest`, and I believe we just decided to wait and see at that point.

davemfish:
This PR makes small modifications to ARGS_SPEC datatypes and adds a utility function to make sure all `ARGS_SPEC` dictionaries are JSON serializable. 

Fixes #617 

# Checklist


src/natcap/invest/ui_server.py
```diff
@@ -202,39 +193,9 @@ def save_to_python():
...
-        # * add extra space to spacing before first item
-        args = args.replace('{', '{\n ')
-        args = args.replace('}', ',\n}')
-        py_file.write(script_template.format(
-            invest_version=natcap.invest.cli.__version__,
-            today=datetime.now().strftime('%c'),
-            modelname=modelname,
-            py_model=pyname,
-            model_args=args))
+    cli.export_to_python(
```
davemfish, line 198: I think we always intended to make use of this `cli` function here, but must have forgotten to make the switch earlier.

src/natcap/invest/ui_server.py
```diff
@@ -25,7 +22,7 @@
     v.pyname: _UI_META(
         run_name=k,
         human_name=v.humanname)
-    for k, v in natcap.invest.cli._MODEL_UIS.items()}
+    for k, v in cli._MODEL_UIS.items()}
```
emlys, line 25: To be nitpicky, `key` and `value` or `val` instead of one-letter variables?

tests/test_args_specs.py
```diff
@@ -311,6 +325,43 @@ def validate_permissions_value(self, permissions):
...
+    def test_model_specs_serialize(self):
+        """ARGS_SPEC: test each ARGS_SPEC can serialize to JSON."""
+        from natcap.invest import spec_utils
+
+        for model_name, metadata in cli._MODEL_UIS.items():
+            model = importlib.import_module(metadata.pyname)
+            try:
+                _ = spec_utils.serialize_args_spec(model.ARGS_SPEC)
+            except TypeError:
+                self.fail(
```
emlys, line 337: Neat, I haven't seen `self.fail` before! Could we also show the original error traceback?

tests/test_ui_server.py
```diff
@@ -34,3 +42,101 @@ def test_get_vector_colnames(self):
...
+            'model_name': 'natcap.invest.carbon'
+        }
+        filepath = os.path.join(self.workspace_dir, 'datastack.json')
+        with open(filepath, 'w') as file:
+            file.write(json.dumps(expected_datastack))
+        response = test_client.post('/post_datastack_file', json=filepath)
+        response_data = json.loads(response.get_data(as_text=True))
+        self.assertEqual(
+            list(response_data),
+            ['type', 'args', 'module_name', 'model_run_name',
```
emlys, line 103: It looks like the `json` module preserves the order of keys, so this should work. But JSON is generally unordered outside of the python implementation, so I think `ui_server.post_datastack_file` could put them in any order without breaking its contract. Maybe use a set here just in case?

tests/test_ui_server.py
```diff
@@ -34,3 +42,101 @@ def test_get_vector_colnames(self):
...
+                'workspace_dir': 'foo'
+            }),
+            'relativePaths': True,
+        }
+        _ = test_client.post('/write_parameter_set_file', json=payload)
+        with open(filepath, 'r') as file:
+            actual_data = json.loads(file.read())
+        self.assertEqual(
+            list(actual_data),
+            ['args', 'invest_version', 'model_name'])
```
emlys, line 125: like above, maybe should be a set? If we do want to guarantee the order of keys in the file produced by `write_parameter_set_file`, lets put that in the docstring

tests/test_ui_server.py
```diff
@@ -34,3 +42,101 @@ def test_get_vector_colnames(self):
...
+
+    def test_get_invest_spec(self):
+        """UI server: get_invest_spec endpoint."""
+        from natcap.invest import ui_server
+        test_client = ui_server.app.test_client()
+        response = test_client.post('/getspec', json='sdr')
+        spec = json.loads(response.get_data(as_text=True))
+        self.assertEqual(
+            list(spec),
+            ['model_name', 'module', 'userguide_html',
```
emlys, line 63: like below, `set`?

tests/test_ui_server.py
```diff
@@ -34,3 +42,101 @@ def test_get_vector_colnames(self):
...
         self.assertEqual(colnames, [])
+
+    def test_get_invest_models(self):
+        """UI server: get_invest_models endpoint."""
+        from natcap.invest import ui_server
+        test_client = ui_server.app.test_client()
+        response = test_client.get('/models')
+        models_dict = json.loads(response.get_data(as_text=True))
+        for model in models_dict.values():
+            self.assertEqual(list(model), ['internal_name', 'aliases'])
```
emlys, line 53: like below, `set`?

emlys (commented):
Thanks @davemfish! Just a few minor comments.

src/natcap/invest/ui_server.py
davemfish: good idea

tests/test_args_specs.py
davemfish: Yeah `self.fail` is new to me too. This seems like the preferred pattern to assert that a specific exception is avoided.

Good idea adding the original error. It's hard (but not impossible) to get the full traceback without ever raising the error. And this `TypeError` should always be the one explicitly raised by `spec_utils.fallback_serializer` so I think it's okay to just include that message without traceback. An example with these changes:

```
E               AssertionError: Failed to avoid TypeError when serializing natcap.invest.carbon.ARGS_SPEC:
E               fallback serializer is missing for <class 'module'>
```

tests/test_ui_server.py
davemfish: Great point! I changed all these tests to use sets instead of lists.

tests/test_ui_server.py
davemfish: order shouldn't matter; changed to a set.

PR #643, opened 2021-09-15
Custom unit definitions file

### LINKED ISSUE #631: Suppress Warnings from pint.util
We have all these warnings logged when we import `spec_utils.py`

```
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'pixel' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'survey_foot' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'sft' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'henry' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'hour' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'h' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'hr' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'year' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'yr' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'a' ()
08/31/2021 08:36:43  pint.util          WARNING  Redefining 'julian_year' ()
```

Since we decided it's a good idea to redefine those units, I think we should suppress the warnings.

emlys:
# Description
Fixes #631 
In addition to avoiding the warnings you get from redefining existing units, having a custom unit definitions file lets us use just the subset of units relevant for invest. 

The downside is that we no longer have the assurance that our unit definitions are correct, since we're not getting them from `pint`. But I labeled the section of definitions that I copied from the pint default file, and they're very unlikely to change.

# Checklist
~- [ ] Updated HISTORY.rst (if these changes are user-facing)~
~- [ ] Updated the user's guide (if needed)~


davemfish:
> In addition to avoiding the warnings you get from redefining existing units, having a custom unit definitions file lets us use just the subset of units relevant for invest.
> 
> The downside is that we no longer have the assurance that our unit definitions are correct, since we're not getting them from `pint`. But I labeled the section of definitions that I copied from the pint default file, and they're very unlikely to change.

@emlys thanks for linking to the docs on this. I'm reading this about whether to append to the default registry or define a completely new one as we're doing here. Does this apply to our situation?
> If you make a translation of the default units or define a completely new set, you don’t want to append the translated definitions so you just give the filename to the constructor:
https://pint.readthedocs.io/en/stable/defining.html

Otherwise I could see the benefit of appending to the default, that way we're not maintaining the copy of all those definitions in our file. What do you think?

davemfish (commented):
Looks like a good choice to move these defs to the file! 

I commented about how to handle the default definitions. 

My other comment is, maybe we could have a `.txt` extension on the new file? I see pint's default comes with that extension so I assume it's okay and it just adds a tiny bit of marginally useful metadata.

And maybe a naive question, but your #644 got me thinking...does this new file make it into our binary builds? I guess there would be some sort of `FileNotFoundError` when importing `spec_utils` if it was missing. Would that get caught anywhere in our build process?

emlys:
@davemfish, this should be good now. 
* renamed `unit_definitions` to `unit_definitions.txt`
* took your suggestion to use `load_definitions` so we can use the defaults provided by `pint`, and keep just our custom definitions in the file.
* added a setting to suppress those warnings. this feature is not mentioned in the documentation but is clearly marked in the code, so i feel ok about using it. https://github.com/hgrecco/pint/blob/16b1385ae3a97f6a93384d06936f99785534f455/pint/registry.py#L199

emlys:
And no changes needed to include `unit_definitions.txt` in the pyinstaller build. The `natcap.invest` hook automatically collects non-python files living in the `natcap.invest` package.

src/natcap/invest/spec_utils.py
```diff
@@ -1,41 +1,15 @@
...
-u.define('henry = weber / ampere')
-u.define('hour = 60 * minute = h = hr')
-# overwrite the year definition to use 'year' rather than 'a' as default symbol
-# the symbol 'yr' is english-specific and the international symbol 'a' may
-# not be well-known, so we will need to translate this
-u.define('year = 365.25 * day = _ = yr = a = julian_year')
-# Use u.none for unitless measurements
-u.define('none = []')
-
+u._on_redefinition = 'ignore'  # don't raise warnings when redefining units
```
davemfish, line 10: Thanks for linking the source code here. If you wanted I think you could do `u = pint.UnitRegistry(on_redefinition='ignore')`. I guess maybe that's the intended usage, but I don't have any reason to suggest it's better.

davemfish (approved):
Thanks for these updates, this looks great to me! I had one really minor change if you'd like to consider it. Otherwise I approve.

src/natcap/invest/spec_utils.py
emlys: Oh good catch! You're right that's probably the intended usage.

PR #650, opened 2021-09-16
Change signtool timestamp option t to tr

emlys:
Missed this the first time around. `/t` needed to change to `/tr`, indicating it's a RFC 3161 time stamp server, not just any old time stamp server. I'm guessing this is required to make sure the hash option in `/td` is supported.

The code signing step does not run in PR workflows, which is how I missed this last time, but you can see that it's working in the checks on my branch.


phargogh (approved):
Let's try it out!

PR #656, opened 2021-09-22
Update Makefile for release

emlys:
Update the sample data, test data, and user's guide hashes to point to the latest revision on `main` or `master` for each. This includes the updates to SDR/NDR UG pages, and removes some test data files that are no longer needed.


dcdenu4 (approved):
Sounds good!

Issue #401, opened 2020-12-07
Water Yield naming inconsistency


dcdenu4:
I think this has always been a weird issue, but I've seen
- Water Yield (model name in startup)
- Hydropower Water Yield (name in model UI and default workspace name)
- Annual Water Yield (UG Name and sample data name)
- hydropower_water_yeild (CLI name)

I think that we've tried to get away from the Hydropower naming convention and adopt the Annual Water Yield one.

emlys:
I've seen this with a lot of the models too and I think it's becoming more of a problem now that we have the workbench. It's been confusing figuring out how to refer to a model. The args spec has `model_name` e.g. 'Wind Energy' and `module` e.g. `natcap.invest.wind_energy`, while `invest list` has two different names: `wind_energy` and `Offshore Wind Energy Production`. It would be great to standardize these.

dcdenu4:
Yep, and for the Wind Energy model, I reckon it should be `offshore_wind_energy`, because that's what the model and data are designed for. But, if we decide to mix some of these up that change the UG or user facing UI, we should try and get confirmation with the science leads.

emlys:
This was fixed in #677.

PR #677, opened 2021-10-05
Single source of truth for model display names

### LINKED ISSUE #625: Standardize how we deal with model "names"
**This issue is about code-style, not about names of invest models.**

Each model has a few different types of "names" that are used in various places. These are catalogued in the `cli._MODEL_UIS` dict. It looks like this:

```
    'carbon': _UIMETA(
        humanname="Carbon Storage and Sequestration",
        pyname='natcap.invest.carbon',
        gui='carbon.Carbon',
        aliases=()),
   'coastal_blue_carbon': _UIMETA(
        ...
```

We end up passing these strings around to various parts of invest (`datastack.py`, `cli.py`, `ui_servery.py` & `invest-workbench`) and it gets confusing for the developer to remember what string we're dealing with when we see a variable named `model_name`, for example. 

So I propose standardizing our variable names across `invest` and `invest-workbench` based on the above dictionary:
* **`model_name`**: to refer to the names used as the keys in the above dictionary. Now, sometimes we use `internal_name` or `modelRunName` to refer to these strings.
* **`pyname`** : to refer to the importable module that has the `execute` method. Now, sometimes we use `model_name` for this.
* **`humanname`** or `model_title`: to refer to the "title" of the model.

Backwards-compatibility issues: 
-----------------------------------
* datastack archives/json files have a "model_name" key that my proposal would change to "pyname". But that would make older datastacks incompatible unless we ensure backwards-compatibility.
* `ARGS_SPEC` have properties that this proposal would change:
    * `model_name` would become `humanname` or `model_title` - But as of a pending workbench PR (https://github.com/natcap/invest-workbench/pull/165), I don't think there are any instances where this property is used, so maybe it can be removed from `ARGS_SPEC` altogether. If we wish to keep it, I would suggest we remove the hardcoded values and reference the lookup in `cli.py`.
    * `module` would become `pyname` - I think this only affects the workbench so it's no trouble to change.

emlys:
# Description
Fixes #625

# Checklist

### DIFF
--- a/src/natcap/invest/carbon.py
+++ b/src/natcap/invest/carbon.py
@@ -17,7 +17,7 @@ from . import utils
 LOGGER = logging.getLogger(__name__)
 
 ARGS_SPEC = {
-    "model_name": "InVEST Carbon Model",
+    "model_name": "Carbon Storage and Sequestration",
     "module": __name__,
     "userguide_html": "carbonstorage.html",
     "args_with_spatial_overlap": {
--- a/src/natcap/invest/cli.py
+++ b/src/natcap/invest/cli.py
@@ -30,148 +30,79 @@ DEFAULT_EXIT_CODE = 1
 LOGGER = logging.getLogger(__name__)
 _UIMETA = collections.namedtuple('UIMeta', 'humanname pyname gui aliases')
 
-_MODEL_UIS = {
-    'carbon': _UIMETA(
-        humanname="Carbon Storage and Sequestration",
-        pyname='natcap.invest.carbon',
-        gui='carbon.Carbon',
-        aliases=()),
-    'coastal_blue_carbon': _UIMETA(
-        humanname="Coastal Blue Carbon",
-        pyname='natcap.invest.coastal_blue_carbon.coastal_blue_carbon',
-        gui='cbc.CoastalBlueCarbon',
-        aliases=('cbc',)),
-    'coastal_blue_carbon_preprocessor': _UIMETA(
-        humanname="Coastal Blue Carbon: Preprocessor",
-        pyname='natcap.invest.coastal_blue_carbon.preprocessor',
-        gui='cbc.CoastalBlueCarbonPreprocessor',
-        aliases=('cbc_pre',)),
-    'coastal_vulnerability': _UIMETA(
-        humanname="Coastal Vulnerability",
-        pyname='natcap.invest.coastal_vulnerability',
-        gui='coastal_vulnerability.CoastalVulnerability',
-        aliases=('cv',)),
-    'crop_production_percentile': _UIMETA(
-        humanname="Crop Production: Percentile Model",
-        pyname='natcap.invest.crop_production_percentile',
-        gui='crop_production.CropProductionPercentile',
-        aliases=('cpp',)),
-    'crop_production_regression': _UIMETA(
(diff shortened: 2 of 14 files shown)


emlys:
@davemfish, you inspired me to put some more effort into this so I added a few things in addition to what you commented on, let me know what you think.
- Remove the `module_name` attribute from `ARGS_SPEC`s (was not used anywhere)
- Move the `userguide_html` attribute from the `ARGS_SPEC`s to a new `userguide` attribute of each `_UIMETA`
- Rename `_MODEL_UIS` to `MODEL_METADATA` and move to `src/natcap/invest/__init__.py`. 
- Get the model title and userguide path for each UI file from the `MODEL_METADATA` (fixing that annoying issue of some names being out of alphabetical order in the launcher)

IMO it makes sense to store this "metadata" as an attribute of the `natcap.invest` package and include the the UI and userguide info together there.

davemfish:
> @davemfish, you inspired me to put some more effort into this so I added a few things in addition to what you commented on, let me know what you think.
> 
> * Remove the `model_name`/`model_title` attribute from `ARGS_SPEC`s, and define it in the `_MODEL_UIS` instead.
> * Remove the `module` attribute from `ARGS_SPEC`s (was not used anywhere)
> * Move the `userguide_html` attribute from the `ARGS_SPEC`s to a new `userguide` attribute of each `_UIMETA`

Hi @emlys , some thoughts before the official review: It seems to make sense to remove these items from the `ARGS_SPEC` and keep them in the new metadata attribute. Though each of them is used by the Workbench so this move would require a Workbench update, probably including a new/different `ui_server` endpoint to get the full metadata attribute instead of just the names returned by `invest list`. If we want to do the `ui_server` update at the same time, we could target `release/3.10` for this PR. These changes might be the right thing to do, but if so I'd appreciate if we could do the Workbench PR in parallel, or at least immediately prioritize it after these changes get merged into `release/3.10`. 

> * Rename `_MODEL_UIS` to `MODEL_METADATA` and move to `src/natcap/invest/__init__.py`.
> * Get the model title and userguide path for each UI file from the `MODEL_METADATA` (fixing that annoying issue of some names being out of alphabetical order in the launcher)
> 
> IMO it makes sense to store this "metadata" as an attribute of the `natcap.invest` package and include the the UI and userguide info together there.

This makes sense to me. I'd appreciate @phargogh taking note of that change too (not the full PR, just the idea to move this metadata to `__init__.py`).

emlys:
> Though each of them is used by the Workbench so this move would require a Workbench update

Oh, I forgot the workbench used these, thanks for pointing that out. Maybe that's reason enough to leave them as is? I'm ambivalent about it.

davemfish:
> One thing I forgot to mention about moving to `__init__.py`: this would need to happen anyway for translations in #645. The reason is that the model titles need to be translated. We translate by importing/reloading the module in the new language. `cli.py` is an entrypoint, and so it can be the `__main__` module, which cannot/should not be reloaded. So `_MODEL_UIS` needs to move to a non-entrypoint module anyway, and `src/natcap/invest/__init__.py` seems as good as anywhere.

Cool, thanks for explaining that. Seems like a smart move!

emlys:
Also renamed Hydropower to Annual Water Yield everwhere that it hadn't already been renamed. Fixes #401

PR #697, opened 2021-10-22
Update codesigning certificate

### LINKED ISSUE #588: Sign Mac Binaries with Stanford Developer ID
I thought there might be an existing issue for this, but I can't seem to find it.  Feel free to close if it does exist!

Since we're not intending on distributing on the Mac App Store, UIT is willing to allow us to use the developer identity to sign and notarize the mac binary.

Assigning to 3.10 since we might be able to look at it on that timeline, but this could also turn out to be a larger issue.

emlys:
Might as well use our new Apple Developer certificate in the next release. Fixes #588 
- Use the new certificate file and password for the mac build. I considered using the Apple certificate for the Windows builds as well but ended up keeping the old cert because (1) the new password has a special character that was massively annoying to escape in the calling-powershell-from-bash command, and (2) the old cert may have ["reputation"](https://www.digicert.com/blog/ms-smartscreen-application-reputation) with Windows Defender that means users don't get a warning. A new cert would set us back to 0 reputation.

- Also removed the old unused target `signcode`, and reorganized some env variables.

Note that the code signing steps don't run in PRs, so check the github actions run on my branch instead: https://github.com/emlys/invest/tree/task/update-mac-certificate

After this is approved, but before it is merged, I will also need to update github secrets in this repo accordingly.

# Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -4,6 +4,10 @@ on:
     push:
     pull_request:
 
+defaults:
+  run:
+    shell: bash -l {0}
+
 jobs:
     check-syntax-errors:
         name: "Check for syntax errors"
@@ -56,7 +60,6 @@ jobs:
                   channels: conda-forge
 
             - name: Install python dependencies
-              shell: bash -l {0}
               run: |
                   conda install nomkl # make sure numpy is w/out MKL
                   conda upgrade -y pip setuptools
@@ -77,7 +80,6 @@ jobs:
                   ./ci/windows-ci-binary-install.ps1
 
             - name: Build userguide, binaries, installer
-              shell: bash -l {0}
               run: |
                   # This builds the users guide, binaries, and installer
                   make windows_installer
@@ -92,7 +94,6 @@ jobs:
             - name: Sign binaries
               # Secrets not available in PR so don't use GCP.
               if: github.event_name != 'pull_request'
-              shell: bash -l {0}
               env:
                 CERT_FILE: Stanford-natcap-code-signing-cert-expires-2024-01-26.p12
                 CERT_PASS: ${{ secrets.WINDOWS_CODESIGN_CERT_PASS }}
@@ -107,7 +108,6 @@ jobs:
             - name: Deploy artifacts to GCS
               # Secrets not available in PR so don't use GCP.
               if: github.event_name != 'pull_request'
-              shell: bash -l {0}
               run: |
                   # setup-gcloud adds 'gsutil' to PATH
                   make GSUTIL="gsutil" deploy
@@ -136,9 +136,6 @@ jobs:
         name: "Build mac binaries"
         needs: check-syntax-errors
         runs-on: macos-10.15
-        defaults:
-          run:
-            shell: bash -l {0}
         env:
             PYTHON_VERSION: 3.8
             PYTHON_ARCH: x64


emlys:
Just got the new Developer ID Application cert.

.github/workflows/binary-applications.yml
```diff
@@ -93,33 +93,22 @@ jobs:
...
+                CERT_PASS: ${{ secrets.WINDOWS_CODESIGN_CERT_PASS }}
               run: |
                 # figure out the path to signtool.exe (it keeps changing with SDK updates)
                 SIGNTOOL_PATH=$(find 'C:\\Program Files (x86)\\Windows Kits\\10' -type f -name 'signtool.exe*' | head -n 1)
                 INSTALLER_BINARY=$(find "$(pwd)/dist" -type f -name 'InVEST_*.exe' | head -n 1)
 
-                # Specify which python version we want to use (it's the one we're using
-                # in actions/setup-python.
-                export CLOUDSDK_PYTHON=$(which python)
-                export CLOUDSDK_GSUTIL_PYTHON=$(which python)
```
emlys, line 104: Pretty sure this has no effect - the default python should be the result of `which python`

.github/workflows/binary-applications.yml
```diff
@@ -210,9 +199,10 @@ jobs:
               if: github.event_name != 'pull_request'
               shell: bash -l {0}
               env:
-                CERT_KEY_PASS: ${{ secrets.STANFORD_CERT_KEY_PASS }}
+                CERT_FILE: 2025-01-16-Expiry-AppStore-App.p12
+                CERT_PASS: ${{ secrets.MACOS_CODESIGN_CERT_PASS }}
                 KEYCHAIN_PASS: ${{ secrets.KEYCHAIN_PASS }}
```
emlys, line 204: I'm not sure whether it's necessary to keep the keychain password a secret -  I'm leaning towards no. If someone had accessed the github actions VM and could use the keychain password, then they'd also have access to the certificate password. I don't think it adds any more security.

Makefile
```diff
@@ -354,45 +356,32 @@ $(MAC_BINARIES_ZIP_FILE): $(DIST_DIR) $(MAC_APPLICATION_BUNDLE) $(USERGUIDE_TARG
...
 	# this is essential to avoid the UI password prompt
-	security set-key-partition-list -S apple-tool:,apple: -s -k '$(KEYCHAIN_PASS)' '$(KEYCHAIN_NAME)'
-	# sign the dmg using certificate that's looked up by unique identifier 'Stanford Univeristy'
-	codesign --timestamp --verbose --sign 'Stanford University' $(MAC_DISK_IMAGE_FILE)
+	security set-key-partition-list -S apple-tool:,apple: -s -k $(KEYCHAIN_PASS) $(KEYCHAIN_NAME)
+	# sign the dmg using certificate that's looked up by unique identifier 'Stanford'
+	codesign --timestamp --verbose --sign Stanford $(MAC_DISK_IMAGE_FILE)
 	# relock the keychain (not sure if this is important?)
-	security lock-keychain '$(KEYCHAIN_NAME)'
+	security lock-keychain $(KEYCHAIN_NAME)
```
emlys, line 378: I don't think this is necessary either.

.github/workflows/binary-applications.yml
dcdenu4: Sorry I'm not very plugged in to this topic, what is the keychain password? Knowing pretty little about this my initial questions are:
- Is this something that would change often?
- Is there a downside from keeping it a secret?
- If someone cloned our repo without the keychain password being a secret to do what they wish with it, would that pose an issue?

Makefile
dcdenu4: Hey Emily, I was searching around trying to learn some things and came across this post: https://stackoverflow.com/questions/57838889/manage-keychain-to-codesign-macos-ios-app-with-xcodebuild-unattended.

Would it make sense to delete the keychain instead of locking it here?

```
# clean up temp keychain we created
security delete-keychain ${tempKeychain}
```

dcdenu4 (commented):
Hey @emlys , thanks for seeing this through! I have little to no experience with the MacOS codesigning procedures so thanks for looking at my comments.

.github/workflows/binary-applications.yml
emlys: On macos, signing certificates need to be accessed through a keychain, which is the password/secret management system. A keychain can contain passwords, signing identities, etc. And keychains themselves are password-protected.

There's a default keychain, but we can't access it because the password is the user login password, which we can't know on Github Actions. We have to create a new keychain with a known password so we can unlock it and import the signing certificate.

No other secrets will be stored in the new keychain, and the signing certificate is still protected by its own password. Someone would have to (1) get the certificate file and (2) get the corresponding password to do anything malicious with it.

emlys: * It would not change often, it's a totally arbitrary password that we make up just so we can create the keychain.
* The downside of keeping it secret is just that it complicates things slightly and we have to store it as a github secret in all forks.

Makefile
emlys: I'm not sure if that is needed either... I'm guessing they suggest that because they're running it locally on their own computer, in which case you wouldn't want stray keychains cluttering things up. On github actions, having that keychain stick around until the end of the workflow should have no effect.

.github/workflows/binary-applications.yml
dcdenu4: @emlys I think I'm convinced it doesn't seem necessary to keep as a GHA secret. Thanks for taking time to write up that rundown.

Makefile
dcdenu4: I see, thanks. So it might just be a choice between letting things get garbage collected or handling the procedure full circle ourselves. If there is seemingly a standard set of operations I guess I'd lean to explicitly doing them, even if it might be redundant or unnecessary. What do you think?

dcdenu4 (changes requested):
Thanks @emlys , I followed up on some of those comments. Let me know what you think!

Makefile
emlys: Since we're now not keeping the password a secret, I'd think there's no reason at all to lock the keychain afterward.

.github/workflows/binary-applications.yml
```diff
@@ -4,6 +4,10 @@ on:
     push:
     pull_request:
 
+defaults:
+  run:
+    shell: bash -l {0}
```
dcdenu4, line 9: Great idea!

dcdenu4 (approved):
Approving and merging because we know the action errors are not from this PR.

PR #710, opened 2021-11-02
Bugfix/709 logfile message logginglevel

### LINKED ISSUE #709: invest processes should always announce the file they log to
`utils.prepare_workspace` creates the user's model workspace and sets up of the logging FileHandler. It announces where it's logging to like this: `LOGGER.info('Writing log messages to %s', logfile)`

I propose to just move this to a level `100` message so that it cannot be filtered out by a user choosing a higher logging level than `INFO`. 

The use-case is the Workbench, where the parent process of invest needs to be told the logfile path. So that message needs to always come through stdout.

davemfish:
Change the log level on a key message that is emitted at the start of every invest job. Please see #709 for the details.

Fixes #709 

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -56,6 +56,8 @@ Unreleased Changes
       when looking for GDAL DLLs.  This fixes an issue where InVEST would not
       launch on computers where the ``%PATH%`` either contained other
       environment variables or was malformed.
+    * invest processes announce their logfile path at a very high logging level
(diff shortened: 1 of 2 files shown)


phargogh (approved):

phargogh:
I think that one windows build failure is unrelated to this PR and appears to be one of those random sorts of errors when installing a chocolatey package.  Merging anyways!

PR #713, opened 2021-11-03
switched to a sourceforge mirror that is responding. #712.

### LINKED ISSUE #712: BUILDS: cannot download zip from sourceforge
We've had trouble downloading one of our build dependencies for a couple days. Switching to a different sourceforge mirror seems to work.

davemfish:
A sourceforge mirror that responds!
Fixes #712 

# Checklist


dcdenu4 (approved):
Seems good to me. I wonder if we should revisit this set up at some point. I think it'd be good to know what we can / can't expect `choco` to install vs what we need to manually grab (like zip).

Issue #722, opened 2021-11-12
Implement Urban Nature Access model


phargogh:
This is a catch-all issue for implementing the Urban Nature Access model.

See https://github.com/natcap/softwareteam/issues/93 for details.

- [x] Add `ARGS_SPEC`
- [x] Implement validation
- [x] Add Qt interface based on `ARGS_SPEC`
- [x] Add sample data to the repository
- [x] Note model addition in `HISTORY.rst` 
- [x] Add assertions to the whole-model test
- [x] Add a note to the User's Guide about when fractional populations may be a good idea and when they may not, but also that the model will not do any rounding on population inputs.
- [x]  _more individual tasks coming soon_

phargogh:
Related to #543 , let's be sure to document the origins of the sample data in a README in the sample data repo.

phargogh:
Well, it's been a while since I've been able to tackle this, but we're back!  I took a stab at the first of the optional modules, split greenspace.  Running the model with the same parameters looks like I haven't broken anything, but running the model using the split greenspace options looks like it may need a little more work.

## Next actionable steps
- [ ] Write a software test for this (and in particular, take a look at whether normalization is needed in convolutions)

phargogh:
- [ ] In the user's guide, emphasize that split populations assume that population groups are spread evenly across the admin units.  See discussion in https://github.com/natcap/invest/pull/1024#discussion_r922494702

phargogh:
We have a PR over on bitbucket (https://bitbucket.org/natcap/invest-sample-data/pull-requests/29) that includes the sample data.  Once that's approved and merged, we need to be sure to include it in the updated release.

phargogh:
UNA has been merged.  Although other issues exist, they are separate from the initial model release.

#1175 #1176 #1180

PR #724, opened 2021-11-15
Urban Nature Access: boilerplate and population resampling

phargogh:
This PR is the first of several relating to #722 .  This PR specifically starts out by:

* Adding an `ARGS_SPEC` (mostly there, not 100% complete yet)
* Implements and tests a function to resample a population raster without gaining or losing significant numbers of people (this is the only serious computation in the model right now)
* Adds the start to a test for the model.

I haven't yet made any notes in the README and there isn't a UG chapter yet, so both of those checklist items are pending a later PR.

I've already created a new target branch (`feature/urban-nature-access`, based on `release/3.10`) that this PR is set to merge into.  If we end up waiting to merge this into `release/3.11` after the 3.10 release, that'll work too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ requires = ['setuptools', 'wheel', 'setuptools_scm', 'numpy', 'cython']
 # Each section of imports will be sorted alphabetically
 force_alphabetical_sort_within_sections = true
 
-# "from osgeo import gdal" and "from osgeo impor osr" are on separate lines
+# "from osgeo import gdal" and "from osgeo import osr" are on separate lines
 force_single_line = true
 
 # This defines the sections used, and clarifies that we don't want to have
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,3 +6,23 @@
 # that we can provide a much easier build experience so long as GDAL is
 # available at runtime.
 requires = ['setuptools', 'wheel', 'setuptools_scm', 'numpy', 'cython']
+
+[tool.isort]
+# isort can be installed from conda-forge
+# flake8-isort can also be used for flake8 linting integration
+#
+# To test isort configuration, run:
+#     isort . --show-config
+# A warning will present if config can't be loaded.
+#
+# Each section of imports will be sorted alphabetically
+force_alphabetical_sort_within_sections = true
+
+# "from osgeo import gdal" and "from osgeo impor osr" are on separate lines
+force_single_line = true
+
+# This defines the sections used, and clarifies that we don't want to have
+# lines between the FIRSTPARTY (absolute natcap.invest imports) and the
+# LOCALFOLDER (relative imports, e.g. "from . import validation) section.
+sections = 'FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER'
+no_lines_before = 'LOCALFOLDER'
--- a/src/natcap/invest/__init__.py
+++ b/src/natcap/invest/__init__.py
@@ -197,6 +197,12 @@ MODEL_METADATA = {
         gui='urban_cooling_model.UrbanCoolingModel',
         userguide='urban_cooling_model.html',
         aliases=('ucm',)),
+    'urban_nature_access': _MODELMETA(
+        model_title='Urban Nature Access',
+        pyname='natcap.invest.urban_nature_access',
+        gui='',  # TODO
+        userguide='',  # TODO,
+        aliases=('una',)),
 }
 
 
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -39,8 +39,7 @@ ARGS_SPEC = {
             'about': (
                 "A map of LULC codes. "
                 "All values in this raster must have corresponding entries "
-                "in the LULC attribute table.  This raster must be linearly "
-                "projected in meters."),
+                "in the LULC attribute table."),
         },
         'lulc_attribute_table': {
             'name': 'LULC attribute table',
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -74,7 +74,7 @@ ARGS_SPEC = {
             'projection_units': u.meter,
             'about': (
                 "A raster representing the number of people who live in each "
-                "pixel. This raster must be linearly projected in meters."
+                "pixel."
             ),
         },
         'admin_unit_vector_path': {
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -39,7 +39,8 @@ ARGS_SPEC = {
             'about': (
                 "A map of LULC codes. "
                 "All values in this raster must have corresponding entries "
-                "in the LULC attribute table."),
+                "in the LULC attribute table.  This raster must be linearly "
+                "projected in meters."),
         },
         'lulc_attribute_table': {
             'name': 'LULC attribute table',
@@ -71,9 +72,10 @@ ARGS_SPEC = {
                 1: {'type': 'number', 'units': u.none}
             },
             'projected': True,
+            'projection_units': u.meter,
             'about': (
                 "A raster representing the number of people who live in each "
-                "pixel. This raster must be linearly projected."
+                "pixel. This raster must be linearly projected in meters."
             ),
         },
         'admin_unit_vector_path': {
@@ -207,6 +209,7 @@ def _resample_population_raster(
     Args:
         source_population_raster_path (string): The source population raster.
             Pixel values represent the number of people occupying the pixel.
+            Must be linearly projected in meters.
         target_population_raster_path (string): The path to where the target,
             warped population raster will live on disk.
         target_pixel_size (tuple): A tuple of the pixel size for the target
@@ -215,7 +218,8 @@ def _resample_population_raster(
             Passed directly to ``pygeoprocessing.warp_raster``.
         target_projection_wkt (string): The Well-Known Text of the target
             spatial reference fro the target raster.  Passed directly to
-            ``pygeoprocessing.warp_raster``.
+            ``pygeoprocessing.warp_raster``.  Assumed to be a linear projection
+            in meters.
         working_dir (string): The path to a directory on disk.  A new directory
             is created within this directory for the storage of temporary files
             and then deleted upon successful completion of the function.
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -23,7 +23,9 @@ ARGS_SPEC = {
     'pyname': MODEL_METADATA['urban_nature_access'].pyname,
     'userguide_html': MODEL_METADATA['urban_nature_access'].userguide,
     'args_with_spatial_overlap': {
-        'spatial_keys': [],
+        'spatial_keys': [
+            'lulc_raster_path', 'population_raster_path',
+            'admin_unit_vector_path'],
         'different_projections_ok': True,
     },
     'args': {
@@ -312,4 +314,5 @@ def _resample_population_raster(
 
 
 def validate(args, limit_to=None):
-    return validation.validate(args, ARGS_SPEC['args'])
+    return validation.validate(
+        args, ARGS_SPEC['args'], ARGS_SPEC['args_with_spatial_overlap'])
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -3,6 +3,9 @@ import logging
 
 import pygeoprocessing
 import taskgraph
+from osgeo import gdal
+from osgeo import osr
+import numpy
 
 from . import validation
 from . import spec_utils
@@ -11,6 +14,8 @@ from .spec_utils import u
 from .. import MODEL_METADATA
 
 LOGGER = logging.getLogger(__name__)
+UINT32_NODATA = numpy.finfo(numpy.uint32).max
+FLOAT32_NODATA = numpy.finfo(numpy.float32).max
 ARGS_SPEC = {
     'model_name': MODEL_METADATA['urban_nature_access'].model_title,
     'pyname': MODEL_METADATA['urban_nature_access'].pyname,
@@ -42,7 +47,10 @@ ARGS_SPEC = {
         'population_raster_path': {
             'type': 'raster',
             'name': 'population raster',
-            'bands': {1: {'type': 'float'}},
+            'bands': {
+                1: {'type': 'number', 'units': u.none}
+            },
+            'projected': True,
             'about': "",  # TODO,
         },
         'admin_unit_vector_path': {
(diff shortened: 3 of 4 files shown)


src/natcap/invest/__init__.py
```diff
@@ -197,6 +197,12 @@ class _MODELMETA:
         gui='urban_cooling_model.UrbanCoolingModel',
         userguide='urban_cooling_model.html',
         aliases=('ucm',)),
+    'urban_nature_access': _MODELMETA(
+        model_title='Urban Nature Access',
+        pyname='natcap.invest.urban_nature_access',
+        gui='',  # TODO
+        userguide='',  # TODO,
+        aliases=('una',)),
```
dcdenu4, line 205: una!

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
+import os
+import logging
+import shutil
+import tempfile
+
+import pygeoprocessing
+import taskgraph
+from osgeo import gdal
+from osgeo import osr
+import numpy
```
dcdenu4, line 10: I think Dave has convinced me that alphabetically ordering the imports within their sections is a nice thing to do. I thought it might be a little early for such a small suggestion, but the more of these small things that get worked out the less there will be later!

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+
+LOGGER = logging.getLogger(__name__)
+UINT32_NODATA = int(numpy.iinfo(numpy.uint32).max)
+FLOAT32_NODATA = float(numpy.finfo(numpy.float32).max)
+ARGS_SPEC = {
+    'model_name': MODEL_METADATA['urban_nature_access'].model_title,
+    'pyname': MODEL_METADATA['urban_nature_access'].pyname,
+    'userguide_html': MODEL_METADATA['urban_nature_access'].userguide,
+    'args_with_spatial_overlap': {
+        'spatial_keys': [],
```
dcdenu4, line 26: Do we know yet which inputs will need to overlap in space?

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+        'population_raster_path': {
+            'type': 'raster',
+            'name': 'population raster',
+            'bands': {
+                1: {'type': 'number', 'units': u.none}
+            },
+            'projected': True,
+            'about': (
+                "A raster representing the number of people who live in each "
+                "pixel. This raster must be linearly projected."
```
dcdenu4, line 74: Not only linearly projected, but meters specifically I think.

dcdenu4:
```suggestion
                "pixel. This raster must be linearly projected in meters."
```

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+            GDAL-compatible land-use/land-cover raster containing integer
+            landcover codes.
+        args['lulc_attribute_table'] (string): (required) A string path to a
+            CSV with the following columns:
+
+            * ``lucode``: the integer landcover code represented.
+            * ``greenspace``: ``0`` or ``1`` indicating whether this landcover
+              code is (``1``) or is not (``0``) a greenspace pixel.
+
+        args['population_raster_path'] (string): (required) A string path to a
```
dcdenu4, line 129: Should the docstring here also reflect that this raster should be projected in meters?

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+    tmp_working_dir = tempfile.mkdtemp(dir=working_dir)
+    population_raster_info = pygeoprocessing.get_raster_info(
+        source_population_raster_path)
+    pixel_area = numpy.multiply(*population_raster_info['pixel_size'])
+    population_nodata = population_raster_info['nodata'][0]
+
+    population_srs = osr.SpatialReference()
+    population_srs.ImportFromWkt(population_raster_info['projection_wkt'])
+
+    # Convert population pixel area to square km
```
dcdenu4, line 235: Because you call out `square km` here, I suggested above that the population raster should be projected in meters.

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+        return out_array
+
+    # Step 1: convert the population raster to population density per sq. km
+    density_raster_path = os.path.join(tmp_working_dir, 'pop_density.tif')
+    pygeoprocessing.raster_calculator(
+        [(source_population_raster_path, 1)],
+        _convert_population_to_density,
+        density_raster_path, gdal.GDT_Float32, FLOAT32_NODATA)
+
+    # Step 2: align to the LULC
```
dcdenu4, line 265: I like the direct note that this is really aligning to the LULC. Should the docstring and/or variable names reflect this intent as well? I guess, is there a reason to keep the variable names ambiguous and something like `lulc_target_projection_wkt`? Curious what you think.

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+        """Convert a population density raster back to population counts.
+
+        Args:
+            density (numpy.array): An array of the population density per
+                square kilometer.
+
+        Returns:
+            A ``numpy.array`` of the population counts given the target pixel
+            size of the output raster."""
+        # We're using a float32 array here because doing these unit
```
dcdenu4, line 293: Great note

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,315 @@
...
+        _resample_population_raster,
+        kwargs={
+            'source_population_raster_path': args['population_raster_path'],
+            'target_population_raster_path': file_registry[
+                'resampled_population'],
+            'target_pixel_size': lulc_raster_info['pixel_size'],
+            'target_bb': lulc_raster_info['bounding_box'],
+            'target_projection_wkt': lulc_raster_info['projection_wkt'],
+            'working_dir': intermediate_dir,
+        },
```
dcdenu4, line 179: I like the named parameter approach. Did you choose this over regular arguments for any particular reason?

tests/test_urban_nature_access.py
```diff
@@ -0,0 +1,134 @@
...
+import unittest
+import tempfile
+import shutil
+import os
+import random
+
+import pygeoprocessing
+import numpy
+from osgeo import gdal
+from osgeo import osr
```
dcdenu4, line 12: Order alphabetically?

tests/test_urban_nature_access.py
```diff
@@ -0,0 +1,134 @@
...
+
+_DEFAULT_ORIGIN = (444720, 3751320)
+_DEFAULT_PIXEL_SIZE = (30, -30)
+_DEFAULT_EPSG = 3116
+
+
+class UNATests(unittest.TestCase):
```
dcdenu4, line 19: una!

tests/test_urban_nature_access.py
```diff
@@ -0,0 +1,134 @@
...
+    """Tests for the Urban Nature Access Model."""
+
+    def setUp(self):
+        """Override setUp function to create temp workspace directory."""
+        # this lets us delete the workspace after its done no matter the
+        # the rest result
```
dcdenu4, line 25:
```suggestion
        # the test result
```

tests/test_urban_nature_access.py
```diff
@@ -0,0 +1,134 @@
...
+
+        args = {
+            'workspace_dir': os.path.join(self.workspace_dir, 'workspace'),
+            'results_suffix': 'suffix',
+            'population_raster_path': os.path.join(
+                self.workspace_dir, 'population.tif'),
+            'lulc_raster_path': os.path.join(self.workspace_dir, 'lulc.tif'),
+        }
+
+        random.seed(-1)  # for our random number generation
```
dcdenu4, line 101: I'm not sure how we'll do the asserting yet, but I was wondering if we could rely on `random.seed` across different types of hardware / OS's?

dcdenu4 (changes requested):
What a great PR size! I just had a couple questions and suggestions.

I did begin to wonder how nit picky to be early on here, but it did concern me that if I did not comment on changes now, I might forget to mention them later because they won't show up as changes in future PR's.

src/natcap/invest/urban_nature_access.py
phargogh: Patched in 442e84e9 using [`isort`](https://pycqa.github.io/isort/index.html) and some configuration for it in `pyproject.toml`.

src/natcap/invest/urban_nature_access.py
phargogh: Pretty sure we do!  The first stab at it given what we know about things right now is in a1cea270

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, I kind of went back and forth on this idea, but after digging into the `GetLinearUnits` source code I think you're right.  I really don't like that `GetLinearUnits` will silently fall back to "oh well, let's just assume meters" and that will for sure cause us more headaches down the line than just requiring a linear projection in meters for the LULC and the population raster.  So that's been added in 32834c43

src/natcap/invest/urban_nature_access.py
phargogh: Added for both of the raster keys in 9ae62145

src/natcap/invest/urban_nature_access.py
phargogh: The main reason I chose to do this was because I'm a bit rusty!  But in general, adding the keyword argument names here just felt more readable to me while I'm getting back into the swing of things.

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, that makes sense.  I just wish we could rely on the `GetLinearUnits` to work as expected.

src/natcap/invest/urban_nature_access.py
phargogh: The reason I had used the `target_` prefix in most of the parameters was because they were being passed directly to `pygeoprocessing.warp_raster` as parameters of the same name.  But you're right, I do think this function should reflect how it's actually being used, and so I've updated the parameter inputs to reflect that they're coming from the LULC.  If and when it becomes useful to abstract this away into a separate function, we can make appropriately general parameter names at that time.

I do think it makes sense to require the particular information needed to be passed in rather than just passing in the LULC.  We're requesting only what's needed to do the operation, which feels right to me.

Patched in 6fa201f4

tests/test_urban_nature_access.py
phargogh: Implemented in one of my recent commits via `isort`!

tests/test_urban_nature_access.py
phargogh: Ha, I do like typing that out.

tests/test_urban_nature_access.py
phargogh: ae68329b

tests/test_urban_nature_access.py
phargogh: Ooo that's a fun question.  I didn't know the answer to it and didn't find anything while googling so I decided to test it out with everything Github Actions allows us to do while I would have loved to have tested every version of python that GHA provides, that wasn't practical (they only allow up to 256 tasks) so I had to settle for a subset, eventually using the latest bugfix (or alpha) releases for each major python version.  [Here are the results](https://github.com/phargogh/experiment-merseinne-twister-reliability/actions/runs/1469434143) and here are my takeaways:

1. There _was_ a change between python 3.1.4 and python 3.2.5 that caused the Merseinne Twister, python's default RNG, to return different results when given the same seed.
2. Since at least python 3.2.5, the python `random` module has returned consistent results
3. The Merseinne Twister (or at least python's implementation of it) appears to not behavior differently in a 32- or 64-bit build of python, even across operating systems.
4. More testing is needed to determine if different processor architectures (such as ARM) produce different results.

So all of that being said, I think this _particular_ use of `random` might be OK because the values themselves don't matter ... since we're asserting the sum of the source array against the sum of the warped array, all we need to know is that the pixel values are numerically close.  And even if we do have a situation where the results differ between runs, the seed is defined, so the only variables there are documented characteristics (python version, package versions, runner OS, etc.) of the GHA environment.

So I _think_ this might actually be OK!  But what do you think about that?

phargogh:
Thanks @dcdenu4 !  I've responded to your comments and have re-requested your review.

tests/test_urban_nature_access.py
dcdenu4: Hey @phargogh , thanks for doing that analysis! That is super fun to scroll through. 

> since we're asserting the sum of the source array against the sum of the warped array, all we need to know is that the pixel values are numerically close.

Of course! I'm not quite sure why I thought we'd be using the random seeded inputs to make direct numerical comparisons against outputs.

>So I think this might actually be OK! But what do you think about that?

I think you're right and that this is fine. Thanks for looking into that though!

pyproject.toml
```diff
@@ -6,3 +6,23 @@
...
+# flake8-isort can also be used for flake8 linting integration
+#
+# To test isort configuration, run:
+#     isort . --show-config
+# A warning will present if config can't be loaded.
+#
+# Each section of imports will be sorted alphabetically
+force_alphabetical_sort_within_sections = true
+
+# "from osgeo import gdal" and "from osgeo impor osr" are on separate lines
```
dcdenu4, line 21:
```suggestion
# "from osgeo import gdal" and "from osgeo import osr" are on separate lines
```

dcdenu4 (changes requested):
Hey @phargogh I just had one minor typo suggestion

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,322 @@
...
+        'n_workers': spec_utils.N_WORKERS,
+        'lulc_raster_path': {
+            **spec_utils.LULC,
+            'projected': True,
+            'projection_units': u.meter,
+            'about': (
+                "A map of LULC codes. "
+                "All values in this raster must have corresponding entries "
+                "in the LULC attribute table.  This raster must be linearly "
+                "projected in meters."),
```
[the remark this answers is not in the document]
dcdenu4: Sorry about the run around here, but I just remembered that with Emily's work with `pint` we were dropping these redundant descriptions since they should be represented from the `projected` and `projected_units` attributes.

src/natcap/invest/urban_nature_access.py
```diff
@@ -0,0 +1,322 @@
...
+            'type': 'raster',
+            'name': 'population raster',
+            'bands': {
+                1: {'type': 'number', 'units': u.none}
+            },
+            'projected': True,
+            'projection_units': u.meter,
+            'about': (
+                "A raster representing the number of people who live in each "
+                "pixel. This raster must be linearly projected in meters."
```
dcdenu4, line 78:
```suggestion
                "pixel."
```

dcdenu4 (approved):
@phargogh looks good, thanks!

PR #728, opened 2021-11-18
Update HQ rarity index calculation

### LINKED ISSUE #720: Habitat Quality Rarity output has confusing negative values
@newtpatrol pointed out that when a current / future LULC has more of a habitat than the baseline lulc, negative numbers are reported. Rarity is currently defined in source as:
```python
numerator = lulc_code_count_x[code] * lulc_area
denominator = lulc_code_count_b[code] * base_area
ratio = 1.0 - (numerator / denominator)
code_index[code] = ratio
```
But I wonder if denominator should really be:

`denominator = (lulc_code_count_b[code] * base_area) + numerator`

This would give values between 0 and 1, where a high numerator (not rare) would give a high ratio, and thus low rarity value. A low numerator would give a low ratio, and thus high rarity value. A current / future pixel area which is the same as the baseline would give a rarity of 0.5. So above 0.5, more rare (less abundant), less than 0.5 less rare (more abundant).

OR, we could simply cap the ratio at 0.0, so we don't see negative numbers. Then, everything above 0.0 would be more rare than the baseline and 0.0 would indicate equal abundance or more abundance.

Either way, it seems desirable that this should be a 0-1 output and better defined in the Users Guide.

- [x] Update Users Guide

dcdenu4:
# Description
This PR changes the output of habitat rarity by constraining the values to a 0-1 index. One thing to note is that 0.5 indicates no abundance change / no rarity change. Values less than 0.5 indicate more abundance and "less" rare. Values greater than 0.5 indicate less abundance and higher rarity. See issue #720 for a complete description.

Fixes #720

Users Guide update is in a PR https://github.com/natcap/invest.users-guide/pull/51 

# Checklist

### DIFF
--- a/tests/test_habitat_quality.py
+++ b/tests/test_habitat_quality.py
@@ -176,15 +176,10 @@ class HabitatQualityTests(unittest.TestCase):
         # this lets us delete the workspace after its done no matter the
         # the rest result
         self.workspace_dir = tempfile.mkdtemp()
-        #dir_path = os.path.join('C:', os.sep, 'Users', 'ddenu', 'Documents',
-        #        'hq-rarity-test-changes')
-        #os.mkdir(dir_path)
-        #self.workspace_dir = dir_path
 
     def tearDown(self):
         """Override tearDown function to remove temporary directory."""
         shutil.rmtree(self.workspace_dir)
-        #pass
 
     def test_habitat_quality_regression(self):
         """Habitat Quality: base regression test with simplified data."""
@@ -238,8 +233,8 @@ class HabitatQualityTests(unittest.TestCase):
                 'deg_sum_f_regression.tif': 33.931896,
                 'quality_c_regression.tif': 7499.983,
                 'quality_f_regression.tif': 4999.9893,
-                'rarity_c_regression.tif': 3333.33335,
-                'rarity_f_regression.tif': 3333.33335}.items():
+                'rarity_c_regression.tif': 3333.3335,
+                'rarity_f_regression.tif': 3333.3335}.items():
             raster_path = os.path.join(args['workspace_dir'], output_filename)
             # Check that the raster's computed values are what we expect.
             # In this case, the LULC and threat rasters should have been
--- a/tests/test_habitat_quality.py
+++ b/tests/test_habitat_quality.py
@@ -176,10 +176,15 @@ class HabitatQualityTests(unittest.TestCase):
         # this lets us delete the workspace after its done no matter the
         # the rest result
         self.workspace_dir = tempfile.mkdtemp()
+        #dir_path = os.path.join('C:', os.sep, 'Users', 'ddenu', 'Documents',
+        #        'hq-rarity-test-changes')
+        #os.mkdir(dir_path)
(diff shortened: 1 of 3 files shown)


tests/test_habitat_quality.py
```diff
@@ -176,10 +176,15 @@ def setUp(self):
         # this lets us delete the workspace after its done no matter the
         # the rest result
         self.workspace_dir = tempfile.mkdtemp()
+        #dir_path = os.path.join('C:', os.sep, 'Users', 'ddenu', 'Documents',
+        #        'hq-rarity-test-changes')
+        #os.mkdir(dir_path)
+        #self.workspace_dir = dir_path
```
davemfish, line 182: Some leftover code here

tests/test_habitat_quality.py
```diff
@@ -176,10 +176,15 @@ def setUp(self):
...
 
     def tearDown(self):
         """Override tearDown function to remove temporary directory."""
         shutil.rmtree(self.workspace_dir)
+        #pass
```
davemfish, line 187: leftovers

tests/test_habitat_quality.py
```diff
@@ -233,8 +238,8 @@ def test_habitat_quality_regression(self):
                 'deg_sum_f_regression.tif': 33.931896,
                 'quality_c_regression.tif': 7499.983,
                 'quality_f_regression.tif': 4999.9893,
-                'rarity_c_regression.tif': 2500.0000000,
-                'rarity_f_regression.tif': 2500.0000000}.items():
+                'rarity_c_regression.tif': 3333.33335,
+                'rarity_f_regression.tif': 3333.33335}.items():
```
davemfish, line 242: I guess it doesn't matter since tests are passing. But I see there's 5 decimals here and 4 on all the subsequent tests where this number appears.

davemfish (changes requested):
Makes sense to me! There's a couple minor things to cleanup. And I think maybe the RTD build is failing on rst syntax in HISTORY?

tests/test_habitat_quality.py
dcdenu4: Oops! Removed in https://github.com/natcap/invest/pull/728/commits/cf5921726822d70ff12f0b95e9d6c27e7d89c821

tests/test_habitat_quality.py
dcdenu4: Removed in https://github.com/natcap/invest/pull/728/commits/cf5921726822d70ff12f0b95e9d6c27e7d89c821

tests/test_habitat_quality.py
dcdenu4: This definitely looks like a classic Vim paste mistake. I went for consistency and removed a decimal place. https://github.com/natcap/invest/pull/728/commits/cf5921726822d70ff12f0b95e9d6c27e7d89c821

dcdenu4:
Thanks @davemfish , I think I addressed everything and the readthedocs seems to be working after adjusting an indent.

davemfish (approved):

PR #730, opened 2021-11-20
Urban Nature Access - Kernels and Convolutions

phargogh:
This PR is centered around kernels and convolutions, and not strictly limited to the Urban Nature Access model.

Changes to UNA in this PR include:

* Added 2 new kernels:
  * "Instantaneous"/"dichotomous" kernel - what we've called a 'disk' in other models.  It's a boolean kernel within a circular search radius.
  * "density" kernel - looks like a sphere if you were to look at it head-on.
* Added a new bit of functionality from the model, the "2-Step Floating Catchment Area" calculation, implemented as 2 convolutions with non-normalized kernels and a raster_calculator call.
* The LULC is now warped to ensure that we're using square pixels (since we're convolving based on a search radius, square pixels are needed)

Changes outside of UNA include:

* ``utils`` now has a gaussian decay kernel function.
* I've refactored GLOBIO and the Scenario Generator:Proximity to use the new gaussian decay kernel function.
* @esoth I updated a number of membership  assertions in the `ARGS_SPEC` test to make it much easier to debug when a key is missing.  Previously `self.assertTrue(a in b)` would just raise an exception saying `"False is not True"`.  Now, using `self.assertIn`, the exception will say something like `"10 could not be found in {'foo': 1}"` where `a=10` and `b={'foo: 1}`, making debugging errors a whole lot easier.

### DIFF
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -106,8 +106,6 @@ ARGS_SPEC = {
             'required': False,
             'options': [
                 'dichotomy',
-                # "exponential" is more consistent with other InVEST models'
-                # terminology.  "Power function" is used in the design doc.
                 'exponential',
                 'gaussian',
                 'density',
@@ -158,6 +156,9 @@ def execute(args):
             number indicating the required greenspace, in m² per capita.
         args['search_radius'] (number): (required) A positive, nonzero number
             indicating the maximum distance that people travel for recreation.
+        args['kernel_type'] (string): (optional) The selected kernel type.
+            Must be one of the keys in ``KERNEL_TYPES``.  If not provided, the
+            ``'dichotomy'`` kernel will be used.
 
     Returns:
         ``None``
@@ -183,6 +184,15 @@ def execute(args):
         # TypeError when n_workers is None.
         n_workers = -1  # Synchronous execution
     graph = taskgraph.TaskGraph(work_token_dir, n_workers)
+
+    if 'kernel_type' not in args:
+        kernel_type = 'dichotomy'
+    elif args['kernel_type'] not in kernel_types:
+        raise ValueError(
+            f'Kernel type "{kernel_type}" is not recognized. '
+            f"Must be one of {', '.join(kernel_types.keys())}")
+    else:
+        kernel_type = args['kernel_type']
 
     # Align the population raster to the LULC.
     lulc_raster_info = pygeoprocessing.get_raster_info(
@@ -251,7 +261,7 @@ def execute(args):
             'target_greenspace_supply_raster_path': (
                 file_registry['greenspace_supply']),
             'search_radius_m': float(args['search_radius']),
-            'kernel_type': str(args['decay_function']),
+            'kernel_type': kernel_type,
             'working_dir': intermediate_dir,
         },
         target_path_list=[''],
@@ -570,25 +580,6 @@ def _two_step_floating_catchment_area(
         target_greenspace_supply_raster_path, search_radius_m, kernel_type,
         working_dir):
     LOGGER.info("Starting 2-Step Floating Catchment Area")
-    # greenspace_raster - float32, greenspace area per pixel
-
-    # sum within the search distance the ratio of:
-    #  * numerator: convolved greenspace (kernel=selected by user)
-    #  * denominator: convolved population (kernel=selected by user)
-
-    # All kernel creation types have the same function signature
-    kernel_types = {
-        'dichotomy': instantaneous_decay_kernel_raster,
-        'exponential': utils.exponential_decay_kernel_raster,
-        'gaussian': utils.gaussian_decay_kernel_raster,
-        'density': density_decay_kernel_raster,
-    }
-    try:
-        kernel_function = kernel_types[kernel_type]
-    except KeyError:
-        raise ValueError(
-            f'Kernel type "{kernel_type}" is not recognized. '
-            f"Must be one of {', '.join(kernel_types.keys())}")
 
     tmp_working_dir = tempfile.mkdtemp(
         dir=working_dir, prefix='UNA-2SFCA')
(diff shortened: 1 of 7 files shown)


src/natcap/invest/urban_nature_access.py
```diff
@@ -165,29 +189,289 @@ def execute(args):
...
+        convolved_population (numpy.array): A numpy array where each pixel
+            represents the total number of people within a search radius of
+            each pixel, perhaps adjusted by a search kernel.
+        greenspace_nodata (int or float): The nodata value of the
+            ``greenspace_area`` matrix.
+        population_nodata (int or float): The nodata value of the
+            ``convolved_population`` matrix.
+
+    Returns:
+        A numpy array with the ratio ``R_j`` representing thew
```
dcdenu4, line 363:
```suggestion
        A numpy array with the ratio ``R_j`` representing the
```

src/natcap/invest/urban_nature_access.py
```diff
@@ -165,29 +189,289 @@ def execute(args):
...
+            * Any non-greenspace pixels will have a value of ``0.0`` in the
+              output matrix.
+    """
+    # ASSUMPTION: population nodata value is not close to 0.
+    #  Shouldn't be if we're coming from convolution.
+    out_array = numpy.full(
+        greenspace_area.shape, FLOAT32_NODATA, dtype=numpy.float32)
+
+    # Avoid divide-by-zero and very small negative values
+    valid_pixels = (convolved_population > 0)
```
dcdenu4, line 378: Is this possible given `_convolve_and_set_lower_bounds_for_population`?

src/natcap/invest/urban_nature_access.py
```diff
@@ -316,6 +600,186 @@ def _convert_density_to_population(density):
     shutil.rmtree(tmp_working_dir, ignore_errors=True)
 
 
+def instantaneous_decay_kernel_raster(expected_distance, kernel_filepath,
+        normalize=False):
+    """Create a raster-based, discontinuous decay kernel.
```
dcdenu4, line 605: Is this the dichotomy decay kernel? If so, would it be better to name it as such over 'instantaneous'?

dcdenu4 (changes requested):
Hey @phargogh this looks really good to me. I just had some very light comments / suggestions.

src/natcap/invest/utils.py
```diff
@@ -407,19 +409,107 @@ def exponential_decay_kernel_raster(expected_distance, kernel_filepath):
...
+    will be square in shape, with a width of ``(sigma * n_std_dev * 2) + 1``
+    pixels.
+
+    Args:
+        sigma (int or float): The distance (in pixels) of the standard
+            deviation from the center of the raster.
+        kernel_filepath (string): The path to the file on disk where this
+            kernel should be stored. If a file exists at this path, it will be
+            overwritten.
+        sigma_multiplier (int or float): The number of times sigma should be
```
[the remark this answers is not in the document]
phargogh: Yep!  Patched in 554229f3

tests/test_utils.py
```diff
@@ -211,6 +212,61 @@ def test_exp_decay_kernel_raster(self):
...
+            https://stackoverflow.com/a/43346070/299084
+            """
+            ax = numpy.linspace(
+                -(kernel_dimension - 1) / 2.,
+                (kernel_dimension - 1) / 2.,
+                kernel_dimension)
+            gauss = numpy.exp(-0.5 * numpy.square(ax) / numpy.square(sigma))
+            kernel = numpy.outer(gauss, gauss)
+
+            dist_from_center = numpy.hypot(
```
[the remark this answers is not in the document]
phargogh: A very useful one!

src/natcap/invest/urban_nature_access.py
phargogh: It is possible!  Population can be 0 in a pixel if no one lives there and it happens to be a long ways from a population center (so outside of the convolution kernel or search radius)

But you're also right that we shouldn't be seeing negative values any more, so I've updated that inline comment in e32404c6

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, I was thinking about that too.  I kind of prefer "instantaneous" decay (or "discontinuous", I'd be happy with that too), but the design doc uses dichotomous, so this probably should too.  I've updated this in 5ae39a8f

src/natcap/invest/utils.py
```diff
@@ -407,19 +409,107 @@ def exponential_decay_kernel_raster(expected_distance, kernel_filepath):
...
+    col_index = numpy.array(range(kernel_size))
+    running_sum = 0.0
+    for row_index in range(kernel_size):
+        distance_kernel_row = numpy.sqrt(
+            (row_index - max_distance) ** 2 +
+            (col_index - max_distance) ** 2).reshape(1, kernel_size)
+        kernel = numpy.where(
+            distance_kernel_row > max_distance, 0.0,
+            (1 / (2.0 * numpy.pi * sigma ** 2) *
+             numpy.exp(-distance_kernel_row**2 / (2 * sigma ** 2))))
```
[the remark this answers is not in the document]
phargogh: Ah yes, you're right.  Taking the square of a square root cancels that out, so we're left with `x**2 + y**2`.

src/natcap/invest/urban_nature_access.py
dcdenu4: For whatever reason I didn't think of the 0 case.

src/natcap/invest/urban_nature_access.py
dcdenu4: Yea, happy to go with which ever you think. I was just trying to match up the kernel functions to the kernel key, name dictionary.

dcdenu4:
Hey @phargogh , I think a few commits you mentioned haven't been pushed yet, is that possible? I'm not seeing all the changes you mention.

src/natcap/invest/urban_nature_access.py
```diff
@@ -165,29 +189,289 @@ def execute(args):
...
+            f"Must be one of {', '.join(kernel_types.keys())}")
+    else:
+        kernel_type = args['kernel_type']
+
     # Align the population raster to the LULC.
     lulc_raster_info = pygeoprocessing.get_raster_info(
         args['lulc_raster_path'])
 
+    squared_lulc_pixel_size = _square_off_pixels(args['lulc_raster_path'])
+    lulc_alignment_task = graph.add_task(
```
dcdenu4, line 219: I was just curious if we'll ever want to make sure that the LULC and population raster align perfectly geospatially by using `align_and_resize_rasters`?

dcdenu4 (changes requested):
Hey @phargogh , I added another rogue comment / question along with not seeing some of those commits you mentioned.

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, this has been on my mind as well.  In the interest of closing out this PR, might a `TODO` suffice?  In case it does, I've added one to 80adcfdf

PR #733, opened 2021-11-22
Bugfix/cv 627 inland holes

### LINKED ISSUE #627: CV: shore points are plotted in inland "holes"
Shore points are plotted on all landmass edges within the AOI, and that includes inland areas like lakes, etc, that appear as holes in the landmass polygon. Many of the algorithms (wind & wave exposure, etc) don't make any sense for those points. If we can, we should exclude those edges by excluding "inner rings" of landmass geometries.

That might be pretty easy if this post is correct that the outer ring is always the first geometry in an ogr Polygon.
https://gis.stackexchange.com/a/95284/112596

davemfish:
Fixes #627 

This PR removes interior rings of the landmass polygon as it's geometries are pre-processed for use in subsequent tasks. So now we don't get shore points in interior holes (i.e. lakes).

The same landmass geometries are also used in the wind exposure routine to measure fetch distances. In that case, the interior edges were always irrelevant so there's no change in the outcome of that routine.

# Checklist

- [x] Updated the user's guide (if needed) I think the UG already reads as if shore points will only be created along the exterior of the landmass.

### DIFF
--- a/tests/test_coastal_vulnerability.py
+++ b/tests/test_coastal_vulnerability.py
@@ -1167,6 +1167,57 @@ class CoastalVulnerabilityTests(unittest.TestCase):
         n_points = layer.GetFeatureCount()
         self.assertTrue(n_points == 6)
 
+    def test_landmass_interior_holes(self):
+        """CV: test no shore points created on interior holes."""
+        workspace_dir = self.workspace_dir
+
+        aoi_path = os.path.join(workspace_dir, 'aoi.geojson')
+        landmass_path = os.path.join(workspace_dir, 'landmass.geojson')
+        polygon_pickle = os.path.join(
+            workspace_dir, 'polygon.pickle')
+        lines_pickle = os.path.join(
+            workspace_dir, 'lines.pickle')
+        lines_rtree = os.path.join(
+            workspace_dir, 'rtree.dat')
+        target_vector_path = os.path.join(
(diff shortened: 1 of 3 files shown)


tests/test_coastal_vulnerability.py
```diff
@@ -1167,6 +1167,57 @@ def test_aoi_multiple_features(self):
...
+            workspace_dir, 'rtree.dat')
+        target_vector_path = os.path.join(
+            workspace_dir, 'shore_points.gpkg')
+        model_resolution = 100
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(26910)  # UTM Zone 10N
+        wkt = srs.ExportToWkt()
+
+        aoi_poly = Polygon([
+            (-200, -200), (200, -200), (200, 200), (-200, 200), (-200, -200)])
```
phargogh, line 1190: Since we're defining a box, I wonder if [`shapely.geometry.box`](https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.box) might be slightly easier for future us to grok?

phargogh: Actually, please disregard ... the use of coordinates here is consistent with the very sensible use of coordinates later in the test.

phargogh (approved):

PR #734, opened 2021-11-29
Feature/cv 581 raster habitats

### LINKED ISSUE #581: CV - accept raster format habitat layers
This was requested by @cybersea and it sounds reasonable to me.

In the past it seemed most typical for coastal & marine habitat data to originate in vector format, so that was the primary case we designed for. In recent times, with advances in remote sensing tech, it's increasingly common for original data sources to be rasters.

If we can, could we document some raster-format data sources that users are likely to want to use before we implement this?

### LINKED ISSUE #628: CV: one intermediate file uses SHP instead of GPKG
It's the intermediate geomorphology projected layer. We should use GPKG to be internally consistent. 

But this also created a bug in `3.9.0` because of now deprecated `Taskgraph` options to `copy_duplicate_artifacts` and the fact that only the `.shp` file is listed in a `target_path_list`. The other associated files that go with the shp are not copied. We've already removed usage of `copy_duplicate_artifacts`, and changing to a GPKG will make the existing `target_path_list` correct.

davemfish:
This PR adds support for accepting raster format habitat layers in the CV model. Habitat maps are only used to represent presence/absence of habitat, so for rasters, I chose to treat 0 and nodata as absence, and anything else as presence.

The new `search_for_raster_habitat` function makes use of the pre-existing utility `_aggregate_raster_values_in_radius`. That function just needed to be made slightly more abstract by taking an `aggregation_op` (inspired by raster calculator) so that it can be shared by the `relief`, `population`, and `habitat` calculations.

I updated test habitat data to replace a shapefile with a raster. This is hard to notice in the test code because that reference is hidden inside a CSV. But `test_habitat_rank` and `test_complete_run` now run with a raster habitat layer. And I updated the sample data in the same manner.

Fixes #581 and fixes #628 

# Checklist

- [x] Updated the user's guide (if needed) -- only a minor update was needed because the details in the Data Needs section now come straight from the `ARGS_SPEC`!

### DIFF
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -1742,7 +1742,7 @@ def calculate_relief_exposure(
 
     def mean_op(array, mask):
         nonlocal missing_values
-        if numpy.count_nonzero(mask) > 0:
+        if numpy.any(mask):
             return numpy.mean(array[mask])
         missing_values += 1
         return numpy.nan
@@ -1871,7 +1871,7 @@ def aggregate_population_density(
 
     def density_op(array, mask):
         nonlocal missing_values
-        if numpy.count_nonzero(mask) > 0:
+        if numpy.any(mask):
             pixel_area_km = abs(pixel_size[0] * pixel_size[1]) / 1e6
             return numpy.mean(array[mask]) / pixel_area_km
         missing_values += 1
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -2744,7 +2744,7 @@ def _aggregate_raster_values_in_radius(
         target_pickle_path (string): path to pickle file storing dict
             keyed by point id.
         aggregation_op (function): takes a signal array and a mask array
-            of same shape. Returns scalar.
+            of same shape. Returns scalar or numpy.nan.
 
     Returns:
         None
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -1750,6 +1750,7 @@ def calculate_relief_exposure(
     _aggregate_raster_values_in_radius(
         base_shore_point_vector_path, positive_dem_path, dem_averaging_radius,
         target_relief_pickle_path, mean_op)
+
     if missing_values:
         LOGGER.warning(
             f'{missing_values} shore points are missing values after '
@@ -2012,6 +2013,7 @@ def search_for_raster_habitat(
         clipped_projected_hab_path)
 
     def habitat_op(array, mask):
+        """Checking for non-zero or non-nodata values in array."""
         if numpy.any(array[mask]):
             return habitat_rank
         return 5  # the rank indicating no habitat protection
@@ -2731,8 +2733,8 @@ def _aggregate_raster_values_in_radius(
             for valid pixels.
         target_pickle_path (string): path to pickle file storing dict
             keyed by point id.
-        aggregation_op (function): takes a signal array, a mask array,
-            and arbitrary *args.
+        aggregation_op (function): takes a signal array and a mask array
+            of same shape. Returns scalar.
 
     Returns:
         None
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -1653,7 +1653,8 @@ def calculate_surge_exposure(
                     "Could not transform feature from %s to spatial reference"
                     "system %s", shelf_contour_path, target_srs_wkt)
                 continue
-            shapely_geom = shapely.wkb.loads(bytes(clipped_geometry.ExportToWkb()))
+            shapely_geom = shapely.wkb.loads(
+                bytes(clipped_geometry.ExportToWkb()))
             if shapely_geom.type == 'MultiLineString':
                 shapely_geom_explode = [
                     shapely.geometry.LineString(x) for x in shapely_geom]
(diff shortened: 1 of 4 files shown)


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -1885,21 +1930,103 @@ def _schedule_habitat_tasks(
...
+                      habitat_row.rank,
+                      habitat_row.id,
+                      base_habitat_path,
+                      target_habitat_pickle_path),
+                target_path_list=[target_habitat_pickle_path],
+                task_name='searching for %s' % habitat_row.id))
+            continue
+        raster = gdal.OpenEx(base_habitat_path, gdal.OF_RASTER)
+        if raster:
+            raster = None
```
phargogh, line 1949: For the spatial type checking (raster/vector), this sounds like a good use case for `pygeoprocessing.get_gis_type()`.  It doesn't do much more than what you have here with the conditionals, but would mean that you wouldn't need to remember to `None` out the raster/vector variables.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2985,10 +3078,14 @@ def _validate_habitat_table_paths(habitat_table_path):
         try:
             pygeoprocessing.get_vector_info(base_habitat_path)
         except ValueError:
-            bad_paths.append(base_habitat_path)
+            try:
+                pygeoprocessing.get_raster_info(base_habitat_path)
+            except ValueError:
+                bad_paths.append(base_habitat_path)
```
phargogh, line 3084: This might be another use case for `pygeoprocessing.get_gis_type`, since `get_raster_info` and `get_vector_info` will both return a bunch of other information that aren't strictly needed.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2599,27 +2724,24 @@ def _aggregate_raster_values_in_radius(
...
         sample_distance (float): radius around each point to search
             for valid pixels.
-        aggregation_mode (string): 'mean' or 'density'. Calculate the mean
-            of valid pixels in search window, or if 'density', divide the mean
-            by the area of valid pixels. 'density' assumes units of meters
-            for sample_distance and raster geotransform.
         target_pickle_path (string): path to pickle file storing dict
             keyed by point id.
+        aggregation_op (function): takes a signal array and a mask array
+            of same shape. Returns scalar.
```
phargogh, line 2739: Or returns `numpy.nan` if no masked signal pixels are found?

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2669,78 +2790,50 @@ def _aggregate_raster_values_in_radius(
...
+            # an empty array.
+            win_xsize = 0
+            win_ysize = 0
+        array = band.ReadAsArray(
+            xoff=pixel_x, yoff=pixel_y, win_xsize=win_xsize,
+            win_ysize=win_ysize)
+        if nodata is not None:
+            kernel_mask &= ~numpy.isclose(array, nodata)
+
+        result[shore_id] = aggregation_op(array, kernel_mask)
```
phargogh, line 2836: I see that in the removed code, there's a check in the `'density'` op about whether there are `nan` pixels.  Is it expected to have `result[shore_id] = numpy.nan`, for the case that one of the ops does end up returning `nan`?

phargogh (changes requested):
Thanks, Dave!  I just had a couple small things I was wondering about.

src/natcap/invest/coastal_vulnerability.py
davemfish: Cool, I did not know that existed

src/natcap/invest/coastal_vulnerability.py
davemfish: Not strictly. The habitat op actually returns 5 in that case, representing no habitat found. Because we're allowing nodata to represent known absence of habitat. As I'm writing that out, it starts to sound like a bad idea. Hmm.

src/natcap/invest/coastal_vulnerability.py
davemfish: Yes, `nan` values are expected to exist in those `result` dicts, if that is your question.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2669,78 +2790,50 @@ def _aggregate_raster_values_in_radius(
...
-            "around these points. You may wish to increase the search radius "
-            "input parameter, or you may wish to view %s and the landmass "
-            "polygon input in a GIS to confirm that both layers are accurate "
-            "and well-aligned. "
-            "\n**************************************************"
-            % (n_missing_data, os.path.basename(base_raster_path),
-                sample_distance, os.path.basename(base_raster_path)))
+            # Not sure how we get here, but clamping to 0 allows
+            # the aggregation_op to decide what to return given
+            # an empty array.
```
davemfish, line 2827: Hmm I should think on this a bit more. If we're completely off the raster, then maybe even the habitat op should return `nan` instead of the rank representing no habitat protection.

davemfish:
Thanks @phargogh . I pushed some changes to use `pygeoprocessing.get_gis_type`. And your other comments are making me want to rethink things a bit. About if/when to assign `nan` as a habitat rank. Or even if we should be more strict with the raster habitat inputs and not allow nodata to represent "known absence" of habitat. Let me think about that.

src/natcap/invest/coastal_vulnerability.py
davemfish: Sorry I was splitting hairs here. I added "or numpy.nan". But I think we can leave it up to the actual `op` to describe when it might return `numpy.nan`, as it might vary by the `op`.

src/natcap/invest/coastal_vulnerability.py
davemfish: There are tests to make sure `nan` is handled properly later on when these intermediate results are used in subsequent equations.

src/natcap/invest/coastal_vulnerability.py
davemfish: I updated this function to do just that.

davemfish:
Okay @phargogh I updated the `_aggregate_raster_values_in_radius` to completely bypass the local `aggregation_op` and immediately assign `result[shore_id] = numpy.nan` when the point + radius are completely outside the bounds of the raster.

I updated `test_aggregate_raster_edge_cases` to demonstrate this.

Prior to this change there was no way for a shore point's habitat value to be `nan`, so I updated and tested `_calc_Rhab` to make sure the `nan` propagates correctly. Then `Rhab` is used in the `geometric_mean` for the final exposure calculation, and that function is already setup to handle `nan` correctly.

phargogh (approved):
Makes sense to me ... thanks, @davemfish !

PR #742, opened 2021-11-30
UCM tests different on re-run

### LINKED ISSUE #615: UCM test: different results when re-run in same workspace
To reproduce:
1. Edit the test `test_ucm_regression_factors` so that `args['workspace_dir']` is not temporary: `args['workspace_dir'] = '/Users/emily/Documents/ucm-test2'`
2. Run the UCM tests (all pass)
3. Run the UCM tests again (now `test_ucm_regression_factors` fails with an error: `AssertionError: 10560639.842782581 != 3520213.280928277 within 4 places (7040426.561854305 difference) : avd_eng_cn should be close to 3520213.280928, actual: 10560639.842783`

You should be able to re-run any model in the same workspace and get the same result, so it's concerning that the aggregate results are different.

phargogh:
This PR fixes #615 by bumping the taskgraph requirement.  The ultimate issue here is the `mtime` bug that was resolved in https://github.com/natcap/taskgraph/pull/75

# Checklist

- [x] Updated the user's guide (if needed) (not needed)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -38,6 +38,9 @@ Unreleased Changes
 ------------------
 * General
     * Update to FontAwesome 5 icons in the QT interface.
+    * Updating the ``taskgraph`` requirement to ``0.11.0`` to resolve an issue
+      where modifying a file within a roughly 2-second window would fool
+      ``taskgraph`` into believing that the file had not been modified.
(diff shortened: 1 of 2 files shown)


emlys (approved):
great, thanks @phargogh!

PR #744, opened 2021-12-01
Urban Nature Access - Under- and Over-Supplied Populations

phargogh:
This is the last PR of the core model functionality before I start on the optional components.  There are two additional statistics that are called for in the design doc:

* `Pund_adm`, the total population in each administrative unit that are underserved by greenspace
* `Povr_adm`, the total population in each administrative unit that are overserved by greenspace

These are not mutually exclusive ... there may be administrative units that contain both underserved and overserved populations.

So this is implemented by a couple `raster_calculator` calls and some `zonal_statistics` before writing out the totals to the new vector.  I also renamed one of the other vector fields to match what's in the design doc.


src/natcap/invest/urban_nature_access.py
```diff
@@ -463,8 +554,11 @@ def _admin_level_supply_demand(
         avg_greenspace_supply_demand = (
             greenspace_stats[feature_id]['sum'] /
             population_stats[feature_id]['sum'])
-
         feature.SetField(supply_sum_fieldname, avg_greenspace_supply_demand)
+        feature.SetField(
+            undersupply_fieldname, undersupplied_stats[feature_id]['sum'])
+        feature.SetField(
+            oversupply_fieldname, oversupplied_stats[feature_id]['sum'])
```
dcdenu4, line 561: If we add anymore stats to the vector aggregate we might want to move this into an iteration. But, I think it's fine at 3.

tests/test_urban_nature_access.py
```diff
@@ -302,5 +302,21 @@ def test_model(self):
...
+        oversupplied_pop = admin_feature.GetField('Povr_adm')
+        numpy.testing.assert_allclose(
+            oversupplied_pop,
+            981.987549)  # From eyeballing the results; random seed = 1
+
+        # The sum of the under-and-oversupplied populations should be equal to
+        # the total population count.
+        numpy.testing.assert_allclose(
+            undersupplied_pop + oversupplied_pop,
+            population_array.sum())
```
dcdenu4, line 322: Nice

dcdenu4 (approved):
Thanks @phargogh , easy peasy

phargogh:
@dcdenu4 this branch is based off of `release/3.10`, so I just pulled in the latest changes from upstream and pushed them here ... we should see the tests updated next, and I'm happy to wait on having this merged until we know that builds/tests will be passing.

dcdenu4 (approved):
Alright, I briefly spotted all those updates coming from 3.10 latest. Seem good and tests are passing.

PR #745, opened 2021-12-02
don't download gnu zip after choco install zip.

davemfish:
There appears to be a redundant install of `zip` in our windows installer workflow. Sorry I didn't create an issue for this, was just trying out a fix and it appears to work. 

Our builds had been failing because that second download and install of `zip` was failing due to unresponsive sourceforge mirror, or something like that.

# Checklist


dcdenu4 (approved):
Thanks for tackling this @davemfish! I thought about testing / doing this the last time we changed the mirror, but got lazy. Glad we seemingly don't have an issue with `choco` now like we must have had in the past.

PR #752, opened 2021-12-04
Deprecate direct setup.py invocation

### LINKED ISSUE #717: setup.py commands deprecation
I was looking at the `setuptools` repo and noticed a recent commit message _"Deprecate setup.py install and easy_install."_ It seems that `setuptools` is slowly deprecating all direct calls to `setup.py` commands - though this isn't well documented or announced anywhere. Need to investigate further and determine if we need to make any changes.

Related setuptools issue: https://github.com/pypa/setuptools/issues/917
Blog post: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

I routinely use `setup.py` install, though our `make install` command already uses `pip install` which may be the correct alternative. Our `setup.py` also provides the `build_ext` command.

phargogh:
This PR replaces direct `setup.py` invocation across the repository with PyPA-recommended alternatives `build` and `pip`.

This also includes the update to `make install` we discussed on the team slack, where it will now install the specific version that is in the CWD.

Fixes #717

# Checklist
- [x] Updated the user's guide (if needed) _no changes needed to the UG_

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -91,13 +91,13 @@ jobs:
               run: |
                   conda install nomkl # make sure numpy is w/out MKL
                   conda upgrade -y pip setuptools
-                  conda install ${{ matrix.numpy }} ${{ matrix.gdal }} toml twine
+                  conda install ${{ matrix.numpy }} ${{ matrix.gdal }} toml twine build
                   conda install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
 
             - name: Build wheel
               shell: bash -l {0}
               run: |
-                  python setup.py bdist_wheel
+                  python -m build --wheel
                   python -m twine check dist/*
 
             - name: Install runtime dependencies
@@ -196,7 +196,7 @@ jobs:
               run: |
                   conda install nomkl # make sure numpy is w/out MKL
                   conda upgrade -y pip setuptools
-                  conda install toml twine
+                  conda install toml twine build
                   conda install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
                   python ./scripts/convert-requirements-to-conda-yml.py \
                     requirements.txt > requirements-all.yml
@@ -205,7 +205,8 @@ jobs:
             - name: Build source distribution
               shell: bash -l {0}
               run: |
-                  python setup.py build_ext sdist
+                  python -m build --wheel
+                  python -m build --sdist
                   python -m twine check dist/*
 
             - name: Install from source distribution
@@ -280,12 +281,12 @@ jobs:
               shell: bash -l {0}
               run: |
                   conda install nomkl # make sure numpy w/out mkl
-                  conda install pytest requests
+                  conda install pytest requests build
                   conda install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
                   conda upgrade -y pip wheel
                   python ./scripts/convert-requirements-to-conda-yml.py requirements.txt > requirements.yml
                   conda env update --file requirements.yml
-                  python setup.py install
+                  make install
 
             - name: Validate sample data
               shell: bash -l {0}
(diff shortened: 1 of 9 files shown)


HISTORY.rst
```diff
@@ -38,6 +38,12 @@ Unreleased Changes
 ------------------
 * General
     * Update to FontAwesome 5 icons in the QT interface.
+    * In response to the deprecation of ``setup.py``-based commands in Python
+      3.10, the recommended way to build python distributions of
+      ``natcap.invest`` is now with the ``build`` package, and installation
+      should be done via ``pip``.  The ``README`` has been updated to reflect
+      this change, and this should only be noticeable for those installing
+      ``natcap.invest`` from source.
```
davemfish, line 46: Could we also note the bugfix in `make install`? e.g. `make install` now always installs the current version of ``natcap.invest``, whereas previously it installed from whichever ``whl`` was found with latest version.

Makefile
```diff
@@ -234,16 +233,16 @@ env:
 # of pip don't think CWD is a valid package.
 install: $(DIST_DIR)/natcap.invest%.whl
 	-$(RMDIR) natcap.invest.egg-info
-	$(PIP) install --isolated --upgrade --no-index --only-binary natcap.invest --find-links=dist natcap.invest
+	$(PIP) install --isolated --upgrade --no-index --only-binary natcap.invest --find-links=dist "natcap.invest==$(VERSION)"
```
davemfish, line 236: It's not in the diff, but making `VERSION` still requires a direct `setup.py` invocation. Is that okay?

davemfish (changes requested):
Thanks for improving our `make install` command, @phargogh ! Everything else looks good too. I just had a suggestion for HISTORY. And a question about still using `setup.py --version`.

Makefile
dcdenu4: I'm curious about this as well, the line is: `VERSION := $(shell $(PYTHON) setup.py --version)`

dcdenu4 (commented):
Thanks for tackling this James. It looks good to me, but I'll just note this as a comment since Dave requested changes.

.github/workflows/build-and-test.yml
```diff
@@ -205,7 +205,11 @@ jobs:
             - name: Build source distribution
               shell: bash -l {0}
               run: |
-                  python setup.py build_ext sdist
+                  # Because we're using PEP518 build requirements, the user's
+                  # computer is guaranteed to have cython available at build
+                  # time.  Thus, it is no longer necessary to distribute the
+                  # .cpp files in addition to the .pyx files.
```
emlys, line 211: Neat!

emlys (approved):
Looks good!

HISTORY.rst
phargogh: Yep!  Patched in c71f02d7

Makefile
phargogh: Long-term, probably not!  But short-term, I think this is probably OK enough because `setuptools` isn't really doing a lot of work here since that's delegated to `setuptools_scm` .... though that delegation is, admittedly, done through a convoluted setuptools-managed system of entrypoints and who knows how stable that is.  So we should eventually get away from invoking `setup.py` even for the version string.

Unfortunately, the `setuptools_scm` devs don't have a good way to do this that bypasses `setup.py` entirely _right now_.  There is an active PR in the works (https://github.com/pypa/setuptools_scm/pull/646), which would enable `python -m setuptools_scm` to read from `pyproject.toml` as it should so that's the option I'd prefer to go with, but I don't know when it'll be released.  Assuming this approach is OK with everyone, I've created https://github.com/natcap/invest/issues/764 and assigned it to the InVEST 3.10.1 milestone so we can be sure to revisit this when we can.

phargogh:
Thanks, all!  I just pushed a couple changes that noted the change to `make install` in HISTORY, but also moved the `setuptools_scm` configuration over to `pyproject.toml` which is now the recommended place for that configuration.

dcdenu4 (approved):
Thanks James, this looks good to me.

Makefile
davemfish: Okay, thank you. Mainly just wanted to confirm it wasn't an oversight to leave it in for now and it sounds like it certainly is not.

davemfish (approved):

Issue #568, opened 2021-06-10
Translations: Add translation infrastructure


emlys:
This will involve:
    - making a way for UI "choose a language" components to communicate the chosen language to invest and reload
        - new endpoint for the workbench
        - ??? for Qt if we do that
    - calling `gettext.translation(...).install()` somewhere (cli.py?) 
    - creating directories in the expected structure to store our translation .po files
    - calling `pybabel compile` as part of the install/build process (converts human-readable .po files to binary .mo files, which are used to look up strings at runtime)

dcdenu4:
I wonder if the choosing the language part should just be upfront in the install process. Maybe that's not doable for our implementation? Or maybe there's examples of other software that let you switch back and forth? I feel like I usually see the language selection once when installing in other applications.

emlys:
This was done in #645

PR #754, opened 2021-12-06
Don't use "millions of" currency units in Wind Energy

### LINKED ISSUE #741: Don't use "millions of" currency units in Wind Energy
In Wind Energy, some inputs are measured in "millions of currency units" while others are in single currency units. The model depends on this and does some conversions using a factor of 1000000. We should standardize to use currency units everywhere, and not measure in "millions of" anything. Using "millions of" for some inputs creates room for confusion, and it's not universal - we've moved away from requiring US$, but not all currencies are sensibly measured in units of millions.

emlys:
# Description
Fixes #741. 
In my opinion the user's guide makes clear enough already that the default data is in millions of US dollars. I'm also editing the wind energy user's guide page with info that I learned from making these changes, but that can happen separately. 

# Checklist


davemfish (commented):
Hi @emlys this looks good to me. Can you just describe what was changed in the test data? Was it just renaming a raster and re-scaling it by 1000000?

And does the sample data need updating as well to change all those parameter values?

davemfish:
> In my opinion the user's guide makes clear enough already that the default data is in millions of US dollars. 

But shouldn't it be changed along with this PR to not use millions? Also there's the change in one output filename that should be updated in the UG. 

> I'm also editing the wind energy user's guide page with info that I learned from making these changes, but that can happen separately.

emlys:
@davemfish oh you're right, I did need to update that file name. I don't think any of the data in the UG needs to be updated, because the tables that have data in millions of $ are not model inputs. They're just illustrating some real-world examples.

emlys:
Also updated the affected parameters in the sample data. 
>  Can you just describe what was changed in the test data? Was it just renaming a raster and re-scaling it by 1000000?

Basically yeah, multiplying it by 1000000 did not work exactly since there was some change in the floating point error that added up. I spot-checked a couple of numbers and they were mostly close enough to the original expected values * 1000000 (relative difference 1e-6 or so). One was different on the order of 1e-4 to 1e-5, not sure why - the original expected data may have been off slightly too. With `float32` I'd expect that the error might be different, but not greater/less, between integers in the millions and those same numbers divided by 1000000 (xxxxxxx.0 vs x.xxxxxx)?

davemfish:
Hi @emlys , that all sounds good about the UG and the test data. ✔️ 

It looks like the sample data json files are failing to validate. And for some reason this model has a whole bunch of them.

emlys:
Ah, I shouldn't have updated the `invest.users-guide` `release/3.10` branch directly for #568.

emlys:
@davemfish this is ready - the mac build is failing with that unrelated issue

davemfish (approved):
Thank you!

PR #757, opened 2021-12-09
Stormwater: aggregate runoff fields

### LINKED ISSUE #756: Stormwater: include runoff ratio and volume in aggregate output
Ben has requested that the model be updated to aggregate (1) the mean runoff ratio and (2) the total runoff volume over each polygon. This complements the existing aggregates for retention and infiltration.

emlys:
# Description
Fixes #756 

I also renamed the aggregate fields to be more understandable. Since the output is a geopackage, we're not limited to 10 character names. The [geopackage spec](https://www.geopackage.org/spec/) recommends keeping identifiers shorter than 30 bytes. The names are still shorter than 30 characters (though could be longer if the user provides a lengthy pollutant name). And they're unique within the first 10 characters, so would still be usable if it was converted to shapefile and names were truncated.

No update to HISTORY.rst needed, since the model hasn't been released yet. User's guide was updated in the draft on google docs.

# Checklist

### DIFF
--- a/src/natcap/invest/stormwater.py
+++ b/src/natcap/invest/stormwater.py
@@ -474,8 +474,10 @@ def execute(args):
     aggregation_task_dependencies = [retention_volume_task]
     data_to_aggregate = [
         # tuple of (raster path, output field name, op) for aggregation
-        (final_retention_ratio_path, 'RR_mean', 'mean'),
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Makes sense and looks good!  Those fields are much more readable.

PR #759, opened 2021-12-09
CV: create 2D geometries instead of 2.5D

### LINKED ISSUE #749: CV outputs 3d point layer
Allison reported that a CV output layer geometry was the Point Z (3D) type, which is unnecessary and troublesome because she needed to convert it back to 2D.

davemfish:
Just a matter of using the `ogr` API correctly: `AddPoint_2D` instead of `AddPoint`. The latter creates a default z=0 when given only an x, y pair.

Fixes #749 

# Checklist

- [ ] ~~Updated the user's guide (if needed)~~


phargogh (approved):
Makes sense, looks great!  Easy fix :)

Issue #504, opened 2021-02-25
Implement the Stormwater Retention Model


phargogh:
The Stormwater Retention model has provisional approval from the PSC\*\*, meaning we can get started on implementing it.  For the model to be complete, the model should:

* Implement the functionality described in the design doc (linked below)
* Adhere to the [InVEST Spec](https://github.com/natcap/invest/wiki/spec_invest)
* Have an implemented Qt interface
* Have a test suite with full coverage where practical
* Have sample data committed to `data/invest-sample-data`
* Be added to the InVEST installer
* Have a User's Guide chapter

Resources:
* [PEP](https://docs.google.com/document/d/1IwFJCvI9wjQX_st5zUU5j5tnLwyhsCzylCPpJNmN-VA/edit)
* [Design Doc](https://drive.google.com/file/d/1rXM49wxmb2g0foo-W2Md27AyalYFVD6_/view)
* [Data](https://drive.google.com/drive/u/1/folders/1uCxfhZ4pGsJMrhoAvVPvmKjTewGlQmnd)

When this model is ready to be merged and we have PSC approval, let's merge this into `release/3.10` for future release.

\*\* NOTE: The PSC's provisional approval is contingent on a final review of one or two minor points.  As a result, it is possible that the model's specification may change slightly and so we should wait to merge it into InVEST until we have that final PSC approval.

emlys:
This is waiting on getting the finalized user's guide from Ben and Perrine, adding that to the UG, and then updating the UG commit hash in invest.

emlys:
Also need to add the sample data - I have emailed Ben and Perrine about which dataset they prefer we use.

davemfish:
Completed in #766

PR #766, opened 2021-12-14
Last stormwater things

### LINKED ISSUE #715: Final decision on name of new Stormwater model
It's possible we already have a decision on this from all the interested parties. But in case not, here's a reminder to do that before the `3.10` release.

Right now we have the user-facing model title as "Stormwater". If we stick with that, what about making it "Urban Stormwater"? 

Perhaps redundant, but it's kind of nice if all the urban models have the "Urban" prefix; it keeps them together in an alphabetized list in the launcher & workbench.

emlys:
# Description
Fixes #715, #504

* Change name to "Urban Stormwater Retention"
* Update user's guide
* Update sample data
* Rename "infiltration" to "percolation" (per discussion with Ben and Perrine, "percolation" is more clearly the water that percolates all the way to groundwater; "infiltration" may include water that is transpired) Accordingly change the IR_X column names to PE_X.
* Change the pollutant load aggregation from mean to sum, as requested by Ben
* Also remove the unused PO files for Spanish and Chinese. They were there to illustrate but I should've removed these in the translation PR (since the contents of `src/natcap/invest/internationalization/locales` serves as the single source of truth for which languages are supported, and we do not yet support other languages)

# Checklist


installer/windows/invest_installer.nsi
```diff
@@ -451,8 +451,8 @@ Section "InVEST Tools" Section_InVEST_Tools
         !insertmacro StartMenuLink "${SMPATH}\DelineateIt" "delineateit"
         !insertmacro StartMenuLink "${SMPATH}\Visitation: Recreation and Tourism" "recreation"
         !insertmacro StartMenuLink "${SMPATH}\Urban Flood Risk Mitigation" "ufrm"
-        !insertmacro StartMenuLink "${SMPATH}\Urban Cooling Model" "ucm"
-        !insertmacro StartMenuLink "${SMPATH}\Urban Stormwater Retention Model" "stormwater"
+        !insertmacro StartMenuLink "${SMPATH}\Urban Cooling" "ucm"
```
emlys, line 454: to be consistent with the names in `natcap/invest/__init__.py`, leaving out "Model"

phargogh (commented):
As far as the changes are concerned, this looks good to me!

I just had two things I wanted to ask about that I didn't have a good place to comment on:

1. Since we're renaming the model, should the python file also be renamed?
2. It looks like removing the `.po` files is causing a package build failure

emlys:
Argh, it turns out git can't deal with having empty directories, so `internationalization/locales` was lost. I put a `.gitignore` file in there as a placeholder. 

I didn't rename `stormwater.py` because though we updated the display name, the internal model name is separate and usually correlates with the file name. IMO `stormwater` still works for the internal model name.

emlys:
@phargogh, sorry I missed a few things in the first round! I'm pretty confident all the checks are on their way to passing now 🤞

phargogh:
Aah, looks like the binaries are failing because of `sphinx-reredirects`, which is my fault ... I added that to the UG repo and forgot to PR that in to the InVEST repo as well under `requirements-dev.txt`.  Could you add that dependency to resolve the build failure?  I'm sorry about that

emlys:
Looks like I did the same thing with `sphinx-intl`!

phargogh (approved):

PR #768, opened 2021-12-14
PR 767 merge conflict resolution

phargogh:
Just a simple merge conflict resolution on setup.py and pyproject.toml from PR #767 .

PR #769, opened 2021-12-14
PR 767 merge conflict resolution

phargogh:
Just resolves merge conflicts from PR #767 (which I also botched in #768 ).


davemfish (approved):
It's just the Qt-mac thing that's failing the build and I see we have a fix for that in the queue already, so I'll go ahead and merge.

Issue #751, opened 2021-12-04
Support HRA inputs in datastack archives


phargogh:
Issue https://github.com/natcap/invest/issues/342 is able to be completed for models that have input tables where fields can be completely described by their ARGS_SPEC, but HRA's input table(s) are so complicated that they can't be properly reflected in the ARGS_SPEC.  So we should look into how we might do this for HRA.

phargogh:
This was included in the HRA rewrite, resolved in #1026 and release with 3.12.

PR #772, opened 2021-12-15
Bundle CSV-linked spatial files in datastack archives

### LINKED ISSUE #342: Datastack archives don't package up data linked to in a CSV
We now have several models (HRA, CBC in the new refactor for sure, HQ in the `release/3.9` branch) that refer to other spatial files, either rasters or vectors, that are used in the model itself.  The datastack archive creation/extraction functions intended to create a self-contained archive of all data needed to run the model are missing these linked spatial files.

If we end up validating specific columns of data as spatial files (https://github.com/natcap/invest/issues/327), then maybe this new information in `ARGS_SPEC` could be used in datastack archiving as well?

phargogh:
This PR is a refactor of how we package up datastack archives so that we can now also package up any spatial files linked in a CSV.  The refactor seemed worth it because we've gotten _much_ clearer on how `args` is allowed to be structured since `datastack` was introduced due to the addition of `ARGS_SPEC`:

* We no longer need to guess at whether a parameter is a file (which may yield false positives)
* We can check whether an arg actually belongs to the model (is it described in `ARGS_SPEC`?)
* We know when a file is a spatial file (requiring special treatment) or is something else like a CSV
* We know when a CSV has spatial files in a column

Although the InVEST Spec does allow nested datatypes, I don't think we actually use them anywhere in InVEST so `datastack` doesn't try to handle them.  Support would be straightforward to add later.

For the HRA case, I've added a hook to `datastack.build_datastack_archive` to check if there's a `build_datastack_archive` function defined in the target model's file, and a stub function is now there that just raises `NotImplementedError` until it's addressed in https://github.com/natcap/invest/issues/751.  I like this approach because it allows us to just define a model-specific function in the model code itself rather than writing some kind of special-case into the datastack module.

Fixes #342 

# Checklist
- [x] Updated the user's guide (if needed) _(I don't think any changes are needed for the User's Guide)_


src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+        elif input_type == 'file':
+            target_filepath = os.path.join(
+                data_dir, f'{key}_{os.path.basename(source_path)}')
+            shutil.copyfile(source_path, target_filepath)
+            LOGGER.debug(
+                f'File copied from {source_path} --> {target_filepath}')
+            target_arg_value = _relpath(target_filepath)
+            files_found[source_path] = target_arg_value
+
+        elif input_type == 'directory':
```
davemfish, line 354: Directories are an interesting case. I wonder if you gave some thought to these issues:

1. Do we want to include the `workspace_dir` and it's contents? It could be very useful to do so. But it might be a little unexpected if the user is expecting to just archive the input data.
2. Could there be directories that contain a lot of other files besides the ones the model will use? E.g. seasonal water yield `et0_dir` looks like it does some pattern matching to find the files within the dir. Maybe it's okay though, to archive extra files. It's unlikely to happen in practice.

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+            target_arg_value = args[key]
+        rewritten_args[key] = target_arg_value
+
+    LOGGER.info('Args preprocessing complete')
+
+    new_args = {
+        'args': rewritten_args,
+        'model_name': model_name,
+        'invest_version': __version__
+    }
```
davemfish, line 402: It may or may not be preferable, but at this point we could probably use `datastack.build_parameter_set`, to handle writing the json file.

davemfish (commented):
Still working on this, but submitting so I don't lose the work.

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+
     args = args.copy()
     temp_workspace = tempfile.mkdtemp(prefix='datastack_')
-    logfile = os.path.join(temp_workspace, 'log')
     data_dir = os.path.join(temp_workspace, 'data')
     os.makedirs(data_dir)
 
-    # For tracking existing files so we don't copy things twice
+    # write a logfile to the archive
+    logfile = os.path.join(temp_workspace, 'log')
```
davemfish, line 209: This comes down to personal preference I guess, but what about `log.txt`?

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+                LOGGER.debug(
+                    f'Rewritten spatial CSV written to {target_csv_path}')
+                dataframe.to_csv(target_csv_path)
+
+            target_arg_value = _relpath(target_csv_path)
+            files_found[source_path] = target_arg_value
+
+        elif input_type == 'file':
+            target_filepath = os.path.join(
+                data_dir, f'{key}_{os.path.basename(source_path)}')
```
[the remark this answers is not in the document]
davemfish: For spatial files I think they retain their original basename, which makes sense to me.

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+                    shutil.copyfile(src_path, dest_path)
+
+            LOGGER.debug(
+                f'Directory copied from {source_path} --> {target_directory}')
+            target_arg_value = _relpath(target_directory)
+            files_found[source_path] = target_arg_value
+
+        elif input_type in spatial_types:
+            # Create a directory with a readable name, something like
+            # "aoi_path_vector" or "lulc_cur_path_raster".
```
davemfish, line 376: I'm curious if there was a functional reason for making a directory for each spatial file? Or is that just the way you prefer to keep things organized? I agree it's nice for organization, especially with shapefiles or ArcGrids floating around. I'm not suggesting a change.

src/natcap/invest/datastack.py
davemfish: The code it would replace is below, not in the diff.

davemfish (changes requested):
Hi @phargogh , this is great. I really like the placeholder solution for HRA. I have a few minor suggestions & questions.

I noticed `n_workers` and `workspace_dir` are not present in the archive or the new json file. They are two distinct cases, but wasn't sure if that was intentional for either/both. I could see the argument either way, for including/excluding them.

I tried this out and it works great!

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+    for key in args:
+        LOGGER.info(f'Starting to archive arg "{key}": {args[key]}')
+        # Possible that a user might pass an args key that doesn't belong to
+        # this model.  Skip if so.
+        if key not in args_spec:
+            LOGGER.info(f'Skipping arg {key}; not in model ARGS_SPEC')
+
+        input_type = args_spec[key]['type']
+        spatial_types = {'raster', 'vector'}
+        file_based_types = spatial_types.union({'csv', 'file', 'directory'})
```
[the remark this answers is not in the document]
phargogh: Whoops, that's an oversight on my part!  Yes, absolutely.  b9d4983af

src/natcap/invest/datastack.py
```diff
@@ -311,77 +192,217 @@ def build_datastack_archive(args, model_name, datastack_path):
...
+                    # (multiple types allowed), so always convert to a set for
+                    # easier comparison.
+                    col_types = col_definition['type']
+                    if isinstance(col_types, str):
+                        col_types = set([col_types])
+                    if col_types.intersection(spatial_types):
+                        spatial_columns.append(col_name)
+            LOGGER.debug(f'Detected spatial columns: {spatial_columns}')
+
+            target_csv_path = os.path.join(data_dir, f'{key}.csv')
```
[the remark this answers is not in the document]
phargogh: Yeah, this modified approach renames most files to the key that they are being used as.  The main reason for this is that it removes a layer of indirection when we're looking at the files themselves and we can just look at the files and have a good idea of which input parameters they're supposed to represent.

And yes!  The contents of the CSVs should also be rewritten to point to new spatial files in the new directory.

src/natcap/invest/datastack.py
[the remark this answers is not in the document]
phargogh: I ended up going for the more verbose version, `{key}_{input_type}_{basename}`, which I don't think will cause problems from length.  What do you think?  5cb94ecc8

Yes, the spatial files will retain their original basename within the directory that contains those file.

src/natcap/invest/datastack.py
phargogh: Ooo good point about the `workspace_dir`.  I just added an explicit exclusion there, lest we one day end up with a massive datastack that has a user's complete workspace in there!  1d85c86d8

> Could there be directories that contain a lot of other files besides the ones the model will use?
Yeah, I used to be really worried about including extra data that shouldn't be there, and I feel fairly confident that most users with a `directory` input created the directory for this purpose while assembling their inputs and I think will be less likely to include unrelated files.

In the SWY ET0 case, we don't currently have an `ARGS_SPEC`-defined way to pick out which files are supposed to be there and which aren't since the pattern matching happens within `execute`.  I suppose this would be a reasonable case for the per-model function there and I'd just need to think a bit more about how this might work.

So I think I'd like to leave it alone for now, but I've created https://github.com/natcap/invest/issues/812 for this because the problem is an interesting one and extends to the various sample datasets that we ship with InVEST and might not want to include in the datastacks.

src/natcap/invest/datastack.py
phargogh: That's a great idea!  It removes the relative path stuff to, since that's all handled within the parameter set code.  eb1fcc739

tests/test_datastack.py
```diff
@@ -285,7 +269,8 @@ def test_duplicate_filepaths(self):
 
         # Collect the file into an archive
         archive_path = os.path.join(self.workspace, 'archive.invs.tar.gz')
-        datastack.build_datastack_archive(params, 'sample_model', archive_path)
+        datastack.build_datastack_archive(
+            params, 'test_datastack_modules.duplicate_filepaths', archive_path)
```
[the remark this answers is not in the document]
phargogh: I'll do one better!  The test now asserts the directory entries and that `data` is a directory.

src/natcap/invest/datastack.py
phargogh: I don't see why not!  07e827c84

src/natcap/invest/datastack.py
phargogh: Yeah, it comes down to those multipart files.  It's not a big deal with ESRI Shapefiles because all of the component files have the vector's basename (`watersheds.shp`, `watersheds.dbf`, etc.).  Arc Binary Grids on the other hand are a known problem because the raster's name is the parent directory but the component files are all pretty much the same as another Arc Binary Grid.  So if a user happened to provide two Arc Binary Grid rasters and we copied them into the data directory, we'd have a conflict, not to mention that we can't just rename `hdr.adf` to `lulc_cur_path` and expect the raster to work!

So given how many file formats GDAL supports, putting all the component files into a directory just seemed the most flexible solution than trying to handle a special case for those raster formats that are a little stranger.

src/natcap/invest/datastack.py
davemfish: Sounds good to me

davemfish (approved):

PR #774, opened 2021-12-15
Update Qt plugins path in pyinstaller binaries

### LINKED ISSUE #773: Mac builds on release/3.10 are failing
We've had a few builds on `release/3.10` failing recently: https://github.com/natcap/invest/commits/release/3.10

and they've all been failing with the message:

```
dist/invest/invest list
qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

phargogh:
This just updates the Qt plugins path to use the _actual_ plugins path.  This path has been broken for well over a year (probably a fair bit longer) and should probably have broken whenever we pinned our builds to always use PySide2 rather than PyQt.  I don't know why this stopped working ... there isn't anything material in the usual suspects (package versions, pyinstaller versions, etc.) that should have caused this failure.  So since that's a mystery, there's always a chance that the builds working on my fork here as a result of this change is just a fluke ... but it _did_ seem to be working.  It's not a particularly satisfying fix, but I'm willing to run with it if you are.

Fixes #773 ?

# Checklist
- ~~[ ] Updated HISTORY.rst (if these changes are user-facing)~~

### DIFF
--- a/exe/qt.conf
+++ b/exe/qt.conf
@@ -1,4 +1,4 @@
 # solves problem with binaries not launching on mac.
 # Suggested in https://github.com/pyinstaller/pyinstaller/issues/2857
 [Paths]
-Prefix = PyQt5/Qt
+Prefix = PySide2/Qt


dcdenu4 (approved):
Hey @phargogh, I'm willing to give this a shot.

PR #775, opened 2021-12-15
Restore accidentally-reverted codesigning updates

emlys:
First time using `git checkout --patch`!
Also deleted the `KEYCHAIN_PASS` and `STANFORD_CERT_KEY_PASS` secrets which are no longer needed.
Note that the codesigning steps don't run in PR workflows, so check that it's passing on [my branch](https://github.com/emlys/invest/tree/bugfix/gha-codesigning-wrong) instead. That's probably how this accidental revert slipped through (in an unrelated PR, where we didn't expect to look for it).
# Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -4,6 +4,10 @@ on:
     push:
     pull_request:
 
+defaults:
+  run:
+    shell: bash -l {0}
+
 jobs:
     check-syntax-errors:
         name: "Check for syntax errors"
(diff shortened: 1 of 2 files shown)


phargogh:
> First time using `git checkout --patch`!

That looks really useful!  And I appreciate that there's a checkout equivalent to `git add --patch`

phargogh (approved):
Yep, looks right to me!  Thanks for putting in a PR so quickly.

I'll mark approval and then merge the PR once the builds/tests finish

phargogh:
RTD looks like their builds for this PR might be deadlocked so since everything else is passing, I'm just going to merge!

Issue #796, opened 2022-01-03
numpy 1.22 compatibility


davemfish:
numpy 1.22 was released a few days ago and it seems to be causing a couple issues.

1. Build failures: `No module named numpy.distutils`
2. Test failure: `test_carbon_totals_precision` fails with a value outside the numerical tolerance

https://github.com/davemfish/invest/actions/runs/1649577411

davemfish:
It looks like the build issue is already fixed in pyinstaller, but not yet released. https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-numpy.py#L53

dcdenu4:
Would it make sense to cap numpy requirements to <1.22 for now, until the Pyinstaller release with this fix comes out?

davemfish:
The test failure is due to a bugfix in RNG https://github.com/numpy/numpy/pull/20314

I verified this by observing that the random floats generated in our test do differ slightly under the different numpy versions. So I will update our regression value.

davemfish:
> Would it make sense to cap numpy requirements to <1.22 for now, until the Pyinstaller release with this fix comes out?

I think it's either this or inject our own numpy hook for pyinstaller, only to remove it later. So I agree it probably makes sense to cap numpy or perhaps try installing Pyinstaller's `develop` branch?

phargogh:
Because the pyinstaller `develop` branch is a moving target (and the branch can sometimes introduce issues), I'd suggest either installing pyinstaller to a specific revision on their `develop` branch (the latest is probably fine) or capping the numpy version.  I'd worry that installing from the pyinstaller `develop` might leave us vulnerable to broken builds.

davemfish:
Okay, we'll restrict numpy for now. And then move this issue to `3.10.1` to update the test so it passes with numpy `1.22`, and also to watch for the Pyinstaller release.

PR #799, opened 2022-01-03
BUILDS: pyinstaller-numpy compatibility & NSIS in GHA

davemfish:
Temporarily addresses #796 by restricting numpy version. That keeps our builds working until Pyinstaller makes a release with an updated numpy hook.

We also had some NSIS issues to resolve:
1. NSIS is already pre-installed in GHA Windows VMs, so we don't need to install it ourselves
2. Apparently we don't have a unicode-compatible installer, so need to set that flag appropriately in the nsis script

Finally, we have very weird python build issues with python 3.9.9, but only on the UI test step. I replaced `make install` with `python -m pip install .` and that worked, but I can't explain why. 

# Checklist


phargogh (approved):
Looks good to me!

PR #800, opened 2022-01-03
only calculate subsurface for nitrogen

### LINKED ISSUE #494: NDR subsurface outputs should only be for nitrogen
A user from the forum asked why their phosphorous output was was always 0. https://community.naturalcapitalproject.org/t/ndr-subsurface-phosphorus-output-is-0/1710

Turns out that phosphorous isn't computed for subsurface results! This is lightly mentioned in the Users Guide, but not in a clear way ( the below statement is buried in a paragraph and data needs inputs for subsurface indicate `n` or `p`).
>Because phosphorus particles are usually sediment bound and less likely to be transported via subsurface flow, the model uses the subsurface option only for nitrogen

Looking at source, there's this comment:
```python
# Perrine says that 'n' is the only case where we could consider a
# prop subsurface component.  So there's a special case for that. 
if nutrient == 'n':                                               
    subsurface_proportion_type = 'proportion_subsurface_n'        
else:                                                             
    subsurface_proportion_type = None                             
```

This makes me think that the Users Guide should be cleaned up such that:
- Data Needs subsurface inputs don't mention phosphorous
- Why phosphorous isn't calculated for subsurface is more clearly mentioned
- Interpreting results removes phosphorous from subsurface outputs
- The model itself stops writing outputs for subsurface phosphorous with 0 values.

emlys:
# Description
Fixes #494 

* Removed inputs `subsurface_critical_length_p` and `subsurface_eff_p`. These were used along with distance to stream to calculate subsurface NDR for phosphorus (`sub_ndr_p.tif`).
* Removed output `sub_ndr_p.tif`. This was calculated legitimately, but IMO it's confusing to include it.
* Removed output `sub_load_p.tif`. This was always filled with 0s.
* Changed the `export` outputs to clarify their meaning:
    * Removed `n_export.tif`, which was the total nitrogen export (surface and subsurface)
    * Removed `p_export.tif`, which was the total phosphorus export (surface and subsurface, though subsurface was always 0)
    * Added `n_surface_export.tif`: surface nitrogen export
    * Added `n_subsurface_export.tif`: subsurface nitrogen export
    * Added `n_total_export.tif`: total nitrogen export (equivalent to the former `n_export.tif`)
    * Added `p_surface_export.tif`: surface phosphorus export (equivalent to the former `p_export.tif`)
* Changed the aggregate vector fields to correspond to the new export outputs.
* Since there are now more field names in the aggregate vector, it was difficult to make them unique and understandable in just 10 characters. I changed the aggregate vector to gpkg format so that longer field names are allowed. 
* Update to modern string formatting syntax

I'm working on the corresponding user's guide changes in a branch off main.

# Checklist


src/natcap/invest/ndr/ndr.py
```diff
@@ -204,18 +181,10 @@
     's_bar_path': 's_bar.tif',
     's_factor_inverse_path': 's_factor_inverse.tif',
     'stream_path': 'stream.tif',
-    'sub_crit_len_n_path': 'sub_crit_len_n.tif',
-    'sub_crit_len_p_path': 'sub_crit_len_p.tif',
-    'sub_eff_n_path': 'sub_eff_n.tif',
-    'sub_eff_p_path': 'sub_eff_p.tif',
-    'sub_effective_retention_n_path': 'sub_effective_retention_n.tif',
-    'sub_effective_retention_p_path': 'sub_effective_retention_p.tif',
```
emlys, line 212: These files weren't used at all.

tests/test_ndr.py
```diff
@@ -314,66 +309,11 @@ def test_validation(self):
...
             'workspace_dir',
             'k_param',
             'watersheds_path',
-            'subsurface_eff_p',
             'subsurface_eff_n',
         ]
         self.assertEqual(set(invalid_args), set(expected_missing_args))
-
-    @staticmethod
-    def _assert_regression_results_equal(
```
emlys, line 329: this wasn't used anywhere

src/natcap/invest/ndr/ndr.py
```diff
@@ -405,7 +362,6 @@ def _validate_inputs(nutrients_to_process, lucode_to_parameters):
 
     # these are used for aggregation in the last step
     field_pickle_map = {}
-    field_header_order_list = []
```
emlys, line 408: This isn't needed since dictionaries now keep their order.

HISTORY.rst
```diff
@@ -75,6 +75,36 @@ Unreleased Changes
...
+      * Removed the inputs ``subsurface_critical_length_p`` and
+        ``subsurface_eff_p``
+      * Removed the output ``sub_ndr_p.tif``. The model no longer calculates
+        subsurface NDR for phosphorus.
+      * Removed the output ``sub_load_p.tif``. All pixels in this raster were
+        always 0, because the model assumed no subsurface phosphorus movement.
+      * Renamed the output ``p_export.tif`` to ``p_surface_export.tif`` to
+        clarify that it only models the surface export of phosphorus.
+      * Renamed the output ``n_export.tif`` to ``n_total_export.tif`` to
+        clarify that is is the total of surface and subsurface nitrogen export.
```
dcdenu4, line 89:
```suggestion
        clarify that it is the total of surface and subsurface nitrogen export.
```

src/natcap/invest/ndr/ndr.py
```diff
@@ -996,6 +990,42 @@ def _mult_op(*array_nodata_list):
...
 
 
+def _sum_rasters(raster_path_list, target_nodata, target_result_path):
+    """Sum the rasters in `raster_path_list`.
+
+    Args:
+        raster_path_list (list): list of single band raster paths.
+        target_nodata (float): desired target nodata value.
+        target_result_path (string): path to float 32 target raster
+            multiplied where all rasters are not nodata.
```
dcdenu4, line 1000: Do you mean __sum__ instead of __multiplied__ here?

src/natcap/invest/ndr/ndr.py
```diff
@@ -996,6 +990,42 @@ def _mult_op(*array_nodata_list):
...
+    def _sum_op(*array_nodata_list):
+        """Multiply non-nodata stacks."""
+        result = numpy.empty(array_nodata_list[0].shape)
+        result[:] = target_nodata
+        valid_mask = numpy.full(result.shape, True)
+        for array, nodata in zip(*[iter(array_nodata_list)]*2):
+            if nodata is not None:
+                valid_mask &= ~numpy.isclose(array, nodata)
+        result[valid_mask] = array_nodata_list[0][valid_mask]
+        for array in array_nodata_list[2::2]:
```
dcdenu4, line 1015: Passing in the nodata values as raw and indexing this way is a neat implementation!

src/natcap/invest/ndr/ndr.py
```diff
@@ -996,6 +990,42 @@ def _mult_op(*array_nodata_list):
...
+    Returns:
+        None.
+
+    """
+    def _sum_op(*array_nodata_list):
+        """Multiply non-nodata stacks."""
+        result = numpy.empty(array_nodata_list[0].shape)
+        result[:] = target_nodata
+        valid_mask = numpy.full(result.shape, True)
+        for array, nodata in zip(*[iter(array_nodata_list)]*2):
```
dcdenu4, line 1011: Could you walk me through `*[iter(array_nodata_list)]*2`? I'm having a hard time understanding how it's unpacking with `zip`? Thanks!

src/natcap/invest/ndr/ndr.py
dcdenu4: Is there a reason to grab the first array into result instead of iterating with `array_nodata_list[0::2]`?

src/natcap/invest/ndr/ndr.py
```diff
@@ -996,6 +990,42 @@ def _mult_op(*array_nodata_list):
...
+        target_nodata (float): desired target nodata value.
+        target_result_path (string): path to float 32 target raster
+            multiplied where all rasters are not nodata.
+
+    Returns:
+        None.
+
+    """
+    def _sum_op(*array_nodata_list):
+        """Multiply non-nodata stacks."""
```
dcdenu4, line 1007: Sum instead of multiply?

src/natcap/invest/ndr/ndr.py
```diff
@@ -1366,10 +1380,12 @@ def _aggregate_and_pickle_total(
...
-    target_vector = driver.CreateCopy(
-        target_vector_path, base_vector)
-    target_vector = None  # seemingly uncessary but gdal seems to like it.
+
+    base_wkt = pygeoprocessing.get_vector_info(
+        base_vector_path)['projection_wkt']
+    # use reproject_vector to create a copy in geopackage format
+    # keeping the original projection
+    pygeoprocessing.reproject_vector(
+        base_vector_path, base_wkt, target_vector_path, driver_name='GPKG')
```
dcdenu4, line 1391: I'm curious if `CreateCopy` doesn't work for GPKG drivers, or if there was another reason to piggy back off of `reproject_vector` here?

dcdenu4 (changes requested):
Hey @emlys thanks for jumping on this issue. I don't even remember making it! I had a few comments and questions. It will also make sense to wait for 3.10 to be merged into main.

src/natcap/invest/ndr/ndr.py
emlys: `CreateCopy` raised an error when I tried to convert shapefile -> gpkg. I suspect it's not really meant for file conversion.

emlys: Would be nice to have a pygeoprocessing function specifically for copying vectors.

src/natcap/invest/ndr/ndr.py
emlys: Oops, yeah

src/natcap/invest/ndr/ndr.py
emlys: I just copied all this from the similar raster multiplication function in NDR haha. I should've looked more closely - this implementation is pretty confusing. Refactored the whole thing.

src/natcap/invest/ndr/ndr.py
emlys: I refactored all this - see above

emlys:
@dcdenu4 I addressed your comments and refactored that whole sum function

src/natcap/invest/ndr/ndr.py
dcdenu4: I believe we did have some functions like that back in a much earlier API implementation. I certainly remember writing a lot of vector copying, clipping, and intersection code when the gdal Python bindings were a lot less robust. When James and Rich refactored that to the modern day pygeoprocessing, I think a lot of that stuff was left behind.

dcdenu4 (commented):
Thanks for those changes @emlys. It looks like there are some conflicts because of the 3.10 release merge. I was also wondering if there was a UG update with these changes as well? And going off our conversation last call, did you happen to run this through the UI?

emlys:
> I was also wondering if there was a UG update with these changes as well?

I updated the UG commit hash for that (it's on a new `release/3.10.2` branch)

> did you happen to run this through the UI?

Good catch! I removed those inputs and ran NDR through the UI and it worked.

emlys:
Thanks for catching that Doug. There was a mix of tabs and spaces in there causing the weird indents.

dcdenu4 (approved):
Looks good to me!

PR #806, opened 2022-01-05
Use GCP auth action

### LINKED ISSUE #791: GHA google cloud warning
I noticed this in the **Setup GCP** step of the binary builds.

>Warning: "service_account_key" has been deprecated. Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. For more details, see https://github.com/google-github-actions/setup-gcloud#authorization

emlys:
# Description
Fixes #791 
The GCP steps don't run in PRs, so be sure to check them on my branch.

# Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -84,7 +84,7 @@ jobs:
                   # This builds the users guide, binaries, and installer
                   make windows_installer
 
-            - name: 'Authenticate GCP'
+            - name: Authenticate GCP
               # Secrets not available in PR so don't use GCP.
               if: github.event_name != 'pull_request'
               uses: 'google-github-actions/auth@v0'
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -258,6 +258,13 @@ jobs:
 
             - run: make sampledata sampledata_single
 
+            - name: Authenticate GCP
+              # Secrets not available in PR so don't use GCP.
+              if: github.event_name != 'pull_request'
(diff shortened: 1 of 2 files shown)


.github/workflows/binary-applications.yml
```diff
@@ -251,9 +261,7 @@ jobs:
             - name: Set up GCP
               # Secrets not available in PR so don't use GCP.
               if: github.event_name != 'pull_request'
-              uses: google-github-actions/setup-gcloud@master
-              with:
-                  service_account_key: ${{ secrets.GOOGLE_SERVICE_ACC_KEY }}
+              uses: google-github-actions/setup-gcloud@v0
```
phargogh, line 264: Do we not need to authenticate in this build job?

phargogh: Actually, looking into the [sampledata build job on your fork](https://github.com/emlys/invest/runs/4720464510?check_suite_focus=true), I see a bunch of errors sprinkled throughout:

```
ServiceException: 401 Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object.
```

So maybe we do need to authenticate?

phargogh (changes requested):
Thanks for updating this @emlys !  I think there's one place where we might also need to authenticate, but looks good otherwise!

.github/workflows/binary-applications.yml
emlys: Oh, looks like I just missed that one!

phargogh (approved):
Good to go!  Thanks @emlys !

PR #809, opened 2022-01-06
Stormwater sampledata and autotest

### LINKED ISSUE #808: Urban Stormwater missing sampledata & autotest
We forgot to include this model's sample data dir in the Makefile's list of zipdirs, and in the NSIS script's list of dirs. So the data is not available from the Windows installer or from our website's "Download individual sample data" link.

We also forgot to add this model to `invest-autotest.py`, otherwise that would have caught the missing data. Did we miss anything else?

And can we remove the `storm_impact` dir from the sampledata? I'm not sure if that was an earlier version of this model's data?

davemfish:
Adds the model's sampledata zip to the Makefile target and NSIS script, so it will be included in the Windows installer. Also adds the model the autotest script.

Fixes #808

# Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -20,9 +20,9 @@
   - Scenic Quality
   - SDR
   - Seasonal Water Yield
-  - Stormwater
   - Urban Cooling
   - Urban Flood Risk
+  - Urban Stormwater Retention
   - Wave Energy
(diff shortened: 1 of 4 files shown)


emlys (approved):
Thanks @davemfish

PR #813, opened 2022-01-11
Resolve shapely 1.8.0 deprecation warnings

### LINKED ISSUE #763: Shapely 1.8 deprecation warnings
It looks like shapely `1.8.0` is out and it's a transitional release in prep for `2.0`. 
https://shapely.readthedocs.io/en/stable/migration.html#migrating-to-shapely-1-8-2-0

There are some new deprecation warnings we can deal with while we're on `1.8.0`. And we should probably restrict our requirements to `<2.0`. I'm not sure exactly when 2.0 will drop.

I observed some warnings when running the CV sample data.

emlys:
# Description
Fixes #763 
* `cascaded_union` -> `unary_union`
* multipart geometries are no longer iterable
* can no longer assign attributes to geometries

# Checklist


src/natcap/invest/utils.py
```diff
@@ -816,18 +822,18 @@ def _assert_vectors_equal(
...
-            # Try comparing geoms exactly equal allowing for different
-            # geometry ordering
-            geoms_equal = expected_geom_shapely.equals(actual_geom_shapely)
+            # confusingly named, `equals` compares numbers exactly but
+            # allows coords to be in a different order.
+            # `equals_exact` compares numbers with a tolerance but
+            # coords must be in the same order.
+            geoms_equal = (
+                expected_geom_shapely.equals(actual_geom_shapely) or
+                expected_geom_shapely.equals_exact(actual_geom_shapely, 1e-6))
```
emlys, line 831: The tolerance argument is absolute not relative. This isn't ideal because coordinate values can span several orders of magnitude. But `shapely` has an [open PR](https://github.com/shapely/shapely/pull/1231) to normalize the coordinate values before comparing. I think `1e-6` will be good enough for us in the mean time.

phargogh (approved):
The changes here make sense to me!

PR #814, opened 2022-01-12
Move over arg spec formatting logic 

### LINKED ISSUE #777: Move args spec formatting code over from UG
The arg spec formatting code used in the `:investspec:` Sphinx extension could live in `invest` rather than `invest.users-guide`. Benefits include:
- Logical place for it to be reused in both `invest.users-guide` and `invest-workbench`
- gettext boilerplate can be consolidated with `invest`'s rather than maintaining separately

I think @davemfish suggested this before, and I'm not sure why I didn't do it then!

emlys:
# Description
Fixes #777. Goes with natcap/invest.users-guide#60. Prerequisite for natcap/invest#854.

* Copied over formatting logic and tests
* Added a test
* Slightly refactored and renamed formatting functions

# Checklist


src/natcap/invest/spec_utils.py
```diff
@@ -217,3 +218,308 @@ def fallback_serializer(obj):
...
+    except AttributeError:
+        raise ValueError(
+            f'Module "{module_name}" does not have an ARGS_SPEC')
+
+    for i, key in enumerate(arg_keys):
+        # convert raster band numbers to ints
+        if arg_keys[i - 1] == 'bands':
+            key = int(key)
+        try:
+            spec = spec[key]
```
davemfish, line 507: This threw me for a minute. Is it worth adding a comment around here reminding us that `arg_keys` is a sequence of nested keys, such that overwriting `spec` in this loop is not actually a problem?

src/natcap/invest/spec_utils.py
```diff
@@ -217,3 +218,308 @@ def fallback_serializer(obj):
...
+        except KeyError:
+            keys_so_far = '.'.join(arg_keys[:i + 1])
+            raise ValueError(
+                f"Could not find the key '{keys_so_far}' in the "
+                f"{module_name} model's ARGS_SPEC")
```
davemfish, line 512: Ah okay, this one seems very useful to add the custom error message with the full `keys_so_far` sequence. I can see how the original `KeyError` could be missing some useful info.

davemfish (changes requested):
Hey @emlys, I had a few comments, mainly about docstrings and exceptions, to help us out in the future.

I also looked at the UG build artifact from this PR and it looks good. But I see the `Makefile` was not updated so I'm guessing the UG is not yet building against the changes in https://github.com/natcap/invest.users-guide/pull/60 so I'm not sure if we yet have GHA evidence that all this is working? Especially because the Actions on the UG PR are not running. What do you think?

src/natcap/invest/spec_utils.py
```diff
@@ -217,3 +218,308 @@ def fallback_serializer(obj):
...
+        if 'columns' not in spec and 'rows' not in spec:
+            first_line += _(
+                ' Please see the sample data table for details on the format.')
+
+    # prepend the indent to each line in the indented block
+    return [first_line] + ['\t' + line for line in indented_block]
+
+
+def document_arg_rst(module_name, *arg_keys):
+    """Get the spec dictionary for a particular arg.
```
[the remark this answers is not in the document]
emlys: Sorry, I meant it as a verb but I see how that's confusing. Renamed this to `describe_arg_from_name`, and the other function to `describe_arg_from_spec`, to clarify that one looks up the arg spec from its name, and the other uses that spec.

src/natcap/invest/spec_utils.py
```diff
@@ -217,3 +218,308 @@ def fallback_serializer(obj):
...
+
+    Returns:
+        arg spec (dict)
+    """
+    # import the specified module (that should have an ARGS_SPEC attribute)
+    try:
+        module = importlib.import_module(module_name)
+        spec = module.ARGS_SPEC['args']
+    except ImportError:
+        raise ValueError(
```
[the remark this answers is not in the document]
emlys: I thought it was a bit clearer when I was updating the user's guide and I got an error from misspelling a module. Since an `ImportError` is usually a problem in the code, but a `ValueError` suggests an incorrect value was provided. But I don't feel strongly about it.

src/natcap/invest/spec_utils.py
emlys: Yeah, I needed this to debug some key names that I misspelled or mixed up.

src/natcap/invest/spec_utils.py
```diff
@@ -217,3 +218,308 @@ def fallback_serializer(obj):
...
+            f'Could not import the module "{module_name}"')
+    except AttributeError:
+        raise ValueError(
+            f'Module "{module_name}" does not have an ARGS_SPEC')
```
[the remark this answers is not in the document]
emlys: Okay, I just took out the try/catch. I found it useful when adding all the `:investspec:`s to the user's guide, but they're not going to change much now, so those errors are unlikely to come up a lot.

src/natcap/invest/spec_utils.py
emlys: Sure, I added a comment.

emlys:
@davemfish, I'm coming back to this now that 3.10.2 is out and the user's guide ReadTheDocs build is working. I think I addressed all your questions.

The ReadTheDocs build should be enough to show it's working, but I'll update the UG revision here so you can see it build on GHA too.

src/natcap/invest/spec_utils.py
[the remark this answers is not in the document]
davemfish: excellent

src/natcap/invest/spec_utils.py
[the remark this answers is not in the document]
davemfish: Yeah I see where you're coming from now. Still I think it's safer with the original exceptions, just in case an `ImportError` arises from some other cause at some point. Of course many other things would also come crashing down if that happened.

src/natcap/invest/spec_utils.py
davemfish: thanks

davemfish:
Thanks @emlys these changes look all good. Just a conflict in the `Makefile` that I think is blocking the actions from running.

emlys:
Hmm thats kind of a circular dependency issue. There is no commit on `natcap/invest.users-guide` that has both the changes for this PR, and the latest changes from main, because that PR isn't merged yet, and it depends on this one. Since the ReadTheDocs build is passing, I think that's enough?

davemfish:
@emlys it looks like maybe the tests needed updating after those latest changes?

emlys:
@davemfish Tests are passing now here and on natcap/invest.users-guide#60

Makefile
```diff
@@ -168,7 +168,7 @@ $(BUILD_DIR) $(DATA_DIR) $(DIST_DIR) $(DIST_DATA_DIR):
 	$(MKDIR) $@
 
 test: $(GIT_TEST_DATA_REPO_PATH)
-	$(TESTRUNNER) tests
+	$(TESTRUNNER) tests/test_spec_utils.py
```
davemfish, line 171: It looks like this is leftover from development and not a change we want, right?

davemfish (changes requested):
Hi @emlys , there's one change in the `Makefile` here that probably doesn't belong.

Makefile
emlys: argh, sorry! Fixed.

davemfish (approved):

PR #820, opened 2022-01-14
Update pull request template

emlys:
# Description
Add a checkbox for testing model UIs like we talked about. Also made the headers smaller.

# Checklist


phargogh (approved):
Looks like only the RTD build remains, but this PR doesn't have anything to do with the RTD docs so I'll merge!

PR #822, opened 2022-01-14
Make better use of GHA caching

### LINKED ISSUE #326: Using github actions caching in workflows
**The Question:**
Should we take another look at how we are currently using caches in our workflows?

**Background Conversations**
>I also noticed several "Restore pip cache" steps in the workflows and I wonder if those can be purged give the very light usage of pip that is left. Or maybe it's just safer & better to leave them in. - @davemfish 

>I implemented the pip cache because it was (at the time) strongly recommended in the github actions docs. In practice, I never got the cache to work reliably such that we would actually use our previously cached items. And it took quite a while to get to that point! - @phargogh 

>In our case, and given how much time caching took up to implement, I'd suggest that we pursue caching only if we know that a significant amount of runtime is spent downloading artifacts from conda or pypi, and the speedup promised from avoiding these downloads outweighs the cost of unzipping the cache. - @phargogh

emlys:
## Description
Fixes #326 

## problems with previous approach
On some jobs, we were caching the pip or conda wheel caches. This saved some time downloading the wheels.
* The wrong directories were being cached on some jobs
* The caches were very outdated, so few of the cached wheels were actually used. Github provides no way to clear or update a cache, only to create a new one. The caches contained wheels for older versions of packages. Since a newer version is available from PyPI, `pip` would download the newer version, and the cache didn't help us.
* Even when done correctly, caching wheels only saves download time. Installing the wheels is still time consuming.

## proposed new approach
* Removed all wheel caching. Instead, cache the entire conda environment. This includes all the installed packages so we don't need to download or install. (https://dev.to/epassaro/caching-anaconda-environments-in-github-actions-5hde)
* Add a week number timestamp (00-53) onto the cache key so that a fresh cache will start each week. Github automatically deletes old caches that haven't been used in 7 days ([docs](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)) 
* The week number increments on Sundays. So we can expect that actions will run in the exact same cached environment all week. Then on the first run on Monday, a new cache is started and all dependencies are installed from scratch. Issues resulting from new versions of dependencies released during the week will show up the next Monday.
* Bonus: added a custom Github Action that wraps the conda setup and caching steps. It can be reused in each job rather than duplicating those steps. See https://docs.github.com/en/actions/creating-actions

Restoring the cache does take some time, but less time than installing everything from scratch. The jobs are running faster.

## Checklist


.github/workflows/binary-applications.yml
```diff
@@ -45,22 +45,16 @@ jobs:
...
-              with:
-                  activate-environment: winbin-env
-                  auto-update-conda: true
-                  python-version: 3.8
-                  channels: conda-forge
+                  key: ${{ env.PYTHON_VERSION }}-windows-binaries
+              id: condacache
 
             - name: Install python dependencies
+              if: steps.condacache.outputs.cache-hit != 'true'
```
emlys, line 55: Only need to install the dependencies if the cache wasn't restored.

.github/workflows/binary-applications.yml
```diff
@@ -45,22 +45,16 @@ jobs:
...
               run: |
+                  conda install python=${{ env.PYTHON_VERSION }}
```
emlys, line 57: Install python with other dependencies, rather than in `setup-miniconda` action. That way the python installation gets cached and reused as part of the environment too.

.github/workflows/build-and-test.yml
```diff
@@ -187,20 +185,20 @@ jobs:
...
 
             - name: Install dependencies
+              if: steps.condacache.outputs.cache-hit != 'true'
               shell: bash -l {0}
               run: |
+                  conda install python=${{ matrix.python-version }}
                   conda install nomkl # make sure numpy is w/out MKL
                   conda upgrade -y pip setuptools
-                  conda install twine build
+                  conda install twine build setuptools_scm
```
emlys, line 201: Need `setuptools_scm` for versioning to work right.

.github/actions/setup_env/action.yml
```diff
@@ -0,0 +1,42 @@
...
+      identifier to make the cache key. This way the cache will refresh each
+      week, or if the requirements change."
+    required: false
+    default: ""
+outputs:
+  cache-hit:
+    description: "True if a cache hit was found, false if not"
+    value: ${{ steps.condacache.outputs.cache-hit }}
+runs:
+  using: "composite"
```
dcdenu4, line 18: What's `composite`?

.github/workflows/binary-applications.yml
```diff
@@ -45,22 +45,16 @@ jobs:
             - name: Fetch git tags
               run: git fetch origin +refs/tags/*:refs/tags/*
 
-            - name: Restore pip cache
-              uses: actions/cache@v2
+            - name: Set up conda environment
+              uses: ./.github/actions/setup_env
```
dcdenu4, line 49: Is the `./` necessary?

.github/workflows/build-and-test.yml
dcdenu4: Can you comment more on this? Is this related to the updates of using` python -m setuptools_scm`?

.github/actions/setup_env/action.yml
```diff
@@ -0,0 +1,42 @@
...
+        echo "CONDA_PREFIX=$CONDA_PREFIX" >> $GITHUB_ENV
+      shell: bash -l {0}
+
+    # NOTE the post step that saves the cache will only run if the job succeeds
+    - name: Restore conda environment cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.CONDA_PREFIX }}
+        key: ${{ inputs.cache-id }}-${{ env.WEEK }}-${{ hashFiles('**/requirements*.txt') }}
+      id: condacache
```
dcdenu4, line 42: General understanding question. Is the `id: condacache` here related and necessary to match the `ids` found in the workflow configs?

dcdenu4 (commented):
Hey @emlys , this is very cool. Really neat to see a custom action for setting up the conda env related parts. I just had a few questions / comments, nothing major.

dcdenu4:
>Restoring the cache does take some time, but less time than installing everything from scratch. The jobs are running faster.

Any metrics to share? Are we talking minimally faster or orders of magnitude faster? Just curious.

.github/workflows/build-and-test.yml
emlys: In another PR I'd removed a bunch of extra dependencies that were being installed but not used in some jobs. Turns out that `setuptools_scm` actually is needed and the versioning fails silently without it (the version will be an empty string). So just correcting that mistake.

.github/actions/setup_env/action.yml
emlys: `composite` is the type of the github action. It means it combines multiple steps that could ordinarily be run directly in a workflow. As opposed to Docker or Javascript actions, which don't have a `steps` section - they point to a Dockerfile or javascript file that contains the action logic.

A `composite` action works here because it's just combining other actions and some shell commands. To do something more complicated, you'd need to use docker or javascript.

.github/actions/setup_env/action.yml
emlys: Indirectly, yes. The `id` here is just used within this file. Then there's another `id` with the same name assigned to the step that calls the custom action.

Basically the custom action is a wrapper around the `actions/cache` action. The `id`s and `output`s are there to propagate the `cache-hit` output from the `actions/cache` up to the custom action so that it's available at the workflow level.

Trying to summarize another way:
1. the workflow step with id `condacache` runs the custom action
2. the custom action step with id `condacache` runs the `actions/cache` action 
3. the `actions/cache` action sets an output `cache-hit` on the custom action step
4. the custom action sets an output `cache-hit` on the workflow step that called it (using the value of `cache-hit` read from the custom action step)
5. future workflow steps can read the `cache-hit` output from the workflow step with id `condacache`

.github/workflows/binary-applications.yml
emlys: Yeah, it doesn't recognize it as a path otherwise. https://github.com/natcap/invest/actions/runs/1831051618

phargogh (approved):
I have no comments!  Looks good, and I'm excited for the addition of our own custom action for de-duplication and also for the faster builds.

PR #824, opened 2022-01-14
pin numpy build version

emlys:
## Description
Needed for natcap/invest.users-guide#62.

## Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -63,14 +63,13 @@ jobs:
               run: |
                   conda install nomkl # make sure numpy is w/out MKL
                   conda upgrade -y pip setuptools
-                  conda install toml requests
-                  conda install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
+                  conda install requests
                   python ./scripts/convert-requirements-to-conda-yml.py \
                         requirements.txt \
                         requirements-dev.txt \
                         requirements-gui.txt > requirements-all.yml
                   conda env update --file requirements-all.yml
-                  python -m pip install .
+                  make install
 
             - name: Install build dependencies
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -87,7 +87,6 @@ def run(self):
     },
     include_package_data=True,
     install_requires=_REQUIREMENTS,
-    setup_requires=['setuptools_scm', 'numpy', 'cython'],
```
emlys, line 90: `setup_requires` [is deprecated](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=setup_requires#package-requirement) in favor of the pyproject.toml `requires`.

.github/workflows/build-and-test.yml
```diff
@@ -291,10 +287,7 @@ jobs:
             - name: Install dependencies
               shell: bash -l {0}
               run: |
-                  conda install nomkl # make sure numpy w/out mkl
-                  conda install pytest requests build
-                  conda install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
```
emlys, line 296: `conda` doesn't understand `pip`'s environment markers that I added in the build-system requirements. But this line isn't needed any more because `build` will automatically install them.

.github/workflows/binary-applications.yml
```diff
@@ -252,10 +243,6 @@ jobs:
               with:
                   python-version: 3.8
 
-            - run: |
-                pip install toml
-                pip install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")
-
```
emlys, line 258: This wasn't actually needed to build the sample data.

phargogh (approved):
Makes sense to me!  Just waiting for the builds to pass.

PR #834, opened 2022-01-26
Fix versioning for uploaded sample data

### LINKED ISSUE #833: Sample data uploads are broken
I broke the "Build sampledata archives" github actions step in [this PR](https://github.com/natcap/invest/pull/824/files#diff-8c84249d542c4dc765611d903da24310d1aba5b01b0bdbd791d653f4ead06741L255-L258). Without the build dependencies, `python setup.py --version` doesn't work. This is how the Makefile gets its `VERSION` variable. It fails silently, `VERSION` is an empty string, and the sample data gets uploaded to `InVEST__sample_data.zip`. For instance: https://github.com/natcap/invest/runs/4941865526?check_suite_focus=true#step:8:32

emlys:
## Description
Fixes #833 
The zip file name now contains the version in the "Deploy artifacts to GCS" step on my fork: https://github.com/emlys/invest/runs/4945648840?check_suite_focus=true

I did not go back to the previous approach reading from the toml file: `pip install $(python -c "import toml;print(' '.join(toml.load('pyproject.toml')['build-system']['requires']))")` because it's very finicky getting it to work with the environment markers. And we'll be removing `setup.py --version` soon anyway.

I wish there was a way to document the makefile's python dependencies better. Or at least exit if one is missing. I guess the error from the shell command in `VERSION := $(shell $(PYTHON) setup.py --version)` doesn't raise an error in the shell `make` is running in. I wasn't able to find any info about errors in variable assignment like this.

## Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -250,6 +250,10 @@ jobs:
               with:
                   python-version: 3.8
 
+            - name: Install dependencies
+              # dependencies of setup.py, needed to get the version string
+              run: pip install babel cython numpy setuptools setuptools_scm wheel
+
             - run: make sampledata sampledata_single
 
             - name: Authenticate GCP


phargogh (approved):
The PRed changes look like they should do the trick!  I see the HRA tests are failing, but I know there's a separate issue #835 about that.

> I wish there was a way to document the makefile's python dependencies better. Or at least exit if one is missing.

One thing I noticed from the [Makefile shell function docs](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html) is that there's a `.SHELLSTATUS` variable that could be checked and an [error](https://www.gnu.org/software/make/manual/html_node/Make-Control-Functions.html) could be raised.

The PR here is complete for the purposes of fixing the `setup.py` imports, but I'll leave this unmerged in case you'd like to add something about `.SHELLSTATUS` in this PR.  If not, merge away!

phargogh:
Actually, never mind about `.SHELLSTATUS` ... it looks like it appeared in Make >= 4 (which isn't ubiquitous)

Issue #835, opened 2022-01-26
HRA tests failing in actions


dcdenu4:
I'm seeing some HRA tests fail in an auto PR for merging `main` into `release/3.10.2. These tests must not have been failing when merging into `main`.  They are failing for Windows and Mac Python 3.8, 3.9. Python 3.7 tests pass okay. https://github.com/natcap/invest/pull/832

This line in HRAs `_zonal_stats_to_csv` function appears to be the culprit:
`    stats_df = pandas.DataFrame(index=overlap_df.index, columns=columns)`

```
2022-01-25T19:17:25.4015294Z =========================== short test summary info ===========================
2022-01-25T19:17:25.4015663Z FAILED tests/test_hra.py::HraRegressionTests::test_hra_no_subregion_multiplicative_exponential
2022-01-25T19:17:25.4016097Z FAILED tests/test_hra.py::HraRegressionTests::test_hra_regression_euclidean_linear
2022-01-25T19:17:25.4016474Z ====== 2 failed, 585 passed, 8 skipped, 15 warnings in 457.84s (0:07:37) ======
```

```
2022-01-25T19:12:49.9538188Z tests/test_hra.py::HraRegressionTests::test_aoi_no_projection PASSED
2022-01-25T19:12:56.0746787Z tests/test_hra.py::HraRegressionTests::test_hra_no_subregion_multiplicative_exponential 01/25/2022 19:12:56  taskgraph.Task     ERROR    Something went wrong when adding task zonal_stats_to_csv (55), terminating taskgraph.
2022-01-25T19:12:56.0747696Z Traceback (most recent call last):
2022-01-25T19:12:56.0748148Z   File "C:\Miniconda\envs\windows-latest-test-env\lib\site-packages\taskgraph\Task.py", line 674, in add_task
2022-01-25T19:12:56.0752229Z     new_task._call()
2022-01-25T19:12:56.0752790Z   File "C:\Miniconda\envs\windows-latest-test-env\lib\site-packages\taskgraph\Task.py", line 1093, in _call
2022-01-25T19:12:56.0753434Z     payload = self._func(*self._args, **self._kwargs)
2022-01-25T19:12:56.0754009Z   File "C:\Miniconda\envs\windows-latest-test-env\lib\site-packages\natcap\invest\hra.py", line 1186, in _zonal_stats_to_csv
2022-01-25T19:12:56.0754603Z     stats_df = pandas.DataFrame(index=overlap_df.index, columns=columns)
[...]

PR #836, opened 2022-01-28
Updating setuptools_scm requirement and noting change in history.

### LINKED ISSUE #764: Don't use python setup.py --version
The `setuptools_scm` devs have a PR in the works (https://github.com/pypa/setuptools_scm/pull/646) which would do this:

```shell
$ python -m setuptools_scm
```

And this would produce a version string that uses our `pyproject.toml`-defined configuration.

The intent here would be to replace `python setup.py --version` with `python -m setuptools_scm`.

This is related to https://github.com/natcap/invest/issues/717, but we couldn't do it in #752 because of the PR above in `setuptools_scm` that hasn't yet been merged.

phargogh:
Fixes #764 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -35,6 +35,10 @@
 Unreleased Changes
 ------------------
 * General
+    * The minimum ``setuptools_scm`` version has been increased to 6.4.0 in
+      order to bypass calling ``setup.py`` for version information.  The
+      version of this project can now be retrieved by calling ``python -m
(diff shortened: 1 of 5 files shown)


emlys (approved):
Hooray! Looks good to me. Just waiting for checks to pass

phargogh:
There are 4 test failures, but these appear to be because of the pandas bug exercised by HRA, described in #835 so I'll merge.

PR #891, opened 2022-02-17
Add a Save Datastack interface to the workbench

### LINKED ISSUE #876: Add an interface to save a datastack
This should cover the functionality existing in `File > Save as` menu of the Qt UI.

davemfish:
Fixes #876 

There's a new flask endpoint. A new UI button for the save. And at the same time I added a new UI alert for feedback on all the "Save..." buttons, especially needed for the datastack because it can take a minute. The feedback necessitated returning data from the requests in `server_requests.js`

I also changed some variable names for consistency across the 3 "Save..." button's and their handlers. And added consistent tooltip text for each.

## Checklist
Hmm we haven't yet tracked any workbench changes in HISTORY. It seems like maybe we can wait until after an official `0.1.0` release.

### DIFF
--- a/src/natcap/invest/ui_server.py
+++ b/src/natcap/invest/ui_server.py
@@ -224,8 +224,7 @@ def build_datastack_archive():
 
     Body (JSON string):
         filepath: string - the target path to save the archive
-        moduleName: string (e.g. natcap.invest.carbon) the python
-            module name for which we are making the datastack.
+        moduleName: string (e.g. natcap.invest.carbon) the python module name
         args: JSON string of InVEST model args keys and values
 
     Returns:


workbench/src/renderer/components/InvestTab/index.jsx
```diff
@@ -226,6 +237,8 @@ export default class InvestTab extends React.Component {
       return (<div />);
     }
 
+    // Alert won't expire during archiving; will expire 2s after completion
+    const alertExpires = (saveAlert === 'archiving...') ? 1e7 : 2000;
```
emlys, line 241: Is the `1e7` just an arbitrary length of time to wait until this `delay` function is called again?

workbench/src/renderer/components/SetupTab/SetupButtons/index.jsx
```diff
@@ -1,22 +1,62 @@
...
+    <OverlayTrigger
+      placement="right"
+      delay={{ show: 250, hide: 400 }}
+      overlay={(
+        <Tooltip>
+          {props.hoverText}
+        </Tooltip>
+      )}
+    >
+      {/* the first child must be a DOM element, not a custom component,
```
emlys, line 20: Neat that you can comment inside JSX like this!

emlys (changes requested):
@davemfish this looks good! I had some comments and questions.

src/natcap/invest/ui_server.py
```diff
@@ -218,6 +218,29 @@ def save_to_python():
     return 'python script saved'
 
 
+@app.route('/build_datastack_archive', methods=['POST'])
+def build_datastack_archive():
+    """.
```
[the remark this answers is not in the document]
davemfish: Thanks!

tests/test_ui_server.py
```diff
@@ -139,6 +139,27 @@ def test_save_to_python(self):
...
+
+        payload = {
+            'filepath': target_filepath,
+            'moduleName': 'natcap.invest.carbon',
+            'args': json.dumps({
+                'workspace_dir': 'foo',
+                'carbon_pools_path': data_path
+            }),
+        }
+        _ = test_client.post('/build_datastack_archive', json=payload)
```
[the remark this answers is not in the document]
davemfish: Definitely. Though to keep it consistent with all the other tests, I made the assignment. Some tests use the return value, some don't. But this way we're consistent and the reader knows that the function returns something, and that we don't particularly care about it.

workbench/src/renderer/components/SetupTab/SetupButtons/index.jsx
```diff
@@ -1,22 +1,62 @@
...
+        hoverText={_("Save model setup to a Python script")}
+      >
+        <SaveFileButton
+          title={_("Save to Python script")}
+          defaultTargetPath="execute_invest.py"
+          func={props.savePythonScript}
+        />
+      </HoverText>
+      <HoverText
+        hoverText={_("Export all input data to a compressed archive")}
```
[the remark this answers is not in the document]
davemfish: I don't have any good alternatives in mind. It may not be a familiar phrase to all, but I do think it's more/less plain english and not technical jargon.

workbench/src/renderer/components/InvestTab/index.jsx
davemfish: Arbitrarily long, yes. ~167 minutes. The idea is to never expire the "archiving..." message, but change it to "completed" whenever it's done, and then display that for 2 more seconds.

workbench/src/renderer/components/SetupTab/SetupButtons/index.jsx
davemfish: Yes! Sublime knows how to do this just by using `ctrl-/` in a jsx block

davemfish:
Thanks, @emlys , I responded to all your comments, but then I discovered some edge cases that aren't handled well, so I have a bit more work to do. Mainly in handling the new alerts, what happens if more then one button is used before the first is complete; and making sure the alert message doesn't re-appear after it expires just because the data is still in state and the component re-renders.

davemfish:
> Thanks, @emlys , I responded to all your comments, but then I discovered some edge cases that aren't handled well, so I have a bit more work to do. Mainly in handling the new alerts, what happens if more then one button is used before the first is complete; and making sure the alert message doesn't re-appear after it expires just because the data is still in state and the component re-renders.

Hi @emlys I addressed these edge cases by using a `key` prop in the `Expire` component, so React knows whether it's rendering a whole new Alert component, or the same one that already rendered and expired. And I added a test to make sure multiple concurrent alerts can display. Back to you!

Oh and in looking at it, the "save alert" data made a lot more sense to manage in the `SetupTab` component rather than `InvestTab` where I had it originally. We already have the `Portal` from the `SetupTab` over to the `InvestTab` sidebar for exactly this purpose.

emlys (approved):
Thanks @davemfish!

PR #895, opened 2022-02-28
Bugfix/894 specutils kwargs parameter ordering

### LINKED ISSUE #894: **spec_utils.ET0 overriding already-defined attributes
In Seasonal Water Yield, we have a case where we're using the standard ET0 information dict from `spec_utils` like so:

```python
ARGS_SPEC = {
    ...
        "[MONTH]": {
            "about": _(
                "Twelve files, one for each month. File names must "
                "end with the month number (1-12)."),
        **spec_utils.ETO
    }
```

Which results in `ARGS_SPEC['args]['ET0_dir']['[MONTH]']['about'] == 'Map of evapotranspiration values.'`, where `'Map of evapotranspiration values'` is being pulled from `spec_utils.ET0['about']`.

This differs from our original assumption about the behavior of the `**kwargs` syntax, but it turns out is, in fact [what python does per the docs](https://docs.python.org/3/library/stdtypes.html#dict):

> If keyword arguments are given, the keyword arguments and their values are added to the dictionary created from the positional argument. If a key being added is already present, the value from the keyword argument replaces the value from the positional argument.

This came up as the result of a [forums question](https://community.naturalcapitalproject.org/t/error-running-model-valueerror-no-precip-found-for-month-1/2529).

This is probably an issue in other places where we use `**spec_utils` as well.

phargogh:
This PR addresses a minor ordering issue that was causing the default `spec_utils.PRECIP` and `spec_utils.ET0` dicts to override the locally-defined about text in SWY.

I also took the opportunity to rename `spec_utils.ETO` to `spec_utils.ET0` in checking with Stacie about what it's supposed to be.

Fixes #894

## Checklist
- ~~[ ] Tested the affected models' UIs (if relevant)~~


src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -51,11 +51,11 @@
             "contents": {
                 # monthly et0 maps, each file ending in a number 1-12
                 "[MONTH]": {
+                    **spec_utils.ET0,
                     "about": _(
                         "Twelve files, one for each month. File names must "
                         "end with the month number (1-12)."),
```
newtpatrol, line 57: Is it only necessary that the filename end in a supported number? Or does it need to have a preceding underscore, like precip_1.tif? I was under the impression that the latter was required, but maybe not.

phargogh: The current version of the SWY source code only requires that the month index is there.  The underscore is optional.  So `foo1.tif` and `foo_1.tif` are both valid and accepted by the model.

I'm trying to construct a degenerate case, but the regular expression is pretty carefully constructed to make it hard to provide ambiguous filenames.

newtpatrol: Ok, good to know. It would be good to provide an example like "precip1.tif or precipitation_1.tif", but that might be fine to add after the ARGS_SPEC entry in the User Guide.

newtpatrol (approved):

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
emlys: @phargogh, my original thought was that examples would go in the UG so as to keep the `ARGS_SPEC` concise. 

I think ideally the `ARGS_SPEC` would have the minimum info necessary to create a correct input. So in this case, if the file names are matched with a regex, we would specify that regex (or a concise human-readable description of the regex pattern).

But as you point out, it's tricky to phrase a description of even a pretty simple regex. Cases like this make me think maybe it's better to require an exact file name rather than do pattern matching. Most people will likely have to rename their files anyway so that they each end with the correct month number. Is it really helpful to allow an arbitrary prefix? If we required that the files are named `evapotranspiration1.tif`, `evapotranspiration2.tif`, and so on, or `january.tif`, `february.tif`, etc., that's easier to document and easier for users to do correctly.

newtpatrol: Seems fine to have the ARGS_SPEC be minimalist, as long as it contains everything necessary. While I'm sure that it's easier on the coding side to require an exact file name, from a user point of view it would be annoyingly limiting. For example, it's common to run multiple scenarios with different climate change inputs for precip and ET0, and if all of those scenario files had to be named the same generic thing, it would be much more complicated to organize input data and know which scenario the different files represent just by looking at their names. So I strongly advocate for retaining flexibility for the rest of the file names.

phargogh: Ok thanks for weighing in @newtpatrol @emlys !  For the moment let's leave this functionality as-is.  I just resolved a merge conflict in `HISTORY`, so @emlys if you're OK with the addition of these couple of examples then I think this might be ready for one last look-over (and assuming the tests pass).

phargogh:
OK @emlys the conflict is resolved and I think this is ready for one last look-over.

emlys (approved):
Thanks @phargogh! Approving, but with the caveat that I still feel the examples fit better in the user's guide vs. the spec. 

IMO the practical distinction between spec vs. user's guide content is that we can import the spec and use its content in UIs, etc. but the user's guide content must be read on the UG page. And so the spec should contain just what we want to show in a UI. I like brief tooltips that can be read at a glance. That's why I lean towards putting examples in the user's guide. But it's totally fine if you disagree.

dcdenu4:
Not to pile onto this discussion, but I think it's an interesting one. From my point of view if a user is familiar with the workings of a model but struggling with getting the UI to work, like with proper filenames, that should be something the UI addresses. I kind of relate this to how a lot of products now come with a "quick setup" guide along with a larger "user manual". If I'm familiar with the product the "quick setup" should get me through. If I'm looking to learn more and have a complete understanding of the product I can open up the "user manual".

In this case I don't feel like the user should have to dig out the User Guide to know how to format an input path. That being said, the original description:
>"Twelve files, one for each month. File names must end with the month number (1-12)."

seems to say exactly what you need to know. I wonder if validation should be responsible with handling any confusion here, where if a file path isn't formatted correctly it can prompt with examples. Or perhaps a more streamlined about text like:
>"Twelve files, one for each month. File names must end with the month number (1-12). Ex: 'et0_1.tif', 'evapotranspiration1.tif'"

Anyway, thanks for listening to my musings about this topic!

phargogh:
I'm sorry for having this PR open for so long, everyone!  It seems like there's still some discussion to be had about where examples should live and/or the degree of customization that the files should have.

I think I'd like to resolve this PR in particular, but I've created https://github.com/natcap/invest/issues/959 for the continued discussion and to arrive at a decision.  Does that sound OK for everyone?

emlys:
Sure, I'll go ahead and merge!

PR #896, opened 2022-03-01
Replace "upstream"/"downstream" with "upslope"/"downslope"

emlys:
## Description
@jagoldstein requested via Slack that we replace "upstream" and "downstream" with "upslope" and "downslope" in the documentation, because the terms apply whether or not a pixel is actually in a stream. I also replaced these throughout the code and comments since that distinction might be helpful when we're thinking about the algorithms.

## Checklist

### DIFF
--- a/src/natcap/invest/delineateit/delineateit.py
+++ b/src/natcap/invest/delineateit/delineateit.py
@@ -152,7 +152,7 @@ def execute(args):
             nearest stream pixel.  If ``True``, ``args['flow_threshold']``
             and ``args['snap_distance']`` must also be defined.
         args['flow_threshold'] (int):  The number of upslope cells that must
-            into a cell before it's considered part of a stream such that
+            flow into a cell before it's considered part of a stream such that
             retention stops and the remaining export is exported to the stream.
             Used to define streams from the DEM.
         args['snap_distance'] (int):  Pixel Distance to Snap Outlet Points
--- a/src/natcap/invest/delineateit/delineateit.py
+++ b/src/natcap/invest/delineateit/delineateit.py
@@ -151,7 +151,7 @@ def execute(args):
         args['snap_points'] (bool): Whether to snap point geometries to the
             nearest stream pixel.  If ``True``, ``args['flow_threshold']``
             and ``args['snap_distance']`` must also be defined.
-        args['flow_threshold'] (int):  The number of upstream cells that must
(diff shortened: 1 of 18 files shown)


jagoldstein:
thanks @emlys ! To be clear, there may be a small handful of mentions of "downstream" that are correct in the UG. Here's an example from [SDR](http://releases.naturalcapitalproject.org/invest-userguide/latest/sdr.html#sediment-retention-services): "_This quantifies the difference in erosion reaching a stream, based on the changes in land cover/climate/etc present in the scenario, which provides a way of evaluating impacts to downstream uses such as reservoirs and drinking water._"

I was specifically referring to the definitions in `ARG_SPECS`, such as for Threshold Flow Accumulation. But either way, a sweeping replacement of these terms would still make the UG more accurate than before.

emlys:
thanks @jagoldstein! I did not make any changes to the user's guide content yet. I meant to ask if you would like to make those UG edits since you're familiar with the appropriate contexts for using "slope" vs "stream"? All the changes I made here in the code & comments seem pretty clearly a "slope" context.

jagoldstein:
no problem @emlys! I will take care of those terms in the body of the UG (which are NOT pulled from `ARG_SPECS`)

src/natcap/invest/delineateit/delineateit.py
```diff
@@ -151,7 +151,7 @@ def execute(args):
         args['snap_points'] (bool): Whether to snap point geometries to the
             nearest stream pixel.  If ``True``, ``args['flow_threshold']``
             and ``args['snap_distance']`` must also be defined.
-        args['flow_threshold'] (int):  The number of upstream cells that must
+        args['flow_threshold'] (int):  The number of upslope cells that must
```
davemfish, line 154: Reading this sentence, there's a word missing: "flow"?

davemfish (changes requested):
Looks good! I noticed one missing word in a docstring we might as well fix now. And the UG builds aren't building because of arg key changes I think. Do we need a change on the UG side?

davemfish:
Oops, this is still in Draft, sorry! Feel free to ask me to review again when you're ready.

emlys:
@davemfish this is ready now! I added that missing word and updated the user's guide.

davemfish (approved):

Issue #898, opened 2022-03-09
Auto-PR into release branches fails when there are no release branches


phargogh:
When we merge a PR into main, the workflow at `workflows/auto-pr-from-main-into-releases.yml` is executed in order to open up PRs to merge the changes just merged into main into each open release branch.  It turns out that if there are no release branches (as we have now), then the workflow fails.  This should be fixed so that the workflow just exits gracefully, without a failure, since this is a perfectly acceptable repo state to be in.

Example failure: https://github.com/natcap/invest/runs/5487825466?check_suite_focus=true

emlys:
This was fixed by #899.

PR #899, opened 2022-03-09
Auto-pr workflow - handle the case where there are no release branches

phargogh:
This case is now logged and the shell will exit with 0 exit status,
avoiding a failure.  RE:#898

I've tested the requisite bash locally, though not in GHA directly.

## Description
Fixes # 898

## Checklist
- ~~[ ] Updated HISTORY.rst (if these changes are user-facing)~~
- ~~[ ] Tested the affected models' UIs (if relevant)~~


emlys (approved):
This makes sense to me! The checks are failing, but they're failing on main too.

PR #903, opened 2022-03-11
Use create-dmg instead of dmgbuild

### LINKED ISSUE #893: DMG fails to build on `macos-latest` runner
As of a few months ago, the `macos-latest` Github Actions runner uses macOS Big Sur (11). Our DMG building utility `dmgbuild` fails to add the license on Big Sur. It does work on Catalina (10.15). We've pinned the runner to 10.15, but ideally we'd keep up with the latest versions. 

This is a known issue with `dmgbuild` but the package doesn't get a lot of attention so this may not get fixed (https://github.com/al45tair/dmgbuild/issues/29). And the maintainer of `dmgbuild` is stepping down (https://pypi.org/project/dmgbuild/). We might need to find a different way to build the DMG in the near future.

emlys:
## Description
Fixes #893

## Checklist

### DIFF
--- a/.github/workflows/binary-applications.yml
+++ b/.github/workflows/binary-applications.yml
@@ -216,6 +216,7 @@ jobs:
 
             - name: Build binaries
               run: |
+                  brew install create-dmg
                   make install
                   make mac_dmg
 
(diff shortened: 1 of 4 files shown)


phargogh (changes requested):
The new DMG looks great!  And it's working on OSX 12 (Monterey).  And thanks for removing those old linux scripts.

Could you also remove the `dmgbuild` line from `requirements-dev.txt`?

I agree, the test failures look unrelated to these changes.

emlys:
Thanks for trying it out @phargogh! Back to you

phargogh (approved):
I'm not too worried about the builds not running (though that is strange), since the builds had passed before except for the issues noted in #905 .  Thanks!

PR #906, opened 2022-03-16
Adding missing f path to the sediment deposition target path list.

phargogh:
## Description

This corrects a minor issue in the SDR task graph where `f.tif` was not in the target path list for the sediment deposition task, though it should have been.

Such a minor bug and there's no issue in the tracker for it, so none are linked.

## Checklist


emlys (approved):
Thanks @phargogh!

PR #910, opened 2022-03-21
Feature/908 delineateit create watersheds with ws id column

### LINKED ISSUE #908: Make delineateit create watersheds with a ws_id column
When watersheds are created in delineateit, they have the feature attribute `ID` to indicate the watershed ID.  For consistency with our routed models which use `ws_id` as their watershed ID fieldnames, it would be nice to just use the `ws_id` fieldname instead.

[This came up on the forums recently](https://community.naturalcapitalproject.org/t/questions-about-running-delineatelt/2577/8?u=jdouglass), where a user didn't know to rename the column from delineateit.

phargogh:
## Description
This PR adds a `ws_id` column in the geometry preprocessing step or the pour point detection step, whichever is selected.  If a `ws_id` field is already in the vector, its values are overwritten.

There is no reference to an identifying field value in the user's guide, so I have not updated the UG.

I did run this through the UI with the sample data and you can see the new attribute and value there.

Fixes #908 

## Checklist


tests/test_delineateit.py
```diff
@@ -487,16 +526,21 @@ def test_check_geometries(self):
...
-            skip_invalid_geometry=True
-        )
+        logger = logging.getLogger('natcap.invest.delineateit.delineateit')
+        with capture_logging(logger) as log_records:
+            delineateit.preprocess_geometries(
+                outflow_vector_path, dem_raster_path, target_vector_path,
+                skip_invalid_geometry=True
+            )
+        self.assertEqual(len(log_records), 6)
+        self.assertIn('named "ws_id". Field values will be overwritten',
```
emlys, line 542: This message could be made a module-level variable in `delineateit` and referenced here. But it's not a lot of duplication.

emlys (commented):
Thanks @phargogh! This definitely solves the problem. I wonder if it might be more direct to add the `ws_id` field directly to the watersheds vector, rather than to the outlets vector first. That could be done in the pygeoprocessing delineation function, or afterwards in delineateit if it should be specific to this model. Let me know what you think.

I know you're adding a test for this in https://github.com/natcap/pygeoprocessing/pull/250, but would it make sense to test that the `ws_id` ends up in the watersheds vector here too? If I'm reading it correctly, only the outlets vector is tested.

HISTORY.rst
```diff
@@ -37,6 +37,11 @@ Unreleased Changes
 * General
     * Updating the ``pyinstaller`` requirement to ``>=4.10`` to support the new
       ``universal2`` wheel architecture offered by ``scipy>=1.8.0``.
+* DelineateIt
+    * Watersheds delineated with this tool will now always have a ``ws_id``
+      column containing integer watershed IDs for easier use within the routed
+      InVEST models.  Existing ``ws_id`` field values will be overwritten if
+      they are present.
```
emlys, line 44:
```suggestion
      InVEST models.  Existing ``ws_id`` field values in the outlets vector
      will be overwritten if they are present.
```

tests/test_delineateit.py
phargogh: Yep, sure can!  Added in a055e59ce

phargogh:
>I wonder if it might be more direct to add the ws_id field directly to the watersheds vector, rather than to the outlets vector first. That could be done in the pygeoprocessing delineation function, or afterwards in delineateit if it should be specific to this model. Let me know what you think.

Yeah, I thought about that as well and started out doing exactly that. I ended up not creating a `ws_id` field in pygeoprocessing because the `ws_id` attribute is an InVEST construct and not something that's core to the idea of watershed delineation.

> I know you're adding a test for this in https://github.com/natcap/pygeoprocessing/pull/250, but would it make sense to test that the ws_id ends up in the watersheds vector here too? If I'm reading it correctly, only the outlets vector is tested.

Yep, makes sense!   Added in ed0d6ad2e

emlys:
Hey @phargogh, it looks like a delineateit test is failing. Not sure if that's related or one of the test issues that's cropped up in the past weeks?

phargogh:
@emlys I'm unsure why that test is failing right now, but I've merged in the latest changes from `main` so let's see if it's still happening on GHA.  I'm not getting the test failure locally.

phargogh:
Oop, never mind, I can reproduce the test failure.  The CLI tests have some side effects for logging which are messing up that delineateit test.

emlys (approved):
Looks good to me @phargogh! It would be nice to update the user's guide to mention this feature, but since the old `point_id` field was also undocumented, no need to hold up the PR over it.

Issue #905, opened 2022-03-16
SDR tests are failing with gdal=3.4.2


emlys:
Three SDR tests are failing on github actions:
```
test_base_regression: AssertionError: these key (actual/expected) errors occured: {'usle_tot': (13.90210914612, 14.25030517578), 'sed_export': (0.55185163021, 0.60502111912), 'sed_dep': (8.80612564087, 9.05251502991)}

test_drainage_regression: AssertionError: these key (actual/expected) errors occured: {'sed_export': (0.97192692757, 1.02959537506), 'usle_tot': (12.68887424469, 12.97211265564)}

test_regression_with_undefined_nodata: AssertionError: these key (actual/expected) errors occured: {'sed_export': (0.55185163021, 0.60502111912), 'usle_tot': (13.90210914612, 14.25030517578)}
```
These passed when I ran them locally. Possibly an updated dependency?

emlys:
Confirmed that the tests fail with GDAL 3.4.2 and pass with 3.4.1. The differences are larger than I'd like to attribute to floating point imprecision.

emlys:
The difference is coming from a change in the aligned LULC raster. Something in the alignment process must have changed. 
Original LULC:
<img width="200" alt="image" src="https://user-images.githubusercontent.com/43770515/161343659-de0a60a5-dcf5-4581-9b8b-51d03dcb0ba3.png">
Aligned LULC using GDAL 3.4.1:
<img width="200" alt="image" src="https://user-images.githubusercontent.com/43770515/161343795-57796cf8-1526-4988-b142-c15bea3f9701.png">
Aligned LULC using GDAL 3.4.2:
<img width="200" alt="image" src="https://user-images.githubusercontent.com/43770515/161343947-8cf01cfe-0677-4c33-80f1-768f39535333.png">

The 3.4.2 version actually looks a lot more accurate (maybe identical) to the original. In the 3.4.1 version lots of pixels are switched to LULC class 2 (gray).

emlys:
This comes down to [this undocumented change](https://github.com/OSGeo/gdal/commit/db9ea60631eda5ef881ae75dc3355f42734fb38c) released in 3.4.2. From what I can tell, nearest neighbor is being used instead of `mode`, because of the new setting `GDAL_WARP_USE_TRANSLATION_OPTIM` that defaults to `YES`. In this particular SDR case, the warp operation is just translating along pixel boundaries (?). And because of that, it's being "optimized" to use `near` instead of `mode`.

[...]

PR #914, opened 2022-03-22
Task/913 sdr clamp sed deposition values

### LINKED ISSUE #913: Clamp sed_deposition values <0 to 0.
A user on the forums is noticing SDR `sed_deposition.tif` values that are numerically 0 but actually ending up as very small negative numbers.  It'll save us time in the future if we just clamp values less than 0 to 0.

phargogh:
## Description

As it says on the tin: just clamping a couple of values to 0 when they go negative (which they shouldn't, except for reasonable numerical noise).

Fixes #913

## Checklist


davemfish:
Hey @phargogh this looks very straightforward to me. I'm inclined to just hold on it though until the SDR tests pass again due to #905 . That way it becomes very obvious if these changes fail any tests and require updated regression data. Lemme know if that's not a good idea for any reason.

phargogh:
@davemfish yep, I completely agree.  Thanks for taking a look!

davemfish (approved):
Everything passes!

PR #918, opened 2022-03-24
Workbench sampledata link-checks bugfix

### LINKED ISSUE #902: Check sampledata registry links is failing
The sample data registry links have been returning 404 recently. This is causing our github actions workflow to fail.

davemfish:
There were a few issues going on here:
1. The link-check test needed to be moved after the deploy step in GHA, so that the files exist at the urls being checked.
2. The reason that test seemed to fail intermittently might just be because it wasn't running at all in PRs - since we don't "deploy" from PRs.
3. The prior build process was overwriting `sampledata_registry.json` in place. Since that's a git-tracked file, modifying it alters the `$VERSION` variable derived by `setuptools_scm` in the `Makefile`, which in turn alters the upload path to our bucket. So in the end the URLs in `sampledata_registry.json` did not match where the data actually lives after deploy.

Solutions:
1. easy fix, re-ordered the workflow steps
2. Not a problem, this should still not run in PRs
3. I updated the process so that a git-tracked file is never overwritten during the build process. So now `sampledata_registry.json` serves as a default registry - it's git-tracked, and the application uses it unless another non-tracked file (`sampledata_registry_production.json`) exists. The production file is created -- using the other as a template -- during the build process (`make sampledata`). 

Basically, the `production` file is used in production, and the original registry is used in devMode. I don't think it matters much if the original is kept current for every release/dev-build. It just needs to be functional and it should be updated whenever a sampledata filename changes, or a new model is added.

Fixes #902 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -512,13 +512,6 @@ jobs:
         with:
           node-version: 14.x
 
-      - name: Check sampledata registry links
-        if: github.event_name != 'pull_request' # no artifacts were deployed in a PR
-        working-directory: workbench
-        run: |
-          yarn add jest
-          yarn run test-sampledata-registry
-
       - name: Upload sample data artifact
         uses: actions/upload-artifact@v2
         with:
@@ -538,3 +531,10 @@ jobs:
       - name: Deploy artifacts to GCS
         if: github.event_name != 'pull_request'
         run: make deploy
+
+      - name: Check sampledata registry links after deploy
+        if: github.event_name != 'pull_request' # no artifacts were deployed in a PR
+        working-directory: workbench
+        run: |
+          yarn add jest
+          yarn run test-sampledata-registry
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -461,7 +461,7 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Workbench-${{ runner.os }}-binary
-          path: workbench/dist/*.{{ matrix.binary-extension }}
+          path: workbench/dist/*.${{ matrix.binary-extension }}
 
       - name: Upload user's guide artifact (Windows)
         if: matrix.os == 'windows-latest'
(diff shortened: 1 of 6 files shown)


emlys (commented):
Thanks for tackling this @davemfish! This looks like a good solution. But I'm a little concerned about the added complexity of having two sample data registry files. Sorry I can't remember the background of `sampledata_registry.json` - could you remind me why we check in the URL for every model? (vs. building it from a base url, model name, and version)?

davemfish:
> could you remind me why we check in the URL for every model? (vs. building it from a base url, model name, and version)?

In one respect it comes down to wanting to know the filesize of each zip. To render, the component needs to know the url and filesize for each zip. It could construct the urls and make requests to find out the filesize (I assume) every time the component mounts. But that would be a lot more network requests, and the data never changes. So instead we create all the data the component needs and package it with the app. 

But also, as of this PR, you don't need to think of the url as being checked-in to the repo. It's really just a template that is checked in. So there will be a lot fewer commits to the checked-in version now than there were in the past. For production, the url for a zip is constructed dynamically from the base url, model name, and version...it just happens during the build process rather than during runtime. 

> But I'm a little concerned about the added complexity of having two sample data registry files

I agree but I kinda failed to come up with anything better. I'm open to better ideas!

emlys:
What about storing the file size data alongside the files in the GCP bucket? Making a separate request to check the size of each sampIe data file could slow things down, but a single request to get a file sizes JSON should be fine.

To handle the dev mode case, we could `try/catch` the request to the file sizes JSON. If it doesn't exist, display "size unknown" for each data set. And the download links would not work. If we need the download links to work in dev mode, maybe storing a single fallback URL template would be enough.

I guess what I'm hung up on is keeping the whole `sampledata_registry.json` with its urls and file sizes that are out-of-date, but still need occasional maintenance, and aren't obviously just placeholders. Maybe some simpler (or obviously fake) placeholder values would be enough?

emlys:
> And that last bullet is a sticking point for doing away with the registry, at least right now.

True, we do need to maintain a list of sample data files until #920 is addressed. And I think it's fine for that list to live in a JSON, if that's easiest. It's specifically the `filesize` and `url` attributes in that JSON that I don't think are necessary. But you're right that we can get this critical fix out now, and revisit later.

emlys (approved):

PR #932, opened 2022-04-05
Bugfix/UI server 928 shutdown

### LINKED ISSUE #928: UI-SERVER: shutdown method is deprecated
Our approach to shutting down the dev server is now deprecated: `request.environ.get('werkzeug.server.shutdown')()`

Also I'm not sure why I ever set things up this way, to have the thread running the flask app try to shut its own server down. It's much easier to have the parent process be responsible for this.

Not shutting things down properly causes workbench tests to hang, and other problems.

davemfish:
Our method for shutting down the flask server is deprecated. This PR makes sure the python process running the flask app and server is shutdown before the Workbench application quits.

Fixes #928 

## Checklist

### DIFF
--- a/src/natcap/invest/ui_server.py
+++ b/src/natcap/invest/ui_server.py
@@ -24,25 +24,10 @@ PYNAME_TO_MODEL_NAME_MAP = {
 }
 
 
-def shutdown_server():
-    """Shutdown the flask server."""
-    func = request.environ.get('werkzeug.server.shutdown')
-    if func is None:
-        raise RuntimeError('Not running with the Werkzeug Server')
-    func()
-
-
 @app.route('/ready', methods=['GET'])
 def get_is_ready():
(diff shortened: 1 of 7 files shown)


davemfish:
@emlys I requested James on this one mainly because I have you in mind for the next couple. But feel free to take a look since you were looking at this today!

The problem -- and the difference between these changes and the prior version that was working -- seemed to be not awaiting the flask shutdown function. So I think the Workbench was exiting, or trying to, before the python process was cleaned up. And depending on the context, that left one or both of those processes open indefinitely.

phargogh (approved):
Looks good to me, and really interesting that the child process would stick around after the main process dies ... that part is bizarre to me.  In any case, the changes here make sense, and thanks for wading through it!

phargogh:
Model tests are failing on the SDR issue mentioned in #905, unrelated to this PR.  Merging!

PR #944, opened 2022-04-13
UNA - Implementing split greenspace

phargogh:
## Description

This PR implements the optional split greenspace feature of UNA, which is mostly a slight refactor of the 2SFCA procedure described in the [UNA design doc](https://docs.google.com/document/d/1lqx75-S605lLxingVjd_yVSGt1tssEFp/edit).

There are a couple of unresolved questions about this feature that I'm waiting to hear back from the science team about, specifically:

1. The design doc describes that `Ai,r` (greenspace supply per greenspace class) is output, but it is not clear if it's important for one such output to be produced for each greenspace classification or if these can be aggregated by a shared search radius as they are now, and are in the base model.
2. The way in which the user defines their various radii for split greenspace doesn't feel great to me in that there's some duplication of the input between a table and a separate UI element.  This may change in the future, and I'd like to ask the science team what they think about it.

This PR also includes all the workbench changes, so sorry for the big diff!  If this is impractical, I'll be happy to do a separate PR that's just `main` into `feature/urban-nature-access`.

## Checklist
- [X] Tested the affected models' UIs (if relevant)


emlys:
Hey @phargogh, it looks like there's a whole bunch of other changes in the diff not related to the purpose of this PR. Maybe need to separately merge `main` into `feature/urban-nature-access`?

phargogh:
Thanks for taking a look @emlys !  I've created https://github.com/natcap/invest/pull/953 for merging `main` into `feature/urban-nature-access`.  Once that's in, this review should become a lot easier.

emlys:
Hey @phargogh I'm still seeing 225 changed files in the diff even though #953 was merged. Do you know if there's a way to refresh the diff?

emlys:
Awesome thank you!

emlys:
> The way in which the user defines their various radii for split greenspace doesn't feel great to me in that there's some duplication of the input between a table and a separate UI element.

I agree. If someone wants to use the same radius for every LULC, it's easy enough to enter the same value for each row in their table. The separate radius input seems unnecessary.

src/natcap/invest/urban_nature_access.py
```diff
@@ -254,103 +282,133 @@ def execute(args):
...
-    convolved_population_task = graph.add_task(
-        _convolve_and_set_lower_bounds_for_population,
+    preprocess_attr_table_task = graph.add_task(
+        _preprocess_lulc_attribute_table,
         kwargs={
-            'signal_path_band': (file_registry['aligned_population'], 1),
-            'kernel_path_band': (file_registry['kernel'], 1),
-            'target_path': file_registry['convolved_population'],
-            'working_dir': intermediate_dir,
+            'source_attr_table_path': args['lulc_attribute_table'],
```
emlys, line 288: I like how you're using all the args as `kwargs` even though they don't have to be. It's nice to see the arg variable name alongside its value.

src/natcap/invest/urban_nature_access.py
```diff
@@ -254,103 +282,133 @@ def execute(args):
...
-            greenspace_reclassification_task,
-        ])
-
-    convolved_greenspace_population_ratio_task = graph.add_task(
-        pygeoprocessing.convolve_2d,
+        target_path_list=[file_registry['attribute_table']],
+        priority=10,  # higher-than-normal priority to address sooner.
+        task_name='Preprocess the attribute table')
+
+    preprocess_attr_table_task.join()
```
emlys, line 296: Does setting the `priority=10` still have any effect if we `join` the task first?

src/natcap/invest/urban_nature_access.py
```diff
@@ -254,103 +282,133 @@ def execute(args):
...
+            _reclassify_greenspace_area,
+            kwargs={
+                'lulc_raster_path': file_registry['aligned_lulc'],
+                'lulc_attribute_table': file_registry['attribute_table'],
+                'target_raster_path': greenspace_pixels_path,
+                'only_these_greenspace_codes': set(matching_landuse_codes),
+            },
+            target_path_list=[greenspace_pixels_path],
+            task_name=(
+                f'Identify area of greenspace in pixels {search_radius_m}'),
```
emlys, line 320: I'm not sure what this is saying, maybe missing a word?

src/natcap/invest/urban_nature_access.py
```diff
@@ -459,6 +517,147 @@ def execute(args):
     LOGGER.info('Finished Urban Nature Access Model')
 
 
+def _sum_rasters(greenspace_supply_list, target_sum_path):
```
emlys, line 520: There are raster sum functions implemented in other models that you could import and reuse here, but I don't know if they're just as good. Given that these type of functions are duplicated a lot in invest, adding this one isn't a big deal. https://github.com/natcap/pygeoprocessing/issues/235

src/natcap/invest/urban_nature_access.py
```diff
@@ -459,6 +517,147 @@ def execute(args):
...
+
+    pygeoprocessing.raster_calculator(
+        [(path, 1) for path in greenspace_supply_list], _sum_op,
+        target_sum_path, gdal.GDT_Float32, FLOAT32_NODATA)
+
+
+def _reclassify_greenspace_area(
+        lulc_raster_path, lulc_attribute_table, target_raster_path,
+        only_these_greenspace_codes=None):
+    """Reclassify LULC pixels into the greenspace area they represent.
```
emlys, line 560: This is a little unclear to me. Maybe another line saying something like "Greenspace pixels' value will be the pixel area, and non-greenspace pixels' value will be 0." (If I'm understanding that correctly)

src/natcap/invest/urban_nature_access.py
```diff
@@ -459,6 +517,147 @@ def execute(args):
...
+        default_radius (number): The default radius to use when
+            "search_radius_m" is not defined.
+        target_attr_table_path (string): The location on disk where the
+            preprocessed attribute table is written.
+
+    Returns:
+        ``None``
+    """
+    attr_dataframe = utils.read_csv_to_dataframe(
+        source_attr_table_path, to_lower=True, sep=None, engine='python',
```
emlys, line 636: `sep=None` and `engine='python'` are the defaults, but feel free to leave them if you think it helps to be explicit here.

src/natcap/invest/urban_nature_access.py
```diff
@@ -617,63 +816,72 @@ def _greenspace_supply_demand_op(greenspace_budget, population):
...
+        if population_nodata is not None:
+            valid_pixels &= ~numpy.isclose(
+                convolved_population, population_nodata)
+
+        if greenspace_nodata is not None:
+            valid_pixels &= ~numpy.isclose(greenspace_area, greenspace_nodata)
+
+        # If the population in the search radius is numerically 0, the model
+        # specifies that the ratio should be set to the greenspace area.
+        population_close_to_zero = numpy.isclose(convolved_population, 0.0)
```
emlys, line 871: The default `isclose` tolerances are rtol=1e-05, atol=1e-08. And I think rtol would actually be 0 because the compared value is 0. I wonder if you might want an even higher atol since in the context of this model, 0.00000008 people is probably negligible. Maybe even as much as 0.5 since that would already double the "effective greenspace" (as I'm understanding this ratio).

tests/test_urban_nature_access.py
```diff
@@ -19,6 +21,78 @@
...
+        'greenspace_demand': 100,  # square meters
+        'admin_unit_vector_path': os.path.join(
+            workspace, 'admin_units.geojson'),
+    }
+
+    random.seed(-1)  # for our random number generation
+    population_pixel_size = (90, -90)
+    population_array_shape = (10, 10)
+    population_array = numpy.array(
+        random.choices(range(0, 100), k=100),
```
emlys, line 44: Neat!

tests/test_urban_nature_access.py
```diff
@@ -194,129 +268,199 @@ def test_greenspace_budgets(self):
...
+            population_array = pygeoprocessing.raster_to_numpy_array(
+                args['population_raster_path'])
+            numpy.testing.assert_allclose(
+                (expected_values['Pund_adm'] + expected_values['Povr_adm']),
+                population_array.sum())
+
+            admin_vector = None
+            admin_layer = None
+
+            shutil.rmtree(args['workspace_dir'])
```
emlys, line 412: Should this go in a `tearDown`?

emlys (changes requested):
@phargogh here's my first round of comments! I know you said you'd have a few more changes to make so I'll look over again when those are ready.

tests/test_urban_nature_access.py
dcdenu4, line 44: `random.choices` - cool!

tests/test_urban_nature_access.py
```diff
@@ -194,129 +268,199 @@ def test_greenspace_budgets(self):
...
+
+        # The split greenspace feature requires an extra column in the
+        # attribute table.
+        attribute_table = pandas.read_csv(args['lulc_attribute_table'])
+        new_search_radius_values = {
+            value: 30*value for value in range(1, 10, 2)}
+        attribute_table['search_radius_m'] = attribute_table['lucode'].map(
+            new_search_radius_values)
+        attribute_table.to_csv(args['lulc_attribute_table'], index=False)
+        # TODO: check with science team about enabling split greenspace.
```
dcdenu4, line 427: That's interesting. I'm not plugged in to the details here, but I wonder if there needs to be an explicit enable/disable switch? With the way I think you have it set up, if the table includes the search radius column with different radii, then that is indicating a `split greenspace`. That seems right to me, with a LOGGER INFO indication of which mode is being used. But yea, it'd have to be articulated well in the Uses Guide.

dcdenu4 (changes requested):
Hey @phargogh , this mostly looks good to me! I just had a few questions and comments. Sorry for the delay on this.

src/natcap/invest/urban_nature_access.py
```diff
@@ -62,6 +63,20 @@
...
                     ),
+                },
+                'search_radius_m': {
+                    'type': 'number',
+                    'units': u.meter,
+                    'about': (
+                        'The distance in meters to use as the search radius '
+                        'for this type of greenspace.  Values are only '
+                        'required in this column if the value of greenspace '
+                        'is 1.  If any greenspace values are left blank, the '
```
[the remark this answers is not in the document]
phargogh: I addressed this particular bit of whitespace in a larger refactor in b1542ce17

src/natcap/invest/urban_nature_access.py
```diff
@@ -104,17 +119,22 @@
...
             'required': False,
             'options': {
-                'dichotomy': {"description": _("dichotomy")},
-                'exponential': {"description": _("exponential")},
-                'gaussian': {"description": _("gaussian")},
-                'density': {"description": _("density")},
+                'dichotomy': {'display_name': 'dichotomy'},
+                'exponential': {'display_name': 'exponential'},
+                'gaussian': {'display_name': 'gaussian'},
+                'density': {'display_name': 'density'},
```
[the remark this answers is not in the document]
phargogh: I took a first stab at this in f611e112b.  I think the choices are pretty clear, though the user's guide should (and the design doc does) include the math for each function.

src/natcap/invest/urban_nature_access.py
```diff
@@ -104,17 +119,22 @@
...
             },
             'about': '',  # TODO
```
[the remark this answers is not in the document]
phargogh: Yep!  Added in f611e112b

src/natcap/invest/urban_nature_access.py
```diff
@@ -62,6 +63,20 @@
...
+                'search_radius_m': {
+                    'type': 'number',
+                    'units': u.meter,
+                    'about': (
+                        'The distance in meters to use as the search radius '
+                        'for this type of greenspace.  Values are only '
+                        'required in this column if the value of greenspace '
+                        'is 1.  If any greenspace values are left blank, the '
+                        'user-defined "search radius" input will be used as '
+                        'a default value.  If this column is missing, all '
```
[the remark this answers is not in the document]
phargogh: In talking with Roy and Hongxiao (and also with your feedback), this column is now required.

src/natcap/invest/urban_nature_access.py
```diff
@@ -104,17 +119,22 @@
             'name': 'search radius',
             'units': u.m,
             'expression': "value > 0",
-            'about': "",  # TODO, will know more about this when I implement.
+            'about': (
```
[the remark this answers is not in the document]
phargogh: This input is now removed.

src/natcap/invest/urban_nature_access.py
```diff
@@ -210,15 +236,17 @@ def execute(args):
     }
     # Since we have these keys defined in two places, I want to be super sure
     # that the labels match.
-    assert sorted(kernel_types.keys()) == (
+    assert sorted(kernel_creation_functions.keys()) == (
```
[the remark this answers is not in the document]
phargogh: Yeah, I thought a bit about this and I think I'd prefer to keep them separate, just with an assertion to guard against programmer error.  I feel like keeping them separate ensures a separation of concerns: ARGS_SPEC info with the ARGS_SPEC, the mapping of kernel functions here in `execute`.

That being said, I did use a few more constants in 2ddbbe6b3 to represent the kernel keys (which cuts out one type of duplication).  The assertion is still useful here though, so I've left it in.

src/natcap/invest/urban_nature_access.py
```diff
@@ -210,15 +236,17 @@ def execute(args):
...
     if 'kernel_type' not in args:
         kernel_type = 'dichotomy'
-    elif args['kernel_type'] not in kernel_types:
+        LOGGER.info(
+            'args["kernel_type"] not provided; defaulting to {kernel_type}')
+    elif args['kernel_type'] not in kernel_creation_functions:
         raise ValueError(
             f'Kernel type "{args["kernel_type"]}" is not recognized. '
-            f"Must be one of {', '.join(kernel_types.keys())}")
+            f"Must be one of {', '.join(kernel_creation_functions.keys())}")
```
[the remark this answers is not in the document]
phargogh: Yep, sure does!  I've removed this check in b2795a8830157bbbd2387fe80cc3d5f56c4179af with the assumption that validation will handle it.  All cases are covered in ARGS_SPEC, so this shouldn't be hard to cover.

src/natcap/invest/urban_nature_access.py
phargogh: Thanks!  I've found it really helpful and clear when writing this out and it helps cut down on issues where I accidentally pass in the wrong thing.

The next nice-to-have would be to have my editor populate the kwargs ... not there yet.

src/natcap/invest/urban_nature_access.py
phargogh: In this case, the issue is addressed by no longer needing to preprocess the table since the `search_radius_m` column is now required.

In the general case, though, yes, setting `priority=10` could still have an effect when the task it `join`ed.  Imagine a case where `n_workers > 0`, which means that the order of task execution is not guaranteed.  It's possible for us to get into a case where all of the other tasks that were added other than this one task to execute before `preprocess_attr_table_task`.  Increasing the priority means that this task will take priority if another lower-priority task also has satisfied its dependencies.

src/natcap/invest/urban_nature_access.py
phargogh: Ha, this really isn't readable at all, is it?  Greatly simplified in  cfa5bbdd9

src/natcap/invest/urban_nature_access.py
phargogh: Yep, I was previously using NDR's implementation and then switched (I don't remember why).  I just switched back, which seems fine.

But yes, I'd really rather see this addressed in a shared utility.

src/natcap/invest/urban_nature_access.py
phargogh: Added a description along these lines in 0dacced0e

src/natcap/invest/urban_nature_access.py
phargogh: I think this function has been removed because this function is no longer needed.

tests/test_urban_nature_access.py
phargogh: Well, really the whole test needed a refactor because the core model now also requires the search radius column, so the line is now removed!  b5a46055a

src/natcap/invest/ui/urban_nature_access.py
```diff
@@ -0,0 +1,71 @@
...
+            **_ui_keys('admin_unit_vector_path'))
+        self.add_input(self.admin_units)
+
+        self.greenspace_demand = inputs.Text(
+            **_ui_keys('greenspace_demand'))
+        self.add_input(self.greenspace_demand)
+
+        self.search_radius = inputs.Text(
+            **_ui_keys('search_radius'))
+        self.add_input(self.search_radius)
```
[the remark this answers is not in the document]
phargogh: Yep, the science team agrees!  And I've removed this UI input in 1dbaa6d68

src/natcap/invest/urban_nature_access.py
```diff
@@ -254,103 +282,133 @@ def execute(args):
...
-        target_path_list=[file_registry['greenspace_population_ratio']],
-        dependent_task_list=[
-            convolved_population_task,
-            greenspace_reclassification_task,
-        ])
-
-    convolved_greenspace_population_ratio_task = graph.add_task(
-        pygeoprocessing.convolve_2d,
+        target_path_list=[file_registry['attribute_table']],
+        priority=10,  # higher-than-normal priority to address sooner.
```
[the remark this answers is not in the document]
phargogh: When it's useful, it's useful!

src/natcap/invest/urban_nature_access.py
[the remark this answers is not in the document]
phargogh: Yeah, @emlys asked the same question, so I hope I offered a good-enough explanation above: https://github.com/natcap/invest/pull/944#discussion_r864411498

But also, we no longer need this join!  So it (and the priority) have been removed.

src/natcap/invest/urban_nature_access.py
```diff
@@ -617,63 +816,72 @@ def _greenspace_supply_demand_op(greenspace_budget, population):
     return supply_demand
 
 
-def _greenspace_population_ratio(
-        greenspace_area, convolved_population, greenspace_nodata,
-        population_nodata):
-    """Calculate the greenspace-population ratio R_j.
+def _calculate_greenspace_population_ratio(
```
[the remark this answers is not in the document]
phargogh: Sure should!  036520324

tests/test_urban_nature_access.py
phargogh: Yeah, there might not need to be any more, if we assume that the presence of multiple search radii triggers the split greenspace mode!

I'll need to work on this a little more and I'll get back to you soon.

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, that's a really interesting point.  Let me think about this for a bit and I'll get back to you soon.

phargogh:
Thanks a ton for your feedback @emlys and @dcdenu4 !  I've responded to most of your comments, but I've got a little more thinking to do on two of your comments:

- [ ] https://github.com/natcap/invest/pull/944#discussion_r859222795
- [ ] https://github.com/natcap/invest/pull/944#discussion_r862898760

I'll get back to you soon and re-request your review.

src/natcap/invest/urban_nature_access.py
phargogh: Yep, I agree, I think a higher atol would be a really good idea.

The main reason is that the population raster coming in to this function is coming out of a non-normalized convolution, and so even within float64 space it wouldn't be that hard for the accumulated numerical error to exceed 1e-8.

In practice, I suspect this case will mostly not be triggered, given this is an urban model and that a single low-population pixel would need to be on the very fringes of a search kernel to get such a low number.  I tried running this on the Paris sample data with atols ranging from 1e-8, 1e-7 ... 1e-1, 1 with no detectable differences between the outputs, which isn't surprising for an urban area, even one surrounded by farmland and open space.

Yep, so you're right that if we had a super low population, we'd end up with a higher ratio since there'd be double (in the case of 0.5 people) the greenspace per person.

So in the interest of caution, how about we bump this up to 0.1?

phargogh: In case you agree, I've done this and added a comment in 5275ee0bd

tests/test_urban_nature_access.py
phargogh: OK! I've implemented this in https://github.com/phargogh/invest/commit/52fbbd565b776ba91b79eab0bf27de957f1d2803.  The behavior now is:

* The `search_radius_m` column is always required in the attributes table
* If the search radii are all the same, we treat that as "unified greenspace" and all greenspace lucodes will be processed together.  An INFO message is logged describing what's up.
* If there are multiple detected search radii, we use "split greenspace" mode and each individual LULC code is processed seprately (per request of the science team).  an INFO message is logged describing what's up.

As a part of this I also factored the kernel creation and population convolution out of the 2SFCA loop since there isn't necessarily a 1:1 mapping of search radii to greenspace lucodes.

phargogh:
OK!  Well one thing led to another since your last kind review, so I'm sorry about the delay.  This is now ready for a re-review, I think, and thanks in advance!

phargogh:
So the latest changes to `main` are also in this PR again.  I've opened up #1012 to help make this review easier to read.  Feel free to review that one first and once it's merged I'll update this PR.

phargogh:
@emlys @dcdenu4 I just pushed a change in 807eb9bc2 that I think might help to clarify the radius/lucode groupings question, and it includes some more comments as requested.  Let me know if there's anything else you'd like to talk through!

src/natcap/invest/urban_nature_access.py
```diff
@@ -617,63 +774,92 @@ def _greenspace_supply_demand_op(greenspace_budget, population):
...
+        out_array = numpy.full(
+            greenspace_area.shape, FLOAT32_NODATA, dtype=numpy.float32)
+
+        # Small negative values should already have been filtered out in
+        # another function after the convolution.
+        # This avoids divide-by-zero errors when taking the ratio.
+        valid_pixels = (convolved_population > 0)
+
+        # R_j is a ratio only calculated for the greenspace pixels.
+        greenspace_pixels = ~numpy.isclose(greenspace_area, 0.0)
```
emlys, line 833: I think this and the other `0.0`s can just be `0` since python 3

src/natcap/invest/urban_nature_access.py
```diff
@@ -617,63 +774,92 @@ def _greenspace_supply_demand_op(greenspace_budget, population):
...
+            valid_pixels &= ~numpy.isclose(greenspace_area, greenspace_nodata)
+
+        # If the population in the search radius is numerically 0, the model
+        # specifies that the ratio should be set to the greenspace area.
+        # Because the convolution happens in float64 space and is
+        # non-normalized, kicking the atol up to 0.1 feels necessary to
+        # eliminate the possibility of accumulation of numerical noise.
+        # JD picked 0.1 out of my back pocket.
+        population_close_to_zero = numpy.isclose(
+            convolved_population, 0.0, atol=0.1)
```
emlys, line 849: As we talked about in person, it might make more sense to set the ratio to the greenspace area for all population values in the range 0-1

phargogh: Patching this in 3f68df5fd and a doc forthcoming!

src/natcap/invest/urban_nature_access.py
phargogh: Yep, they sure don't: f8087c3ae

src/natcap/invest/urban_nature_access.py
```diff
@@ -200,25 +240,23 @@ def execute(args):
...
         sorted(ARGS_SPEC['args']['decay_function']['options']))
 
     if 'kernel_type' not in args:
         kernel_type = 'dichotomy'
-    elif args['kernel_type'] not in kernel_types:
-        raise ValueError(
-            f'Kernel type "{args["kernel_type"]}" is not recognized. '
-            f"Must be one of {', '.join(kernel_types.keys())}")
+        LOGGER.info(
+            'args["kernel_type"] not provided; defaulting to {kernel_type}')
```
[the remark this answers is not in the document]
phargogh: I agree, this optional behavior just came back to bite me right now, where I was using the correct key names everywhere _except_ in `execute`, which led to incorrect regression values having been committed to the test.  I've fixed this and there's really no reason why I can't just require the kernel name too, so I am now.  See rev b2a1f7a59

src/natcap/invest/urban_nature_access.py
```diff
@@ -62,6 +68,14 @@
                         "1 if this landcover code represents greenspace, 0 "
                         "if not."
                     ),
+                },
+                'search_radius_m': {
+                    'type': 'number',
+                    'units': u.meter,
+                    'about': (
+                        'The distance in meters to use as the search radius '
+                        'for this type of greenspace. Values must be >= 0.'
```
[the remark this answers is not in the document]
phargogh: Patched in 03c91cd54

src/natcap/invest/urban_nature_access.py
phargogh: Here's a link to the doc for future us (stanford auth needed): https://docs.google.com/document/d/1CdzGT9YyeedtjaXQ17ZBGkg2KrsD8HTPUZ2sS7GZ3Uo/edit?usp=sharing

dcdenu4:
I've been slow to respond and haven't allowed #1012 to be merged yet! Sorry! Once that is done this should be quick to merge in. (mostly note to self).

src/natcap/invest/urban_nature_access.py
phargogh: We heard back from the urban team and here's Lisa's response:

> James and Emily, I just chatted with Roy Remme (who had lost access to Slack, but is back now), and Anne and Eric, too, and we all agree your solution looks like a great one.And we really appreciate that you noticed the funky discontinuity implied by the original model!

So!  The approved solution is included in this PR already.

dcdenu4:
Hey @phargogh, just a heads up that #1012 has been merged producing a few conflicts in this PR.

phargogh:
OK!  Conflicts have been resolved and these branches are now much more up-to-date with the latest `main`.  Re-requesting review.

src/natcap/invest/urban_nature_access.py
```diff
@@ -263,103 +289,161 @@ def execute(args):
...
+        LOGGER.info(
+            'Multiple search radii found. '
+            'Running 2SFCA in split greenspace mode, calculating greenspace '
+            'supply per LULC class.')
+
+        # If there are varying search radii, we run 2SFCA once for each
+        # radius:lucode combination so that modelers can see the contribution
+        # of each landcover code to the total greenspace supply.  This comes at
+        # the cost of more convolutions and more runtime (and disk space), but
+        # it's what the UNA science team wants.
```
dcdenu4, line 357:
```suggestion
        # the cost of more convolutions and more runtime (and disk space).
```

dcdenu4: I suggested a removal of that last bit because I'm not sure it's helpful. Let me know what you think!

src/natcap/invest/urban_nature_access.py
```diff
@@ -263,103 +289,161 @@ def execute(args):
...
+        radius_to_lucodes = [
+            (search_radius_m, [lucode])
+            for ((search_radius_m, lucode), _) in
+            greenspace_attrs.groupby(['search_radius_m', 'lucode'])
```
dcdenu4, line 361: `groupby`, nice

dcdenu4 (changes requested):
Thanks @phargogh , I just had one very minor suggestion on a comment. Otherwise this looks good to me!

src/natcap/invest/urban_nature_access.py
phargogh: Yeah, that `groupby` is really kind of interesting ... it's yielding a generator of dataframes, which is _not_ immediately clear.

src/natcap/invest/__init__.py
```diff
@@ -34,6 +32,11 @@
 def set_locale(locale_code):
     """Set the `gettext` attribute of natcap.invest.
 
+    This is the locale that will be used for translation. The `gettext`
```
emlys, line 35: Possibly something amiss with the merge? This looks correct but I'm surprised this file is in the diff

emlys (approved):

dcdenu4 (approved):
Thanks for all of this @phargogh! I'm going to approve but let you do the final merge since the binary tests are failing (seemingly for different reasons), and Emily had a comment about something `gettext` related that raised a flag.

phargogh:
So there are 2 issues for sure in the binary builds, one with the UG not building (pending https://github.com/natcap/invest.users-guide/pull/69, https://github.com/natcap/invest.users-guide/pull/68) and one with a bug in my workbench ui config for UNA.  I've pushed a fix for the workbench ui config and am waiting on tests now.

phargogh:
OK!  We have 2 approvals and I think the remaining issue is related to the forums tag and will need to be figured out in due time.  We have 2 approvals now and these test failures appear minor so I'll go ahead and merge.

PR #945, opened 2022-04-14
Workbench - check all external links are okay

### LINKED ISSUE #849: Fix & test User Guide & Forum links
Each model links directly to its User Guide #DataNeeds section and it's forum tag page. I just found one broken UG link for Urban Flood Risk Mitigation. We should have a test that catches that for us.

This issue actually cuts across the invest / workbench interface. UG html filenames are stored in `natcap.invest.MODEL_METADATA` (and that's where the UFRM filename needs to be fixed). But the root URL of the latest UG is now hardcoded in the Workbench, along with the `#data-needs` suffix. 

A dictionary of model forum tags is maintained in the workbench in `ResourcesLinks`.

davemfish:
This PR adds tests that check all of the external links (i.e. `<a href="...">`) in the UI. These include one forum and one userguide link on each model's `InvestTab`. And one userguide link for each arg input in the "info" modal. 

Fixes #849 

## Checklist


workbench/src/renderer/components/ResourcesLinks/index.jsx
```diff
@@ -3,36 +3,40 @@ import React from 'react';
...
+import UG_ROOT from '../../userguideURL';
 import { ipcMainChannels } from '../../../main/ipcMainChannels';
 
 const { ipcRenderer } = window.Workbench.electron;
 
-const UG_ROOT = 'http://releases.naturalcapitalproject.org/invest-userguide/latest';
 const FORUM_ROOT = 'https://community.naturalcapitalproject.org';
 
 // map model names to forum tags:
 const FORUM_TAGS = {
```
emlys, line 14: Instead of hard-coding this mapping, could we derive the forum tag from the model name? (in most cases, just replace `_` with `-`). Or at least default to that and only hard-code the exceptions. Also consider retroactively changing the tags that don't follow the pattern.

workbench/src/renderer/components/ResourcesLinks/index.jsx
```diff
@@ -52,23 +56,24 @@ function handleClick(event) {
...
-    userGuideURL = `${UG_ROOT}/${docsName}#data-needs`;
-  }
   if (tagName) {
-    forumURL = `${FORUM_ROOT}/tags/${tagName}`;
+    forumURL = `${FORUM_ROOT}/tag/${tagName}`;
+  }
+
+  let userGuideURL = `${UG_ROOT}/index.html`;
+  if (docs) {
+    userGuideURL = `${UG_ROOT}/${docs}#data-needs`;
```
emlys, line 69: When would `docs` be undefined?

workbench/src/renderer/userguideURL.js
```diff
@@ -0,0 +1 @@
+export default 'https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest';
```
emlys, line 1: I like this pattern of having a separate file to export a single hard-coded value. Curious if this is an "official" pattern that you've heard of, or your invention?

emlys (commented):
Thanks @davemfish! Just had a couple questions. I was also wondering about the approach of testing the URL by itself, rather than testing clicking on the link. I briefly tried to test by clicking on the link in #901 and got stuck because `shell.openExternal` wasn't available in the test environment. Is that the reason you went with checking the URL directly?

workbench/src/renderer/components/ResourcesLinks/index.jsx
davemfish: We could. The only other thing to consider is that some models don't have forum tags because there are no questions. By leaving them out of this map completely, we let the link default to the forum homepage for those models instead of going to a tag that doesn't exist, or a tag that does exist but has no threads (maybe the latter would be okay). 

I think we should whatever we think is lowest maintenance. Maybe that is to proactively create tags for all models following a strict pattern and then remove the hardcoded tags like you suggest?

Also, crop productions - they share a tag which I think is okay, they also share a UG and many of the questions pertain to both models.

workbench/src/renderer/components/ResourcesLinks/index.jsx
davemfish: Probably never? But maybe that wasn't true when this was written. I updated this. It seems safe to not guard against this since we are literally validating `userGuideURL` for every model in this PR.

workbench/src/renderer/userguideURL.js
davemfish: I think it's conventional. At an abstract level it's not really different than exporting a function or anything else that needs to be shared across more than one module.

But if we accumulated a lot of these shared constants, I think it's personal preference whether we have them each in their own module or have many exported from the same module.

workbench/src/main/setupOpenExternalUrl.js
```diff
@@ -8,7 +8,6 @@ import { ipcMainChannels } from './ipcMainChannels';
 export default function setupOpenExternalUrl() {
   ipcMain.on(
     ipcMainChannels.OPEN_EXTERNAL_URL, (event, url) => {
-      event.preventDefault();
```
davemfish, line 11: This doesn't belong. This `event` is a node `Event` instance, not a browser `Event`. We already did `event.preventDefault()` on the browser event earlier in the renderer right before sending the IPC here.

workbench/src/renderer/components/ResourcesLinks/index.jsx
emlys: Okay, lets leave it as it is

emlys (approved):
Looks great, thanks @davemfish!

PR #947, opened 2022-04-19
replace _ with gettext to avoid name conflicts

### LINKED ISSUE #941: gettext '_' makes it hard to use invest in the python REPL
```python
>>> from natcap.invest import spec_utils
>>> spec_utils.format_type_string('csv')
'`CSV <input_types.html#csv>`__'
>>> x= spec_utils.format_type_string('csv')
  File "C:\Users\dmf\projects\invest\env\lib\site-packages\natcap\invest\spec_utils.py", line 362, in format_type_string
    'boolean': _('true/false'),
TypeError: 'str' object is not callable
```

In this case the REPL assigns to `_` implicitly on the second line. But we also have cases in invest where we assign to `_` following the convention of using that variable name for a throwaway object. Won't that also interfere with gettext trying to call `_()`?

emlys:
## Description
Fixes #941 

The translation function `gettext(...)` (alias of `_(...)`) can be thought of as a state variable whose value changes during execution when the user chooses a language. This PR moves that "state" from the global scope into the `natcap.invest` module to avoid conflicts with other globals of the same name (`_` as the previously returned value in the REPL; `gettext` as the standard library module).

* No longer "install" the gettext function globally. Instead, add a `gettext` attribute to `natcap.invest`. This is the translation function to be used by all `natcap.invest` modules.
* Any module that needs to translate strings will explicitly define the translation function within its own scope by importing: `from natcap.invest import gettext`. Note that we could call it `_`, or any other name, but gettext is more explicit and won't conflict with `_` as a placeholder variable.
* To switch languages during execution: before, `natcap.invest.install_language`. This function is replaced with `natcap.invest.set_locale`, which updates the value of `natcap.invest.gettext`.
* Same as before, all affected modules must be reloaded after calling `set_locale` for the strings to get translated.
* Moved the `MODEL_METADATA` into its own module, `natcap.invest.model_metadata`. This was needed because `natcap.invest.__init__.py` needs to set the locale to the default ('en') to make `natcap.invest.gettext` available right away to other modules.
But there are translated strings in the `MODEL_METADATA` so it needs to get reloaded when the language changes. So we need to be able to re-evaluate `MODEL_METADATA` without re-evaluating `natcap.invest.__init__.py`. So I moved it into a separate module.
* Simplify setup, teardown, and mocking in the translation tests.
* Fix a bug in the translations tests where the `language` query parameter wasn't being used. The tests were passing because the global state was accidentally persisting between tests.

Note that `pybabel` automatically extracts strings wrapped in `gettext(...)` as well as `_(...)`, so no changes needed to the message extraction command.

## Checklist


tests/test_translation.py
```diff
@@ -108,57 +108,54 @@ def test_invest_validate(self):
...
+            'api/validate', json=payload,
+            query_string={'language': TEST_LANG})
         results = json.loads(response.get_data(as_text=True))
         messages = [item[1] for item in results]
         self.assertIn(TEST_MESSAGES[missing_key_msg], messages)
 
     def test_translate_formatted_string(self):
-        with patch('natcap.invest.LOCALE_DIR', self.test_locale_dir):
-            install_language('en')
+        set_locale('ll')
```
davemfish, line 154: It looks like `TEST_LANGUAGE` is 'll', so could use that variable? But it's curious that the language is the only change to this test and it still passes? This one also looks a lot like the previous test, are they both needed?

davemfish (changes requested):
@emlys this looks good, thanks for the clear explanation! I just had a couple comments in the tests.

tests/test_translation.py
emlys: I changed the test language for all tests from "en" to "ll" (the foo of language codes) to make it more explicit that we're using a fake, test language.

The first three tests are for translation through the CLI for the 3 commands where it's relevant: `list`, `getspec`, `validate`.
The next three tests are for translation through the 3 corresponding server endpoints: `api/models`, `api/getspec`, `api/validate`.
This last one tests translation more directly, not through either interface. That's why it's the only one that calls `set_locale` directly. The CLI or server endpoint calls `set_locale` in the other tests. Its purpose is testing the translation functionality independent of the CLI or server.
It also tests that translation works on an f-string, which is important because formatting an f-string triggers evaluation of the string, which calls the translation function.

emlys:
@davemfish I removed those `print`s, added the docstring, and replaced the hard-coded string with the `TEST_LANG` variable.

davemfish (approved):
Great, thank you!

PR #952, opened 2022-04-25
Bugfix for workbench data-download progress bar

### LINKED ISSUE #926: WB sampledata: progress bar state is incorrect if download fails
The progress bar never changes from "Downloading 1 of ..." if the download failed. We just don't expect that to ever happen. But just like we switch to "Download Complete" when they are all downloaded, we can clear the progress counter and briefly display a "Download Failed" if it fails.

davemfish:
The progress bar did not accurately reflect the state of a failed download, which would happen on a 404, or presumably a loss of the network .

Fixes #926 

## Checklist


emlys (approved):
Looks good!

PR #954, opened 2022-04-25
Validation: bugfix in pattern-matching on freestyle strings

### LINKED ISSUE #897: Validate results_suffix does not include path separators
For example, `2022/3/5` as a `results_suffix` passes our validation, but is quite problematic when models look for intermediate files that don't exist at paths like `intermediate\\raster_aligned_2022\\3\\5.tif`.

davemfish:
In order for a string to be valid, the whole string needs to fulfill the regex. Otherwise `re.findall('[a-zA-Z0-9_-]*', '4/20')` will find two matches: ['4', '20'] when we want it to find none.

I also removed the default `re.IGNORECASE` because I checked for all the uses of a "regex" in `ARGS_SPEC` and neither of the two uses have a need for ignoring case. It seems better to explicitly include the cases in the regex. And I updated a docstring that seemed out of date.

Fixes #897 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -39,6 +39,8 @@ Unreleased Changes
       ``universal2`` wheel architecture offered by ``scipy>=1.8.0``.
     * Expose taskgraph logging level for the cli with
       ``--taskgraph-log-level``.
+    * Fixed bug in validation of ``results_suffix`` so that special characters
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Yep, this looks great!  Thanks @davemfish

PR #956, opened 2022-04-26
SDR: type checking for lucode column of table

### LINKED ISSUE #890: SDR: validate the lucode column of the biophysical table
A table like this, with the `lucode` and `description` swapped yielded a really cryptic error way down the line.

```
Description;lucode;usle_c;usle_p
0;Class1_Prosopis;0.35;0
1;Class2_Acacia_Tort;0.38;0
...
```

The `ARGS_SPEC` describes this field as an `integer`, but we're not using that info in validation. We could do the validation in `sdr.validate` though.

davemfish:
This PR adds type-checking for the "lucode" column of the biophysical table, to check that the values are numbers. I decided to do this during `execute` instead of `validate` because this model already does some other value-checking of this table there. And it's potentially slow to read the whole table in `validate`. 

I also removed a reference to an `arg` no longer in the `ARGS_SPEC`.

Fixes #890 

## Checklist

### DIFF
--- a/src/natcap/invest/sdr/sdr.py
+++ b/src/natcap/invest/sdr/sdr.py
@@ -264,14 +264,13 @@ def execute(args):
                 float_value = float(table[table_key])
                 if float_value < 0 or float_value > 1:
                     raise ValueError(
-                        'Value should be within range 0..1 offending value '
-                        'table %s, lulc_code %s, value %s' % (
-                            table_key, str(lulc_code), str(float_value)))
+                        f'{float_value} is not within range 0..1')
             except ValueError:
                 raise ValueError(
-                    'Value is not a floating point value within range 0..1 '
-                    'offending value table %s, lulc_code %s, value %s' % (
-                        table_key, str(lulc_code), table[table_key]))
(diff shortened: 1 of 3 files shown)


src/natcap/invest/sdr/sdr.py
```diff
@@ -241,35 +241,36 @@ def execute(args):
...
+                    f'biophysical table is not a number. Please check the '
+                    f'formatting of {args["biophysical_table_path"]}')
             try:
                 float_value = float(table[table_key])
                 if float_value < 0 or float_value > 1:
                     raise ValueError(
-                        'Value should be within range 0..1 offending value '
-                        'table %s, lulc_code %s, value %s' % (
-                            table_key, str(lulc_code), str(float_value)))
+                        f'{float_value} is not within range 0..1')
```
dcdenu4, line 267: Pairing this down to just the information being tested makes a lot of sense.

dcdenu4 (approved):
Thanks @davemfish , looks good to me! I also think it's okay to do the type checking here.

PR #957, opened 2022-04-26
Add support for python 3.10 and drop 3.7

### LINKED ISSUE #672: Test against Python 3.10
Python 3.10 us supposed to be released today (Oct 4, 2021) so let's make sure everything is working as expected.

emlys:
## Description
Fixes #672 
* Updated pyproject.toml as mentioned in the issue
* Updated classifiers in setup.py
* Updated versions in github actions workflow
* Discovered a bug where model UI windows would not open with python 3.9.0-3.9.3. This was fixed by removing the buggy conditional that was only there to support 3.7.

We don't have to drop 3.7, but I think it's nice to keep the github workflow matrix from growing.

## Checklist


src/natcap/invest/ui/model.py
```diff
@@ -687,16 +687,7 @@ def set_title_attr(self, name, value):
         """
         LOGGER.info('__setattr__: %s, %s', name, value)
         old_attr = getattr(self, name, 'None')
-        # Python Core and Builtins were updated in Python 3.8.4
-        # that handle __setattr__ differently. In 3.7, the super()
-        # implementation causes a `TypeError: can't apply this __setattr__
-        # to object object` when running `invest run carbon`. In Python 3.8.4+
-        # with the object implementation the error `TypeError: can't apply
-        # this __setattr__ to Carbon object`
-        if (sys.version_info.minor >= 8) and (sys.version_info.micro >= 4):
```
emlys, line 696: For versions 3.9.0 - 3.9.3 and 3.10.0-3.10.3, this would evaluate to `False`, and so the wrong `__setattr__` was used.

emlys:
I'm not sure what's up with the mac build failure - it's happening pretty regularly, but it didn't reproduce locally and I doubt it's related to python 3.10.

src/natcap/invest/ui/model.py
phargogh: Whoops!  Thanks for catching this.

phargogh (approved):
I support this change, but I'd like to run this by @dcdenu4 in case he has any concerns about dropping 3.7.  Any objections, Doug?

Abandoning 3.7 seems like a pretty clear move for us.  According to the [python dev docs](https://devguide.python.org/#status-of-python-branches), 3.7 is in bugfix mode now (as is 3.8 for that matter, but I think it's reasonable to continue to support it until we add 3.11).  Even `debian:stable` is on 3.9!

Should we move forward this, I'd just like to mention 2 additional tasks that will become necessary whenever we actually require 3.8 for functionality:
- [ ] The Rec Server's python is currently 3.7.9, so we may want to update that to 3.10
- [ ] We'll need to update our conda-forge package for `natcap.invest` to use `python>=3.8`.

dcdenu4:
@phargogh it looks like Python 3.7, 3.8 is in `Security` status, no (https://devguide.python.org/#status-of-python-branches)? I think historically we've tried to support the latest 3 Python versions, so with adding 3.10 support, dropping 3.7 would keep with that loose practice.

I think it's fair to drop 3.7, and thanks for bringing up those adjacent changes that need to be made!

phargogh:
@dcdenu4 yes, I also saw that python 3.7 _and_ 3.8 had security status, which just made me think that we could do with only supporting the more recent of the two :)

So I think we're clear to merge this, then!  I'll press the button.

phargogh:
As Emily  mentioned, the mac binaries are failing and they appear to be related to unmounting a disk image and not with the application itself, so these changes should be clear to merge.

PR #958, opened 2022-04-27
turn on logging again for the invest server. #563.

### LINKED ISSUE #563: ui_server.py excessive logging causes server to not respond
This came up in a very specific workbench context where jest + puppeteer are running a workbench Windows binary and entering text in a model setup form. Requests to the `/validate` endpoint would work as expected for the first few dozen requests, and then the flask server would stop responding, but would not crash or emit any stderr.

Removing most of the `LOGGER.debug` calls within the `/validate` endpoint function avoids the problem. Could it be that the LOGGER's handlers are not threadsafe and the repeated requests were invoking logging from multiple threads? 

The hang/freeze was very reliable, but only on Windows and only with jest + puppeteer. I could not reproduce it manually with the same exact binaries. 

None of these specific logging statements are important. They were all leftover from debugging during development.

davemfish:
We had turned off logging because of #563 . I've turned it back on and confirmed that the server-not-responding is no longer an issue. Possibly because we now make far fewer calls to validation (and subsequent logging) during typing. I chose `--debug` level because these messages only go to a file during production, so the extra noise is not noticeable to the user. And it's potentially useful to have as much logging as possible in those files.

Also removed debug-level logging from `flask-cors`, that was only meant to be there during development.

Fixes #563

## Checklist

### DIFF
--- a/src/natcap/invest/ui_server.py
+++ b/src/natcap/invest/ui_server.py
@@ -16,7 +16,6 @@ from natcap.invest import spec_utils
 from natcap.invest import usage
 
 LOGGER = logging.getLogger(__name__)
-logging.getLogger('flask_cors').level = logging.DEBUG
 
 PREFIX = 'api'
 app = Flask(__name__)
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Looks good to me!

PR #964, opened 2022-04-28
Workbench: formatting and layout for About menu pages

### LINKED ISSUE #961: WB: professionalize the About menu pages
The "About invest" page and the "Report a problem" page both have completely unstyled html. A little work would go a long way to making those look more professional.

davemfish:
These pages were updated to look like this:

Fixes #961 

## Checklist


workbench/about.html
```diff
@@ -1,129 +1,147 @@
...
+      <img
+        src="/src/renderer/static/invest-logo.png"
+        width=191
+        height=167
+      />
+      <div id='invest-version'>
+      <p>version:</p>
+      <span id='version-string'></span>
+    </div>
+      <p id='invest-copywrite'>
```
phargogh, line 24: This is trivial, but would it be worth using `invest-copyright` as the ID instead?  It'd be same number of letters

workbench/about.html
```diff
@@ -1,129 +1,147 @@
...
+                href='http://github.com/Toblerity/Shapely'
+              >
+                http://github.com/Toblerity/Shapely
+              </a>
+            </td>
+          </tr>
+      </table>
+      <p>
+        <em>The source code for GPL'd components are included as an extra
+        component on your installation medium.</em>
```
phargogh, line 142: Yeah, we should probably do this, though maybe linking to the software projects themselves is sufficient?

Technically speaking, we're responsible for making available the source code of any GPLed software we distribute, so including them in the installer would be a nice way to fulfill our obligations under the GPL for everyone obtaining the binaries.  But it also adds a lot of overhead to actually include all that software.  Just in a quick spot-check of what other projects do, I see slack links to a [page with their dependencies](https://slack.com/libs/desktop), and Atom includes a textfile with everything concatenated together, along with the node package that it references.  So maybe what we have in the dialog is sufficient?

Do we distribute GPL'd source code in the workbench's distribution medium?

Although we might be obligated to do this (and I know the Qt interface says we do this), we might want to not include this extra sentence until we do offer it.

phargogh (changes requested):
Looks nice!  I had one trivial comment about an ID name and then a recurring thought about software licenses.  What do you think?

workbench/about.html
davemfish: Thank you, yes spelling words correctly is generally a good thing :)

workbench/about.html
davemfish: I also just noticed we're not including our `LICENSE.txt` in the distribution. It's linked from from the About page, but should probably be included in the distribution.

phargogh: Thanks for including our LICENSE!  I just created an issue for generating a list of python packages here: https://github.com/natcap/invest/issues/979, as we were talking about on the coffee call this morning.

phargogh:
We talked about the build issue and there'll be a separate PR for the upcoming release (3.11.0) with the rtree version restriction since the issue appears to be upstream.

PR #965, opened 2022-04-28
Workbench: better visual cues for clickable navigation

### LINKED ISSUE #962: WB: Sidebar tabs need better visual cue as navigation buttons.
Just got a report from Peter, and I agree, the "Setup" and "Log" tabs in the sidebar that we want people to click for navigation are not obviously clickable until you mouse over them.

davemfish:
Trying to make it more obvious that "Setup" and "Log" are clickable, and not just headers. I added `>` icons to attract the mouse, and enhanced the "on-hover" styles. In this image my mouse isn't visible, but it's hovering over "Log".

tabs

Log is still disabled and has no hover effects until after a model starts running. In this image, I already ran the model.
Fixes #962 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/OpenButton/index.jsx
+++ b/workbench/src/renderer/components/OpenButton/index.jsx
@@ -38,8 +38,6 @@ export default class OpenButton extends React.Component {
     const tipText = _('Browse to a datastack (.json) or InVEST logfile (.txt)');
     return (
       <OverlayTrigger
-        // defaultShow={this.props.defaultShow}
-        // flip={this.props.defaultShow}
         placement="left"
         delay={{ show: 250, hide: 400 }}
         overlay={<Tooltip>{tipText}</Tooltip>}
--- a/workbench/src/renderer/components/OpenButton/index.jsx
+++ b/workbench/src/renderer/components/OpenButton/index.jsx
@@ -38,6 +38,8 @@ export default class OpenButton extends React.Component {
     const tipText = _('Browse to a datastack (.json) or InVEST logfile (.txt)');
     return (
       <OverlayTrigger
+        // defaultShow={this.props.defaultShow}
+        // flip={this.props.defaultShow}
         placement="left"
         delay={{ show: 250, hide: 400 }}
         overlay={<Tooltip>{tipText}</Tooltip>}
--- a/workbench/src/renderer/styles/style.css
+++ b/workbench/src/renderer/styles/style.css
@@ -40,12 +40,16 @@ body {
 .navbar-middle {
 	flex-shrink: 1;
 	min-width: 0;
+	padding-left: 0;
+	padding-right: 0;
 }
 
 .navbar-right {
 	/*max-width: fit-content;
 	margin-left: 0.5rem;*/
 	margin-right: -15px;
+	padding-left: 0;
+	padding-right: 0;
 }
 
 .navbar-brand .nav-link {
--- a/workbench/src/renderer/styles/style.css
+++ b/workbench/src/renderer/styles/style.css
@@ -6,26 +6,18 @@ body {
 	--invest-green: #148F68;
 	background-color: rgba(0,0,0,0);
 	margin: 0;
-	/*padding: 1rem;*/
 	overflow-y: hidden;
 }
 
 /*Top Navigation*/
 
 .navbar {
-	/*height: 3rem;*/
-	/*align-items: end;*/
-	/*padding-bottom: 0;*/
 	border-bottom: 3px solid var(--invest-green);
-	/*background-color: #fff*/
 	padding: 0;
 }
 
 .navbar .row {
 	align-items: end;
-	/*padding-bottom: 0;
-	border-bottom: 3px solid var(--invest-green);
-	background-color: #fff*/
 }
 
 .navbar-brand {
@@ -172,7 +164,6 @@ exceed 100% of window.*/
 .invest-list-container {
 	height: 87vh;
 	overflow-y: auto;
-	/*padding-left: 0;*/
 }
 
 .invest-list-group {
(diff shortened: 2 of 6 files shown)


emlys (changes requested):
Thanks for taking on this tricky UI design problem @davemfish! This is definitely an improvement, but to my eye, still looks somewhat like a header.

I think what's really throwing me off is the placement of the "load parameters" and "save ..." options that are indented below Setup. That makes me think header and not button. What if instead of 
```
Setup
    Load parameters from file
    Save to JSON
    Save to Python script
    Save datastack
Log
User's Guide
Frequently Asked Questions
```
it was more like
```
Setup
Log
-------------
Load parameters from file
Save to JSON
Save to Python script
Save datastack
User's Guide
Frequently Asked Questions
```
?

emlys:
I was looking at navigation sidebars in other apps to see how they do it. Here are some examples:
Screen Shot 2022-04-28 at 11 11 51 AM
Screen Shot 2022-04-28 at 11 12 10 AM
Screen Shot 2022-04-28 at 11 12 33 AM
Screen Shot 2022-04-28 at 11 12 57 AM

Differences I see are:
- not a lot of styling is needed to indicate items are clickable - just a faint highlight on hover.
- each item has squared corners and takes up the full height and width of its bbox
- each item has an associated icon on the left hand side

Based on that I would try
- extending the Setup and Log boxes all the way to the left edge of the window
- adding icons on the left side (may not be necessary)
- removing the underline on hover

I might make the grayed-out Log a lighter gray to match the color of disabled input labels. Or consider removing the Log until there's a log to show.

davemfish:
> ```
> Setup
> Log
> -------------
> Load parameters from file
> Save to JSON
> Save to Python script
> Save datastack
> User's Guide
> Frequently Asked Questions
> ```
> 
> ?

Yes, I agree all those Save buttons are what make Setup look like a header. So this layout is better for that. I worry that it becomes slightly less obvious what you are saving with those Save buttons when they're not nested under Setup. But that's a much smaller problem than someone not immediately knowing how to navigate the app. Good suggestion, here's what it looks like now:

Capture

davemfish:
> I was looking at navigation sidebars in other apps to see how they do it. Here are some examples: Screen Shot 2022-04-28 at 11 11 51 AM Screen Shot 2022-04-28 at 11 12 10 AM Screen Shot 2022-04-28 at 11 12 33 AM Screen Shot 2022-04-28 at 11 12 57 AM

Sorry I didn't see this until I already made the recent changes. By the way the original design was inspired by the sidebar on Stack Overflow: https://stackoverflow.com/

> Differences I see are:
> 
> * not a lot of styling is needed to indicate items are clickable - just a faint highlight on hover.

This is what we had originally.

> * each item has squared corners and takes up the full height and width of its bbox

I think we have this.

> * each item has an associated icon on the left hand side

Well I went with the right-hand side, but yeah. I don't think there are conventional icons for "Setup" and "Log" so I'm not sure it's a good idea to give them unique icons. I went with the `>` because I thought it evoked navigation.

davemfish:
> * removing the underline on hover

I agree. We don't need this now with the new layout. I made the highlighted gray more subtle too. But it still needs to have some contrast with the one already active, since they butt up against each other.

> I might make the grayed-out Log a lighter gray to match the color of disabled input labels. 

Yes, couldn't hurt.

> Or consider removing the Log until there's a log to show.
I don't like it when stuff appears and disappears in a UI, especially when it moves other things around. So I'd rather not mess with this option.

emlys:
Thanks for making those changes, I think the navigation is pretty unmistakable now. One minor thing is that the vertical green bar on the selected item pushes the > over to the left slightly. It would be nice if they stayed aligned. But not worth it if the fix isn't obvious.

workbench/src/renderer/styles/style.css
```diff
@@ -300,29 +308,35 @@ body {
 	border-radius: 0;
 }
 
-.sidebar-setup {
+.sidebar-row {
+	display: flex;
+	flex-direction: column;
+	font-size: 1.0rem;
+	padding-left: 1.0rem;
```
emlys, line 315: Is the `.0` needed?

emlys (changes requested):

workbench/src/renderer/styles/style.css
davemfish: No. Lots of times a `rem` value will not be a whole number though.

davemfish:
> Thanks for making those changes, I think the navigation is pretty unmistakable now. One minor thing is that the vertical green bar on the selected item pushes the > over to the left slightly. It would be nice if they stayed aligned. But not worth it if the fix isn't obvious.

Yes I noticed that too. I fixed it.

@emlys  while we're on the subject. Do you think we should have a "home" icon next to the big "InVEST" button? To make it obvious that you can get back to the home screen without closing the invest model tab?

emlys:
> Do you think we should have a "home" icon next to the big "InVEST" button?

Sure, it couldn't hurt!

workbench/src/renderer/components/OpenButton/index.jsx
```diff
@@ -38,6 +38,8 @@ export default class OpenButton extends React.Component {
     const tipText = _('Browse to a datastack (.json) or InVEST logfile (.txt)');
     return (
       <OverlayTrigger
+        // defaultShow={this.props.defaultShow}
+        // flip={this.props.defaultShow}
```
emlys, line 42: Can these commented lines be deleted?

emlys (changes requested):
I tried it out and it looks good! Just one place where some commented code was left.

workbench/src/renderer/components/OpenButton/index.jsx
davemfish: Oh thanks! I was trying something out but ultimately made no changes to this.

emlys (approved):
Thanks @davemfish!

PR #971, opened 2022-05-03
WB: switch to Setup from Log after Load Parameters

### LINKED ISSUE #948: WB: Load Parameters button should toggle display to Setup Tab
The "Load Parameters" button in the sidebar is usable when viewing "Setup" or "Log". If I use it while viewing "Log", it would be nice to automatically flip to "Setup" when the parameters load, so I can see the data I just loaded.

I don't think the same is true for the "Save..." buttons

davemfish:
This PR adds a test and functionality to toggle the UI to the Setup tab after the Load Parameters button is used. Since that button can be used from the Log tab, it's nice to automatically toggle to Setup to make it clear that the parameters did in fact load.

Fixes #948 

## Checklist

### DIFF
--- a/workbench/tests/invest/flaskapp.test.js
+++ b/workbench/tests/invest/flaskapp.test.js
@@ -227,7 +227,7 @@ describe('Build each model UI from ARGS_SPEC', () => {
         sidebarSetupElementId="foo"
         sidebarFooterElementId="foo"
         executeClicked={false}
-        setSaveAlert={() => {}}
+        switchTabs={() => {}}
       />
     );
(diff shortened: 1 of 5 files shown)


workbench/tests/invest/flaskapp.test.js
```diff
@@ -227,7 +227,7 @@ describe('Build each model UI from ARGS_SPEC', () => {
         sidebarSetupElementId="foo"
         sidebarFooterElementId="foo"
         executeClicked={false}
-        setSaveAlert={() => {}}
```
davemfish, line 230: `setSaveAlert` was never a prop passed to `SetupTab`.

PR #975, opened 2022-05-04
Workbench: use all available vertical space.

### LINKED ISSUE #972: WB: extra whitespace at the bottom of window
I can't quite remember why that is; maybe a very crude way to avoid a scroll bar? There's got to be a better way.

Capture

davemfish:
It seems that containers that should scroll on overflow require a well-defined height -- not a percentage height. So this PR removes some hardcoded values and adds CSS variables to calculate the viewport height available and make sure all of it gets used.

Fixes #972 

## Checklist

### DIFF
--- a/workbench/src/renderer/styles/style.css
+++ b/workbench/src/renderer/styles/style.css
@@ -5,11 +5,11 @@
 
 /*
 --header-height: fixed height applied to the top navbar.
---content-height: fixed height of the remaining viewport.
+--content-height: height of the remaining viewport.
   Apply this height to any container that should fill the
   available window height and also have a scrollbar appear
   when the content overflows. I could never get a scrollbar
-  to appear with height: 100%, even when parent has this fixed height;
+  to appear with height: 100%, even when parent has a fixed height;
 */
 body {
 	--invest-green: #148F68;
@@ -308,7 +308,6 @@ exceed 100% of window.*/
 	flex:  1 1 0;
 	padding-left: 0;
 	padding-right: 0;
-	height: var(--content-height);
 }
 
 .invest-main-col .tab-content {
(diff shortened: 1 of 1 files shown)


workbench/src/renderer/styles/style.css
```diff
@@ -2,8 +2,20 @@
...
 }
 
+
+/*
+--header-height: fixed height applied to the top navbar.
+--content-height: height of the remaining viewport.
+  Apply this height to any container that should fill the
+  available window height and also have a scrollbar appear
+  when the content overflows. I could never get a scrollbar
+  to appear with height: 100%, even when parent has a fixed height;
```
emlys, line 12: Thanks for these comments!

emlys (changes requested):
Thanks @davemfish! I never noticed that empty space before but it does look better now. Could we also make the first column of models extend all the way to the bottom?
image

davemfish:
> Thanks @davemfish! I never noticed that empty space before but it does look better now. Could we also make the first column of models extend all the way to the bottom? 

Thanks! And probably, though I gave it a try and it's tricky.

The easy layouts are,
* one single column that scrolls
  * this looks nice and is probably a safe choice, but not all models fit on screen without scrolling, which is a bit unfair to the end of the alphabet
* two columns of equal length, that only need to scroll when the window is made very short (the current design)
  * this looks slightly less nice (I think this was your motivation for the change), but it allows all models to fit on screen without a need for scrolling in most cases, which is nice.

The hard layout,
* two columns where the first column goes all the way to the bottom
  * this one would suggest wrapping over to the second column dynamically, but then we need to somehow prevent further wrapping to a 3rd & 4th column (the whole thing gets too wide or the buttons too narrow) and instead start scrolling. This one has me stumped for now.
  * It would be easier to not dynamically wrap and scroll instead, whenever the window is too short for the whole first column. But I don't think this is better than the current layout.

@emlys What do you think? We could make another issue for it?

emlys (approved):

PR #982, opened 2022-05-09
Bugfix for validation spatial overlap message

### LINKED ISSUE #980: WB: node package `os` still used in renderer
```
Uncaught Error: Module "os" has been externalized for browser compatibility and cannot be accessed in client code.
    at Object.get (__vite-browser-external:os:3:11)
    at filterSpatialOverlapFeedback (index.jsx:38:28)
    at ArgInput.render (index.jsx:152:27)
```

The `filterSpatialOverlapFeedback` function isn't used often, so I just noticed this bug. We should look into whether we can get tests to cover this case. We do have unit tests for that function, the problem is our test env is different from the vite dev-mode env.

### LINKED ISSUE #981: Validation: Message about 'bounding box intersections' is incomplete
Here's some results of `validate()`

```
Validation warnings: [(['soil_group_path',
   'lulc_path',
   'aggregate_areas_path',
   'road_centerlines_path',
   'precipitation_path'],
  'Bounding boxes do not intersect:')]
```

In the past this message also listed the bounding boxes of each layer. That's useful for identifying the offending layer.

davemfish:
Fixes #981 

This PR re-instates the complete feedback message when bounding boxes do not intersect. We were just missing the formatted list of bounding boxes.

Tests needed updating to use the new `kwarg` added to the string template. In one test, I actually fetched the bounding boxes and repeated the formatting step to assert the complete message is present. In other tests, I simply passed in a placeholder empty string, as the tests were focused on other things.

Fixes #980 

I removed the instance of the `os` module; there's no need to use an OS-specific newline character when we're rendering the text in the browser.

I failed to find an easy way to catch this problem during builds or tests: the case where a node module is imported on the renderer side. Maybe a solution will reveal itself as we get more familiar with vite.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -49,6 +49,8 @@ Unreleased Changes
       ``--taskgraph-log-level``.
     * Fixed bug in validation of ``results_suffix`` so that special characters
       like path separators, etc, are not allowed.
+    * Fixed a bug in validation where a warning about non-overlapping spatial
+      layers was missing info about the offending bounding boxes.
 * Crop Production
     * Fixed a bug in both crop production models where the model would error if
       an observed yield raster had no nodata value.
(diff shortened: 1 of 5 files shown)


emlys (approved):
Thanks for fixing that @davemfish!

PR #985, opened 2022-05-11
Allow error in create-dmg

### LINKED ISSUE #984: create-dmg non-zero exit breaks Mac builds
The problem is in `create-dmg`'s cleanup step. https://github.com/create-dmg/create-dmg/issues/133

It seems the DMG is created just fine. So in the interest of passing builds, we can probably force our workflow to continue past that failure.

davemfish:
There's consistently an error from `create-dmg` when it tries to unmount a volume. Until that's fixed upstream, we can make sure our workflows continue past this step.

Fixes #984 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -376,6 +376,10 @@ jobs:
         run: make install
 
       - name: Build binaries
+        # Due to a create-dmg issue, sometimes we wait a very long time
+        # for a disk to unmount, and it fails to do so. Instead of waiting
+        # for an inevitable failure, set timeout-minutes.
+        timeout-minutes: 15
         run: make ${{ matrix.binary-make-command }}
 
       - name: Install Node.js
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -378,8 +378,10 @@ jobs:
       - name: Build binaries
         # Due to a create-dmg issue, sometimes we wait a very long time
         # for a disk to unmount, and it fails to do so. Instead of waiting
-        # for an inevitable failure, set timeout-minutes.
+        # for an inevitable failure, set timeout-minutes and allow continuation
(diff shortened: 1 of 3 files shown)


emlys:
@davemfish this is great. I wonder if we might also add a shorter timeout to the dmg building step? Looks like it took over 30 minutes, I assume most of that is waiting for the detach which is going to fail anyway.

davemfish:
> @davemfish this is great. I wonder if we might also add a shorter timeout to the dmg building step? Looks like it took over 30 minutes, I assume most of that is waiting for the detach which is going to fail anyway.

Great idea! Looking back at past builds, I saw some taking ~10 minutes. So I set the timeout for `make binaries` at 15.

.github/workflows/build-and-test.yml
```diff
@@ -376,6 +376,10 @@ jobs:
         run: make install
 
       - name: Build binaries
+        # Due to a create-dmg issue, sometimes we wait a very long time
+        # for a disk to unmount, and it fails to do so. Instead of waiting
+        # for an inevitable failure, set timeout-minutes.
+        timeout-minutes: 15
```
emlys, line 382:
```suggestion
        timeout-minutes: 15
        continue-on-error: true
```

emlys (changes requested):
Ah I guess the timeout causes it to fail at the workflow level, maybe `continue-on-error` will fix it?

davemfish:
> Ah I guess the timeout causes it to fail at the workflow level, maybe `continue-on-error` will fix it?

Ah thanks for noticing that. I added `continue-on-error` and removed the Makefile changes so we're only addressing this in one place. Presumably we'll always timeout now before `create-dmg` raises the error anyway.

davemfish:
Woah 30 mins to restore the node modules cache for Mac, 30 seconds for Windows!

emlys (approved):

PR #987, opened 2022-05-13
reverting changes to use the Makefile solution. #984.

### LINKED ISSUE #984: create-dmg non-zero exit breaks Mac builds

davemfish:
Fixes a bug introduced in #985 . We didn't want to wait for the inevitable `create-dmg` timeout, but setting a time limit & allowing `continue-on-error` for `make binaries` meant we sometimes ended up with an incomplete DMG. 

This PR moves the error-toleration to the Makefile. So we may have to wait for the full timeout from `create-dmg`, but we won't accidentally deploy broken DMGs.

Fixes #984 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -376,12 +376,6 @@ jobs:
         run: make install
 
       - name: Build binaries
-        # Due to a create-dmg issue, sometimes we wait a very long time
-        # for a disk to unmount, and it fails to do so. Instead of waiting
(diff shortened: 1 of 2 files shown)


davemfish:
Closing this because the timeout actually does prevent the final DMG from being created, so it's not a solution to just tolerate the timeout.

PR #990, opened 2022-05-17
use PEP 420: implicit namespace package

### LINKED ISSUE #967: Use PEP 420: implicit namespace packages
`natcap` is a namespace package. We indicate that with the file `src/natcap/__init__.py` which contains
```
import pkg_resources
pkg_resources.declare_namespace(__name__)
```
Today I learned that that's outdated. If I'm understanding correctly, we can just delete `src/natcap/__init__.py` and it will work the same way.
https://stackoverflow.com/a/27586272/14451410
https://peps.python.org/pep-0420/
This isn't really an enhancement, just keeping up with changes in the python standards.

emlys:
## Description
Fixes #967 
- Removed `src/natcap/__init__.py` to make `natcap` an implicit namespace package
- Removed the deprecated `namespace_packages` parameter to `setup`. Since we already list out all packages, I don't think we need to change anything else: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
- Update `doc/api-docs/conf.py` to work with the implicit namespace package, and some minor refactoring:
    - Update the text at the top of the model entrypoints page
    - Build the list of model entrypoints from the installed location of `natcap.invest`, not the build/lib* location
    - Name the model entrypoints doc sections after the model title, not the first line of the execute docstring 

## Checklist

### DIFF
--- a/doc/api-docs/conf.py
+++ b/doc/api-docs/conf.py
@@ -188,8 +188,8 @@ InVEST Model Entry Points
 
 All InVEST models share a consistent python API:
 
-    - Every InVEST model has a corresponding module in the ``natcap.invest``
-      package
+    - Every InVEST model has a corresponding module or subpackage in the 
+      ``natcap.invest`` package
     - The model modules contain a function called ``execute``
     - The ``execute`` function takes a single argument (``args``), a dictionary
       that stores all data inputs and configuration options for the model
--- a/doc/api-docs/conf.py
+++ b/doc/api-docs/conf.py
@@ -69,7 +69,7 @@ version = release.split('+')[0]
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/doc/api-docs/conf.py
+++ b/doc/api-docs/conf.py
@@ -12,17 +12,17 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+from datetime import datetime
 import importlib
 import itertools
 import os
 import pkgutil
 import subprocess
 import sys
-
-from datetime import datetime
-from sphinx.ext import apidoc
 from unittest.mock import MagicMock
 
+import natcap.invest
+from sphinx.ext import apidoc
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -30,15 +30,8 @@ from unittest.mock import MagicMock
 DOCS_SOURCE_DIR = os.path.dirname(__file__)
 INVEST_ROOT_DIR = os.path.join(DOCS_SOURCE_DIR, '..', '..')
 INVEST_BUILD_DIR = os.path.join(INVEST_ROOT_DIR, 'build')
-
-# get name of build/lib directory; this depends on the OS and python version
-# e.g. lib.macosx-10.9-x86_64-3.8
-for path in os.listdir(INVEST_BUILD_DIR):
-    if path.startswith('lib'):
-        INVEST_LIB_DIR = os.path.join(INVEST_BUILD_DIR, path)
-        break
-else:
-    raise ValueError(f'lib directory not found in {INVEST_BUILD_DIR}')
+# get the directory that the natcap package lives in
+INVEST_LIB_DIR = os.path.dirname(os.path.dirname(natcap.invest.__file__))
 
 # -- General configuration ------------------------------------------------
 
@@ -67,7 +60,6 @@ master_doc = 'index'
 project = 'InVEST'
 copyright = f'{datetime.now().year}, The Natural Capital Project'
 
-import natcap.invest
 # The full version, including alpha/beta/rc tags.
 release = natcap.invest.__version__
 version = release.split('+')[0]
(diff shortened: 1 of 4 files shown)


doc/api-docs/conf.py
```diff
@@ -235,45 +230,33 @@ class MockQGroupBox:
...
+for _, name, _ in pkgutil.walk_packages(path=[INVEST_LIB_DIR],
+                                        prefix='natcap.'):
+    if any(x in name for x in EXCLUDED_MODULES):
         continue
 
-    try:
-        module = importlib.import_module(name)
-    except Exception as ex:
-        print(ex)
-        continue
```
emlys, line 257: Removed this because I think we want to raise an error if a module doesn't import.

emlys:
It turns out that this has some implications for our package data (currently just the placeholder directories for translation catalogs). See discussion: https://github.com/pypa/setuptools/discussions/3323

* PEP 420 is standard and the python interpreter looks for namespace packages by default
* By design, any directory within a package is also a package, even if it's empty or contains no python files.
* Therefore, any "data" directories included in a package are really packages
* `natcap.invest.internationalization`, `natcap.invest.internationalization.locales`, and `natcap.invest.internationalization.locales.en` are packages. The interpreter already finds them:
```
>>> import natcap.invest.internationalization.locales.en
>>> natcap.invest.internationalization.locales.en.__name__
'natcap.invest.internationalization.locales.en'
```
* Therefore, we should stop trying to include them as "data" (with the `package_data` kwarg) and instead let them be packages.

doc/api-docs/conf.py
```diff
@@ -77,7 +69,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
```
emlys, line 72: This was causing an error in the ReadTheDocs build, probably due to a sphinx update.

emlys:
@phargogh this is finally ready! The RTD build issues fixed themselves in time.

doc/api-docs/conf.py
```diff
@@ -12,33 +12,26 @@
...
 
+import natcap.invest
+from sphinx.ext import apidoc
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 DOCS_SOURCE_DIR = os.path.dirname(__file__)
 INVEST_ROOT_DIR = os.path.join(DOCS_SOURCE_DIR, '..', '..')
 INVEST_BUILD_DIR = os.path.join(INVEST_ROOT_DIR, 'build')
```
phargogh, line 32: Are these used anywhere else now that we're properly defining `INVEST_LIB_DIR`?  Might they be able to be removed?

doc/api-docs/conf.py
```diff
@@ -202,29 +195,32 @@ class MockQGroupBox:
 
 All InVEST models share a consistent python API:
 
-    1) The model has a function called ``execute`` that takes a single python
-       dict (``"args"``) as its argument.
-    2) This arguments dict contains an entry, ``'workspace_dir'``, which
-       points to the folder on disk where all files created by the model
-       should be saved.
+    - Every InVEST model has a corresponding module in the ``natcap.invest``
+      package
```
phargogh, line 199:
```suggestion
    - Every InVEST model has a corresponding module or subpackage in the 
      ``natcap.invest`` package
```

doc/api-docs/conf.py
phargogh: I agree, better to fail loudly.

doc/api-docs/conf.py
```diff
@@ -235,45 +231,33 @@ class MockQGroupBox:
 EXCLUDED_MODULES = [
     '_core',  # anything ending in '_core'
     'recmodel_server',
-    'recmodel_workspace_fetcher'
+    'recmodel_workspace_fetcher',
+    'natcap.invest.ui'
```
phargogh, line 235: Based on the RTD-rendered docs (https://invest--990.org.readthedocs.build/en/990/api.html for this PR) it looks like maybe this exclusion isn't working.  Maybe it never worked?  In any case, it would probably be nice to exclude packages, or else just have everything in there since it _is_ in the source code.

doc/api-docs/conf.py
```diff
@@ -235,45 +231,33 @@ class MockQGroupBox:
...
-    try:
-        module_title = module.execute.__doc__.strip().split('\n')[0]
-        if module_title.endswith('.'):
-            module_title = module_title[:-1]
-    except AttributeError:
-        module_title = None
-    all_modules[name] = module_title
+    # all modules with an execute function should have an ARGS_SPEC
+    model_title = module.ARGS_SPEC['model_name']
+    all_modules[model_title] = name
```
phargogh, line 252: This is so much better than our prior reliance on module-level docstrings!

setup.py
```diff
@@ -69,19 +69,7 @@ def run(self):
...
-        'natcap.invest.coastal_blue_carbon',
-        'natcap.invest.delineateit',
-        'natcap.invest.ui',
-        'natcap.invest.ndr',
-        'natcap.invest.sdr',
-        'natcap.invest.recreation',
-        'natcap.invest.scenic_quality',
-        'natcap.invest.seasonal_water_yield',
-    ],
+    packages=find_namespace_packages('src'),
```
phargogh, line 72: Great!  It was not immediately clear to me from this function name that normal packages within `natcap.invest` would also be captured, but that does seem to be the behavior.

doc/api-docs/conf.py
```diff
@@ -12,33 +12,26 @@
...
 import pkgutil
 import subprocess
 import sys
-
-from datetime import datetime
-from sphinx.ext import apidoc
 from unittest.mock import MagicMock
 
+import natcap.invest
+from sphinx.ext import apidoc
```
phargogh, line 25: Could you just check these imports and make sure that we aren't importing anything we no longer need?

phargogh (changes requested):
Very exciting!  I just had a few very small comments and suggestions, but overall this is looking good!

doc/api-docs/conf.py
emlys: You're right, a couple of them aren't needed any more

setup.py
emlys: Yeah, it's confusing! https://github.com/pypa/setuptools/discussions/3427

doc/api-docs/conf.py
emlys: You're right, they can be removed

doc/api-docs/conf.py
emlys: I think this is working as expected - it's excluding those from the model entrypoints file `models.rst`. But they're still in the API docs.
The list of excluded modules was unnecessary, though. I refactored so that any module with an `ARGS_SPEC` is included in `models.rst`.

emlys:
@phargogh I'd like to squash these commits before merging

phargogh (approved):
Looks great, thanks@emlys !

PR #993, opened 2022-05-23
CV: Handling unset nodata in bathymetry masking.

### LINKED ISSUE #992: CV bathymetry masking does not handle unset nodata
A [user on the forums](https://community.naturalcapitalproject.org/t/coastal-vulnerability-model-not-running-due-to-nonetype-error/2813) is reporting an error that's occurring because the nodata value is not defined (and we can't set a numpy array value to be `None`).

Here's the minimum reproducible issue:

```python
import numpy
a = numpy.array((4, 4), dtype=numpy.int32)
a[:] = None
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

phargogh:
This PR handles the case where the bathymetry raster could have an unset nodata.

## Description
Fixes #992

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -57,6 +57,10 @@ Unreleased Changes
 * Crop Production
     * Fixed a bug in both crop production models where the model would error if
       an observed yield raster had no nodata value.
+* Coastal Vulnerability
+    * Fixed a bug that would cause an error if the user's bathymetry layer did
+      not have a defined nodata value.  The user's bathymetry layer should now
+      be correctly preprocessed with or without a nodata value.
 * DelineateIt
     * Watersheds delineated with this tool will now always have a ``ws_id``
       column containing integer watershed IDs for easier use within the routed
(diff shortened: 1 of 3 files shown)


davemfish (commented):
Hey @phargogh , thanks for fixing this. As I look at it I wonder if it makes sense to simply define our own nodata value for the masked raster when it's created in `warp_and_mask_bathymetry`? Instead of trying to pass along the original input's nodata and accommodating `None`?

We're creating a raster that should have only negative values, so would be easy to set the nodata as any positive value.

And when this raster is used in `calculate_wind_exposure` it's expected to contain only negative values or nodata, and I'm not sure that's true if we instantiate with `numpy.empty_like` and then don't pre-fill the array (the case where nodata is `None`).

What do you think?

phargogh:
Sounds good to me @davemfish !  I just pushed a change that would handle this case.

This is also a case that reminded me of #1005 and #1009 , where a possible solution is to handle this when warping rasters.  More of a note to self than anything else, since as you suggest, we have a whole range of values >0 that are suitable for nodata.

davemfish (approved):
Thanks James, looks great!

PR #1011, opened 2022-05-27
WB: Simplify the rules for adding to the recent runs

### LINKED ISSUE #999: WB: Icon on a recent run card is unintuitive
We display a ✅  on the recent run cards when they represent a completed model run with no errors. That was unintuitive as it looks too much like a checkbox that can be toggled on/off. Instead we can print "Model Complete".

Also for some reason this icon is cropped when I view it on Mac, so safer to just use text.

### LINKED ISSUE #1010: Workbench recent jobs cards - replacement rules
Each new run of an invest model should add to the list of recent runs. Previously I thought there should be one exception:

If the same model, workspace, and suffix are used, it might make sense to replace the existing recent run card because the actual results of the run will also get overwritten. But actually, I still thinks it's okay to list these as separate runs. Even if the results no longer exist for the prior run, someone might want to load it's parameters again.

Also, the design of this feature should get simpler if we always just append to the list and never replace.

Now there's also a bug where the card is replaced if the user never closes the model tab before re-running the model. Even if they change the workspace and suffix, the card will be replaced instead of appended.

davemfish:
The new rule is, anytime an invest model run exits, a new job is appended to the list of the recent runs. 

Also some minor enhancements to style and layout of the cards. Here's the new look:
Capture

Fixes #1010
Fixes #999

## Checklist

### DIFF
--- a/workbench/src/renderer/components/HomeTab/index.jsx
+++ b/workbench/src/renderer/components/HomeTab/index.jsx
@@ -143,7 +143,6 @@ class RecentInvestJobs extends React.Component {
               <span className="text-heading">{'Suffix: '}</span>
               <span className="text-mono">{job.argsValues.results_suffix}</span>
             </Card.Title>
-            <Card.Text>{job.description || <em>no description</em>}</Card.Text>
             <Card.Footer className="text-muted">
               <span className="timestamp">{job.humanTime}</span>
               <span className="status">
--- a/workbench/src/renderer/components/HomeTab/index.jsx
+++ b/workbench/src/renderer/components/HomeTab/index.jsx
@@ -128,7 +128,7 @@ class RecentInvestJobs extends React.Component {
         <Card
           className="text-left recent-job-card"
           as="button"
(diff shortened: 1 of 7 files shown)


workbench/src/renderer/components/HomeTab/index.jsx
```diff
@@ -143,13 +143,12 @@ class RecentInvestJobs extends React.Component {
...
             </Card.Title>
-            <Card.Text>{job.description || <em>no description</em>}</Card.Text>
             <Card.Footer className="text-muted">
               <span className="timestamp">{job.humanTime}</span>
-              <span className="status-traceback">
+              <span className="status">
                 {(job.status === 'success'
-                  ? '\u{2705}'
-                  : <em>{job.finalTraceback || ''}</em>
+                  ? <span className="status-success">{_('Model Complete')}</span>
```
emlys, line 150: Agree it looks better without the checkbox icon

emlys (approved):
Thanks @davemfish!

Issue #911, opened 2022-03-22
SDR: sed_deposition has large swaths of nodata in Gura sample data


phargogh:
Reported by Stacie:

> Hey James, following up from the SDR meeting... To reproduce my concern, just run SDR with the Gura sample data and look at the sed_deposition.tif output. There are large swaths of NoData. The only thing they correspond obviously with is the F intermediate output, no other intermediate files are missing data there, nor does it correspond obviously with LULC. I'd love to figure that out/fix it, since our sample data should not produce weird output. Thanks!

After running SDR, I overlayed the `sed_deposition.tif` on top of `what_drains_to_stream.tif` to show that these regions do indeed flow to the stream.

![image](https://user-images.githubusercontent.com/4583897/159570295-a533e561-cc92-49fe-b145-bbe894e09cb7.png)

I can confirm that this is happening on InVEST 3.10.2 with the sample data.

newtpatrol:
It's the DEM causing the problem. In a desperate attempt to be able to use the SDR sample data for training, I downloaded a different DEM for the Gura sample data area, ran it with SDR version 3.12.0, and the output rasters do not have these weird holes! The original DEM data was SRTM, the new DEM data is HydroSHEDS (resampled from 90m to 15m to match the LULC, which is something I wouldn't typically do, but it's fine for this purpose).

emlys:
Thanks for the update @newtpatrol! Does this behavior still seem like a software bug, or do we just need to update the sample data?

emlys:
Investigating this now, some notes:
- the unexpected nodata pixels are not being explicitly set to nodata. They are not being visited by the sediment deposition algorithm at all.
- At least one of the nodata areas appears to originate at a pixel where flow direction is defined, but e_prime is not

emlys:
The sediment deposition function takes three inputs: flow direction, sdr, and e_prime. flow_dir and sdr have valid data in the larger area (same as the LULC and DEM), while `e_prime` has valid_data in the smaller area (same as erosivity). The swaths of nodata originate from points along the edge of the watershed where `flow_dir` is defined and `e_prime` is undefined. 

Because `f` is defined in terms of `e_prime`, `sed_dep` and `f` are undefined where `e_prime` is undefined.
Because `sed_dep` and `f` are defined recursively in terms upslope neighbors, they are undefined for all pixels that are downslope from a pixel where `e_prime` is undefined. The nodata propagates downslope from the edge of the watershed.

[...]

PR #1016, opened 2022-06-28
Rework SDR outputs from the recent PEP "Revising SDR Indicators"

### LINKED ISSUE #601: Clarification of SDR "sediment retention" and "sediment retention index"
SDR provides two sediment retention outputs, and it's not clear how to interpret them. @newtpatrol says:
> Regarding sediment retention and sediment retention index, YES that should be clarified. I still do not understand the difference, and maintain that if we're providing something that is an "index" it should be on a scale of 0-1. Given that I don't currently understand the difference, I'd prefer to have a single sediment retention index provided (not sed_retention, but an index), on a scale of 0-1, which makes it clear that it is not to be interpreted quantitatively, which we've had to deal with on the forum many times.

Need to figure out what these outputs mean, and if that's not possible, consider deprecating one/both, and/or replacing with a 0-1 index (would need science team engagement).

### LINKED ISSUE #887: Deprecate SDR retention outputs
Proposal
---------
This issue is to log the proposal of deprecating `sed_retention.tif`, `sed_retention_index.tif`, and watershed result `sed_retent`. This should be accompanied by a user guide update as well.

- [ ] Deprecate the outputs and related code in source
- [ ] Update the users guide to reflect these changes
- [ ] Have Stacie and Rafa review the users guide update
- [ ] Have Stacie and Rafa review a development version of the model

Background
-------------
There has been recurring confusion over the SDR outputs `sed_retention.tif` and `sed_retention_index.tif`. From a [user forum post](https://community.naturalcapitalproject.org/t/interpreting-sdr-results/1649/9) awhile ago Rich indicated that these outputs were from a bygone era and that they probably should be deprecated in favor of `sed_deposition.tif`. The message from Rich:

>The sed_retention_index.tif and sediment_retention.tif analyses were left over from very old versions of InVEST that was our only “what is the benefit” scenario at the time and before we derived the sediment deposition analysis. In reality, comparing the current state to a state where the entire landscape is bare exposed dirt isn’t really useful. I suspect in all practical cases folks, and you too, will want to use sediment_deposition_[Suffix].tif in lieu of either of the sed_retention rasters.

 We seem to have some agreement from Stacie and Rafa in a Slack channel reflecting this desire. 
From Rafa:
>I'd strongly advocate for deprecating those bare-earth indices as proposed by Rich. Indeed, now that the model calculates sediment retention, deposition those indices are redundant and confusing.

I confirmed the desire was to remove the `sed_retention_index.tif` and `sed_retention.tif` outputs. Stacie pointed out that `sed_retent` should be removed from the watershed output as well.

phargogh:
## Description

This PR includes the changes described in the [_Revising SDR Indicators_ PEP](https://docs.google.com/document/d/1DRWYmOuyprQSXp4cn1M8d5biS8D0P3xsAeBYIqdVLr8/edit) that was recently approved.   Everything requested is, to the best of my knowledge, implemented as requested, but I would really appreciate your careful review to make sure everything in there is as it should be.

I do see that `total_retention.tif` appears to be propagating the issue with `sed_deposition.tif` described in #911  and I have not addressed that in this PR.

The user's guide has not yet been updated.

Fixes #887 
Fixes #601 by way of removing those files

## Checklist


HISTORY.rst
```diff
@@ -42,6 +42,25 @@ Unreleased Changes
       in the list of recent runs.
 * NDR
     * Added parameters to the sample data to support nitrogen calculations.
+* SDR
```
dcdenu4, line 45: Given the magnitude of changes here, I wonder if we should be able to point to a "clean" version of the PEP somewhere? I think we talked about this awhile ago, but it might be nice to track clean PEPs either in the InVEST repo directly or in a new PEP repo? I think it would be useful to link to that PEP here in the history for anyone who wants to know more about the justification for the changes. It would also help us in having something to point forum users to.

Does this or a different form of what I'm getting at here make sense?

HISTORY.rst
```diff
@@ -42,6 +42,25 @@ Unreleased Changes
...
+    * We have made a significant update to the SDR model's outputs:
+      * Legacy outputs ``sed_retention_index.tif`` and ``sed_retention.tif``
+        have been removed from the model.
+      * Two new output rasters have been added:
+        * ``total_retention.tif``, indicating vegetation's contribution to
```
dcdenu4, line 50: In the PEP document it says the 2 new outputs should be:
1) Avoided Export
2) Avoided Erosion

Is `total_retention.tif` the inverse of `avoided_export`?

src/natcap/invest/sdr/sdr.py
```diff
@@ -154,13 +153,13 @@
...
     'rkls_path': 'rkls.tif',
     'sed_export_path': 'sed_export.tif',
-    'sed_retention_index_path': 'sed_retention_index.tif',
-    'sed_retention_path': 'sed_retention.tif',
     'sed_deposition_path': 'sed_deposition.tif',
     'stream_and_drainage_path': 'stream_and_drainage.tif',
     'stream_path': 'stream.tif',
     'usle_path': 'usle.tif',
     'watershed_results_sdr_path': 'watershed_results_sdr.shp',
+    'total_retention_path': 'total_retention.tif',
```
dcdenu4, line 161: Again, inconsistency between PEP naming and code naming. Just want to make sure our source code is aligned with the UG.

src/natcap/invest/sdr/sdr.py
```diff
@@ -566,14 +565,24 @@ def execute(args):
...
-    _ = task_graph.add_task(
-        func=_calculate_sed_retention_index,
+    avoided_erosion_task = task_graph.add_task(
+        func=_calculate_avoided_erosion,
+        args=(
+            f_reg['rkls_path'], f_reg['usle_path'],
+            f_reg['avoided_erosion_path']),
+        dependent_task_list=[rkls_task, usle_task],
+        target_path_list=[f_reg['avoided_erosion_path']],
+        task_name='calculate local retention')
```
dcdenu4, line 575: Is this task name the inverse from the function call name? Would it make sense for the task name to reflect the function name at all? I feel like they are the same thing just said differently.

HISTORY.rst
dcdenu4: Okay, I also see the PEP refer to Avoided Export as Total Retention. I guess it would just be nice to be consistent with whatever the UG goes with.

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -633,28 +633,38 @@ def calculate_sediment_deposition(
...
+                    # dr_i, t_i, and f_i given in the docstring
+                    if sdr_i == 1:
+                        # This reflects property B in the user's guide and is
+                        # an edge case to avoid division-by-zero.
+                        dr_i = 1
+                    else:
+                        dr_i = (downslope_sdr_weighted_sum - sdr_i) / (1 - sdr_i)
+
+                    # Lisa's modified equations
+                    t_i = dr_i * f_j_weighted_sum  # deposition
```
dcdenu4, line 652: What about adding "trapped sediment" to the comment as well?

`# deposition | trapped sediment`

dcdenu4 (changes requested):
Hey @phargogh, this looks pretty good to me and aligns with the PEP from what I can tell. Most of my comments were around terminology for the "total retention" / "avoided export" output.

I also had a larger comment about how we might track PEPs.

dcdenu4:
@phargogh I was also wondering if we want this to go into `main` or if we should be sending this to a `release/3.12` branch?

HISTORY.rst
phargogh: Yeah, I really like the idea of a repo for tracking PEPs.  A separate repo would clarify that it's for the platform in the abstract and not just for InVEST and I suspect we could use other repo features there too if we wanted to, like issues for submitting the PEP, should the PSC approve.  It also hints a little at what we were chatting about earlier, about natcap being an open-source org in some ways, and having a public repo for these would be a part of that.

Python does this as well in their peps repo: https://github.com/python/peps

src/natcap/invest/sdr/sdr_core.pyx
phargogh: Sure!  Added in a4a0af346

src/natcap/invest/sdr/sdr.py
phargogh: Whoops, that's from a botched find-and-replace.  I think at one point we were calling this "local retention", but settled on "avoided erosion".  I added it in the task name. cca41e494

HISTORY.rst
dcdenu4: >to what degree should the PSC be a part of the decision to make such a repo public?

Yep, this is a good point. I will note to bring this up on the next PSC call.

For this item, we could either pass and wait until that future PSC convo or we could ping Lisa to see if she'd give a quick thumbs up and use this as an example for the next PSC call and "ask for forgiveness" in the sense that we went ahead with it this once...

phargogh: I'll just do it as a proof-of-concept and we can ask for forgiveness later.  For a minute there I was worried about how we'd render math in github since that's been difficult in the past, but wait!  [Github added mathjax support just in the last month or two](https://github.blog/2022-05-19-math-support-in-markdown/)!  So I'll make a github markdown page and just post that.  Easy.

phargogh: Added in https://github.com/natcap/peps/blob/main/pep-0010.md.

HISTORY.rst
phargogh: Yeah, they're synonyms.  I re-read the slack thread at https://stanford-natcap.slack.com/archives/CECFUU1DW/p1649109737853559?thread_ts=1647968772.949939&cid=CECFUU1DW and settled on:

* "Avoided Export" instead of "Total Retention"
* "Avoided local erosion" instead of "Avoided erosion"

src/natcap/invest/sdr/sdr.py
phargogh: I've updated this to be more consistent and according to the slack discussion at https://stanford-natcap.slack.com/archives/CECFUU1DW/p1649109737853559?thread_ts=1647968772.949939&cid=CECFUU1DW.

phargogh:
OK @dcdenu4 I'm at a loss about what to do with these test failures.

The problem here is that the PR process merges `task/sdr-retreat...` into `release/3.12` before running the tests, which causes the new regression values to be overwritten with the ones on `release/3.12` and is therefore causing test failures.  My branch is all updated and everything, and I can't think of any other merges I can do to force the right values to be merged into the release branch.

So, if you're OK with it, here's what I propose:
1. Once you're happy with the contents of the PR, let's merge this PR even though these 3 tests are failing
2. If the tests are still failing after the merge (which they probably will), then I'll submit a new PR with you as a reviewer to fix the regression results.

How does that sound?

dcdenu4 (approved):
Thanks @phargogh . That plan sounds good to me. I'll approve and let you go ahead with the merge.

phargogh:
OK.  Merging and then we'll see what happens in the tests.  Will report back here with the status.

phargogh:
[Tests are still failing](https://github.com/natcap/invest/actions/runs/2618354826), but the strange thing is that the SDR tests _do_ have the correct regression values in them, which have been updated from what's on `main`.

PR #1024, opened 2022-07-08
Urban Nature Access - Implementing split populations

phargogh:
## Description

Alright, we've got a shorter one here!  This PR implements split populations in the Urban Nature Access model as described in the [UNA Design Doc](https://docs.google.com/document/d/1lqx75-S605lLxingVjd_yVSGt1tssEFp/edit).  This is a simple calculation where 2 new rasters (indicating greenspace undersupply and oversupply) for each population group are created and summarized in the output vector.  The current implementation does not actually summarize the calculated vectors, as the field values of the summary vector can be derived without another `zonal_statistics` call.

The design doc is a little unclear about what `g`, `gn` and `Rp,gn` mean.  After checking with Roy, I've settled on this structure:
* The user may provide any number of `pop_*` columns in their administrative units vector, e.g. `pop_adult` and `pop_youth`.  The model does not assert that their field values sum to 1, and so the collection of `pop_*` columns could represent an arbitrary number of distinct population groups that may or may not be related.
* The values of `pop_*` are assumed to be in the range 0-1.
* The summary vector now includes a new field each for undersupply and oversupply of greenspace for each population group.

@dcdenu4 @emlys I've requested a review from you because you were on the last UNA PR and so have greater context about the current state of the model.

There's also some minor refactoring, mostly around the admin units vector, that came up with implementing the above.


src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +623,148 @@ def execute(args):
...
+    vector = gdal.OpenEx(target_path, gdal.GA_Update)
+    layer = vector.GetLayer()
+    field = ogr.FieldDefn(id_fieldname, ogr.OFTInteger)
+    layer.CreateField(field)
+
+    layer.StartTransaction()
+    field_id = 0
+    for feature in layer:
+        feature.SetField(id_fieldname, field_id)
+        field_id += 1
```
emlys, line 657:
```suggestion
    for field_id, feature in enumerate(layer):
        feature.SetField(id_fieldname, field_id)
```

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +623,148 @@ def execute(args):
...
+        layer.SetFeature(feature)
+    layer.CommitTransaction()
+    layer = None
+    vector = None
+
+
+def _reclassify_and_multiply(
+        admin_units_raster_path, reclassification_map, supply_raster_path,
+        target_raster_path):
+    """Reclassify the admin units and multiply by greenspace supply.
```
emlys, line 667: I'd suggest clarifying the "before" and "after" of the reclassification, something like
```suggestion
    """Reclassify AOI ids to population group ratios and multiply by greenspace supply.
```

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +623,148 @@ def execute(args):
...
+        key_field (string): The string key field within the vector.
+            ``key_field`` must exist within the vector at ``vector_path``.
+            ``key_field`` is case-sensitive.
+        value_field (string): The string value field within the vector.
+            ``value_field`` must exist within the vector at ``vector_path``.
+            ``value_field`` is case-sensitive.
+
+    Returns:
+        attribute_map (dict): A dict mapping ``key_field`` to
+            ``value_field`` value.
```
emlys, line 731:
```suggestion
        attribute_map (dict): A dict mapping each ``key_field`` value to
            the corresponding ``value_field`` value.
```

tests/test_urban_nature_access.py
```diff
@@ -384,3 +389,72 @@ def test_split_greenspace(self):
 
         admin_vector = None
         admin_layer = None
+
```
emlys, line 392: It might be worth a unit test for `_reclassify_and_multiply`.

src/natcap/invest/urban_nature_access.py
```diff
@@ -664,24 +882,32 @@ def _admin_level_supply_demand(
...
+    driver = ogr.GetDriverByName('GPKG')
+    driver.CopyDataSource(source_vector, target_admin_unit_vector_path)
+    source_vector = None
 
     target_vector = gdal.OpenEx(target_admin_unit_vector_path, gdal.GA_Update)
     target_layer = target_vector.GetLayer()
 
     supply_sum_fieldname = 'SUP_DEMadm_cap'
     undersupply_fieldname = 'Pund_adm'
     oversupply_fieldname = 'Povr_adm'
```
emlys, line 895: Since we're making a geopackage, these field names could be more verbose, if you want

src/natcap/invest/urban_nature_access.py
```diff
@@ -704,10 +930,18 @@ def _admin_level_supply_demand(
...
+        feature.SetField(undersupply_fieldname, undersupply)
+        feature.SetField(oversupply_fieldname, oversupply)
+        for pop_groupname in population_groupnames:
+            group_proportion = feature.GetField(f'pop_{pop_groupname}')
+            feature.SetField(
+                f'{oversupply_fieldname}_{pop_groupname}',
+                oversupply * group_proportion)
+            feature.SetField(
+                f'{undersupply_fieldname}_{pop_groupname}',
+                undersupply * group_proportion)
```
emlys, line 944: As I'm understanding it, this method assumes that the different population groups are evenly distributed across the admin unit. Let's be sure to emphasize that in the user guide.

tests/test_urban_nature_access.py
```diff
@@ -384,3 +389,72 @@ def test_split_greenspace(self):
...
+                _read_and_sum_raster(group_supply_raster_path),
+                rtol=1e-6
+            )
+
+        # verify additional rasters exist and have reasonable values relative
+        # to the total rasters
+        #   undersupply of greenspace
+        #   opversupply of greenspace
+        # Also, fields must exist in output vector and have correct values,
+        # whicih should also match the raster.
```
emlys, line 460: Want to keep this comment?

emlys (changes requested):
thanks @phargogh I had a few small suggestions!

src/natcap/invest/urban_nature_access.py
phargogh: Since the line with this change was > 80 chars, I added a little more detail in the body of the docstring in 2bd03a6c5

src/natcap/invest/urban_nature_access.py
```diff
@@ -285,6 +290,20 @@ def execute(args):
         target_path_list=[file_registry['aligned_population']],
         task_name='Resample population to LULC resolution')
 
+    admin_unit_reprojection_task = graph.add_task(
+        _reproject_and_identify,
+        kwargs={
+            'base_vector_path': args['admin_unit_vector_path'],
```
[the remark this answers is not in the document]
phargogh: Sure, we can do that!  The catch about AOIs is that the model expects them to be non-overlapping geometries, so this has been noted in the ARGS_SPEC.  59d33fa02

src/natcap/invest/urban_nature_access.py
phargogh: This one is a little tricky because of the duplication of `'value'`, since `key` and `value` are the conventional words for the dictionary accessor and accessed data, respectively.  How about this instead?

```
attribute_map (dict): A dict mapping each ``key_field`` key to 
    the corresponding ``value_field`` value.
```

In case that works, I've added it in 02a52e981

tests/test_urban_nature_access.py
phargogh: Nope!  It's already implemented. Removed in 24f891118

src/natcap/invest/urban_nature_access.py
phargogh: Yes, that's true.  As Roy explained it over slack:

> This step is necessary because data on the general population numbers is often available at high resolution, whereas for other demographic statistics this data often is only available at neighborhood level at best.

I'll add a note to this to the UNA issue (#722 ) so I'm sure to tackle this in the UG.

tests/test_urban_nature_access.py
phargogh: 356f8e722

src/natcap/invest/urban_nature_access.py
phargogh: Yes, that's true for the geopackage specification!  In this case, this is how the UG defines these fieldnames, and I'm not yet sure that folks will be happy with using a geopackage in the first place.  So for the time being, I'd like to leave them as-is, but then revisit just before the end, probably when doing the UG.  Does that sound OK?

I've added a note to self about this in https://github.com/natcap/invest/issues/722#issuecomment-1186003044

src/natcap/invest/urban_nature_access.py
emlys: that does read better to me, thanks!

src/natcap/invest/urban_nature_access.py
emlys: Sounds good!

src/natcap/invest/urban_nature_access.py
```diff
@@ -97,19 +100,34 @@
...
-        'admin_unit_vector_path': {
+        'aoi_vector_path': {
             'type': 'vector',
             'name': 'administrative boundaries',
             'geometries': spec_utils.POLYGONS,
-            'fields': {},  # TODO, complete required fields (if any)
-            'about': "",  # TODO, will know more about this when I implement.
+            'fields': {
+                "pop_[POP_GROUP]": {
+                    "type": "number",
```
emlys, line 109: To be more specific, this could be `"ratio"` instead of `"number"`, and then we could remove the explanation `(numeric values in the range 0-1)` in the about text.

src/natcap/invest/urban_nature_access.py
```diff
@@ -97,19 +100,34 @@
...
+                    "required": False,
+                    "about": gettext(
+                        "The proportion (numeric values in the range 0-1) of "
+                        "the population within this region belonging to the "
+                        "identified population group."
+                    ),
+                }
+            },
+            'about': gettext(
+                "Non-overlapping regions (typically administrative "
```
emlys, line 120: Out of curiosity, which operation requires that the polygons are non-overlapping?

emlys: And is that something that's worth validating in `validate`?

emlys (changes requested):
thanks @phargogh! Just had a couple more comments. It looks like the workbench tests are failing since that spec key changed. And the readthedocs build is failing with an outdated import in `src/natcap/invest/ui/urban_nature_access` -  `from natcap.invest import MODEL_METADATA` should be `from natcap.invest.model_metadata import MODEL_METADATA`

src/natcap/invest/urban_nature_access.py
phargogh: Makes sense!  Patched in 0ff3baffd

src/natcap/invest/urban_nature_access.py
phargogh: Currently, the reason for this is that we rasterize a population group field all at once in [`_rasterize_aois`](https://github.com/natcap/invest/pull/1024/files#diff-3c80dfa184fdd56cb7d4dba74d5215d5f1f4ab52284cd104e1dff04d8035a952R572), so each pixel can only belong to a single polygon.

But the truth is that this is really the only function in the whole model that actually relies on geometries being non-overlapping, and there's no real reason why this should have to be the case ... supporting overlapping admin units would just be a little more looping.  So, I'll work this in.

And then that will remove the need to validate that constraint.

- [ ] Refactor to support overlapping admin units
- [ ] Remove this note from the `ARGS_SPEC`

phargogh: OK @emlys !  I've refactored this in bca7408db, but the outputs definitely do not look particularly good since we're using pygeoprocessing's definition of disjoint, where polygons can't even be touching (see below).  So I'm going to ask what the science team would like to do about this, and if they want something specific I can PR that in separately.

In any case, this slight refactor is ready for review!

Screen Shot 2022-07-18 at 10 15 47 PM

phargogh: OK I was dissatisfied with my solution, because while it's the best general-case solution, it leaves users with AOIs that touch but do not overlap in the lurch with outputs that are harder to interpret.  667e1b4d7 does a little more work to determine whether we need to do the fracturing above.  If we do, then we do and you'll get the disjoint sets above.  If not, then the whole admin units vector is treated as a single layer, as below:
Screen Shot 2022-07-18 at 10 57 23 PM

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +655,376 @@ def execute(args):
...
+    Args:
+        base_vector_path (string): The string path to the source vector.
+        target_projection_wkt (string): The WKT of the target projection.
+        target_path (string): The string path to where the new vector should be
+            saved.
+        driver_name (string): The GDAL driver name of the target vector.
+        id_fieldname (string): The name of the ID field.  A new field with this
+            name and an integer type will be created in the target vector.
+            Each feature in the target vector will be assigned a unique integer
+            ID.
```
dcdenu4, line 859: Awesome, I like this workflow in general when handling vectors that don't have a specified unique ID.

src/natcap/invest/urban_nature_access.py
```diff
@@ -520,21 +558,91 @@ def execute(args):
             population_alignment_task,
         ])
 
-    aggregate_admin_units_task = graph.add_task(
+    # only do split population if there are population group fields in the
+    # admin units vector
+    split_population_fields = list(
+        filter(lambda x: re.match(POP_FIELD_REGEX, x),
+               validation.load_fields_from_vector(
+                   file_registry['reprojected_aois'])))
```
dcdenu4, line 566: Cool.

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +655,376 @@ def execute(args):
...
+    all_geometries = []
+    vector = ogr.Open(aoi_vector_path)
+    layer = vector.GetLayer()
+    area_sum = 0
+    all_fids = []
+    for feature in layer:
+        geom = feature.GetGeometryRef()
+        all_geometries.append(shapely.wkt.loads(geom.ExportToWkt()))
+        area_sum += geom.Area()
+        all_fids.append(feature.GetFID())
```
dcdenu4, line 682: I think in `_reproject_and_identify` we added our own ID field. Should we be relying on that here instead of FID?

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +655,376 @@ def execute(args):
...
+        aoi_vector_path, ok_for_geometries_to_touch=True)
+
+
+def calculate_disjoint_polygon_set(
+        vector_path, layer_id=0, bounding_box=None,
+        ok_for_geometries_to_touch=False):
+    """Create a sequence of sets of polygons that don't overlap.
+
+    Determining the minimal number of those sets is an np-complete problem so
+    this is an approximation that builds up sets of maximal subsets.
```
dcdenu4, line 703: Smart.

dcdenu4 (changes requested):
Hey @phargogh, this looks great. There's a lot of cool algorithms and things happening in here. I had a few comments and questions.

I'm curious, do you think we should be unit testing the disjoint functions?

I also had some confusion about the admin vector input and how to think about that. So we have multiple polygons representing admin units with some pop_* attributes. For each disjoint polygon we want to calculate under/over stats for each pop_*. If we're just doing status for each polygon do we care if there is overlap? Thanks for helping me see what I'm missing!

src/natcap/invest/urban_nature_access.py
phargogh: OK!  So I'm just getting back to this and this really turned into an interesting conversation!

After discussing tradeoffs with the science team, they decided that we should not go through the extra effort of supporting possibly-overlapping polygons.  Rev 53af0c709 reverts the state of the codebase to before I started trying to make the above work reasonably.  In case the user provides a vector with admin units that overlap (greater than a reasonable numerical tolerance), I'll add a function that logs a warning.

That being said, it seems like a useful thing for cases like this, where we may not actually have disjoint polygons that can be rasterized in one pass, so we'll handle this in https://github.com/natcap/pygeoprocessing/issues/269.

src/natcap/invest/urban_nature_access.py
phargogh: Yes!  It's always nice to be able to uniquely identify the AOIs, and in this case it turns out to be necessary for when we rasterize them a little later on in `execute`.

src/natcap/invest/urban_nature_access.py
phargogh: Ah this was within the disjoint polygon function, I think.  Probably, yes (might as well!), but I've removed the disjoint polygon handling per science team request.

src/natcap/invest/urban_nature_access.py
phargogh: Ha, I really can't claim credit for this ... this is Rich's handiwork: https://github.com/natcap/pygeoprocessing/blob/main/src/pygeoprocessing/geoprocessing.py#L2232

it _is_ a cool algorithm, though, and very handy!  I copied this function in here so that I could modify the logic determining which features are disjoint since we have an interesting case where we expect our admin geometries to be nonoverlapping.

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +655,376 @@ def execute(args):
...
+
+            if intersect_poly_fid == poly_fid or polygon.intersects(
+                    shapely_polygon_lookup[intersect_poly_fid]):
+                poly_intersect_lookup[poly_fid].add(intersect_poly_fid)
+        polygon = None
+    LOGGER.info(
+        'poly intersection feature lookup 100.0%% complete on %s',
+        os.path.basename(vector_path))
+
+    # Build maximal subsets
```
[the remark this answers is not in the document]
phargogh: Also Rich!

tests/test_urban_nature_access.py
```diff
@@ -384,3 +432,65 @@ def test_split_greenspace(self):
 
         admin_vector = None
         admin_layer = None
+
+    def test_split_population(self):
```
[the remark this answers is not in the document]
phargogh: Yeah, for the sake of being thorough it might be good to, but now that the overlapping functionality isn't there, what do you think about this?

phargogh:
Thanks for your reviews @emlys @dcdenu4 !  I've pushed a couple new changes that:

1. Undoes the disjoint polygon set stuff per science team request (https://github.com/natcap/invest/pull/1024/commits/53af0c70982eb880a97181caad7f24818ec3e1d4)
2. Adds a function to test for polygon overlap in a layer which is then used to log a warning (https://github.com/natcap/invest/pull/1024/commits/eb747989144cdb23a7397729dcc7d72afed9aff2)

So this also should simplify the PR considerably, moving the disjoint polygon logic update over to https://github.com/natcap/pygeoprocessing/issues/269.

tests/test_urban_nature_access.py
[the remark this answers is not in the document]
dcdenu4: No need!

src/natcap/invest/urban_nature_access.py
```diff
@@ -520,21 +556,84 @@ def execute(args):
             population_alignment_task,
         ])
 
-    aggregate_admin_units_task = graph.add_task(
+    # only do split population if there are population group fields in the
+    # admin units vector
```
dcdenu4, line 560: Somewhat nit-picky but it seems like we've been moving away from `admin` to `aoi`. Would it make sense for this to be "in the aoi vector"?

dcdenu4 (approved):
This is looking good to me @phargogh . Do we have sample data for this model? One thing I wanted to do with this PR was test it through the UI.

dcdenu4:
Looks like a few HRA tests are failing.

src/natcap/invest/urban_nature_access.py
phargogh: Makes sense to keep it consistent!  I've updated this to reflect both possibilities in 032cf92b9.

src/natcap/invest/urban_nature_access.py
```diff
@@ -547,6 +646,182 @@ def execute(args):
...
 
 
+def _geometries_overlap(vector_path):
+    """Check if the geometries of the vector's first layer overlap.
+
+    Args:
+        vector_path (string): The path to a GDAL vector.
+
+    Returns:
+        overlaps (bool): Whether there's numerically significant overlap
```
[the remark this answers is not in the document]
phargogh: Yeah, I _think_ this is how things used to be written in the Google style, but when I looked at [the updated docs](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) I had it wrong!  So fffcb1447 updates it to be correct.

But you're right, this is very like type annotations and can in part be replaced by type annotations.

phargogh:
Well @dcdenu4 @emlys I merged in `release/3.12` to try to address the HRA failures forgetting that I hadn't yet merged `release/3.12` into `feature/urban-nature-access` so now there are a whole lot of changes in this PR.  Hopefully we're close enough to the end of this PR where it's not a big deal and I'm sorry in advance!

emlys (approved):

phargogh:
With approvals in the books, let's merge this in!  Thanks @emlys @dcdenu4 !

PR #1042, opened 2022-07-29
Bugfix CV landmass-AOI intersection

### LINKED ISSUE #1040: CV: edge case in landmass, AOI intersection
This turned up here: https://community.naturalcapitalproject.org/t/coastal-vulnerability-typeerror-cannot-unpack-non-iterable-nonetype-object/2949/7

```
2022-07-27 11:11:13,542 (natcap.invest.utils) utils.prepare_workspace(167) ERROR Exception while executing natcap.invest.coastal_vulnerability
File “natcap\invest\utils.py”, line 165, in prepare_workspace
File “natcap\invest\cli.py”, line 486, in main
File “natcap\invest\coastal_vulnerability.py”, line 476, in execute
File “natcap\invest\coastal_vulnerability.py”, line 807, in prepare_landmass_line_index_and_interpolate_shore_points
File “shapely\ops.py”, line 113, in linemerge
File “shapely\geometry\multilinestring.py”, line 49, in init
File “shapely\geometry\multilinestring.py”, line 141, in geos_multilinestring_from_py
TypeError: cannot unpack non-iterable NoneType object
```

In this case the landmass polygon has two vertices that are basically identical. The model converts the polygons to lines and then intersects each line with the AOI polygon. We end up trying to intersect a polygon with a line where the start and end coordinate pairs are the same, like this:

```python
from shapely.geometry import Point, LineString
polygon = Point(0, 0).buffer(2)
polygon.intersection(LineString([(1, 1), (1, 1)]))
```

This returns an object of the expected type `shapely.geometry.linestring.LineString`, but it's `coords` property is an empty sequence. 

The `LineString` from the user's data has these coordinates, but the return value from the intersection is the same as the simple example above.
```
(420275.798461657,  6042939.285698998)
(420275.7984616568, 6042939.285698998)
```

We actually already try to guard against a `LineString` like this in CV, probably because duplicate vertices are not that uncommon in GIS data. But in this case we probably just need to be using `numpy.isclose` to do so.

davemfish:
I think the bug is well-defined in #1040 . This PR adds a test case and a fix. I had a hard time re-creating the bug with synthetic data, but ultimately was able to by using precisely the same vertices that appeared in the user's data. 

Fixes #1040 

## Checklist

### DIFF
--- a/tests/test_coastal_vulnerability.py
+++ b/tests/test_coastal_vulnerability.py
@@ -1283,28 +1283,32 @@ class CoastalVulnerabilityTests(unittest.TestCase):
             workspace_dir, 'shore_points.gpkg')
         model_resolution = 1
         srs = osr.SpatialReference()
-        srs.ImportFromEPSG(26910)  # UTM Zone 10N
+        srs.ImportFromEPSG(32634)  # UTM Zone 34N
         wkt = srs.ExportToWkt()
 
-        aoi_poly = Point(0, 0).buffer(10)
+        # Nearly identical vertices are from real data that raised an error
+        landmass_coords = [
+            (420274.0, 6042938.0),
+            (420275.7984616568, 6042939.285698998),  # nearly identical
+            (420275.7984616570, 6042939.285698998),  # nearly identical
+            (420276.0, 6042938.0),
+            (420274.0, 6042938.0)]
+        landmass_geometry = Polygon(landmass_coords)
+        aoi_poly = Point(420274.0, 6042938.0).buffer(10)  # contains landmass
         pygeoprocessing.shapely_geometry_to_vector(
             [aoi_poly], aoi_path, wkt, 'GeoJSON')
-
-        coord_list = [
-            (-2, -2),
-            (-2, 2),
-            (-2.0000000002, 2.0000000002),
-            (2, 2),
-            (2, -2),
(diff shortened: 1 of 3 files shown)


tests/test_coastal_vulnerability.py
```diff
@@ -1267,6 +1267,49 @@ def count_shore_points(landmass_geometries):
...
+        srs.ImportFromEPSG(32634)  # UTM Zone 34N
+        wkt = srs.ExportToWkt()
+
+        # Nearly identical vertices are from real data that raised an error
+        landmass_coords = [
+            (420274.0, 6042938.0),
+            (420275.7984616568, 6042939.285698998),  # nearly identical
+            (420275.7984616570, 6042939.285698998),  # nearly identical
+            (420276.0, 6042938.0),
+            (420274.0, 6042938.0)]
```
phargogh, line 1295: This is a really nicely constructed bit of sample data, and I particularly like the `# nearly identical` highlight to emphasize what is relevant to the test.

phargogh (approved):
This is a really neat bug, a nice fix and a good test.

👍

phargogh:
There's still one test running, but it's been running for over 3 hours as I'm writing this so I'm going to just take a chance and say it's probably a fluke and merge anyways.

PR #1046, opened 2022-08-03
Task/reformat ndr effective retention

phargogh:
This PR is a refactor of the code that represents NDR equation 41 to make it match the UG more closely (see [UG chapter for reference](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/ndr.html#surface-ndr)).

As far as I can tell, the current implementation is working fine and has been for some time, but the implementation doesn't look like what's in the UG, which makes it a headache to pull apart for maintenance.  Practically, I was messing with this code for a science support request for Eric Lonsdorf and @chrisnootenboom where they wanted to know which of the 3 options in equation 41 is selected for each pixel, and the approach for this isn't as clear with the 3.11 implementation.

I did compare the output produced by this updated code against InVEST 3.11 and have found the outputs to be identical:
```bash
$ md5sum effective_retention_p_*.tif
508225763f15b4cd243fc247135f3b62  effective_retention_p_311-main.tif  # <-- this is run against natcap/invest@b57ab2049ab603d876c087217d43ff8b63abfb46
508225763f15b4cd243fc247135f3b62  effective_retention_p_lonsdorf.tif  # <-- this is the updated version
```

In case I'm wrong, I added a note to `HISTORY.rst` about this refactor so that future us has a reference point.

## Checklist


src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -529,18 +529,24 @@ def ndr_eff_calculation(
 
                     neighbor_effective_retention = (
                         effective_retention_raster.get(ds_col, ds_row))
-                    if neighbor_effective_retention >= retention_eff_lulc:
-                        working_retention_eff += (
-                            neighbor_effective_retention) * outflow_weight
-                    else:
+
+                    # Case 1: downstream neighbor is a stream pixel
+                    if neighbor_effective_retention == 0:
```
emlys, line 534: To be even more similar to the user's guide, this could directly check the stream raster instead: `if stream_raster.get(ds_row, ds_col) == 1`. But the way you have it is more efficient. I think it's good either way.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -529,18 +529,24 @@ def ndr_eff_calculation(
 
                     neighbor_effective_retention = (
                         effective_retention_raster.get(ds_col, ds_row))
-                    if neighbor_effective_retention >= retention_eff_lulc:
```
emlys, line 532: I thought there might be an edge case here if `neighbor_effective_retention` and `retention_eff_lulc` are both 0, i.e. both pixels are part of a stream. But the result is still the same! 🎉

emlys (commented):
Thanks @phargogh! Just one minor suggestion about wording.

src/natcap/invest/ndr/ndr_core.pyx
phargogh: Yeah, comparing `int`s here should be faster than possibly incurring a cache miss on the stream managed raster.  That being said, I think a module constant here would help clarify that we're checking for a stream, so I've added one in 13fac9f51

phargogh:
Thanks @emlys, back to you!

emlys (approved):
thanks!

PR #1051, opened 2022-08-18
Merge `release/3.12` into `feature/urban-nature-access`

phargogh:
Related to #1024 , this PR merges `release/3.12` into `feature/urban-nature-access`.  A separate PR is needed because of some merge conflicts.


phargogh:
@dcdenu4 I think this PR is fine even with the 2 failing workbench tests ... that could easily be because the port we're trying to use is already in use elsewhere on the OS.  All of the other tests are passing.

emlys (approved):

dcdenu4 (approved):
Thanks James!

PR #1053, opened 2022-08-18
Merge `main` into `release/3.12`

phargogh:
This PR merges `main` into `release/3.12`.

PR #1056, opened 2022-08-19
Bugfix/hq 1039 unnecessary warning

### LINKED ISSUE #1039: HQ: a warning about undefined nodata is too alarming
As discussed here: https://community.naturalcapitalproject.org/t/warning-raster-has-undefined-nodata-value-in-invest-3-11-0-workbench/2937

Here's the warning and it's context:

```python
def _raster_values_in_bounds(raster_path_band, lower_bound, upper_bound):
    """Check raster values are between ``lower_bound`` and ``upper_bound``.

    Check that the raster has values ``lower_bound`` <= x <= ``upper_bound``.
    Nodata values are ignored.

    Args:
        raster_path_band (tuple): a 2 tuple for a GDAL raster path with
            the form (filepath, band index) to the raster on disk.
        lower_bound (int): integer for the lower bound of raster values,
            inclusive.
        upper_bound (int): integer for the upper bound of raster values,
            inclusive.

    Returns:
        True if values are within range and False otherwise.
    """
    raster_info = pygeoprocessing.get_raster_info(raster_path_band[0])
    raster_nodata = raster_info['nodata'][0]

    if raster_nodata is None:
        LOGGER.warning(
            f"Raster has undefined NODATA value for {raster_path_band[0]}.")
        # If raster nodata is None then set to _OUT_NODATA to use for masking
        # where in this case nodata_mask will be all False.
        raster_nodata = _OUT_NODATA

    values_valid = True

    for _, raster_block in pygeoprocessing.iterblocks(raster_path_band):
        nodata_mask = ~utils.array_equals_nodata(raster_block, raster_nodata)
        if ((raster_block[nodata_mask] < lower_bound) |
                (raster_block[nodata_mask] > upper_bound)).any():
            values_valid = False
            break

    return values_valid
```

I was going to suggest just changing it to `debug`, but actually I think this function might work even better if `raster_nodata` is left as `None` so long as `utils.array_equals_nodata` can handle `None`. And it just so happens that James is making this possible in https://github.com/natcap/invest/pull/1026

So now I think my suggestion is to wait for that PR, and then remove this whole chunk:
```python
    if raster_nodata is None:
        LOGGER.warning(
            f"Raster has undefined NODATA value for {raster_path_band[0]}.")
        # If raster nodata is None then set to _OUT_NODATA to use for masking
        # where in this case nodata_mask will be all False.
        raster_nodata = _OUT_NODATA
```

As it is, there is the _extremely unlikely_ case where a value of `_OUT_NODATA` was actually present in the raster, but will now get masked out allowing this function to return `True` when it should not.

What do you think @dcdenu4 ?

davemfish:
This PR removes a check and warning about `nodata is None` because as of a recent HRA PR, `utils.array_equals_nodata` handles the `None` case for us.

Also, it turns out that it's okay for a threat raster to have undefined nodata. The only requirement for those rasters is that they only have values between 0 and 1.

Fixes #1039

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -46,6 +46,9 @@ Unreleased Changes
     * Fixed a bug where using unaligned rasters in the preprocessor would cause
       an error.  The preprocessor will now correctly align input landcover
       rasters and determine transitions from the aligned rasters.
+* Habitat Quality
+    * Removed a warning about an undefined nodata value in threat rasters
+      because it is okay for a threat raster to have an undefined nodata value.
 * HRA
     * Fixed an issue with risk calculations where risk values would be much
       lower than they should be.  Risk values are now correctly calculated.
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -994,14 +994,6 @@ def _raster_values_in_bounds(raster_path_band, lower_bound, upper_bound):
     """
     raster_info = pygeoprocessing.get_raster_info(raster_path_band[0])
     raster_nodata = raster_info['nodata'][0]
-
-    if raster_nodata is None:
-        LOGGER.warning(
-            f"Raster has undefined NODATA value for {raster_path_band[0]}.")
-        # If raster nodata is None then set to _OUT_NODATA to use for masking
-        # where in this case nodata_mask will be all False.
-        raster_nodata = _OUT_NODATA
-
     values_valid = True
 
     for _, raster_block in pygeoprocessing.iterblocks(raster_path_band):


davemfish:
Hmm the SDR tests are failing here. I expect that will be resolved maybe by merging #1053

davemfish:
This is now waiting on #1061 if we want to see passing RTD.

davemfish:
> This is now waiting on #1061 if we want to see passing RTD.

I don't know how to re-run the RTD build, it's outside the control of GHA. But I think it's okay to proceed with this PR anyway.

PR #1058, opened 2022-08-22
WB: keep the important part of filepaths visible in text inputs

### LINKED ISSUE #855: Dealing with overflow in text input fields
Input text fields should default to display the right-most end of their text content when they contain text that is too long to display completely. That way the more important part of the filepaths is always visible. 

Inputs should also not scroll automatically in either direction when a user moves focus off of the input.

davemfish:
Fixes #855 

At first I thought this would be doable with CSS. But these simple-ish solutions didn't seem to work for `<input>` elements. https://stackoverflow.com/questions/38222696/issue-with-the-direction-rtl-css-property

Instead I went for a javascript-based horizontal scrolling approach, but that turned out to get complicated and almost makes we want to look at CSS again. But since this is working, I'll open a PR for it.

Since these are "controlled" inputs, their values are controlled by parent component props. So we have to be careful to trigger the scroll _after_ the value is updated. But we also don't want to scroll after any value change. We don't want to trigger scrolls while someone is typing. We only want to scroll after the input is populated by some means other than typing. That meant adding some more component state, props, and a `useEffect` hook.

This now seems to work for,
* loading text via the `input's` browse button
* drag-and-dropping a file into the text box
* using the Load Parameters button from an already open invest model tab
* opening an invest model tab with pre-loaded arg values (i.e. a Recent Run, or the Open button on the home screen)

I'm not sure how to cover this sort of thing by tests, since it's not functional, only visual.

## Checklist


workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -125,174 +125,188 @@ function dragLeavingHandler(event) {
...
-    // Messages with this pattern include validation feedback about
-    // multiple inputs, but the whole message is repeated for each input.
-    // It's more readable if filtered on the individual input.
-    const pattern = 'Bounding boxes do not intersect';
-    if (validationMessage.startsWith(pattern)) {
-      validationMessage = filterSpatialOverlapFeedback(
-        validationMessage, value
-      );
-    }
+export default function ArgInput(props) {
```
davemfish, line 128: This diff of this function is extra "different" because the indent changed on this block. I'll drop a note where the important new bits are.

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -125,174 +125,188 @@ function dragLeavingHandler(event) {
...
-      );
+  // Occasionaly we want to force a scroll to the end of input fields
+  // so that the most important part of a filepath is visible.
+  // scrollEventCount changes on drop events and on use of the browse button.
+  // Also depend on [value] so we don't scroll before the value updated.
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.scrollLeft = inputRef.current.scrollWidth;
     }
+  }, [scrollEventCount, value]);
```
davemfish, line 157: new hook.

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -125,174 +125,188 @@ function dragLeavingHandler(event) {
...
-          as="select"
           name={argkey}
-          value={value}
+          type="text"
+          placeholder={placeholderText}
+          value={value || ''} // empty string is handled better than `undefined`
           onChange={handleChange}
-          onFocus={handleChange}
+          onFocus={handleFocus}
+          onBlur={(e) => e.target.scrollLeft = e.target.scrollWidth}
```
davemfish, line 272: new `onBlur`

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -125,174 +125,188 @@ function dragLeavingHandler(event) {
...
+        {dropdownOptions.map(
+          (opt) => <option value={opt} key={opt}>{opt}</option>
+        )}
+      </Form.Control>
+    );
+  } else {
+    form = (
+      <React.Fragment>
         <Form.Control
+          ref={inputRef}
```
davemfish, line 264: new `ref`

emlys (approved):
thanks @davemfish! I tried it out and it works for me. I did notice in the original issue you said 
> Inputs should also not scroll automatically in either direction when a user moves focus off of the input.

I think that somewhat conflicts with the current implementation? I like the way it works now, just want to make sure we're not missing something.

davemfish:
> thanks @davemfish! I tried it out and it works for me. I did notice in the original issue you said
> 
> > Inputs should also not scroll automatically in either direction when a user moves focus off of the input.
> 
> I think that somewhat conflicts with the current implementation? I like the way it works now, just want to make sure we're not missing something.

Oh right. I agree with you and disagree with my past self on this. I think it's okay to scroll to the end when focus moves out. Thanks for trying it out! I'll go ahead and merge since you approved.

PR #1062, opened 2022-08-24
Update SDR filenames for consistency with ug chapter

phargogh:
This updates filenames in SDR for consistency with the SDR UG chapter per recent discussions on the filenames for improved readability.


dcdenu4 (approved):
Thanks @phargogh!

Issue #819, opened 2022-01-14
HRA risk reclassification always turning up low-risk


phargogh:
Jade is trying to reproduce a prior analysis with HRA and noticed that HRA is producing noticeably different outputs from the version of HRA used back in 2013/2014 for the analysis.

Looking in to the model a bit more, the risk reclassification function `_reclassify_risk_op` does not appear to work the way it's described in the user's guide.

Working on this in `bugfix/819-hra-risk-reclassification-incorrect-for-jade`

phargogh:
Also related, https://github.com/natcap/invest/issues/835 identified an issue resulting from a pandas version bump that was causing some silent failures.  Perhaps there's an issue with how we're loading or using pandas dataframes in HRA (outside of that particular issue) that's causing toms issues to the risk classification?

phargogh:
Well, the HRA rewrite feels like it's nearly there. Jade's seeing good results compared with the fixes that we had made to the current HRA implementation and the model appears to be producing reasonable results.

# Next actionable steps:
- [ ] implement the datastack archiving function
- [ ] Update the user's guide to better describe the mathy bits like how distance weighting and high/med/low risk classifications are calculated.  See [this doc](https://docs.google.com/document/d/1WbLLU-knitvFX1sQdp8_3TNMQbAYCzZJ6-M7TAdsxJE/edit) for the start of this.
- [ ] Complete the suite of tests for the refactored HRA

phargogh:
I'm catching up on the state of HRA before submitting it as a PR for review.  I'm partway through correcting some unexpected test failures and ensuring we have good test coverage.

For the sake of completing this PR, I'd like to finish the datastack functionality later on.

## Next actionable steps:

- [ ] Complete tests, with coverage
- [ ] Make sure target filepaths match what we expect
- [ ] Update the UG chapter in a PR for review.
- [ ] Go through the model and address any remaining TODOs

phargogh:
I just pushed a new change to add  R_%{HIGH,MEDIUM,LOW} columns to the summary table in response to https://community.naturalcapitalproject.org/t/hra-summary-statistic-missing-values-and-0/2755/4.  Still need to summarize the table across all stressors to a given habitat.

davemfish:
fixed in #1026

PR #1064, opened 2022-08-25
Handling a ZeroDivisionError case in HRA.

phargogh:
This case occurs when a subregion might not overlap a particular habitat
or stressor.  RE:#819

I didn't update HISTORY because this HRA rewrite hasn't yet been released.  Tests have been updated to cover this case, and have been confirmed to trigger the issue in the absence of the fix.


davemfish (approved):
Looks good, thanks @phargogh

PR #1083, opened 2022-09-15
Correct HRA summary statistics table

### LINKED ISSUE #1080: HRA summary statistics table not producing correct outputs
Jade reported that the 3.12.0 version of HRA doesn't produce correct outputs for the percentages of risk classifications for HIGH, MEDIUM, and LOW (`R_%{HIGH,MEDIUM,LOW}`).  I have confirmed that this is an issue on the Belize inputs that Jade sent me.

The summary statistics table function is pretty complicated and could be greatly simplified once https://github.com/natcap/pygeoprocessing/issues/270 is implemented.

phargogh:
## Description

This PR remains a work in progress until https://github.com/natcap/pygeoprocessing/pull/271 is resolved and released.

Fixes #1080 

The user's guide will need an update for this change.

## Checklist


phargogh:
Tests are currently failing pending #1134

tests/test_hra.py
```diff
@@ -982,6 +911,147 @@ def test_exception_invalid_decay(self):
...
+
+        # For the sake of testing this function more rigorously, creating a new
+        # classification path for the per-habitat summary classification
+        # raster.
+        per_habitat_classifications = {
+            pairwise_raster_dicts[0]['habitat']: os.path.join(
+                self.workspace_dir, 'summary_classes.tif')
+        }
+        pygeoprocessing.numpy_array_to_raster(
+            numpy.array([[2, 3, 2, 3]], dtype=numpy.int8),
```
davemfish, line 949: Are these values in the array directly related to the expected values of the summary stats? If so, is there a way -- either a variable or just a comment -- to make that connection?

tests/test_hra.py
```diff
@@ -982,6 +911,147 @@ def test_exception_invalid_decay(self):
...
+                aoi_vector_path, pairwise_raster_dicts,
+                per_habitat_classifications, target_summary_csv_path)
+            expected_records = [
+                {**std_record,
+                 **{'SUBREGION': 'first region',
+                    'STRESSOR': '(FROM ALL STRESSORS)'},
+                    'R_%HIGH': 50.0,
+                    'R_%MEDIUM': 50.0,
+                    'R_%LOW': 0,
+                    'R_%NONE': 0,
```
davemfish, line 1000: I'm a bit confused by these numbers. Somehow the % from all stressors are different than the % from one stressor, even though in this test there is only one stressor, right?

davemfish (commented):
Hey @phargogh , I had a couple questions in the test to try to clarify why the expected values are what they are. Thanks for taking a look!

HISTORY.rst
```diff
@@ -53,6 +53,13 @@ Unreleased Changes
     * Fixed a bug where uncaught exceptions in the React tree would result in
       a blank browser window.
       (`#1119 <https://github.com/natcap/invest/issues/1119>`_)
+    * Fixed a bug where the Workbench would become unresponsive during an
+      InVEST model run if the model emitted a very high volume of log messages.
+    * Fixed a bug where the Workbench could crash if there was too much
+      standard error emitted from an invest model.
+    * Added a new "Save as" dialog window to handle different save options, and
+      allow the option to use relative paths in a JSON datastack
+      (`#1088 <https://github.com/natcap/invest/issues/1088>`_)
```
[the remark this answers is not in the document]
phargogh: Huh, wonder how that happened!  Patched in 395a757ef

src/natcap/invest/hra.py
```diff
@@ -1110,153 +1095,128 @@ def _read_block(raster_path, block_info, mask_array):
...
-                    stats['R_N_ANY'] += classes_block.size
-                    subregion_stats[subregion_id] = stats
-
-        for subregion_id, stats in subregion_stats.items():
-            # Avoid dividing by zero.  If this is 0, then the other counts must
-            # all also be 0.
-            reclassified_count = max(stats['R_N_ANY'], 1)
+                        subregion_stats[fieldname] = reduce_func([
+                            subregion_stats[fieldname],
+                            stats_under_feature[opname.lower()]])
```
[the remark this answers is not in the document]
phargogh: Yes, the intent is to support multiple features having the same name being treated as the same subregion.

I'm not sure I understand what you mean by accumulating them further ... since `subregion_stats_by_name` is a `collections.defaultdict()`, we're initializing the values the first time a subregion is identified so this `reduce_func` accumulation should be the only accumulation that's needed.

The only other sort of accumulation that should be needed would be if no subregion names are provided, in which case we default to all regions having the name `"Total Region"`.  In this case, the same `reduce_func` accumulation applies, we're just accumulating over all the features' stats.

I _think_ we're handling these cases as expected in the tests, too, under `test_summary_stats()`, but I could absolutely be missing something.

Could you elaborate a little on the further accumulation?

tests/test_hra.py
phargogh: Yes, absolutely.  In edaa069f7 I reworked the test so that variable names could be more clear and also to derive test values from the arrays they're generated from.  I hope that'll help make the connection ... I think this should make the test easier to maintain too, but let me know what you think!

tests/test_hra.py
phargogh: Yes, if we were running the model with only one habitat/stressor pair, we would expect the cumulative risk to match the pairwise risk.  In this case in this test, I'm putting in different values for the cumulative risk classes to test the summary table construction only -- these values are definitely not real or reasonable model outputs!

I've added a note to clarify this (and improved variable names a bit) in 43a37ec52.

src/natcap/invest/hra.py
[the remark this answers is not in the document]
davemfish: > I'm not sure I understand what you mean by accumulating them further 

I was referring to the `zonal_stats` operation as the first accumulation/reduction -- across pixels in a feature. So I was trying to process why we would still need to be calling a a reducing function _after_ we already found the `min, max, etc` from `zonal_stats`. But you've answered that...it's the need to reduce (or accumulate, to re-use my poorly chosen term) across features with the same name.

tests/test_hra.py
davemfish: Thanks that's very helpful!

tests/test_hra.py
davemfish: Gotcha, thanks for adding that clarification in the test

davemfish (approved):
Some unrelated test is still failing simply because the Actions here are not checking out the latest commit on `main`, and triggering a re-run doesn't seem to make that happen. I'm confident all these will pass on the merge commit, so I'm going to merge.

PR #1086, opened 2022-09-23
Output spec rough drafts

emlys:
## Description
Changes include:
- a rough-draft-quality output spec for every model
- renamed ARGS_SPEC to MODEL_SPEC throughout invest, workbench, and tests
- a test to verify the output spec structure (which is _not_ passing yet)

The goal of this PR is to set up the `feature/output-spec` branch with a basic spec for each model. That will make it easier to see the actual changes

The output spec tests in `test_model_spec.py` are expected to be failing now, mostly due to missing attributes. I have intentionally _not_ yet tracked down the info which wasn't clearly documented in the user's guide. For the sake of smaller diffs, that will go in future PRs.
I have also not bothered to wrap the about text to 80 characters yet because it's tedious and a lot of the text is going to change again. 

The diff is large but probably doesn't need a detailed review, since all the content is taken from existing documentation. I would focus on `tests/test_model_spec.py`, since that enforces the design from the design doc.

A couple changes since we last talked - 
- After writing all these, I totally agree that the `type` attribute is unneeded for files and we should instead leave the file extensions on the file names.
- Renamed the `created` property to `created_if` since it sounded more natural
- There were enough different output sub-directories that it was clearly the better solution to describe them with nested dictionaries. The `intermediate` boolean property I originally proposed would not work.
- I have not included the log files because I realized they're more of a special case - they're not created by the model itself. I'm leaning towards saying they should not be part of the output spec, but we can discuss.

## Checklist

### DIFF
--- a/src/natcap/invest/carbon.py
+++ b/src/natcap/invest/carbon.py
@@ -268,7 +268,10 @@ MODEL_SPEC = {
         },
         "intermediate": {
             "type": "directory",
-            "contents": {**CARBON_OUTPUTS}
+            "contents": {
+                **CARBON_OUTPUTS,
+                "_taskgraph_working_dir": spec_utils.TASKGRAPH_DIR
+            }
         }
     }
 }
--- a/src/natcap/invest/carbon.py
+++ b/src/natcap/invest/carbon.py
@@ -20,7 +20,29 @@ from . import gettext
 
 LOGGER = logging.getLogger(__name__)
 
-ARGS_SPEC = {
+CARBON_OUTPUTS = {
+    f"c_{scenario}_{pool}.tif": {
+        "type": "raster",
+        "about": (
+            f"Raster of {pool_name} carbon values in the {scenario_name} "
+            "scenario, mapped from the Carbon Pools table to the LULC."),
+        "bands": {1: {
+            "type": "number",
+            "units": u.metric_ton/u.pixel
+        }}
+    } for pool, pool_name in [
+        ('above', 'aboveground'),
+        ('below', 'belowground'),
+        ('soil', 'soil'),
+        ('dead', 'dead matter')
+    ] for scenario, scenario_name in [
+        ('cur', 'current'),
+        ('fut', 'future'),
+        ('redd', 'REDD')
+    ]
+}
+
+MODEL_SPEC = {
     "model_name": MODEL_METADATA["carbon"].model_title,
     "pyname": MODEL_METADATA["carbon"].pyname,
     "userguide": MODEL_METADATA["carbon"].userguide,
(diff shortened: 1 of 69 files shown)


tests/test_model_specs.py
```diff
@@ -57,35 +57,184 @@ def test_model_specs_are_valid(self):
...
                             {'spatial_keys', 'different_projections_ok'}))
 
             # validate that each arg meets the expected pattern
             # save up errors to report at the end
-            for key, arg in model.ARGS_SPEC['args'].items():
+            for key, arg in model.MODEL_SPEC['args'].items():
                 # the top level should have 'name' and 'about' attrs
                 # but they aren't required at nested levels
-                self.validate(arg, f'{model_name}.{key}')
+                self.validate(arg, f'{model_name}.args.{key}')
```
dcdenu4, line 75: Should we update this function to `validate_args`?

tests/test_model_specs.py
```diff
@@ -57,35 +57,184 @@ def test_model_specs_are_valid(self):
...
+            else:
+                file_extension = key.split('.')[-1]
+                if file_extension == 'tif':
+                    t = 'raster'
+                elif file_extension in {'shp', 'gpkg', 'geojson'}:
+                    t = 'vector'
+                elif file_extension == 'csv':
+                    t = 'csv'
+                elif file_extension in {'json', 'txt', 'pickle', 'db', 'zip',
+                                        'dat', 'idx'}:
```
dcdenu4, line 114: Look at all the formats we output!

dcdenu4: Would `html` fall under here as well? Does Carbon output an `html` file?

tests/test_model_specs.py
```diff
@@ -57,35 +57,184 @@ def test_model_specs_are_valid(self):
...
+                pass
+
+            # iterate over the remaining attributes
+            # type-specific ones have been removed by this point
+            if 'about' in attrs:
+                self.assertIsInstance(spec['about'], str)
+                attrs.remove('about')
+            if 'created_if' in attrs:
+                # required value may be True, False, or a string that can be
+                # parsed as a python statement that evaluates to True or False
```
dcdenu4, line 215: Maybe, just a value that can be evaluated to a python boolean?

dcdenu4 (changes requested):
Hey @emlys, thanks for laying out this PR and the expectations. I just had a few comments, suggestions, and questions!

src/natcap/invest/carbon.py
```diff
@@ -20,7 +20,28 @@
...
+        ('above', 'aboveground'),
+        ('below', 'belowground'),
+        ('soil', 'soil'),
+        ('dead', 'dead matter')
+    ] for scenario, scenario_name in [
+        ('cur', 'current'),
+        ('fut', 'future'),
+        ('redd', 'REDD')
+    ]
+}
```
phargogh, line 42: Nice job making this rather complex nested comprehension more readable!

src/natcap/invest/hra.py
```diff
@@ -214,11 +214,116 @@
...
+                        "R_%LOW": {
+                            "about": "the percentage of low risk areas.",
+                            "type": "percent"
+                        }
+                    }
+                }
+            }
+        },
+        "visualization_outputs": {
+            "type": "directory",
```
phargogh, line 254: HRA is fresh in my mind, but I think this should also have a "created_if" flag, pointing at the optional argument of whether to create viz outputs (and maybe that's on the docket for later since this is a draft)

src/natcap/invest/scenario_gen_proximity.py
```diff
@@ -110,6 +110,33 @@
...
         }
+    },
+    "outputs": {
+        "nearest_to_edge.tif": {
+            "about": "LULC raster for the scenario of conversion nearest to the edge of the focal habitat."
+        },
+        "farthest_from_edge.tif": {
+            "about": "LULC raster for the scenario of conversion farthest from the edge of the focal habitat."
+        },
+        "nearest_to__edge.csv": {
```
phargogh, line 121: Bonus underscore in the key?

src/natcap/invest/sdr/sdr.py
```diff
@@ -147,6 +147,169 @@
...
+                "what_drains_to_stream.tif": {
+                    "about": "Map of which pixels drain to a stream. A value of 1 means that at least some of the runoff from that pixel drains to a stream in stream.tif. A value of 0 means that it does not drain at all to any stream in stream.tif."
+                },
+                "weighted_avg_aspect.tif": {
+                    "about": "average aspect weighted by flow direction(in eq. (69))"
+                },
+                "ws_inverse.tif": {
+                    "about": "Inverse of the thresholded cover-management factor times the thresholded slope (in eq. (74))"
+                },
+                "churn_dir_not_for_humans": spec_utils.TASKGRAPH_DIR
```
phargogh, line 310: I think SDR is somewhat unique in that this taskgraph directory also contains some preprocessing files.  Should those also be included here?

tests/test_model_specs.py
phargogh: I think we removed that and the rest of `natcap/invest/reporting/`, didn't we?

phargogh (commented):
This is looking really good!  The overall structure makes a lot of sense to me, and I don't have any major comments on that.  I left a few minor comments, but nothing that should be a breaker since it sounds like the intent is to go back and flesh out each model one-by-one (which I know I for one would be very happy to help with, particularly on models I wrote).

Two other comments that came up for me that didn't really apply to one line were:

1. is the intent to also add `gettext` wrappers for about when going back through each models?
2. Does `"created_if"` define a python expression in the same way that the ARGS_SPEC `required` can evaluate to an expression?  Or is the intent to just define an args key or some other string?

Thanks!

src/natcap/invest/hra.py
```diff
@@ -1977,6 +2082,7 @@ def _parse_criteria_table(criteria_table_path, target_composite_csv_path):
         records, columns=['habitat', 'stressor', 'criterion', 'rating', 'dq',
                           'weight', 'e/c'])
     overlap_df.to_csv(target_composite_csv_path, index=False)
+    print(target_composite_csv_path)
```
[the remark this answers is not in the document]
emlys: oops!

src/natcap/invest/hra.py
emlys: ah, good catch!

tests/test_model_specs.py
emlys: Carbon does still produce an html file: https://github.com/natcap/invest/blob/b437cb8d0cf74f53ce98b5386eb0f2d4a26f4342/src/natcap/invest/carbon.py#L575-L640

phargogh: Ah, got it.  So it's writing its own HTML report anot using the `reporting` subpackage.

tests/test_model_specs.py
emlys: done!

src/natcap/invest/carbon.py
```diff
@@ -176,6 +197,73 @@
...
+        "npv_redd.tif": {
+            "about": "Rasters showing the economic value of carbon sequestered between the current and the REDD landscape dates.",
+            "bands": {1: {
+                "type": "number",
+                "units": u.currency/u.pixel
+            }},
+            "created_if": "lulc_redd_path"
+        },
+        "intermediate": {
+            "type": "directory",
```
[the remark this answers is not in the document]
emlys: In `test_model_specs` we assume it's a file unless `"type"` is set to `"directory"`.
- If the `type` key exists, use that
- Otherwise, infer the type from the file extension
- If there's no valid file extension, raise an error
https://github.com/emlys/invest/blob/7a2fc54d07562d5ad979e0f84f64429bc2a724a2/tests/test_model_specs.py#L103-L119

emlys:
> is the intent to also add gettext wrappers for about when going back through each models?

Yeah, I'm saving that for later because I'll need to go back and edit those strings anyway

> Does "created_if" define a python expression in the same way that the ARGS_SPEC required can evaluate to an expression? Or is the intent to just define an args key or some other string?

I think it's just an args key. I can't think of any situations where an output is conditional on anything more complicated.

tests/test_model_specs.py
emlys: Tried to rephrase this better

phargogh (approved):

dcdenu4 (approved):
Thanks for addressing my comments! Looks good to me so far.

PR #1089, opened 2022-10-12
Bugfix for rendering high-volume invest logging to workbench

### LINKED ISSUE #1023: Workbench not responding after logging 10000s gdal warnings
Jade reported this and I reproduced it. We ran the rec model with some data that generated many warnings like this:

```
2022-07-01 14:18:15,992 (osgeo) utils._log_gdal_errors(98) WARNING [errno 1] Value 21650600 of field area_m2 of feature 1267 not successfully written. Possibly due to too larger number with respect to field width
```

There are almost 40000 of these in the invest log. I'm not sure how many got logged to the Workbench UI before it stopped responding. I re-ran the same data with logging level set to `ERROR` and avoided the problem. That's the main reason I think the UI logging is the problem.

Even when the workbench stopped responding, invest continues running in the background so the model completed successfully.

[InVEST-natcap.invest.recreation.recmodel_client-log-2022-07-05--17_24_30.txt](https://github.com/natcap/invest/files/9049905/InVEST-natcap.invest.recreation.recmodel_client-log-2022-07-05--17_24_30.txt)

davemfish:
This PR improves on some pretty naive text rendering we were doing for displaying invest logging to the Workbench. 

The fixes are:

1. Store log data in an array instead of one big string, that way react can create a DOM node for each text chunk, and not have to re-render all the text on every update.
2. Create a cache on the react side to ingest the invest stdout data and only update the component that renders that data every 10ms or so. This reduces the overall number of renders by about 10x for cases like the flood of gdal warnings.

I tested this with the data linked in #1023 which generates 100ks of gdal warnings. And I observe that the browser is responsive the whole time and the browser display updates just as promptly (earlier actually) than when I watch the invest logfile in my terminal with `tail -f`.

Fixes #1023 

## Checklist


workbench/src/renderer/components/LogTab/index.jsx
```diff
@@ -74,12 +76,35 @@ export default class LogTab extends React.Component {
...
+   * Cache incoming logger messages in real-time, then move them to react state
+   * in batches, at some longer interval. This limits the number of renders in
+   * the event of very high volume logging, while still allowing the browser to
+   * appear that it is updating in real-time.
+   */
+  debouncedLogUpdate(data) {
+    if (this.validationTimer) {
+      clearTimeout(this.validationTimer);
+    }
+    this.cache = [...this.cache, data];
```
phargogh, line 105: Isn't this creating a whole new array each time we want to append to it?  If so, is this behavior preferable to simply appending to the existing array?

I see above on line 92 that we're clearing the cache each time we update state from the cache data (totally makes sense), so maybe this really isn't likely to be an issue in practice since the arrays are probably fairly small.

phargogh (changes requested):
Thanks @davemfish !  This looks very, very good to me.  I really like the architectural change here and the approach makes a lot of sense.  I did have one question about how the cache array is appended to, so please let me know what you think about that.  Otherwise, this looks great!

workbench/src/renderer/components/LogTab/index.jsx
davemfish: Thanks, @phargogh , good catch. My best guess is that I was just in the mode of not mutating data, as that is important for data in react's `state`. But this cache is deliberately not in react state. I changed it to push to the existing array.

phargogh (approved):
Thanks, @davemfish !

PR #1097, opened 2022-10-21
Update gettext f-strings and add empty Spanish PO file

phargogh:
Minor updates to ARGS_SPEC-based f-strings so that pybabel can read them and write out POT/PO files.

CC: @emlys


dcdenu4 (approved):
Thanks @phargogh!

Issue #778, opened 2021-12-15
Make an all-encompassing translation pipeline script


emlys:
In my first iteration of setting up translation, translation mechanisms and the corresponding documentation are spread out across 3 repos (`invest`, `invest-workbench`, `invest.users-guide`). While the translation boilerplate has to be located with the code being translated, the documentation and process doesn't have to be. 

In practice we'll most likely want to update _all_ translations for a particular language at one time. One or more centrally-located scripts could partially* automate this process for the various sets of translated messages.

\* Files will need to be manually sent to and received from the translator, unless we use a translation service that provides an API.

PR #1098, opened 2022-10-25
Adding grep/sed generation of POT file body.

phargogh:
This PR doesn't fix anything specifically, but it does add a first stab at a custom PO file generation via grep and sed in a bash script to extract gettext strings from the javascript source.  It's not perfect, but I figure it'll get us most of the way there.  The updated POT and PO files for es and zh are also included as presented from the result of `msgmerge`.

RE:#778


emlys (approved):
Thanks @phargogh! Having the PO files with `msgstr` filled in for Spanish and Mandarin is a bit misleading since those aren't actually translations - I shouldn't have left those in. It's okay for now though.

PR #1101, opened 2022-11-01
Urban Nature Access greenspace/population values are too high

phargogh:
## Description

While in Stockholm and helping Diana run the Stockholm data on the in-development UNA model, it looked like the greenspace supply values (the only main raster output of the model) far too high.  Turns out, they were.

This is a relatively small fix to improve a few minor things in the UNA model that are outside of the regular feature iteration cycle:
* Fixing a masking issue in a `raster_calculator` call, where we could have very large ratio values where they should not be.
* Fixing a masking issue in the same `raster_calculator` call where numerical error in convolved pixels could lead to small negative values.  These are now masked to 0.
* Using `utils.read_csv_to_dataframe` for consistency with other models' behavior.

Any changes that are outside of these are sneak previews of the in-progress final module of this model, or merges in from other fixes in this repo.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -39,6 +39,10 @@ Unreleased Changes
 * General
     * Update python packaging settings to exclude a few config files and the
       workbench from source distributions and wheels
+    * Updating SDR test values due to an update in GDAL's mode resampling
+      algorithm. See https://github.com/natcap/invest/issues/905
+    * Updated our ``scipy`` requirement to fix a bug where invest crashed
+      if a Windows user had a non-Latin character in their Windows username.
 * Workbench
     * Fixed a bug where some model runs would not generate a new item
       in the list of recent runs.
--- a/requirements-docs.yml
+++ b/requirements-docs.yml
@@ -12,7 +12,7 @@ channels:
 dependencies:
 - chardet>=3.0.4
 - Cython
-- GDAL>=3.1.2,!=3.3.0,<3.4.1
+- GDAL>=3.4.2
 - Flask
 - flask_cors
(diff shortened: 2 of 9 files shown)


HISTORY.rst
```diff
@@ -39,6 +39,10 @@ Unreleased Changes
 * General
     * Update python packaging settings to exclude a few config files and the
       workbench from source distributions and wheels
+    * Updating SDR test values due to an update in GDAL's mode resampling
```
dcdenu4, line 42: I'm assuming this is in here because we haven't merged the latest `main` into `natcap:feature/urban-nature-access`

dcdenu4 (approved):
This looks okay to me. When thinking about numerical 0 with convolution results, I'm wondering if this is something the new convolution framework in pygeoprocessing should be able to handle? Or if the caller of that function is responsible for checking / handling that case.

HISTORY.rst
phargogh: Yep!  That's all this is.

phargogh:
> When thinking about numerical 0 with convolution results, I'm wondering if this is something the new convolution framework in pygeoprocessing should be able to handle?

@dcdenu4 are you thinking of https://github.com/natcap/pygeoprocessing/issues/268?  If so, that module or subpackage would just be a series of utility functions for creating kernels and wouldn't change anything else about `convolve_2d`.

RE: numerical 0, in this case I think the convolution is performing as expected ... there aren't any pixel values that are numerically close to 0 in the convolved output.  I think the issue in this calculation was numerical error in the division, so this `raster_calculator` call should be responsible for handling that case, which it now does.

PR #1107, opened 2022-11-03
Habitat Quality reprojecting inputs

### LINKED ISSUE #1093: Habitat Quality: validate or handle inputs in different projections
Habitat Quality is one of several models that require all spatial data to use the same SRS. #445 .

But the model does not validate this is true for the inputs. `align_and_resize` is called without doing any reprojection, but with a target bounding box from the LULC input. So rasters with a different SRS will appear to not intersect the target bounding box, even if they do.

Here's an example: https://community.naturalcapitalproject.org/t/habitat-quality-alignment-error/3081/9

HQ could either reproject it's inputs, or validate that all SRS are the same. But this validation is not easy (https://github.com/natcap/invest/issues/558), so maybe easier to reproject everything?

dcdenu4:
## Description
Habitat Quality required all spatial inputs to be projected and to have the same projection as the `lulc_cur_path` input, which is the current lulc. However, there was no validation being done of this which resulted in a bug (#1093 ) where rasters not projected the same as the current lulc would appear to not intersect the lulc bbox, even if they did.

This PR now reprojects all spatial inputs to `lulc_cur_path`, including threat rasters and the access vector. The access vector is no longer required to be projected, but `lulc_cur_path`, `lulc_fut_path`, `lulc_base_path` are.

The UG does state that all inputs need to be in the same projection. I created a new branch and made the necessary changes [here](https://github.com/natcap/invest.users-guide/blob/task/hq-projection-update/source/habitat_quality.rst).

No tests were added, since we don't have a standard for testing models with different SRS inputs.

Fixes #1093

## Checklist


davemfish (changes requested):
Hey @dcdenu4 , this looks great to me! Just one thing:

In `ARGS_SPEC`, I think `args_with_spatial_overlap` will now need to also specify `different_projections_ok: True` so that validation knows to transform bounding boxes before testing for overlap.

davemfish:
> No tests were added, since we don't have a standard for testing models with different SRS inputs.

I guess we could update one of the basic regression tests to include an input with a different SRS. The results might change slightly due to warping/interpolation, but assuming they are very close, we just update the regression data accordingly.

It's up to you, all the new code is being run by the existing tests, so I don't think it's crucial.

dcdenu4:
Hey Dave, I added a test per your suggestion. It takes a regression test and reprojects the access vector into wgs84 and one of the threats into another projection. Also made the ARGS_SPEC update. Back to you!

dcdenu4:
Oh, I added the issue number in HISTORY.rst as well!

davemfish (commented):
Hey @dcdenu4 , thanks for adding that test! I notice that the regression values are exactly the same as in `test_habitat_quality_regression` -- good news! Do you think we gain anything by keeping that old test around? It seems this new test covers everything that old one does, and more.

dcdenu4:
@davemfish , good point! I just removed the redundant test.

davemfish (approved):

PR #1109, opened 2022-11-03
Refactor monthly quickflow function

### LINKED ISSUE #1318: SWY: Monthly quickflow implementation is unclear and unmaintainable
Unsure if there's a bug here or not - but either way, we should refactor and document the `_calculate_monthly_quick_flow` function. It's not maintainable if we don't understand what it's doing.

- Is the original equation correct? Is it expected to produce these very large and very small numbers?
- If so, can we preemptively threshold the inputs to prevent over/underflow?
- Document and refactor the code so that it will be maintainable going forward.

## QF details

According to the user's guide, QF is defined in terms of two cases:

1. For stream pixels: $`\text{QF}_{stream,m} = \ P_{stream,m}`$

2. For non-stream pixels: 
$`\text{QF}_{i,m} = n_{m} \times \left( \left( a_{i,m} - S_{i} \right)\exp\left( - \frac{0.2S_{i}}{a_{i,m}} \right) + \frac{S_{i}^{2}}{a_{i,m}}\exp\left( \frac{0.8S_{i}}{a_{i,m}} \right)E_{1}\left( \frac{S_{i}}{a_{i,m}} \right) \right) \times \left( 25.4\ \left\lbrack \frac{\text{mm}}{\text{in}} \right\rbrack \right)`$

This equation can't be directly implemented in code because some of the terms overflow or underflow on valid data.

The edge cases happen in this term of the equation:
$\frac{S_{i}^{2}}{a_{i,m}}\exp\left( \frac{0.8S_{i}}{a_{i,m}} \right)E_{1}\left( \frac{S_{i}}{a_{i,m}} \right)$

a. Where $s_i = 0$, this term should equal 0 (per conversation with Rafa). But, evaluating in numpy you get NaN and a warning because $E_{1}\left( 0 \right) = \inf$.

Solution: Preemptively set the result to 0 where s_i = 0 in order to avoid calculations with infinity.

b. When the ratio $\frac{S_{i}}{a_{i,m}}$ becomes large, this term approaches 0. [NOTE: I don't know how to prove this mathematically, but it holds true when I tested with reasonable values of s_i and a_im]. The $exp$ term becomes very large, while the $E1$ term becomes very small. When $\frac{S_{i}}{a_{i,m}}$ > 880 ish, $\exp\left( \frac{0.8S_{i}}{a_{i,m}} \right)$ will overflow and round up to infinity. Meanwhile, $E_{1}\left( \frac{S_{i}}{a_{i,m}} \right)$ will become so small that it rounds down to 0.

Values of `s_i / a_im` that make `exp` overflow aren't common, but are totally possible with valid input data, for example:
```
CN = 30
P = 6 (millimeters of rain per month)
n = 10 (number of rain events per month)

s_i = 1000/CN - 10 = 23.33
a_im = P/(n * 25.4) = 8/(10 * 25.4) = 0.023

s_i / a_im = 23.33 / 0.023 = 1014

exp(0.8 * 1014) = exp(811) = 1.6 * 10^352, which overflows float64. And E1 underflows:
>>> scipy.special.exp1(811)
0.0
```
Solution: Catch the overflow warning, and set the term to 0 anywhere that overflow happens.

emlys:
## Description
Fixes #1318 
This PR refactors the monthly quickflow function with the goal of making it clear how the different cases and edge cases are handled. I reorganized the function around 3 main cases, one of which has a couple edge cases. I tried to make the code line up as closely as possible with the quickflow equation in the user's guide. I hope that the big comment block clarifies what's going on.

The functional changes from #1318 are:
- When the s_i / a_im ratio is greater than 100 (an arbitrary value I chose as per discussion with Lisa and Rafa), set QF to 0
- When QF is negative, set it to 0

And a few trivial changes:
- remove trailing zeros left over from python 2 era
- use the same TARGET_NODATA variable for all output nodatas
- remove 2 unused parameters of the monthly quickflow function

## Checklist


src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
         None.
 
     """
-    qf_nodata = -1
-
     def qfi_sum_op(*qf_values):
         """Sum the monthly qfis."""
-        qf_sum = numpy.zeros(qf_values[0].shape)
-        valid_mask = ~utils.array_equals_nodata(qf_values[0], qf_nodata)
```
emlys, line 887: We were assuming that all 12 monthly quickflow rasters would have nodata in the same areas, but that's not necessarily true.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
     """Calculate quick flow for a month.
 
     Args:
-        precip_path (string): path to file that correspond to monthly
-            precipitation
-        lulc_raster_path (string): path to landcover raster
-        cn_path (string): path to curve number raster
-        n_events_raster_path (string): a path to a raster where each pixel
+        precip_path (string): path to monthly precipitation raster
+        n_events_path (string): a path to a raster where each pixel
```
emlys, line 903: These two parameters were not used.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
-        # on large valid_si/a_im this number will be zero and the latter
-        # exponent will also be zero because of a divide by zero. rather than
-        # raise that numerical warning, just handle it manually
-        E1 = scipy.special.expn(1, valid_si / a_im)
-        E1[valid_si == 0] = 0
-        nonzero_e1_mask = E1 != 0
-        exp_result = numpy.zeros(valid_si.shape)
-        exp_result[nonzero_e1_mask] = numpy.exp(
-            (0.8 * valid_si[nonzero_e1_mask]) / a_im[nonzero_e1_mask] +
-            numpy.log(E1[nonzero_e1_mask]))
```
emlys, line 982: The old implementation took advantage of exponent math to avoid overflow: By rewriting `exp(0.8 * s_i / a_im) * E1(s_i / a_im)` as the equivalent `exp(0.8 * s_i / a_im + log(E1(s_i / a_im)))`, the result of `exp` wouldn't get so large. But then, there's another edge case where E1 = 0 and log(E1) = infinity. That was handled with the `nonzero_e1_mask` above.

I replaced that with handling the overflow warning, below. It's closer to the original equation, and IMO it's clearer what's going on.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+            )
+        )
 
         # this handles some user cases where they don't have data defined on
         # their landcover raster. It otherwise crashes later with some NaNs.
         # more intermediate outputs with nodata values guaranteed to be defined
-        qf_im[utils.array_equals_nodata(qf_im, qf_nodata) &
-              ~utils.array_equals_nodata(stream_array, stream_nodata)] = 0.0
+        qf_im[utils.array_equals_nodata(qf_im, TARGET_NODATA) &
+              ~utils.array_equals_nodata(stream, stream_nodata)] = 0
```
emlys, line 1043: This part seems suspicious, but I'm going to make another issue about it.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+        # qf_im is the quickflow at pixel i on month m
+        qf_im = numpy.full(p_im.shape, TARGET_NODATA, dtype=numpy.float32)
+
+        # case 1: there is no precipitation where both p_im and n_m are
+        # defined and equal to zero.
+        case_1_mask = ~precip_mask
+        qf_im[case_1_mask] = 0
+
+        # case 2: there is precipitation where both p_im and n_m are
+        # defined and greater than zero.
```
dcdenu4, line 993:
```suggestion
        # defined and greater than zero, and we're on a stream.
```

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+        # don't raise a warning on overflow
+        with numpy.errstate(over='ignore'):
+            exp_result = numpy.exp((0.8 * s_i[case_3_mask]) / a_im)
+
+        # edge case 3a: set the whole term to 0 when s_i = 0
+        subcase_a_mask = s_i[case_3_mask] == 0
+        term_result[subcase_a_mask] = 0
+
+        # edge case 3b: set the whole term to 0 when exp() overflows
+        subcase_b_mask = numpy.isinf(exp_result)
```
dcdenu4, line 1017: Interesting, so ignoring the `OverflowError` will leave `exp_result` with an infinity value?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+        # defined and greater than zero.
+        case_2_mask = precip_mask & stream_mask
+        qf_im[case_2_mask] = p_im[case_2_mask]
+
+        # case 3
+        # only where all four inputs are defined
+        case_3_mask = valid_mask & precip_mask & ~stream_mask
+
+        term_result = numpy.full(
+            qf_im[case_3_mask].shape, TARGET_NODATA, dtype=numpy.float32)
```
dcdenu4, line 1002: Is `qf_im[case_3_mask].shape` == `qf_im.shape`? Is there a reason to mask here?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
dcdenu4: This is odd. This is setting nodata values to 0 where `qf_im == Nodata & stream != Nodata`... This would indicate only Stream nodata values would persist through as nodata. Huh.

dcdenu4 (changes requested):
Hey @emlys, thanks for looking into this and being thorough. I like your approach in breaking out the components and the documentation that accompanies it. I think the handling of the OverflowError is works well, but I do wonder if there's a way to constrain the problem further up somehow. I'll be interested to hear what Rafa says.

Oh, I didn't write this as a comment, but is there a reason you chose to "ignore" the error vs surround it in a try / except block?

Thanks!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
emlys: Yes:
```
>>> numpy.exp(800)
<stdin>:1: RuntimeWarning: overflow encountered in exp
inf
```
To your question above
> is there a reason you chose to "ignore" the error vs surround it in a try / except block?

Because it's a warning not an error, and a RuntimeWarning could be caused by other things whereas the numpy context manager lets you specifically ignore overflows

dcdenu4: Oh interesting! I was testing `math.exp(800)` and getting:
```python
>>> math.exp(900)
OverflowError: math range error
```
That makes sense.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
emlys: Masking gets you the shape of where the mask is true, i.e. a 1-d array whose length is `sum(mask)`.
```
>>> a = numpy.array([True, False, True, False])
>>> b = numpy.array([1, 1, 1, 1])
>>> b.shape
(4,)
>>> b[a].shape
(2,)
```

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+        # 3b. When the ratio s_i / a_im becomes large, this term approaches 0.
+        #     [NOTE: I don't know how to prove this mathematically, but it
+        #     holds true when I tested with reasonable values of s_i and a_im].
+        #     The exp() term becomes very large, while the E1() term becomes
+        #     very small. When (s_i / a_im) > 880 ish, exp(0.8 * s_i / a_im)
+        #     will overflow and round up to infinity. Meanwhile, E1(s_i / a_im)
+        #     will become so small that it rounds down to 0.
+        #
+        #     Solution: Catch the overflow warning, and set the term to 0
+        #     anywhere that overflow happens.
```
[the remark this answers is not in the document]
emlys: Values of `s_i / a_im` that make `exp` overflow aren't common, but are totally possible with reasonable input data, for example:
```
CN = 30
P = 8 (millimeters of rain per month)
n = 10 (number of rain events per month)

s_i = 1000/CN - 10 = 23.33
a_im = P/(n * 25.4) = 8/(10 * 25.4) = 0.031

s_i / a_im = 23.33 / 0.031 = 752

exp(752) = 3.8 * 10^326, which overflows float64.
```

as far as numerical constraints, `exp(x)` overflows when `x` is about 709.78. I'm not sure if this might vary on different systems.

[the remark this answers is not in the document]
emlys: > We're setting to 0 because if the exp has overflowed, then we know the E1 term is 0?

yes, the E1 will be extremely close to 0 and "underflow" to 0. for example:
```
>>> scipy.special.exp1(200)
6.885226106307636e-90
>>> scipy.special.exp1(400)
4.776013586420972e-177
>>> scipy.special.exp1(700)
1.406518766234033e-307
>>> scipy.special.exp1(800)
0.0
```

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -879,133 +875,178 @@ def _calculate_annual_qfi(qfm_path_list, target_qf_path):
...
+        #
+        # 3c. Otherwise, evaluate the term as usual.
+
+        # qf_im is the quickflow at pixel i on month m
+        qf_im = numpy.full(p_im.shape, TARGET_NODATA, dtype=numpy.float32)
+
+        # case 1: there is no precipitation where both p_im and n_m are
+        # defined and equal to zero.
+        case_1_mask = ~precip_mask
```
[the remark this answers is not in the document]
emlys: because precip_mask is defined above as:
```
# precip mask: both p_im and n_m are defined and greater than 0
precip_mask = valid_p_mask & valid_n_mask & (p_im > 0) & (n_m > 0)
```
it could include pixels that have nodata in the `s_i` or `stream` arrays, but not in the precip or n_events arrays. 

That's because you don't need to know `s_i` or `stream` to calculate `QF = 0 when P = 0`. I think you could argue that it should be nodata anyway, but this is consistent with the past implementation.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
schmittrjp: Why would that not be true? If a pixel does have no quickflow in a month, I'd expect it to be 0. If its outside of the model domain it would be always nan?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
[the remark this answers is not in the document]
schmittrjp: @emlys thanks for making this example. It makes numerically sense, but let's look at it from a process perspective. basically we have no precip (0.8 mm per rain is basically no rain) meeting a very permeable soil (very low curve number). As I see from the above, if we just use the threshold value of exp(709) then QF would be 0, right? Probably we can check that somehow bottom up? Calculate for each pixel what values of x would result in an overflow and set quickflow to 0 there? There  might e a more elegant way. Let's think about it after thanksgiving.

emlys:
@dcdenu4 I've updated this with the latest from our conversation with @lmandle and @schmittrjp last week!

dcdenu4 (changes requested):
Thanks @emlys,

I had a few comments / suggestions.

I was also wondering if we talked about updating the Users Guide at all to better reflect what the model was doing under some of these instances? I can't remember if during our discussion with Rafa or Lisa that came up.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -1039,109 +1035,154 @@ def qfi_sum_op(*qf_values):
...
+                ((a_im[case_3c_mask] - s_i[case_3c_mask]) *
+                 numpy.exp(-0.2 * s_i[case_3c_mask] / a_im[case_3c_mask])) +
+                (s_i[case_3c_mask] ** 2 / a_im[case_3c_mask] *
+                 numpy.exp(0.8 * s_i[case_3c_mask] / a_im[case_3c_mask]) *
+                 scipy.special.exp1(s_i[case_3c_mask] / a_im[case_3c_mask]))
+            )
+        )
+
+        # case 3d: set any negative values to 0
+        qf_im[qf_im < 0] = 0
```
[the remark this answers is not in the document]
emlys: Thanks for catching that!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -1039,109 +1035,154 @@ def qfi_sum_op(*qf_values):
...
-            valid_mask &= ~utils.array_equals_nodata(
-                stream_array, stream_nodata)
-
-        valid_n_events = n_events[valid_mask]
-        valid_si = s_i[valid_mask]
-
+        valid_mask = (
+          valid_p_mask &
+          valid_n_mask &
+          ~utils.array_equals_nodata(stream, stream_nodata) &
```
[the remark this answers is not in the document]
emlys: Yeah, the stream mask is separate

dcdenu4 (approved):
Thanks @emlys, let's roll this in!

PR #1099, opened 2022-10-30
CVE-2007-4559 Patch

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

PR #1114, opened 2022-11-08
Bugfix/1113 patch CVE-2007-4559

### LINKED ISSUE #1113: Patch CVE-2007-4559
The Trellix security team submitted a patch for CVE-2007-4559 in https://github.com/natcap/invest/pull/1099, which we declined for some good reasons.

After thinking about this, I think it would actually be very wise for us to patch this exactly because of our use of tarfiles in datastacks.  The intent of a datastack archive is to allow someone to share their data with us, and a maliciously-constructed datastack archive would be a very easy way to take advantage.  So we might as well patch it, even if just in case.

phargogh:
## Description
After thinking about it some more, datastack archives are a possible point of vulnerability for us because of the intended use of these archives: for folks to share data with us.  So while the code might run in a trusted environment, we can't say the same about data.  So I figured it might be worth going ahead with the patch anyways.

Originally reported in #1099 but declined for some good reasons.

Also, python should _really_ still fix this upstream.

Fixes #1113

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -36,6 +36,10 @@
 
 Unreleased Changes
 ------------------
+* General
+    * Fixed a possible path traversal vulnerability when working with datastack
+      archives.  This patches CVE-2007-4559, reported to us by Trellix.
+      https://github.com/natcap/invest/issues/1113
 * Workbench
     * Fixed a bug where the Workbench would become unresponsive during an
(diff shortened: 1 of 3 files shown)


dcdenu4 (changes requested):
Thanks @phargogh, I do see a `History.rst` conflict has popped up.

dcdenu4 (approved):
Thanks!

PR #1121, opened 2022-11-21
Fix rasterization of spatial criteria relative to InVEST 3.9

### LINKED ISSUE #1120: HRA regression when rasterizing spatial criteria
A [user on the forums](https://community.naturalcapitalproject.org/t/habitat-risk-assessment-problems-with-risk-calculation-and-outputs/3122/15) noticed a difference in how spatial criteria were being represented between InVEST 3.9 (former HRA) and InVEST 3.12 (the new HRA implementation), specifically around how input vectors are being rasterized.

After digging into this for a while, it appears that the InVEST 3.9 implementation makes the following distinction that I missed when rewriting HRA:
* Habitat and stressor vectors are rasterized with `ALL_TOUCHED=TRUE`
* Spatial criteria vectors are rasterized with `ALL_TOUGHED=FALSE`

The current HRA does not make this distinction, but probably should, at least for backwards compatibility.

For what it's worth, I could not find this distinction in the InVEST 3.3.0 HRA.

phargogh:
This PR restores a bit of functionality from the InVEST 3.9 HRA implementation that I had missed in the rewrite.  In InVEST 3.9, habitat and stressor vectors are rasterized with `ALL_TOUCHED=TRUE`, while spatial criteria vectors are rasterized with `ALL_TOUCHED=FALSE`.  The HRA rewrite was rasterizing everything as `ALL_TOUCHED=TRUE`, leading to the behavior discussed in https://community.naturalcapitalproject.org/t/habitat-risk-assessment-problems-with-risk-calculation-and-outputs/3122/14.

The HRA 3.9 behavior for spatial criteria seems reasonable to me, so I've restored it here and made sure that it's being tested.

No changes in the user's guide are needed for this since we're just restoring existing functionality.

Fixes #1120 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -45,6 +45,12 @@ Unreleased Changes
       now reprojected to the ``lulc_cur_path`` raster. This fixes a bug where
       rasters with a different SRS would appear to not intersect the
       ``lulc_cur_path`` even if they did. (https://github.com/natcap/invest/issues/1093)
+* HRA
+    * Fixed a regression relative to InVEST 3.9.0 outputs where spatial
+      criteria vectors were being rasterized with the ``ALL_TOUCHED=TRUE``
+      flag, leading to a perceived buffering of spatial criteria in certain
+      cases.  In InVEST 3.9.0, these were rasterized with ``ALL_TOUCHED=FALSE``.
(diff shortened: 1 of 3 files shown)


davemfish (approved):
This looks good to me

PR #1125, opened 2022-12-01
Support relative paths in datastacks with new Save As modal

### LINKED ISSUE #998: WB: Save buttons could have better default filenames
From Stacie: 

> When saving a datastack, seems like it would be useful to include the model name in the .tgz filename. Same with saving to JSON and python script.

👍

### LINKED ISSUE #1088: WB: saved parameter sets should support relative paths
We have this option in the QT UI. JSON parameter sets can have relative or absolute paths to files. It's simply an argument to `datastack.build_parameter_set`.

The Workbench needs to expose this option in the interface.

emlys:
## Description
Fixes #1088 
Fixes #998 

This PR moves the three "save as" options into a new modal window and adds a "use relative paths" option to the "save as JSON" button. It also includes the model name in the default filenames.

Screen Shot 2022-12-01 at 11 31 10 AM

The actual file browsing alert window is OS-specific and there's no way to add a "use relative paths" option to that alert. So the "use relative paths" option had to go either in the main setup sidebar, or a new component entirely. I chose to make a new modal component - I think it looks nicer, and it made space for some more help text about the different save options.

## Checklist


davemfish (changes requested):
@emlys , this looks awesome! Good call on adding the modal to handle the extra option. And nice additions of the text descriptions. I have only a few minor comments.

The failing macos puppeteer test is annoying. Everything seems to be working there, but the "Run" button still reads "Running" when the test is expected it to have already switched to "Cancel Run". Maybe we can chalk it up to the general sluggishness of macos VMs on Actions?

Did you run that test locally by chance?

davemfish:
I just noticed https://github.com/natcap/invest/issues/998, might be nice to knock that one out here

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -75,6 +73,20 @@ function initializeArgValues(argsSpec, uiSpec, argsDict) {
   });
 }
 
+export function RunButton(props) {
```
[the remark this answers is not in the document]
emlys: Good call, I'll put it inline in the render. I had just copied it unchanged from `SetupButtons`.

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -245,10 +257,10 @@ export default class SetupTab extends React.Component {
     const payload = {
       filepath: datastackPath,
       moduleName: pyModuleName,
-      relativePaths: false,
+      relativePaths: relativePaths,
       args: JSON.stringify(args),
     };
-    const response = await writeParametersToFile(payload);
+    const response = await writeParametersToFile(payload, relativePaths);
```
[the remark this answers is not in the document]
emlys: good catch!

workbench/tests/renderer/investtab.test.js
```diff
@@ -223,9 +230,12 @@ describe('Sidebar Buttons', () => {
     const mockDialogData = { filePath: 'data.tgz' };
     ipcRenderer.invoke.mockResolvedValue(mockDialogData);
 
-    const { findByText, findByRole, getByRole } = renderInvestTab();
-
-    const saveButton = await findByText('Save datastack');
+    const { findByText, findByLabelText, findByRole, getByRole} = renderInvestTab();
+    const saveAsButton = await findByText('Save as...');
+    userEvent.click(saveAsButton);
+    const datastackOption = await findByLabelText((content, element) => content.startsWith('Parameters and data'));
```
emlys, line 236: @davemfish I think this is why the test was flaky. I'd forgotten to update this line for this test case so it was actually testing the python script option, not the datastack option. The "archiving..." message would still appear briefly, but it wasn't getting the 100ms timeout defined above.

davemfish: @emlys Hmm, I do see that test is failing again actually, just on Mac. I'm playing with the timeout length in the mocked Promise and it seems like when it's too short, the alert has already updated to "saved", so the test never sees "archiving". On Windows I had to go down to 10ms to get the failure. But again, the slow Mac VMs might be main problem. Going up to 1000ms seems fine to me.

emlys: ah okay. That test is reliable for me locally, so I'll just increase the timeout assuming it's a VM slowdown.

davemfish (approved):
Nice, all passing now!

PR #1127, opened 2022-12-02
Update LULC and Biophysical Table descriptions

### LINKED ISSUE #1077: LULC and Biophysical table should mention LULC codes are unique
@newtpatrol mentioned in a Slack thread that:

>... the User Guide description for LULC and biophysical table do not specify that the integer codes must be unique. Since this seems like a quality that is common across all models that use an LULC and biophysical table, and these are now defined in the UG with investspec, I'd like to request that this be added to their description. This might involve changes to different inputs in different models (like current, future and baseline land cover in Habiat Quality), so will take a bit of thoroughness to check each model.

For LULC that should be an easy add in `spec_utils` where we define a LULC about section.

dcdenu4:
## Description
Updating `LULC.about` text in `spec_utils` to add suggestion of mentioning each LULC class type should be represented as a unique integer. Stopping short of mentioning how this relates to the Biophysical table as most models do this separately and some model specs with `LULC.about` aren't restricted to have a Biophysical table.

Updating consistency of Biophysical table `about` sections in model specs.

Fixes #1077 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -40,6 +40,8 @@ Unreleased Changes
     * Fixed a possible path traversal vulnerability when working with datastack
       archives.  This patches CVE-2007-4559, reported to us by Trellix.
       https://github.com/natcap/invest/issues/1113
+    * Updating descriptions for LULC about text and biophysical table for
+      clarity in model specs. https://github.com/natcap/invest/issues/1077
 * Habitat Quality
     * All spatial inputs including the access vector and threat rasters are
       now reprojected to the ``lulc_cur_path`` raster. This fixes a bug where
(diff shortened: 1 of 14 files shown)


dcdenu4:
Hey @emlys, I thought about adding some constants to `spec_utils` instead of repeating the `lucode` text everywhere but decided against it because this wasn't an established practice in `spec_utils` for just text. Let me know what you think.

emlys (changes requested):
@dcdenu4 it looks good to me, I think if you want to make a spec_utils constant for the commonly repeated phrases that would be neat. Looks like the tests are just failing with a missing parentheses.

dcdenu4:
Thanks @emlys, I just put the repeated biophysical LULC column about text in a constant in spec_utils. Let me know what you think!

emlys (approved):
Looks good!

PR #1129, opened 2022-12-05
Add ErrorBoundary to catch exceptions in react tree

### LINKED ISSUE #1119: Workbench uncaught exceptions in React
A few people have reported the Workbench starting up with a "blank screen":
image

I think this probably indicates an uncaught exception in a react component. The browser hasn't crashed and the main nodejs process seems fine also, looking at the logs.

Jade reports the Workbench was working fine for a while, then crashed one time, and then this started happening every time she starts the workbench, even after re-installing it.

I'm not sure what problem would persist between sessions, though there are a couple things the workbench looks for on the filesystem on startup.

First step is to setup a react `ErrorBoundary` to catch and log exceptions that we might be missing now. Especially since devtools is disabled in production.

davemfish:
This fixes the "blank screen problem" that has been reported a couple times. Any UncaughtExceptions in the React tree will now go to a fallback UI known as an Error Boundary. And since we're catching them now, we can also log the exceptions. 

There's also a fix here for the bug in `HomeTab` that revealed the need for this Error Boundary.

Fixes #1119 

Capture

## Checklist


workbench/vite.config.js
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
   build: {
     mode: process.env.MODE,
     sourcemap: 'inline',
-    minify: process.env.MODE !== 'development',
+    minify: false,
```
emlys, line 18: Is this related to the issues addressed in this PR?

emlys (commented):
@davemfish this is really neat! The "something went wrong" screen looks very professional.

I just had a question about the vite config change. And also, is it expected that `job.argsValues` is missing sometimes? When would that occur?

davemfish:
> @davemfish this is really neat! The "something went wrong" screen looks very professional.

Thanks!
> 
> I just had a question about the vite config change. And also, is it expected that `job.argsValues` is missing sometimes? When would that occur?

It's not expected that it would be missing, but it did happen and that led to the uncaught exception that was the impetus for this whole error boundary. I'm not exactly sure how it happened, but I'm guessing the application was closed or crashed at just the wrong moment, when the `job` was being written to the local database. Then next time the app was opened, it tried loading a `job` with no `argsValues`. 

It's probably good practice to do this type of safeguarding any time we're reading data from the local system. Because who knows what could happen to that data when we're not looking.

workbench/vite.config.js
davemfish: Yeah, good question. It is related because now that we're logging error traceback from browser exceptions, it's much nicer to have un-minified code in that traceback. Otherwise we have to figure out how to decode it with source maps.

Plus the size difference in the bundle is completely negligible in the context of all the invest & electron binaries. Minification is really just meant for app's distributed over the network, I think.

emlys (approved):
thanks @davemfish!

PR #1144, opened 2023-01-05
Add tooltip to model tabs

### LINKED ISSUE #1071: Many workbench tabs make the labels unreadable
While testing the workbench for the InVEST 3.12 release, I was running all of the models, one after another, leaving all of the tabs open as I went.  This resulted in the tabs becoming unreadable by the end:

Screen Shot 2022-08-31 at 1 06 15 PM

Related, I noticed that since some of our models share prefixes (e.g. "Habitat Quality" and "Habitat Risk Assessment"), only the prefixes are visible when there are enough tabs:

Screen Shot 2022-08-31 at 12 57 53 PM

And then a related issue is that since the tab is the only place apart from the logfile where the model name can be read, it's easy to lose track of which model I'm looking at.  Maybe there's another place where the model name could also be displayed?

emlys:
## Description
Fixes #1071 
I don't think there's any way around the model tabs being unreadable when there are many tabs open, but having tooltips should help. This is similar to how Firefox handles a lot of tabs.

## Checklist

### DIFF
--- a/workbench/tests/renderer/app.test.js
+++ b/workbench/tests/renderer/app.test.js
@@ -97,7 +97,7 @@ describe('Various ways to open and close InVEST models', () => {
   });
 
   test('Clicking an invest model button renders SetupTab', async () => {
-    const { findByText, findByRole } = render(
+    const { findByText, getByTestId, findByRole } = render(
       <App />
     );
 
@@ -110,6 +110,9 @@ describe('Various ways to open and close InVEST models', () => {
     const setupTab = await findByText('Setup');
     expect(setupTab.classList.contains('active')).toBeTruthy();
     expect(getSpec).toHaveBeenCalledTimes(1);
+    const navTab = await findByRole('tab', { name: MOCK_MODEL_TITLE });
+    userEvent.hover(navTab);
+    getByTestId('tab-tooltip');
   });
 
   test('Clicking a recent job renders SetupTab', async () => {
--- a/workbench/tests/renderer/app.test.js
+++ b/workbench/tests/renderer/app.test.js
@@ -98,7 +98,7 @@ describe('Various ways to open and close InVEST models', () => {
   });
 
   test('Clicking an invest model button renders SetupTab', async () => {
-    const { findByText, findByTestId, findByRole } = render(
+    const { findByText, findByRole } = render(
       <App />
     );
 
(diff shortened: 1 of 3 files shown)


workbench/tests/renderer/app.test.js
```diff
@@ -110,6 +110,9 @@ describe('Various ways to open and close InVEST models', () => {
     const setupTab = await findByText('Setup');
     expect(setupTab.classList.contains('active')).toBeTruthy();
     expect(getSpec).toHaveBeenCalledTimes(1);
+    const navTab = await findByRole('tab', { name: MOCK_MODEL_TITLE });
+    userEvent.hover(navTab);
+    getByTestId('tab-tooltip');
```
davemfish, line 115: Does the `Tooltip` element have role `tooltip`? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

Basically I'm wondering if we can avoid using a test ID to find the element, as that would make the test a bit more resilient during refactors.

Along those lines...Tooltips often show up with a delay. I don't think you have one set here, but changing from `getBy` to ` await findBy` would allow us to adjust the delay at any point in the future and not break the test.

davemfish (commented):
Hey @emlys , this looks good. I had a couple ideas related to the test that you could consider, but not too critical. 

I also see that this test happened to timeout in the Windows VM and raised a warning about "overlapping act() calls". I don't see how any of these changes would have caused that, though I don't really know the internals of `OverlayTrigger`. It's possible there is async stuff happening in there causing that warning. I've sometimes found `awaiting` the `userEvent.*` to resolve stuff like that.

workbench/tests/renderer/app.test.js
emlys: Good idea, the `tooltip` role seems to work. I did try to avoid using a test ID, but I couldn't find it documented anywhere what the role of the overlay is, and inspecting it in the browser didn't work.

emlys:
@davemfish I got a warning that something should be wrapped in `act()`, so I did that, and it seems to be working now! And thanks for pointing out that the `tooltip` role exists.

davemfish:
> @davemfish I got a warning that something should be wrapped in `act()`, so I did that, and it seems to be working now! And thanks for pointing out that the `tooltip` role exists.

Okay cool. I don't think I've tried using `act()` directly before, but I think it's a fine way to resolve that warning. I think `act` is baked into most/all of the `testing-library` APIs, so I don't fully understand when we would need to use it ourselves. Anyway, this test looks good to me!

davemfish (approved):

PR #1157, opened 2023-01-24
Updates to the autorelease script

### LINKED ISSUE #658: Auto release scripts are broken
I have run into a few issues with the auto release scripts (`scripts/release-1-initiate.sh`, `scripts/release-2-commit.sh`, `scripts/release-3-publish.sh`). Some typos and some may be differences with running on macOS. It will be worth going back through once the release is done and making sure we have working scripts for windows and mac.

emlys:
## Description
Fixes #658. This is a step along the way to #1160.

- combine the 3 release scripts into 2
- move all release-related scripts into one folder
- replace `hub` commands with `gh` commands (see https://github.com/cli/cli/blob/trunk/docs/gh-vs-hub.md)
- the script now downloads the release artifacts for you
- the autorelease process no longer updates the UG rev, it is assumed that it's already updated (discussed this with @phargogh on slack). Removed the python script that updated the Makefile.
- Replaced the python script that updated HISTORY.rst with a one-line perl command

There is a `gh run watch` command that lets you wait for a github actions run to complete, so it would be possible to combine the two scripts into one. But I found it's too easy to forget that it's running, and close the tab or change branches or something. These two scripts can become two workflows in #1160: the second will be triggered once the first one succeeds.

## Checklist


scripts/release/autorelease-step-2.sh
```diff
@@ -0,0 +1,76 @@
...
+#   - Attach the build artifacts
+# - Create a pull request from the autorelease branch into main
+# - Create a pypi release
+#   - Attach the build artifacts
+set -e  # Exit the script immediately if any subshell has a nonzero exit code.
+
+VERSION=$1
+GH_RUN_ID=$2
+GITHUB_REPO="natcap/invest"
+PYPI_REPO="testpypi"
```
emlys, line 14: I am leaving this set to testpypi for now so we don't accidentally create a permanent release on pypi. To run the script for real, just change this to "pypi"

dcdenu4: I like the idea of having this be "testpypi", but then after the script completes maybe we can have instructions saying something like:

```sh
echo "Release has been created using testpypi, to release officially run:"
echo "twine upload -r pypi dist/natcap.invest.*"
```

scripts/release/autorelease-step-1.sh
```diff
@@ -0,0 +1,100 @@
...
+git add HISTORY.rst
+git commit -m "Committing the $VERSION release."
+git tag "$VERSION"
+git push https://github.com/${GITHUB_REPO}.git $VERSION $AUTORELEASE_BRANCH
+
+echo "Waiting for Github Actions run to start..."
+sleep 5
+# from the list of recent github actions runs,
+# extract the run ID corresponding to the release tag
+RUN_ID=$(gh --repo $GITHUB_REPO run list --branch $VERSION --json databaseId --jq ".[].databaseId")
```
dcdenu4, line 82: I wonder, is it possible through the GH CLI to know when / if actions has completed successfully? Then we could just have one release script?

https://cli.github.com/manual/gh_run_watch

```
# Run some other command when the run is finished
gh run watch [run-id] --exit-status && notify-send "run is done!"
```

dcdenu4 (changes requested):
Thanks @emlys, I just had a few questions / suggestions.

scripts/release/autorelease-step-1.sh
dcdenu4: Oops, I just read your comment about why you chose NOT to do this. Sorry, I should have read the summary thoroughly first! This seems like a subjective choice and I'm fine leaving it as you have it, knowing we'll be iterating in #1160.

scripts/release/autorelease-step-2.sh
emlys: Sounds good

dcdenu4 (approved):
Hey @emlys, this looks good to me! I'm sure some things will shake out when we try it for the first time.

Issue #755, opened 2021-12-08
Deprecate the Qt UI


davemfish:
We're planning on doing this at some point. Here's an issue where we can talk about when & how.

### Criteria that should be met before we do it:
- [x] the Workbench is stable on Windows & Mac
- [x] the Workbench has feature-parity with the Qt UI
- [x] the Workbench is equally accessible to partners and users around the world as Qt.
- [x] Workbench-invest repo management is decided and we are confident in the WB release process
- [x] Post on the forum saying the next release will not include the classic UI. So please try the Workbench and report any issues now.

### todo:
- [x] Replace the invest-autotest script that runs the Qt UI for all models. This could mean doing #1008 . Or it could mean doing #74. Because running all the sampledata through the CLI, plus running the existing Workbench test suite (it builds the UI for all models but does not run them) might be equivalent to actually running all sampledata through the Workbench.
- [x] Remove GUI functionality from the invest CLI. i.e. all `invest run` will be "headless"; remove `invest launch`.
- [x] Replace the Workbench's interim icon with the standard invest logo icon
- [x] Remove `natcap.invest.ui` source
- [x] Remove `requirements-gui.txt`, all references to it
- [x] Remove `natcap.invest.ui` references in `setup.py`
- [x] Remove `ui_tests/ and associated targets in `Makefile`, steps in GHA workflows
- [x] Review `README.rst`
- [x] Remove `installer/` scripts & resources, unless any are still used by the electron build process
- [x] Remove installer build steps from GHA workflows
- [ ] Remove "classic invest" links from our website auto-updater thing

davemfish:
It's been several months since we released the Workbench with `3.11.0`. But we had a bug in usage logging so we won't have usage stats for the workbench until the next release. Sometime after that we can look at the ratio of "workbench" / "classic" invest runs and decide if anyone is really relying on classic invest anymore.

davemfish:
Now that `3.12.0` is released, we'll look at the usage logging stats later this month and then decide when is a good target for deprecation. And then we'll announce an estimated deprecation date that is at least a few more months out.

davemfish:
We looked at usage stats for the `3.12.0` release and found >80% of invest runs for this version were from the workbench. With that many runs, we've only really seen one outstanding Workbench issue on the forum that we don't have a fix for yet (because we don't know the source of the problem). https://community.naturalcapitalproject.org/t/why-does-invest-3-12-stay-in-a-completely-white-state-after-opening-it/3110

[...]

PR #1158, opened 2023-01-24
Removing GUI functionality from CLI & autotest

### LINKED ISSUE #74: Run InVEST against its sample data
As maintainers of InVEST, we have a suite of software tests that ensure the model works as expected on precisely-constructed sample data.  As @dcdenu4 mentioned today on a call, it would be responsible for us to ensure that the model works as expected on the sample data that we ship with InVEST in addition to the tests for software functionality.

One aspect of this form of testing (_do binaries crash when running the model on sample data?_) is currently contained in `scripts/invest-autotest.py`, but perhaps it would be better to restructure this as a suite of tests that also can assert some summary statistics of the output files?  And, if so, how might we include these tests as a regular part of our development cycle in a way that doesn't dramatically extend the runtime of our tests?

This would be worth designing and reviewing before implementing in full.

davemfish:
This PR is one step towards deprecating the Qt UI #755 .
* Updates `invest-autotest.py` to only run headless tests
* Updates `cli.py` to always `invest run` in headless mode; also removes the `invest launch` command
* Adds a GHA step to run `invest-autotest` after building binaries, only on pushes to `main` or `release*` branches
* Added a `make invest_autotest` target

Here's an example of the new autotest step running on my branch (I temporarily allowed it to run on a push to that branch).
https://github.com/davemfish/invest/actions/runs/4006930987/jobs/6880339563

Otherwise, this new step won't run in this PR, but should run once it's merged into `release/3.13.0`.

Fixes #74 

## Checklist


src/natcap/invest/cli.py
```diff
@@ -275,15 +272,13 @@ def main(user_args=None):
     listmodels_subparser.add_argument(
         '--json', action='store_true', help='Write output as a JSON object')
 
-    subparsers.add_parser(
-        'launch', help='Start the InVEST launcher window')
-
     run_subparser = subparsers.add_parser(
         'run', help='Run an InVEST model')
+    # Recognize '--headless' for backwards compatibility.
```
emlys, line 277: What would you think about gradually deprecating this option so we don't have to keep it around forever? We could warn the user that it's deprecated now and will be removed in a future release

.github/workflows/build-and-test.yml
```diff
@@ -378,6 +378,12 @@ jobs:
       - name: Build binaries
         run: make ${{ matrix.binary-make-command }}
 
+      - name: Run invest-autotest with binaries
+        if : |
+          github.event_name == 'push' &&
+          (contains(github.ref, 'release') || github.ref == 'refs/heads/main')
```
emlys, line 384: neat!

emlys (approved):

src/natcap/invest/cli.py
davemfish: @emlys good idea! Just pushed the change.

davemfish: I tested it manually with all the variations of the command. I didn't bother adding a test to check for the logging since this isn't a feature we plan to keep around for long.

emlys:
Looks good! Test failures are unrelated.

PR #1162, opened 2023-01-25
Bugfix/1161

### LINKED ISSUE #1161: A few workbench strings are not translating
I noticed a few Workbench strings that are not translating to either chinese or spanish. One on the `Home` tab that's only visible when there are no recent runs to display:
>Set up a model from a sample datastack file (.json) or from an InVEST model's logfile (.txt): 

And a few in the `Settings` modal under the `more info` toggle. 

The "more info" string itself is just missing the `_()` wrapper. But for the others, I think what they have in common is that the strings are quoted with backticks _( \`some text\` ) instead of regular quotes. Could that be screwing up the gettext parsing? If so maybe we also don't have the translations for these strings.

I think the backticks were only used to appease the linter because the strings are long I didn't want to exceed the 80 char line width. There are other ways to deal with that though.

emlys:
## Description
Fixes #1161

## Checklist


workbench/package.json
```diff
@@ -40,11 +40,13 @@
   "dependencies": {
     "@babel/runtime": "^7.13.10",
     "electron-log": "^4.3.5",
-    "gettext-parser": "^4.2.0",
-    "gettext.js": "^1.0.1",
+    "i18next": "^22.4.9",
+    "i18next-conv": "^13.1.0",
```
emlys, line 44: these should be dev dependencies

workbench/tests/main/main.test.js
```diff
@@ -174,7 +174,7 @@ describe('createWindow', () => {
       ipcMainChannels.SHOW_OPEN_DIALOG,
       ipcMainChannels.SHOW_SAVE_DIALOG,
       ipcMainChannels.IS_FIRST_RUN,
-      ipcMainChannels.SET_LANGUAGE,
+      ipcMainChannels.CHANGE_LANGUAGE,
```
emlys, line 177: Should we test that the menu bar and about / report a problem windows are translated?

workbench/src/renderer/components/DataDownloadModal/index.jsx
```diff
@@ -152,6 +153,7 @@ export class DataDownloadModal extends React.Component {
       selectedLinksArray,
       dataRegistry,
     } = this.state;
+    const { t } = this.props;
```
davemfish, line 156: I was expecting to see `t` passed as a prop to this component in the `App`'s render method. Otherwise `t` will be `undefined` here right?

workbench/src/renderer/components/DownloadProgressBar/index.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Alert from 'react-bootstrap/Alert';
+import { withTranslation } from 'react-i18next';
```
davemfish, line 4: `withTranslation` not used here I think. Hooks for the win.

workbench/src/renderer/components/DataDownloadModal/index.jsx
davemfish: Ahhh I get it, the parent component is now `withTranslation()`, which must be passing `t` to it's child.

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -295,7 +296,7 @@ export default class SetupTab extends React.Component {
       this.triggerScrollEvent();
     } else {
       alert( // eslint-disable-line no-alert
-        _(`Datastack/Logfile for ${datastack.model_human_name} does not match this model.`)
+        t(`Datastack/Logfile for ${datastack.model_human_name} does not match this model.`)
```
davemfish, line 299: I think this component is also missing `const { t } = this.props;`

workbench/src/renderer/menubar/report.jsx
```diff
@@ -0,0 +1,76 @@
...
+        {t('please see the guidelines here for reporting problems: ')}
+        <a
+          href="https://community.naturalcapitalproject.org/t/guidelines-for-posting-software-support-questions/24"
+        >
+          {t('Guidelines for posting software support questions')}
+        </a>
+      </p>
+      <p>
+        <b>{t('If the problem is related to this User Interface, ')}</b>
+        {t('rather than with a specific InVEST model,')}
```
davemfish, line 32: I hope it's okay to translate these two strings separately. I like the visual hierarchy the page has by bolding the first line, but it's clearly a tradeoff.

workbench/tests/jest.overrides.js
```diff
@@ -23,6 +25,25 @@ if (global.window) {
...
+      ll: {
+        translation: {
+          Open: 'σρєи',
+          Language: 'ℓαиgυαgє'
+        }
+      }
+    },
+  });
+
+jest.mock('../src/shared/i18n', () => mockI18n);
```
davemfish, line 45: Cool, if this is moved to `main` and loaded via preload, I think there will be a chance to mock it right in this file around line 16.

davemfish (changes requested):
Thanks for these updates @emlys , I added a few minor comments and I'll take a look again if we move the new `shared` javascript over to `main` and `preload` it.

workbench/tests/main/main.test.js
emlys: nah, we're not testing those windows anyway

workbench/src/renderer/components/DataDownloadModal/index.jsx
emlys: Exactly!

workbench/src/renderer/menubar/report.jsx
emlys: Yeah, it's probably ok

emlys:
@davemfish I spent a bit of time trying to move i18n into the preload script, and while it still might be a good option, what would you think about saving it for another issue? My concerns are
- It's not really how electron intends the preload script to be used: it's supposed to be to "securely expose privileged APIs into the renderer process". Whereas we're interested in sharing a complete instance, not an API, between processes. And I worry that could introduce some race conditions and things that we're not expecting.
- Moving i18n.js and the json translation files into preload/ broke the imports, I assume because vite isn't serving the json files the same way (?)
- The implementation would be different if we decide to do https://github.com/natcap/invest/issues/1168. Maybe it could wait until then?

davemfish:
> @davemfish I spent a bit of time trying to move i18n into the preload script, and while it still might be a good option, what would you think about saving it for another issue? My concerns are
> 
> * It's not really how electron intends the preload script to be used: it's supposed to be to "securely expose privileged APIs into the renderer process". Whereas we're interested in sharing a complete instance, not an API, between processes. And I worry that could introduce some race conditions and things that we're not expecting.

Yeah, good call.

> * Moving i18n.js and the json translation files into preload/ broke the imports, I assume because vite isn't serving the json files the same way (?)
> * The implementation would be different if we decide to do [Restart the app when language changed #1168](https://github.com/natcap/invest/issues/1168). Maybe it could wait until then?

@emlys I think the new plan you laid out in the issue is great. 

There is another problem to figure out in the meantime though -- revealed by failing pptr tests -- which is a missing dependency in the `main` process. `shared/i18n.js` is trying and failing to import `react`. 

electron-builder & vite share responsibility for packaging the dependencies, and we deliberately have react and other frontend dependencies listed only in `devDependencies` so that electron-builder will ignore them, letting vite take care of them instead.

A quick solution could be to move react back up to `dependencies`, but then we would end up bundling two copies of it, which feels kind of crude.

Otherwise I think this all gets sorted out by your new plan where there is only one instance of the `i18n` object, instantiated by the main process, and then we probably drop `react-i18next` altogether, right? I'll add some other thoughts to the new issue.

workbench/package.json
```diff
@@ -15,11 +15,10 @@
     "test-electron-app": "jest --runInBand --testMatch **/tests/binary_tests/*.test.js",
     "test-sampledata-registry": "jest --runInBand --testMatch **/tests/sampledata_linkcheck/*.test.js",
     "postinstall": "electron-builder install-app-deps",
-    "build-main": "babel --delete-dir-on-start src/main -d build/main --verbose --config-file ./babel.config.js --ignore **/__mocks__/* -s both",
+    "build-main": "babel --delete-dir-on-start src/main -d build/main --verbose --config-file ./babel.config.js --ignore **/__mocks__/* -s both --copy-files",
```
emlys, line 18: Copies the json files over to the build dir

workbench/babel.config.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
   ],
   plugins: [
     '@babel/plugin-transform-runtime',
+    '@babel/plugin-syntax-import-assertions',
```
emlys, line 8: This lets us directly import the json files in main/i18n

workbench/package.json
davemfish: Unfortunately it looks like we also need to add `--no-copy-ignored` for reasons: https://github.com/babel/babel/issues/6226

Jest was warning us about a duplicated mock because it found `__mocks__/logger.js` in `build/` and in `src/`.

davemfish (changes requested):
@emlys thanks for working through all of this. Just a couple minor things I noticed after looking with fresh eyes:

1. See my comment in `package.json`
2. I think the warning during tests about the uninitialized `i18next-react` comes down to exactly that: when we run a test like `openbutton.test.js`,  `renderer/i18n/i18n.js` is never imported or initialized anywhere. So I think if we add an
   `import '../src/renderer/i18n/i18n';` 
to `jest.override.js` then we solve that problem!

PR #1171, opened 2023-02-08
Use micromamba for building and cacheing python env in GHA

### LINKED ISSUE #1169: GHA: setup-miniconda failures
The failures are intermittent, but I've seen this a number of times now:
https://github.com/natcap/invest/actions/runs/4087567947/jobs/7048238625

Not clear what goes wrong. Logging from the conda downloads gets very strange, and the process exits with a failure code. 

It might be worth trying `micromamba` instead. It may or may not avoid this issue, but should generally be faster than `conda` and looks very easy to setup. I'm going to give it a try. It supports cache-ing too.
https://github.com/marketplace/actions/provision-with-micromamba

davemfish:
This PR switches from using `setup-miniconda` to using `provision-with-micromamba` to create python environments for all tests & builds. https://github.com/marketplace/actions/provision-with-micromamba

Hopefully we stop seeing the intermittent failure during conda installs noted in #1169. And I think we should also benefit from faster installs. 

Install times without using a cached env:
`conda` (5 samples): `[10, 11, 8, 13, 6]` minutes
`mamba` (7 samples): `[4.5, 2, 6, 3.5, 3, 3, 4]` minutes

In a couple places I had to modify a `conda list --export` command to `micromamba list`, since it's cli does not have the `--export` option. `export` does some text formatting to aid machine-reading. But for our purposes it's all the same to use the more human-readable format. It should not change the results of a `diff`.

Fixes #1169 

## Checklist

### DIFF
--- a/.github/actions/setup_env/action.yml
+++ b/.github/actions/setup_env/action.yml
@@ -70,4 +70,4 @@ runs:
 
     - name: List conda environment
       shell: bash -l {0} # conda only available in login shell
-      run: mamba list
+      run: micromamba list
--- a/.github/actions/setup_env/action.yml
+++ b/.github/actions/setup_env/action.yml
@@ -33,12 +33,6 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
 
-    - name: Setup conda environment
-      uses: conda-incubator/setup-miniconda@v2
-      with:
-        auto-update-conda: true
-        channels: conda-forge
-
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Looks good to me, let's give it a go!  Here's hoping it resolves the issue as well, and even if it doesn't, the faster build times should at least make it faster to re-run failed builds.

PR #1184, opened 2023-02-15
Address Github Actions warnings

### LINKED ISSUE #1183: Address Github Actions warnings
A bunch of warnings and a few errors (that don't fail the tests) have accumulated in our GHA runs. E.g. https://github.com/natcap/invest/actions/runs/4127028065 (scroll to the bottom). This includes some warnings about [deprecated commands that will be removed on June 1st 2023](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). We aren't using those commands directly, so we probably need to update other actions that we're using.

emlys:
## Description
Fixes #1183
Causes of the error and warning annotations were:
- We were using old versions of some actions. I bumped these up.
- The diff command was causing an error annotation when differences were found because then the exit code is 1. I changed the command so the exit code is always 0.
- The Qt UI tests raise a test exception, which causes an error annotation, probably because it isn't handled in the best way. Since we're deprecating Qt really soon I did not bother to fix it.

You can see that the annotations are gone, and that the 3 Qt-related errors remain, on the "Summary" page of the checks tab: https://github.com/natcap/invest/actions/runs/4189242762

## Checklist
- [ ] ~Updated HISTORY.rst and link to any relevant issue (if these changes are user-facing)~

### DIFF
--- a/.github/actions/setup_env/action.yml
+++ b/.github/actions/setup_env/action.yml
@@ -29,7 +29,7 @@ runs:
     # after the environment is updated, 'python' will point to the
     # conda-managed python and this python won't be used again
     - name: Set up python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ inputs.python-version }}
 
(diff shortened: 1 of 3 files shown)


davemfish (approved):
Looks good, thank you!

PR #1194, opened 2023-02-20
Error message for SWY when invalid soil group pixel found

### LINKED ISSUE #1193: Improve the error message in SWY if an invalid soil group is found
A [user on the forums](https://community.naturalcapitalproject.org/t/bounding-boxes-do-not-intersect-in-seasonal-water-yield-model/3387/13) encountered a `KeyError` when they provided a soil group pixel with a value of 6, which is outside the valid set of values: `{1, 2, 3, 4}`.  This exception should be made more readable.

phargogh:
This PR adds a human-readable exception raised in Seasonal Water Yield when a user provides a soil group that isn't permitted.  Only values of 1, 2, 3, 4 and nodata are permitted.

Fixes #1193 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -44,6 +44,10 @@ Unreleased Changes
 * Workbench
     * Added tooltips to the model tabs so that they can be identified even when
       several tabs are open (`#1071 <https://github.com/natcap/invest/issues/1088>`_)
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Looks good, thanks @phargogh !

PR #1195, opened 2023-02-20
Urban Nature Access: add sample data

phargogh:
This relatively simple PR adds sample data to the Urban Nature Access model and complements https://bitbucket.org/natcap/invest-sample-data/pull-requests/29.

RE: #722


pyproject.toml
```diff
@@ -52,5 +52,25 @@ build-backend = "setuptools.build_meta"
 version_scheme = "post-release"
 local_scheme = "node-and-date"
 
+[tool.isort]
```
dcdenu4, line 55: Excellent!

dcdenu4 (changes requested):
Hey @phargogh , looks like this PR is set up to merge directly into `main`!

phargogh:
Whoops, you're right!  I just updated the base.

pyproject.toml
phargogh: I've really liked having `isort` in there ... it's the kind of thing where I just put an import somewhere up top and I don't have to think about it at all!  `isort` just puts it in the right place.  Although it did take a little fiddling to define what the correct place is, since there's apparently some variety across python projects.

dcdenu4 (approved):

phargogh:
@dcdenu4 it looks like a couple builds are failing because of the recent deprecation of globio, so not strictly because of this PR.  Maybe I'll submit a separate PR to merge in the latest changes on the release branch?  I suspect that should mostly resolve these build issues.

dcdenu4:
Hey @phargogh, let me know if I can help with this at all! A few of those binary failures seem odd.

phargogh:
Thanks @dcdenu4 ! I think I've resolved it (I merged in the release branch, hence the large number of changed files now), but let's see what the builds say.

dcdenu4:
@phargogh, hmmm I took a look at the failing binaries and it's not clear to me what is failing.

phargogh:
OK!  @dcdenu4 I think we're probably OK now as far as the builds go.  There's just the failing mac builds now.

PR #1112, opened 2022-11-05
Output spec part 2

emlys:
## Description
Fixes #

## Checklist
- [ ] Updated HISTORY.rst and link to any relevant issue (if these changes are user-facing)
- [ ] Updated the user's guide (if needed)
- [ ] Tested the affected models' UIs (if relevant)

PR #1216, opened 2023-03-01
Output spec

### LINKED ISSUE #596: Equivalent of args_spec for model outputs
As we've discussed related to the recent ARGS_SPEC changes, it would be nice to have a similar data structure for each model's intermediate and final outputs.

This would store similar information, and we could reuse most/all of the ARGS_SPEC design:
- file name and corresponding variable name
- description
- units
- projection (which input projection does it inherit)
- vector geometries and fields

Outputs only use a subset of the possible input types and don't need validation, so really it will be simpler.

As an example of why this would be useful, today I discovered the SDR user's guide doesn't mention a dozen of the intermediate outputs. Without a central source of information, it's pretty tedious to figure out what they represent. We could generate this documentation like we are starting to do for the model inputs. 

Another use case would be in a workbench data viewer, we would need to access units and descriptions mapped to output file names. This information currently doesn't exist in any structured format.

Unlike the recent ARGS_SPEC updates, this is not a priority to happen before translation, because this info is not duplicated (exists only in the user's guide, if anywhere).

emlys:
## Description
Merging the output spec branch into release/3.13.0. These changes were already reviewed in #1086 and #1112. Closes #596 

## Checklist


phargogh (approved):
Looks good to me!  I didn't read through it _super_ closely, but I did go through everything and created a couple extra issues for things that we could take a look at in the future - nothing that should hold up this PR, though :)

* #1227
* #1228
* #1229
* #1230

PR #1226, opened 2023-03-02
Replace radio buttons with toggle switch for boolean inputs

emlys:
## Description
A toggle switch is language-neutral. It's also a little simpler to implement - no need to convert strings to booleans.
image

## Checklist


workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -220,30 +225,15 @@ export default function ArgInput(props) {
...
-        />
-      </React.Fragment>
+      <Form.Check
+        inline
+        type="switch"
+        id={argkey}
+        name={argkey}
+        onChange={() => updateArgValues(argkey, !value)}
+        disabled={!enabled}
+        bsCustomPrefix="form-switch"
```
davemfish, line 235: I think we need the `checked` prop here same as before. That way the state can be controlled by interactions other than just clicks, like when the app passes a set of initial arg values from a datastack file. 

And I suppose we should cover that by a test too. Sorry we didn't do it earlier! Maybe these ideas will help... In `setuptab.test.js` we could maybe add/modify tests in `describe('Arguments form input types'` to also expect a value for the input.  Something like 

```javascript
test.only('render a toggle with a value', async () => {
    const spec = baseArgsSpec('boolean');
    const { findByLabelText } = renderSetupFromSpec(spec, UI_SPEC, { arg: true });
    const input = await findByLabelText(`${spec.args.arg.name}`);
    // for some reason, the type is still checkbox when it renders as a switch
    expect(input).toHaveAttribute('type', 'checkbox');
    expect(input).toBeChecked();
  });
```
Where we added a third optional parameter to `renderSetupFromSpec`:
```javascript
function renderSetupFromSpec(baseSpec, uiSpec, initValues = undefined)
```
so we can pass a default args-dict object `{ arg: true }` to it and use it like:
```jsx
<SetupTab
      argsInitValues={initValues}
```

davemfish (changes requested):
Thanks, @emlys , this is much, much nicer this way! 

I caught one missing `prop` in the new `Form.check` and I'm sorry our tests were not really adequate to catch that going into this small refactor. So I suggested a way to improve them.

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
emlys: Nice, thanks for catching that!

Fortunately this was as simple as adding `checked={value}`. I made the test changes pretty much exactly as you suggested.

davemfish (approved):
Looks great, thank you!

PR #1233, opened 2023-03-03
Workbench bugfix for missing image files in production

### LINKED ISSUE #1220: Workbench images in About and Report menus are missing
I think this was an overlooked bug we created recently, in the `release/3.13.0` branch.

When moving the content of `about.html` and `report.html` to `.jsx` we broke a reference to a few image sources:

``

Previously, `vite` did its magic by finding & replacing this path for us during the build process, to point to the comparable location in `build/`. But I'm guessing that only works for html documents, not `.jsx`. So now the production javascript still points to a `src/` folder which doesn't exist.

davemfish:
Fixes #1220 

Apparently it works, and is good practice, to just `import` resources like images. Then `vite` will modify the import path for production builds. 

I confirmed this works in dev & production environments. 

## Checklist


emlys (approved):
Cool!

Issue #1217, opened 2023-03-01
Dropdown menus display their values, not the `display_name` from MODEL_SPEC


phargogh:
When I load up a model that has a dropdown menu in the workbench, the menu shows the option's key (which is the string that the model expects) rather than the `display_name`.  Here's an example from scenic quality, compared with what's in the MODEL_SPEC:

<img width="869" alt="Screen Shot 2023-03-01 at 10 37 41 AM" src="https://user-images.githubusercontent.com/4583897/222234231-26ddcf3c-0cda-4322-a3b2-2428c90915b3.png">
<img width="596" alt="Screen Shot 2023-03-01 at 10 37 54 AM" src="https://user-images.githubusercontent.com/4583897/222234240-3b0382d4-641e-4e58-bbfd-7a2f0e178e9a.png">

For us English speakers this isn't an issue because we can just make the keys (what the model expects) more human-readable.  But that also means that the dropdown menu options won't be translated.

emlys:
Fixed in #1235

PR #1235, opened 2023-03-03
Use option display names correctly

### LINKED ISSUE #1163: Option strings not rendering in RST
This turned up when working on the Urban Nature Access model, but I see that it's an issue with other models as well.

Anywhere we have an ARGS_SPEC dict with an "option_string", the user's guide rendering of that option string is unformatted:

Here's an example from Scenic Quality:
Screen Shot 2023-01-30 at 2 57 13 PM

And another from Coastal Vulnerability:
Screen Shot 2023-01-30 at 2 58 32 PM

emlys:
## Description
Fixes #1163 and #1217 

## Checklist


davemfish (changes requested):
Hi @emlys , this looks good. My only request is that we cover the Workbench bug with a test.

It looks like `setuptab.test.js` already tests with the array & dict forms of the options, but we're only testing for the expected `value` of the input, not the expected display content. Could we add an `expect(...).toHaveTextContent()` or something like that?

Oh, and there's a history conflict which I probably created with a merge a few minutes ago.

emlys:
@davemfish good suggestion. I added an assertion on to the current dropdown test to verify the display name.

davemfish (approved):
Thank you!

PR #1240, opened 2023-03-08
Fix cropped icon on windows

### LINKED ISSUE #1236: Workbench desktop icon is cropped/squashed
After replacing the interim icon with our standard `InVEST.ico` icon, the icon - as it appears in the Windows tray, or in the upper-left corner of the Workbench window - is not even close to square.

emlys:
## Description
Fixes #1236 
It turns out that both operating systems are very picky about icon sizes. Using a 256x256 icon resolved the issue on windows, but then the mac build failed with `ERR_ICON_TOO_SMALL`. So now we have separate icons for each.

## Checklist


workbench/electron-builder-config.js
```diff
@@ -42,12 +42,12 @@ const config = {
   artifactName: ARTIFACT_NAME,
   mac: {
     category: 'public.app-category.business',
-    icon: 'resources/InVEST-2.ico',
+    icon: 'resources/InVEST-2-574x574.ico',
```
dcdenu4, line 45: Woah, 574 is a fun dimension size. Was 512x512 too small as well or was there a particular reason for going 574x574?

dcdenu4 (commented):
Hey @emlys, I just had a quick question about the chosen size for the mac icon.

workbench/electron-builder-config.js
emlys: 574x574 was the size of the original image before this PR, so I just stuck with that.

dcdenu4 (approved):
Looks good Emily! Sorry random @ e person!

Issue #1239, opened 2023-03-07
Workbench Windows test failure


dcdenu4:
There's a new failure that popped up for just the Windows Workbench tests. I'm not sure if it's related to PR #1222 or a new separate issue.

https://github.com/natcap/invest/actions/runs/4327838148/jobs/7608191125#step:8:19

davemfish:
I tried to resolve this race condition in #1222 but apparently failed.

dcdenu4:
Currently skipping effected test on Windows and I think we're okay living with that for now. Closing.

PR #1241, opened 2023-03-08
for now, skipping a test on Windows. #1239.

davemfish:
Feel free to reject this PR as it does *not* address #1239 

But right now the test in question does more harm than good. Unfortunately I don't expect to have time this week to address it properly, so in the meantime I thought folks might appreciate having passing checks anyway.


dcdenu4 (approved):
I'm fine skipping this for now to allow more fluent PRs. Being a big minor release we'll be doing a lot of manual testing as well.

PR #1242, opened 2023-03-09
Urban Nature Access: Fixing out-of-range supply values

### LINKED ISSUE #1218: Urban Nature Access: sample data updates
From Stacie:

> Yay sample data metadata!
>
> Sample data - the "class" lulc descriptions that are included in the metadata should be added as a column to lulc-attributes.csv. I went to do this, and see that there's a class of 0 in the LULC raster that covers a lot of the area, but it does not have a class description, and does not appear to just be NoData - so what is it? This needs a class description too.
>
> Once we know what the 0 value is, I'll also be happy to add a column to the LULC raster attribute table with the description information, for those whose GIS can use it.
>
> It would be helpful to list the other provided layers (like tables) in the metadata, at least saying which input they're used for in the model, preferably also saying where their data came from, even if it's "we made it up for testing purposes, don't use this for your own analysis".

### LINKED ISSUE #1219: Urban Nature Access: `urban_nature_supply.tif` has out-of-range values on sample data
From Stacie:

> The results of running the sample data using the .json file inputs are mostly reasonable, except for urban_nature_supply.tif. It has some very extreme high values (10^15) that seem incorrect (and some small negative values to -0.05), and is very hard to symbolize in either ArcGIS or QGIS. What's up with this result?  I tried changing the radii to be much smaller, similar result. Tried changing the Decay function, similar result for all. If I use a Search Radius Mode of radius_per_urban_nature_class, the values are much more reasonable (20 - 14,000) and visualize easily in GIS.
>
> In trying to figure out what's up with that result (and having no inspiration), I wondered what happens with the double-counting that occurs between some of the population classes, like when some of the pop_female are also counted in pop_old?  Maybe it doesn't matter for the way things are calculated, and it's up to the user to know to compare pop_old with pop_young and pop_male with pop_female?

phargogh:
This PR fixes the out-of-range supply values that were the result of a failure to have a mutually agreed-upon nodata mask when doing a weighted sum.

This also fixes some filenames (no issue associated with this change).  @lmandle had requested that the "decay_" in the names of some files be renamed to something less macabre, so I went with "distance_weighted_" as an alternate prefix.

Some additional textual changes are also included that are needed for the user's guide and better suited to be stored here in the model source code than in the UG chapter itself.

The sample data are also updated to offer a few more notes on the sample tables and which values we made up.

Fixes #1219 
Fixes #1218

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ GIT_TEST_DATA_REPO_REV      := 29d8da596ff197d3cc6e355e7cd4313945b89b71
 
 GIT_UG_REPO                 := https://github.com/natcap/invest.users-guide
 GIT_UG_REPO_PATH            := doc/users-guide
-GIT_UG_REPO_REV             := 20bce284cb58ff4508a634cd2becde6ad175f7e1
+GIT_UG_REPO_REV             := 49fcb5b30b0f0336e932e6e32e662284a3e711af
 
 
 ENV = "./env"
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 DATA_DIR := data
 GIT_SAMPLE_DATA_REPO        := https://bitbucket.org/natcap/invest-sample-data.git
 GIT_SAMPLE_DATA_REPO_PATH   := $(DATA_DIR)/invest-sample-data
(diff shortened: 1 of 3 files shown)


dcdenu4 (changes requested):
One quick suggested edit. Otherwise looks good!

dcdenu4 (approved):

PR #1246, opened 2023-03-13
Minor Workbench installer configuration

### LINKED ISSUE #996: WB: make Windows desktop shortcut optional
Request from Stacie to ask the user before creating a desktop shortcut for the Workbench during installation. Seems like a good idea. 

I guess we could use `createDesktopShortcut` to stop creating the shortcut altogether. Would anyone miss it?
https://www.electron.build/configuration/nsis

Otherwise, if we want to ask the user their preference, we might need to use a custom NSIS script. https://www.electron.build/configuration/nsis#custom-nsis-script

### LINKED ISSUE #1069: The Workbench has 3 license files
@newtpatrol discovered recently that the workbench (at least on Windows) has 3 licenses, one for InVEST, one for Electron and one for Chromium.  We should clarify which one is which, maybe with an `InVEST_LICENSE.txt` prefix or something.

### LINKED ISSUE #1078: Allow the user to select the workbench's install location
@newtpatrol  mentioned this recently since it's convenient to sometimes have all  your InVEST installs in a single place that isn't `C:\Program Files`.  The current Workbench installed (tested on InVEST 3.12.0) does not allow the user to configure the install location.  I've confirmed that this is the case for both "all users" and "only this user".

davemfish:
* Adds a header icon to the NSIS window
* NSIS wizard asks user for the install location
* invest license file renamed to distinguish from electron & chromium license files
* NSIS installer no longer creates a desktop icon
* hide the package.json version string from the user by overriding some electron-builder defaults

I tested the Mac and Windows installers with all these changes: https://github.com/davemfish/invest/actions/runs/4407476799

Fixes #996 
Fixes #1069 
Fixes #1078 

## Checklist

### DIFF
--- a/workbench/electron-builder-config.js
+++ b/workbench/electron-builder-config.js
@@ -35,7 +35,7 @@ const config = {
   ],
   extraFiles: [{
     from: '../LICENSE.txt',
-    to: 'LICENSE.txt',
+    to: 'LICENSE.InVEST.txt',
   }],
   appId: APP_ID,
   productName: PRODUCT_NAME,
--- a/workbench/electron-builder-config.js
+++ b/workbench/electron-builder-config.js
@@ -19,9 +19,16 @@ const APP_ID = `NaturalCapitalProject.Invest.Workbench.${investVersion}`;
 const PRODUCT_NAME = `InVEST ${investVersion} Workbench`;
 const ARTIFACT_NAME = `invest_${investVersion}_workbench_${OS}_${ARCH}.${EXT}`;
 
+// this version appears as a footer in the NSIS installer & uninstaller.
+// It includes package.json version by default, but that is meaningless for us.
+// We can override that, but also must maintain semver compliance, so
+// always trim to this format.
+const installerVersion = investVersion.match(/[0-9]+\.[0-9]+\.[0-9]+/)[0];
+
 const config = {
   extraMetadata: {
     main: 'build/main/main.js',
+    version: installerVersion,
   },
   extraResources: [
     {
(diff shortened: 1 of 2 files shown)


workbench/electron-builder-config.js
```diff
@@ -19,9 +19,16 @@ const APP_ID = `NaturalCapitalProject.Invest.Workbench.${investVersion}`;
 const PRODUCT_NAME = `InVEST ${investVersion} Workbench`;
 const ARTIFACT_NAME = `invest_${investVersion}_workbench_${OS}_${ARCH}.${EXT}`;
 
+// this version appears as a footer in the NSIS installer & uninstaller.
+// It includes package.json version by default, but that is meaningless for us.
+// We can override that, but also must maintain semver compliance, so
+// always trim to this format.
+const installerVersion = investVersion.match(/[0-9]+\.[0-9]+\.[0-9]+/)[0];
+
```
phargogh, line 27: Interesting!  It does look like [SemVer 2.0 does support build metadata in the version string](https://semver.org/#spec-item-10), but not in quite the format that we've had it.  Would the extra version metadata be useful to us?

phargogh (commented):
Thanks @davemfish !  Looks good to me, and I just had one question about the `installerVersion` and whether it would be useful to include some of the additional build metadata.

workbench/electron-builder-config.js
davemfish: Oh interesting. So the place that this appears is actually at the very end of a string that already includes the fuller version string with build metadata.
Capture

I would have replaced this with nothing if I could have!

davemfish: Basically, we construct a custom `productName` and `artifactName` using invest's string. But `electron-builder` still always appends the version it finds in `package.json`. And `package.json` requires a `version` property. We have a fairly bespoke project here, with node + python.

phargogh: Huh, I wouldn't have expected both versions in there!  In that case, let's just go with what you have here, that sounds great.  Thanks!

phargogh (approved):

PR #1251, opened 2023-03-16
HRA fix for when the user skips all criteria in E or C

### LINKED ISSUE #1250: HRA outputs always high-risk when E/C pairs are entirely 0 for a habitat
A user on the forums ([link to private convo](https://community.naturalcapitalproject.org/t/hra-outputs/3495) with @phargogh and @davemfish ) has an interesting setup for their run where all of the criteria for all stressors affecting a single habitat are being ignored by setting those table values to 0.  This is resulting in some invalid values (numpy `WARNING`s from divide-by-zero), which I suspect is then throwing off the results.

phargogh:
This PR handles a case where the user may request to skip all habitat/stressor overlap pairs when calculating exposure or consequence.  The current behavior is that you'll get a bunch of division-by-zero warnings, which then percolates up to the risk classification where `inf` values are being interpreted as valid, which they are not.

This PR fixes the issue by handling this case and using 0 as the calculated score, which works within the existing model structure.

Holding in draft until the user confirms the fix, and (realistically) until after the 3.13.0 release.

Fixes #1250 

## Checklist


phargogh:
We haven't heard back from the user about this issue, specifically (he says he'll try it out soon), but I think this is a pretty clear fix that's needed.

davemfish (approved):
Looks good. There's a weird readthedocs error, but not related. We'll see if it keeps coming up.

PR #1268, opened 2023-03-29
Force the InVEST python application to run in UTF-8 mode

### LINKED ISSUE #1167: GDAL-level `SystemError` for some users
We have had a couple of issues on the forums now involving some kind of low-level exception in GDAL where we have almost no extra information.  Examples include:

* https://community.naturalcapitalproject.org/t/coastal-vulnerability-systemerror-built-in-function-len-returned-a-result-with-an-exception-set/3362
* https://community.naturalcapitalproject.org/t/invest-habitat-quality-systemerror/2694
* https://community.naturalcapitalproject.org/t/urban-cool-cooling-model-sample-error/899
* https://community.naturalcapitalproject.org/t/i-cannot-run-any-modules/2121/4

The issue appears to be happening at various times, including while reading a raster and while registering drivers.

phargogh:
This PR addresses issues arising from #1167, in which a cryptic `UnicodeDecodeError` would present in some users computers but we were unable to reproduce it locally.

My best understanding about the problem is that there is a discrepancy in the user's locale and the user's filesystem encoding, which, on Windows, would cause python to use the locale's encoding _as_ the filesystem encoding unless we specify otherwise.  This would ultimately cause problems when filepaths made it down to GDAL, where paths must be UTF-8.  Although I cannot pinpoint precisely which GDAL function is failing, I strongly suspect that it is this difference between expected and real encodings that is causing the error.

This is only an issue on Windows.  Mac and Linux are unaffected by the bug, as these are both POSIX systems.

As a solution, I have the workbench forcing the python application to start up in [UTF-8 mode on Windows](https://docs.python.org/3/using/windows.html#utf-8-mode).  This does have a side effect of causing text files to be opened as UTF-8 files by default, but this should not be a problem for InVEST because we generally tell people to use UTF-8 anyways.  [UTF-8 mode is scheduled to become the default in another 4 years, with the release of python 3.15.](https://peps.python.org/pep-0686/)

While debugging this, I also discovered how to enable GDAL's debug logging, so I have added a workbench setting to enable this in case it's useful in the future.  Note that to make full use of this logging, the user will also need to lower the logging threshold to DEBUG, which is then passed to the python application as an environment variable.

@emlys since the GDAL setting adds a new string to translate, is there anything else I will need to do to make sure it is ready for future translation?

@davemfish @emlys I took a stab at implementing this new setting, but please let me know if there is a better way to do any of this!

Fixes #1167 

## Checklist


emlys (changes requested):
Neat, I like that the GDAL debug setting is now exposed in the workbench! The translated string looks correct, we don't need to do anything else with it. I just had one minor code style suggestion. Also looks like there are a couple of files in the diff that are unrelated to this issue?

workbench/src/renderer/components/SettingsModal/index.jsx
```diff
@@ -164,6 +164,23 @@ class SettingsModal extends React.Component {
...
+              <Col sm="6">
+                <Form.Control
+                  id="gdal-logging-select"
+                  as="select"
+                  name="gdalLoggingLevel"
+                  value={investSettings.gdalLoggingLevel}
+                  onChange={this.handleChange}
+                >
+                  <option value="NORMAL" key="NORMAL">NORMAL</option>
+                  <option value="DEBUG" key="DEBUG">DEBUG</option>
```
davemfish, line 180: We should wrap the displayed strings to get translated `<...>{t('NORMAL')}</...>`

src/natcap/invest/utils.py
```diff
@@ -69,34 +69,41 @@ def _log_gdal_errors(*args, **kwargs):
...
+        # We had a case on the forums where GDAL was calling the error handler
+        # with no args or kwargs at all - it isn't even clear why the error
+        # handler was called in the first place..  Handling it here at the
+        # DEBUG level to avoid exceptions being logged to the logfile.
+        if len(args) + len(kwargs) == 0:
+            LOGGER.debug("_log_gdal_errors was called with no args/kwargs. "
+                         "Skipping.")
+            return
+    except Exception as e:
+        LOGGER.exception(str(e))
```
davemfish, line 82: How do we get here, some improperly typed `args` or `kwargs`? If so, do we want to proceed to with logging, or should we `return/raise` here?

davemfish (changes requested):
Hi @phargogh , thanks for explaining all this in so much detail. I think this is a nice solution!

I'm not sure if the changes in `utils.py` are causing this somehow, but `test_utils.py` seems like it never properly exits. I re-ran the jobs once and they all seem to be hanging again.

workbench/src/renderer/components/SettingsModal/index.jsx
phargogh: patched in bd9a9b64c

src/natcap/invest/utils.py
phargogh: It is super unclear because I'm having a tough time reproducing it on my system, but the user I'm working with on the forums was seeing something in `len` raising a `UnicodeDecodeError`.  And because of the nature of this error, I'm having a really hard time actually logging the object itself, or even the type!  It's very bizarre.

I added a note about this in c883dc196 and am now returning from the logging function.  You're right, if we're seeing this same error, we probably won't be able to do much about the rest of the logging.

src/natcap/invest/pollination.py
```diff
@@ -1,25 +1,25 @@
...
-import itertools
 import collections
-import re
-import os
-import logging
 import hashlib
 import inspect
+import itertools
+import logging
+import os
```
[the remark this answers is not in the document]
phargogh: Whoops!  Yes, these were left over from a few changes I made to pollinaton, and `isort` sorted the imports.  Unpatched in 05651604d

phargogh:
@davemfish @emlys thank you both for your review and good comments!  I have addressed them and fixed the builds hanging issue, but I'm still trying to understand why the user's getting the error she's getting.  So I'll mark this as a draft for now while I'm working that out and ping you both when it's ready for review.

PR #1272, opened 2023-04-04
Remove XLS format from HRA

### LINKED ISSUE #1271: Update to pandas 2.0.0: remove .xls
pandas 2.0.0 was released yesterday and it's causing one HRA test to break. The error is
```
for filename, func in [('target.csv', source_df.to_csv),
                       ('target.xls', source_df.to_excel),
                       ('target.xlsx', source_df.to_excel)]:
    full_filepath = os.path.join(self.workspace_dir, filename)
>   func(full_filepath, index=False)
...
ValueError: No engine for filetype: 'xls'
```
I couldn't find this mentioned in the release notes, but it looks like they [removed support for the .xls format](https://github.com/pandas-dev/pandas/commit/bcb8346e8106be4267ec77dfc603d0d77a3fda81) (it's been deprecated for a while, and was provided by another library that's no longer maintained). We should accordingly remove the .xls test from HRA - if pandas isn't supporting it, I doubt there's much demand for this outdated format.

Also an opportunity to revisit whether we still want HRA to support excel files, since it's the only model that does so.

emlys:
## Description
Fixes #1271
No changes needed to the user guide, it mentions Excel files but not the format specifically
The mac binary build failure is unrelated: #1273 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -42,11 +42,12 @@ Unreleased Changes
     * Fixed a bug in HRA where the model would error when all exposure and
       consequence criteria were skipped for a single habitat. The model now
       correctly handles this case. https://github.com/natcap/invest/issues/1250
+    * Tables in the .xls format are no longer supported. This format was
+      deprecated by ``pandas``. (`#1271 <https://github.com/natcap/invest/issues/1271>`_)
 * Scenic Quality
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Thanks Emily!

Issue #1276, opened 2023-04-07
Scenic Quality fieldnames are expected to be uppercase, but this is not documented


phargogh:
A user on the forum wrote to me in a DM:

> i found the issue (but it is not clear in the documentation), the columns radius weight height in the csv are required to be in uppercase. and the distance for RADIUS requires to be in negative numbers

It'd be worth verifying this and seeing how the model's fieldnames and/or documentation can be improved.

phargogh:
Well, it turns out that fields _are_ required to be upper case, but I'm not so sure about the negative values ... the model calls `math.fabs` on the field values, so we should be fine there.

PR #1277, opened 2023-04-07
Make SQ more tolerant of leading/trailing whitespace and case

phargogh:
SQ fieldnames are now more tolerant of leading/trailing whitespace and case.  It seemed easier on us in the long run to fix this rather than to just update the documentation to specify precise fieldnames.

Fixes: #1276

## Checklist


dcdenu4 (approved):
Easy, thanks @phargogh . I also like being case insensitive and handling trailing whitespace.

PR #1294, opened 2023-05-02
Urban Nature Access: Assume a default nodata value if one is not provided

### LINKED ISSUE #1293: Urban Nature Access assumes the LULC has a defined nodata value
A [user on the forums](https://community.naturalcapitalproject.org/t/nature-access-error/3701) reported this exception, which points to an undefined nodata value when masking out certain classes:

```python
File “natcap\invest\urban_nature_access.py”, line 781, in execute
File “natcap\invest\urban_nature_access.py”, line 2615, in _mask_raster
File “pygeoprocessing\geoprocessing.py”, line 517, in raster_calculator
File “natcap\invest\urban_nature_access.py”, line 2612, in _mask
TypeError: int() argument must be a string, a bytes-like object or a real number, not ‘NoneType’
```

phargogh:
Urban Nature Access has a masking step where, given a mask raster, convert pixel values to nodata outside of the mask.  This breaks if the user does not have a nodata value defined.  To get around this, this PR assumes a datatype-specific default nodata value if one has not already been provided.  I don't particularly like assuming values like this, but masking seems like an interesting possible case where it just doesn't make sense without having a nodata value defined or assumed.

Fixes #1293 

## Checklist

### DIFF
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -2606,17 +2606,32 @@ def _mask_raster(source_raster_path, mask_raster_path, target_raster_path):
     """
     source_raster_info = pygeoprocessing.get_raster_info(source_raster_path)
     source_raster_nodata = source_raster_info['nodata'][0]
+    source_raster_dtype = source_raster_info['datatype']
+
+    if source_raster_nodata is None:
+        source_numpy_dtype = source_raster_info['numpy_type']
+        try:
+            target_nodata = numpy.finfo(source_numpy_dtype).min
+        except ValueError:
+            # ValueError: When the numpy dtype is an int, finfo raises this.
+            target_nodata = numpy.iinfo(source_numpy_dtype).max
+        target_nodata = float(target_nodata)
+        LOGGER.warning(
+            'No nodata value defined, assuming a value of '
+            f'{target_nodata} on {source_raster_path}')
+    else:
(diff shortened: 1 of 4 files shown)


src/natcap/invest/urban_nature_access.py
```diff
@@ -2606,17 +2606,32 @@ def _mask_raster(source_raster_path, mask_raster_path, target_raster_path):
...
     def _mask(array, valid_mask):
         array = array.copy()
-        array[valid_mask == 0] = source_raster_nodata
+        array[valid_mask == 0] = target_nodata
         return array
 
     pygeoprocessing.raster_calculator(
         [(source_raster_path, 1), (mask_raster_path, 1)], _mask,
         target_raster_path,
-        datatype_target=source_raster_info['datatype'],
```
emlys, line 2618: Would it make sense to leave this in-line like it was before? Looks like the `source_raster_dtype` variable is only used once

emlys (commented):
@phargogh this looks good. I just wonder would it make sense to require that the input rasters have a defined nodata value? Maybe that should be something we require for all invest inputs.

phargogh:
In talking with Yingjie about this, we decided that the right solution for this case was to fix the data issue.

There's still the outstanding case that @emlys you're alluding to, which is whether we should require nodata values in general, or what should be done about nodata handling and consistently across models.

Closing this PR in lieu of the future discussion.

PR #1314, opened 2023-06-02
Fix natcap/invest#1105: mask qf sum using nodata from all months

### LINKED ISSUE #1105: SWY: negative quickflow values on edges
Two users have reported negative quickflow values at the edges of the AOI:
https://community.naturalcapitalproject.org/t/there-are-negative-values-in-quickflow/3173
https://community.naturalcapitalproject.org/t/seasonal-water-yields-negative-results/1633/8

emlys:
## Description
Fixes #1105 

Breaking these changes out from the other SWY issues.

The issue was that we were assuming all 12 monthly quickflow rasters had nodata in the same place. We were masking all of them based on the nodata mask for January. This was a wrong assumption - each monthly quickflow raster is derived from a user-provided monthly precipitation raster, which can have different nodata areas.

This bug was the root cause of at least one user complaint that they had negative values on the edges of their quickflow raster. They had a negative nodata value that was being treated as valid data, and summed together with real data, resulting in a negative sum.

This PR masks the quickflow sum based on all 12 monthly quickflow rasters. The quickflow sum will have nodata where any of the 12 monthly rasters has nodata. The quickflow sum will have valid data only where all 12 monthly rasters have valid data.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -79,6 +79,10 @@ Unreleased Changes
     * Fixed an issue with sediment deposition progress logging that was
       causing the "percent complete" indicator to not progress linearly.
       https://github.com/natcap/invest/issues/1262
+* Seasonal Water Yield
+    * Fixed a bug where monthy quickflow nodata pixels were not being passed
+      on to the total quickflow raster, which could result in negative values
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Thanks @emlys!

PR #1321, opened 2023-06-09
Remove GDAL cache limit in SWY

### LINKED ISSUE #1320: SWY's limits on the GDAL Cache Max is limiting
SWY is unique among the InVEST models in that it makes a call to `gdal.SetCacheMax`.  We had a [support case over slack](https://stanford-natcap.slack.com/archives/CECFUU1DW/p1685049456935449) where, given a Float32 DEM of the entire country of Pakistan, the model was just outrageously slow.  When given an Int16 version of the same raster, the model was quite performant.  Removing the one line that set the cache limit, the model completed in a very reasonable amount of time, several orders of magnitude faster than with the cache limit in place.

By default, [GDAL uses 5% of the system's installed memory](https://gdal.org/user/configoptions.html#performance-and-caching) for caching, which is significantly more than what SWY was restricting it to.  In this case, we were likely invoking a pile of extra disk reads because of cache misses during pitfilling that were avoided when we set the cache limit.

phargogh:
This PR removes the GDAL cache limit.  The current default is 5% of the system's installed memory.  This was much more performant when testing this on Lingling's computer, running on the whole country of Pakistan for the BELA project.

Fixes #1320 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -83,6 +83,9 @@ Unreleased Changes
     * Fixed a bug where monthy quickflow nodata pixels were not being passed
       on to the total quickflow raster, which could result in negative values
       on the edges (`#1105 <https://github.com/natcap/invest/issues/1105>`_)
+    * Removed the GDAL cache size limit on this model, which means that, by
+      default, the model will use up to 5% of installed memory.
+      https://github.com/natcap/invest/issues/1320
 
 3.13.0 (2023-03-17)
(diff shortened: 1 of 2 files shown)


emlys (approved):
Neat! I'm curious why we set the cache limit lower originally, but whatever reason, it doesn't seem to be important now. The one failing check looks unrelated.

PR #1322, opened 2023-06-10
`utils.build_lookup_from_csv`: pass kwargs to `pandas.read_csv`

### LINKED ISSUE #1319: Expose the `dtype` parameter of `pandas.read_csv` in `utils.build_lookup_from_csv`
Underlying `utils.build_lookup_from_csv`, we use `pandas.read_csv` to read in the file. [`read_csv`](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html?highlight=read_csv) has many options, including `dtype`, which lets you set a data type for each column. `pandas` will interpret each value in the column as an object of that type, or raise an error if it can't do so.

Exposing this option would be useful to address #1278 and prevent similar bugs in other models. Because we're not specifying the dtype, pandas infers it. The resulting dtype can change depending on the user data. In #1278, the user had all integer values in the ID column, so `pandas` set the dtype to `int`. If there were any non-numeric values, `pandas` would set the dtype to `str`. We should enforce a `str` type, an `build_lookup_from_csv` is an easy place to do that.

emlys:
## Description
Fixes #1319 

## Checklist

### DIFF
--- a/src/natcap/invest/hra.py
+++ b/src/natcap/invest/hra.py
@@ -1845,8 +1845,7 @@ def _open_table_as_dataframe(table_path, **kwargs):
         return excel_df
     else:
         return utils.read_csv_to_dataframe(
-            table_path, sep=None, to_lower=True, engine='python',
-            expand_path_cols=['path'], **kwargs)
+            table_path, to_lower=True, expand_path_cols=['path'], **kwargs)
 
 
 def _parse_info_table(info_table_path):
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -582,7 +582,7 @@ def build_file_registry(base_file_path_list, file_suffix):
 
 def build_lookup_from_csv(
         table_path, key_field, column_list=None, to_lower=True,
-        expand_path_cols=[]):
+        expand_path_cols=[], **kwargs):
     """Read a CSV table into a dictionary indexed by ``key_field``.
 
     Creates a dictionary from a CSV whose keys are unique entries in the CSV
@@ -607,6 +607,7 @@ def build_lookup_from_csv(
             columns that contain paths to expand. Any relative paths in these
             columns will be expanded to absolute paths. It is assumed that
             relative paths are relative to the CSV's path.
+        **kwargs: additional kwargs will be passed to ``pandas.read_csv``
 
     Returns:
         lookup_dict (dict): a dictionary of the form
@@ -631,7 +632,7 @@ def build_lookup_from_csv(
 
     table = read_csv_to_dataframe(
         table_path, to_lower=to_lower, sep=None, index_col=False,
-        engine='python', expand_path_cols=expand_path_cols)
+        engine='python', expand_path_cols=expand_path_cols, **kwargs)
 
     # if 'to_lower`, case handling is done before trying to access the data.
     # the columns are stripped of leading/trailing whitespace in
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -581,8 +581,7 @@ def build_file_registry(base_file_path_list, file_suffix):
 
 
 def build_lookup_from_csv(
-        table_path, key_field, column_list=None, to_lower=True,
-        expand_path_cols=[], **kwargs):
+        table_path, key_field, column_list=None, to_lower=True, **kwargs):
     """Read a CSV table into a dictionary indexed by ``key_field``.
 
     Creates a dictionary from a CSV whose keys are unique entries in the CSV
@@ -603,11 +602,7 @@ def build_lookup_from_csv(
         to_lower (bool): if True, converts all unicode in the CSV,
             including headers and values to lowercase, otherwise uses raw
             string values. default=True.
-        expand_path_cols (list[string])): if provided, a list of the names of
-            columns that contain paths to expand. Any relative paths in these
-            columns will be expanded to absolute paths. It is assumed that
-            relative paths are relative to the CSV's path.
-        **kwargs: additional kwargs will be passed to ``pandas.read_csv``
+        **kwargs: additional kwargs will be passed to ``utils.read_csv_from_dataframe``
 
     Returns:
         lookup_dict (dict): a dictionary of the form
@@ -631,8 +626,7 @@ def build_lookup_from_csv(
         col_list.append(key_field)
 
     table = read_csv_to_dataframe(
-        table_path, to_lower=to_lower, sep=None, index_col=False,
-        engine='python', expand_path_cols=expand_path_cols, **kwargs)
+        table_path, to_lower=to_lower, index_col=False, **kwargs)
 
     # if 'to_lower`, case handling is done before trying to access the data.
     # the columns are stripped of leading/trailing whitespace in
@@ -706,8 +700,8 @@ def expand_path(path, base_path):
 
 
 def read_csv_to_dataframe(
-        path, to_lower=False, sep=None, encoding=None, engine='python',
-        expand_path_cols=[], **kwargs):
+        path, to_lower=False, expand_path_cols=[], sep=None, engine='python',
+        encoding='utf-8-sig', encoding_errors='replace', **kwargs):
     """Return a dataframe representation of the CSV.
 
     Wrapper around ``pandas.read_csv`` that standardizes the column names by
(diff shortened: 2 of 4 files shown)


src/natcap/invest/utils.py
```diff
@@ -603,10 +602,7 @@ def build_lookup_from_csv(
         to_lower (bool): if True, converts all unicode in the CSV,
             including headers and values to lowercase, otherwise uses raw
             string values. default=True.
-        expand_path_cols (list[string])): if provided, a list of the names of
```
emlys, line 606: This doesn't need to be documented here, it's passed directly to `read_csv_to_dataframe`, so it's just part of `**kwargs`.

src/natcap/invest/hra.py
```diff
@@ -1845,8 +1845,7 @@ def _open_table_as_dataframe(table_path, **kwargs):
         return excel_df
     else:
         return utils.read_csv_to_dataframe(
-            table_path, sep=None, to_lower=True, engine='python',
-            expand_path_cols=['path'], **kwargs)
+            table_path, to_lower=True, expand_path_cols=['path'], **kwargs)
```
emlys, line 1848: This was redundant (setting kwarg to its default value).

src/natcap/invest/wind_energy.py
```diff
@@ -1856,8 +1856,7 @@ def _read_csv_wind_parameters(csv_path, parameter_list):
     # this doesn't benefit from `utils.read_csv_to_dataframe` because there
     # is no header to strip whitespace
     # use sep=None, engine='python' to infer what the separator is
-    wind_param_df = pandas.read_csv(
-        csv_path, header=None, index_col=0, sep=None, engine='python')
+    wind_param_df = pandas.read_csv(csv_path, header=None, index_col=0)
```
emlys, line 1859: This was redundant (setting kwarg to its default value).

src/natcap/invest/wind_energy.py
```diff
@@ -1975,7 +1974,7 @@ def _read_csv_wind_data(wind_data_path, hub_height):
             to dictionaries that hold wind data at that location.
 
     """
-    wind_point_df = utils.read_csv_to_dataframe(wind_data_path, to_lower=False)
+    wind_point_df = utils.read_csv_to_dataframe(wind_data_path)
```
emlys, line 1977: See above

tests/test_utils.py
```diff
@@ -856,6 +856,27 @@ def test_csv_dialect_detection_semicolon_delimited(self):
         self.assertEqual(lookup_dict[4]['header3'], 'bar')
         self.assertEqual(lookup_dict[1]['header1'], 1)
 
+    def test_csv_utf8_encoding(self):
```
davemfish, line 859: This test looks identical to `test_csv_utf8_bom_encoding` unless I'm missing something subtle.

src/natcap/invest/utils.py
```diff
@@ -715,41 +710,36 @@ def read_csv_to_dataframe(
...
-        encoding = 'utf-8-sig'
     try:
         dataframe = pandas.read_csv(
-            path, engine=engine, encoding=encoding, sep=sep, **kwargs)
+            path, sep=sep, engine=engine, encoding=encoding,
+            encoding_errors=encoding_errors, **kwargs)
     except UnicodeDecodeError as error:
         LOGGER.error(
-            f'{path} must be encoded as utf-8 or ASCII')
+            f'{path} must be encoded as UTF-8 or ASCII')
```
davemfish, line 742: To test my understanding...is it possible to get here with the default of `encoding_errors='replace'`? Or is this only catching errors if we were to use `encoding_errors='strict'`?

src/natcap/invest/utils.py
```diff
@@ -705,8 +700,8 @@ def expand_path(path, base_path):
 
 
 def read_csv_to_dataframe(
-        path, to_lower=False, sep=None, encoding=None, engine='python',
-        expand_path_cols=[], **kwargs):
+        path, to_lower=False, expand_path_cols=[], sep=None, engine='python',
+        encoding='utf-8-sig', encoding_errors='replace', **kwargs):
```
davemfish, line 704: From your description Emily, it sounds like there should be no difference in behavior with these new defaults for `encoding` and `encoding_errors`. And I see you added a test to show the replacement happening now. But I don't think we were testing for that before so I'm not sure if the replacement was happening or if an error was raised.

In that context I'm wondering about the case where a column contains a filepath that happens to include a non-ASCII character (e.g. in a person's Windows username) and it gets replaced such that the resulting filepath cannot be opened. Is this a concern? @phargogh do you have any thoughts since you were dealing with such usernames a lot recently?

It seems like we would rather have an exception than do the replacement in this specific case. But again, if these changes are not introducing any new behavior here then I support them.

davemfish (changes requested):
Thanks @emlys , I have a question about the new `encoding` defaults, but it might not warrant any changes. 

And I think the new test looked identical to an existing one?

src/natcap/invest/utils.py
emlys: @davemfish the new test would fail without these changes, but based on the pandas documentation and comments in our code, I think it's in line with the original intent.

See these notes from the documentation for the `encoding` argument:
> Changed in version 1.2: When `encoding` is `None`, `errors="replace"` is passed to `open()`. Otherwise, `errors="strict"` is passed to `open()`. This behavior was previously only the case for `engine="python"`.

> Changed in version 1.3.0: `encoding_errors` is a new argument. `encoding` has no longer an influence on how encoding errors are handled.

And from the `read_csv_to_dataframe` docstring:
```
encoding (string): name of encoding codec to pass to `pandas.read_csv`.
            Defaults to None. Setting engine='python' when encoding=None allows
            a lot of non-UTF8 encodings to be read without raising an error.
            Any special characters in other encodings may get replaced with the
            replacement character.
            If encoding=None, and the file begins with a BOM, the encoding gets
            set to 'utf-8-sig'; otherwise the BOM causes an error.
engine (string): kwarg for pandas.read_csv: 'c', 'python', or None.
            Defaults to 'python' (see note about encoding).
```
So it seems like this was originally implemented pre-pandas 1.2, and we were using the combination of `engine='python'` and `encoding=None` to get the character replacement behavior. But that was never tested, so with pandas>=1.3.0, you get an error instead of replacement.

src/natcap/invest/utils.py
emlys: I think it would only happen with `encoding_errors='strict'`.

tests/test_utils.py
emlys: Removed the comment on line 864.

src/natcap/invest/utils.py
phargogh: I recommend that we raise errors instead of replacing invalid characters, at least for filepaths.  My main concern here is that the replacement will give us a different filepath entirely ... and if a file _does_ happen to exist at that location, then the issue is even tougher to grok from the logfile alone.  By replacing characters, we pass the exception down the road to GDAL or another function that won't be able to find and open the file, which is more cryptic than it needs to be.  Furthermore, rewriting the filepath could open the doors to a security vulnerability if a library were exploited and a filepath planted in the right place.  We can avoid these various concerns by simply being clear about when errors are encountered in filepaths and raising those exceptions to the user.

It's times like these where I wish we had a table importing interface like what you find in openoffice and QGIS that would offer an opportunity to review how the table is being interpreted and to make any tweaks to things like encoding, separators, etc. before the table is actually loaded.

tests/test_utils.py
phargogh: Just to confirm, defaulting to `utf-8-sig` will handle both cases, where the BOM is present and BOM is not present?  I think that's what I'm reading here, but wanted to check (fantastic if so!)

emlys: Yep, it works in both cases. From https://docs.python.org/3/library/codecs.html#standard-encodings:
> On encoding the utf-8-sig codec will write 0xef, 0xbb, 0xbf as the first three bytes to the file. On decoding utf-8-sig will skip those three bytes if they appear as the first three bytes in the file.

src/natcap/invest/utils.py
emlys: I agree that replacement makes sense for some data but not for filepaths. How about we change the default `encoding_errors` to `'strict'` (in keeping with what the undocumented, untested behavior has been since pandas 1.3)? And if we get complaints about `UnicodeDecodeError`s because people are using non-UTF8 encodings, then revisit the replacement thing?

phargogh: Sounds good to me!

emlys: Ok, I made this change and added a test asserting that the `UnicodeDecodeError` is raised on non-UTF8 special characters.

davemfish: Great, thanks for adding that test!

tests/test_utils.py
```diff
@@ -856,6 +856,27 @@ def test_csv_dialect_detection_semicolon_delimited(self):
...
+        """utils: test that CSV read correctly with UTF-8 encoding."""
+        from natcap.invest import utils
+
+        csv_file = os.path.join(self.workspace_dir, 'csv.csv')
+        # writing with utf-8-sig will prepend the BOM
```
davemfish, line 864: We want to remove this comment, right?

tests/test_utils.py
```diff
@@ -856,6 +856,27 @@ def test_csv_dialect_detection_semicolon_delimited(self):
...
+            file_obj.write(textwrap.dedent(
+                """
+                header1,HEADER2,header3
+                1,2,bar
+                4,5,FOO
+                """
+            ).strip())
+        lookup_dict = utils.build_lookup_from_csv(
+            csv_file, 'header1')
+        # assert the BOM prefix was correctly parsed and skipped
```
davemfish, line 875: And this one

tests/test_utils.py
emlys: Sorry, I thought I'd removed these already. Lost that change somehow I guess

davemfish (approved):

PR #1329, opened 2023-06-14
SWY: don't fill in nodata pixels in quickflow

### LINKED ISSUE #1317: SWY: Quickflow function sets nodata pixels to 0
In the monthly quickflow calculation, some pixels that seemingly should be nodata get intentionally set to 0. This is probably the root cause of [this issue Adrian reported](https://community.naturalcapitalproject.org/t/handling-of-nodata-values-in-the-swy-model-soil-raster/3720). 

Why are these pixels being set to 0, and do we still think that's the correct approach?

https://github.com/natcap/invest/blob/061b25bf00e4d4f8a6292965d42cfbfbdb5fe7f9/src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py#L1136-L1140

emlys:
## Description
Fixes #1317 
Just removing that line of code. I tested this out on the sample data and the results are basically unchanged (there are like 4 more nodata pixels in the result).

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -86,6 +86,8 @@ Unreleased Changes
     * Removed the GDAL cache size limit on this model, which means that, by
       default, the model will use up to 5% of installed memory.
       https://github.com/natcap/invest/issues/1320
+    * Fixed a bug where nodata pixels in quick flow were being set to 0
+      (`#1317 <https://github.com/natcap/invest/issues/1317>`_)
 
 3.13.0 (2023-03-17)
 -------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -86,8 +86,10 @@ Unreleased Changes
     * Removed the GDAL cache size limit on this model, which means that, by
       default, the model will use up to 5% of installed memory.
       https://github.com/natcap/invest/issues/1320
-    * Fixed a bug where nodata pixels in quick flow were being set to 0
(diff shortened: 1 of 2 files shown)


HISTORY.rst
```diff
@@ -86,6 +86,8 @@ Unreleased Changes
     * Removed the GDAL cache size limit on this model, which means that, by
       default, the model will use up to 5% of installed memory.
       https://github.com/natcap/invest/issues/1320
+    * Fixed a bug where nodata pixels in quick flow were being set to 0
```
dcdenu4, line 89: Hey @emlys, I'm thinking of future us and wondering if we could overly expand on some context here. Mostly, because the prior implementation appears to have been a design decision by someone.  Maybe something like:

>Allowing monthly quick flow nodata values to propagate instead of being set to 0. The design choice to set to 0 was not well documented and was causing confusion about why some outputs had nodata values where other did not. By allowing nodata to propagate instead we're erroring on the side of not assuming data where this is not which could lead to misinterpretations of the results.

Not the best wordsmithing!

dcdenu4 (changes requested):
Thanks @emlys , just one quick suggestion!

HISTORY.rst
emlys: I expanded the note along these lines. I think the link to the issue also provides some context.

dcdenu4 (approved):

Issue #1286, opened 2023-04-28
Workbench - handle unexpected data in localStorage


davemfish:
We should safeguard against loading bad data from localStorage, where we keep settings & recent runs data. It seems there are instances where bad data is written if a crash happens. For example, https://community.naturalcapitalproject.org/t/invest-3-13-0-crashing-on-start-on-macos-12-4/3684

We should add safeguards to avoid writing bad data...for example, why would an invest exception like the one from this user lead to bad data being written to the "recent runs" store?

And also safeguards when reading data from the database. It would be better to clear out the bad items, or even reset the whole database, rather than letting the error propagate to the `ErrorBoundary`.

davemfish:
This is related to #1168 in the sense that that issue proposes moving persistent data out of the browser and onto the filesystem where it can be managed by the electron main process. If we plan to do that, then we should not go further down the road of adding complexity to the `localStorage` or `IndexedDB` browser storage that we're using now, such as by adding schema that could validate the read & writes to those databases.

Issue #1275, opened 2023-04-07
Make sure two different workbench versions can run on the same system


davemfish:
Because of `requestSingleInstanceLock`, we don't expect multiple clients to run at the same time, but they should be able to run independently, and might not because they share some resources on the filesystem:

Workbench settings & recent jobs are stored in the browser's localstorage. If the schema for these data changes between workbench versions, something might be broken when the other version is opened.

I don't know of this happening in the wild yet, but I think it's worth testing to find out if it does. And generally adding some error-handling around corrupt data in localstorage is a good idea.

davemfish:
A simple solution here could be to keep all data & settings isolated by version. The downside is that it probably makes sense for the settings to be shared across versions whenever they are compatible, which should be almost always I would hope.

For the data (recent invest runs), maybe it does _not_ make sense to share across versions. Because I think the use-case for having multiple versions of the Workbench would be to compare models (inputs & outputs) across invest versions. So it's probably expected that the "Recent Runs" available in the Workbench are from that same version of invest.

So maybe recent runs data should always be stored & requested by invest version.

And for settings, maybe something like, 
1. a newly installed Workbench tries to load settings from an existing store tagged by workbench version
2. if such a store does not exist, try loading from a store tagged with an older version,
3.  if schemas mismatch (or no stores existed at all), then instantiate a new settings store that is tagged by workbench version. 

So settings would carry over from an older installation to a newer one. But the older installation would still load settings from it's own store. And changing settings would only update the store associated with that workbench version.

We should also confirm that uninstalling the Workbench then only removes data related to that version.

davemfish:
> We should also confirm that uninstalling the Workbench then only removes data related to that version.

By default no user data is removed during uninstall. I confirmed this on Windows by observing that all contents of `C:\Users\dmf\AppData\Roaming\invest-workbench` still exists after uninstalling.

davemfish:
> For the data (recent invest runs), maybe it does not make sense to share across versions. Because I think the use-case for having multiple versions of the Workbench would be to compare models (inputs & outputs) across invest versions. So it's probably expected that the "Recent Runs" available in the Workbench are from that same version of invest.

We back-tracked on this and decided that the expected behavior would be to populate recent runs from prior versions of invest, but to skip any runs for deprecated models. We think this will be a convenient feature, compared to the alternative of always having "Recent Runs" be an empty list whenever a user installs the latest Workbench version.

PR #1330, opened 2023-06-14
Restart the app on change of language setting

### LINKED ISSUE #1168: Restart the app when language changed
Tell users they must restart the app (or automate restarting it) when the language setting is changed.

- This would eliminate some complexity where the renderer process has to tell the main process to change its language. Instead, the language setting would be constant throughout the life of each process.
- It would fix an (unimportant) loophole where changing the language setting does not affect the "About"/"Report a problem" windows if they are already open. Because they are parallel renderer processes, it would be complicated to communicate the change of language and reload those windows.

The downside is that changing the language setting is a larger interruption to using the app, but 
- we already reload the renderer process, which closes all your tabs
- requiring restart for settings to take effect is a common pattern
- we don't expect users to change this setting very often

### current process
(on startup)
1. main process starts
2. renderer process starts
3. renderer process reads language setting from storage
4. renderer sets its i18n instance to that language
5. via IPC, renderer tells main to change its language
6. main sets its i18n instance to that language
 
(on change of the language setting)
1. renderer process updates the language setting in storage
2. renderer process restarts (`window.reload`)
3. repeat steps 3 - 6 above

## alternative
(on startup)
1. main process starts
2. main process reads language setting from storage
3. main process sets its i18n instance to that language
4. renderer process starts
5. renderer process sets its i18n instance to that language (could be read from a variable exposed in preload)

(on change of the language setting)
1. renderer process (tells main process to) updates the language setting in storage
2. user quits and restarts the app
3. repeat steps 1 - 5 above

emlys:
## Description
Fixes #1168 

## Checklist


workbench/src/main/main.js
```diff
@@ -102,14 +110,6 @@ export const createWindow = async () => {
       menuTemplate(mainWindow, ELECTRON_DEV_MODE, i18n)
     )
   );
-  // when language changes, rebuild the menu bar in new language
```
emlys, line 105: No longer need to do this because we'll quit and restart in the new language

davemfish:
By the way @emlys a week ago I was trying out `electron-store` in the context of #1275 and hit the same issue you have here in the failing tests: `electron-store` only works in an `electron` runtime. I started down the path of mocking the whole module, by simply mocking a `Store` as a `Conf`, since `electron-store` seems to be a very light wrapper around https://github.com/sindresorhus/conf.

I never got that working; there were issues with how `conf` was getting transpiled by jest/babel, I think. At that point I stopped and did the electron & jest updates instead. Anyway we should chat about how to proceed!

Sorry that I didn't make it obvious that I was working on similar things, that's my bad. There is some potentially some overlap between issues like #1275 and #1168 and #1286

davemfish (changes requested):
All tests passing! This is working great and having that language value stored in the main process seems like a much cleaner solution, codewise. I have a few minor comments/suggestions below. And then some thoughts after trying it out here.

Actually the restart works even better than I expected! It's working dev mode, which I wasn't expecting given our earlier conversation. And it restarts very quickly.

I'm noticing that the python subprocesses do not exit...that could be really useful, except that new ones are launched during the restart. So my task manager shows multiple `python` and `invest` processes running. I think that comes down to calling `app.exit()` instead of `app.quit()`. You can see we do some cleanup in `app.on('before-quit')` in `main.js`. It's nice to have the fast restart! And leaving the python processes running definitely helps with that, but I'm not sure it's worth the effort to manage all that since a user will probably only ever go through this once. Maybe easiest to just do the full quit & startup routines. What do you think?

Similar story for the node processes that launch on `yarn start`, like `vite` and `esbuild`. These also don't exit on restart, and again that makes for very fast startup. I'm not sure why they don't exit when electron exits, but it's somewhat convenient that they don't. And unlike the python processes, no new processes are launched, so that's good. The only downside here is that finally closing the workbench after a restart does not close those processes. I don't consider this a big deal since it won't come up very much. What do you think?

Also worth noting that after a restart, the electron app is just generally not associated with the shell where it was opened the first time, so logging and stuff doesn't appear there. Again, not a big deal in my opinion. But if we use this feature in dev mode, there will be some background processes left behind.

workbench/src/renderer/components/SettingsModal/index.jsx
```diff
@@ -124,18 +136,18 @@ class SettingsModal extends React.Component {
...
-                    </Form.Text>
                   </Form.Label>
                   <Col sm="4">
                     <Form.Control
                       id="language-select"
                       as="select"
                       name="language"
-                      value={investSettings.language}
-                      onChange={this.handleChange}
+                      value={window.Workbench.LANGUAGE}
```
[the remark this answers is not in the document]
emlys: Yeah, this is intentionally kind of a fake dropdown - like you describe in your second comment. I'll update the modal to confirm which language you're switching to.

workbench/src/renderer/components/SettingsModal/index.jsx
```diff
@@ -264,6 +276,26 @@ class SettingsModal extends React.Component {
             <span>{t('no invest workspaces will be deleted')}</span>
           </Modal.Body>
         </Modal>
+        <Modal show={showConfirmLanguageChange} className="confirm-modal" >
+          <Modal.Header>
+            <Modal.Title>Warning</Modal.Title>
```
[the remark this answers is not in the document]
emlys: Sure! Made it a `h5`.

workbench/src/main/setupChangeLanguage.js
```diff
@@ -1,16 +1,25 @@
...
+  });
+
   ipcMain.handle(
     ipcMainChannels.CHANGE_LANGUAGE,
     (e, languageCode) => {
       logger.debug('changing language to', languageCode);
-      i18n.changeLanguage(languageCode);
+      store.set('language', languageCode);
+      app.relaunch();
+      app.exit();
```
davemfish, line 22: Hey @emlys I think you may have missed my big comment about this `app.exit()` vs. `app.quit()`. Sorry I didn't comment in-line, it all occurred to me as I was testing things out after I had reviewed the code. Please see my big comment and let me know what you think about mitigating leftover processes. Thanks!

workbench/src/renderer/components/SettingsModal/index.jsx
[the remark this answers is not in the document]
davemfish: Cool, that makes sense

workbench/src/main/setupChangeLanguage.js
emlys: Ah sorry I missed that. I agree that it would be better to `quit()` and have the cleanup run, but in my testing `quit()` does not actually close the window. You end up with duplicate workbench windows. (At least on mac - the latest binary artifacts on this branch have that change, if you'd like to try it). I can look into why that's happening, but it seemed like something that might be outside our control.

davemfish: @emlys Ahh interesting, so maybe it is not so simple as switching `exit()` for `quit()`. Nevertheless, I think we do need to come with an approach to make sure the python processes are cleaned up, or not duplicated on startup if we wish to leave them open.

davemfish: BTW we have a `/ready` endpoint for checking if the flask app is running or not, so that could be handy.

emlys: @davemfish I think this is a bug in the existing `before-quit` handler. When I quit the app normally, the python processes are not cleaned up. It gets stuck in an infinite loop of `shutdownPythonProcess` calls. Same thing is happening when the app quits programatically.

emlys: You were right, this was happening because of an old python process. When there's only the expected process, the cleanup succeeds.

davemfish (approved):

Issue #1325, opened 2023-06-13
Workbench: upgrade to React18


davemfish:
React 18 has significant breaking changes and enhancements. https://react.dev/blog/2022/03/29/react-v18

It seems like the performance enhancements could be really nice. A brief attempt at upgrading revealed some issues though.

1. A number of new warnings from `testing-library` about `act()` calls. Some of these were caused by conflicting versions of `testing-library/dom`. That specific dependency should be removed from `package.json` so that other libraries (e.g. `testing-library/react-testing-library`) manage getting the version they need.
2. The new behavior around batched-state updates seems to conflict with our approach for debounced state updates, for example in `LogTab.jsx`. Perhaps we will no longer need to manage that debouncing ourselves. Or perhaps we still should, and also manage state updates -- avoiding the default batching -- with something like this:
```javascript
  updateState() {
    ReactDom.flushSync(() => { // force the state update to happen now
      this.setState((state) => ({
        logdata: state.logdata.concat(this.cache)
      }));
    });
    this.cache = [];
  }
```

Issue #862, opened 2021-07-16
ExtensionLoadWarnings on electron startup


davemfish:
```
(node:3046) ExtensionLoadWarning: Warnings loading extension at /Users/emily/Library/Application Support/invest-workbench/extensions/fmkadmapgofadopljbjfkapdkoienihi:
  Unrecognized manifest key 'browser_action'.
  Unrecognized manifest key 'minimum_chrome_version'.
  Unrecognized manifest key 'update_url'.
  Cannot load extension with file or directory name _metadata. Filenames starting with "_" are reserved for use by the system.

(Use `Electron --trace-warnings ...` to show where the warning was created)
```
These warnings are related to loading the react-devtools extension, which was a recent change. There's been a lot of discussion of them and my conclusion was that it's simply not worth the effort to try to silence them. Or rather, the bug is upstream and silencing them would require each of us to jump through some manual hoops in our local filesystem, which I never bothered trying. I should have warned you all though!

https://github.com/MarshallOfSound/electron-devtools-installer/issues/143
https://stackoverflow.com/questions/23918539/extension-installed-from-chrome-webstore-throws-error-package-is-invalid-detai

davemfish:
These warnings have gotten more verbose, but it still sounds like there is no reasonable fix. See https://github.com/electron/electron/issues/36545

davemfish:
Not only do the warnings exist, but the devtools don't work either. I'll remove the extension and we can add it back if the upstream electron issue is resolved.

PR #1331, opened 2023-06-15
Workbench: update to electron v25, jest v29

### LINKED ISSUE #1223: Do major node package upgrades for Workbench
Lots are outdated: `yarn outdated`

It would be nice to upgrade, but lots of breaking changes are possible, so I think we should do this after the `3.13.0` release.

davemfish:
The main thing here is upgrading to electron 25. Most of the changes were necessitated by breaking changes in v20, related to even more restrictions on using `node` packages in the preload script. Specifically, we no longer import `electron-log` in preload or renderer. Instead we do all logging from the renderer via IPC to main. 

At the same time, something broke in the `babel` transformations we were using for the preload module. I decided to move away from `babel` and use `vite` instead, especially since the preload script is now even more targeted to the browser than in the past. We could consider doing the same change for `main` module code, but so far that still just works. 

Some relatively minor changes related to the jest upgrade. `mockRestore` behavior somehow changed. And they _finally_ exposed an option to print out all the failing tests at the end of the summary.

I also scoped updating React to v18, but decided to that separately (see #1325 ). 

Fixes #1223 

## Checklist


workbench/src/main/main.js
```diff
@@ -90,7 +90,7 @@ export const createWindow = async () => {
   setupRendererLogger();
   await getFlaskIsReady();
 
-  const devModeArg = ELECTRON_DEV_MODE ? '--devMode' : '';
+  const devModeArg = ELECTRON_DEV_MODE ? '--devmode' : '';
```
davemfish, line 93: this was silently getting lowercased anyway, which was messing with our args parsing in `preload/api.js`

workbench/src/preload/api.js
```diff
@@ -18,17 +18,17 @@ const ipcRendererChannels = [
...
 // As well as being loaded by the electron preload script,
 // this module is loaded by the jest setup script in order to
 // mock APIs that are normally setup during electron preload.
 // In the jest case, we do not have userPaths data.
 const userguidePath = (!isDevMode && userPaths)
-  ? `file:///${userPaths.resourcesPath}/documentation`
+  ? `file:///${userPaths.resourcesPath}/documentation` // as a URL
   : ''; // In DevMode, local UG is served at the root path
 const electronLogPath = (userPaths)
-  ? `${userPaths.userData}/logs/main.log`
```
davemfish: These forward slashes made for an invalid path on Windows when mixed with backslashes in `userPaths.userData`. So I moved this whole construction over to the main process where we can use `path.join`.

The userguide path, which is actually read by a browser as a URL, works fine this way. And that URL is tested by puppeteer, so definitely works in dev & production on all OS.

davemfish:
@emlys I made those few changes we discussed, and also created #1333.

FYI, if you run these changes you might notice there are even more warnings on startup emanating from the react-devtools-extension. So I re-opened #862 .

emlys (approved):
looks good!

Issue #1328, opened 2023-06-14
Proposal: use info from `MODEL_SPEC` to read CSVs and add `index_col` key to CSV specs


emlys:
In light of #1278 and #1319, we might want to start specifying dtypes for table columns when we read CSVs with `build_lookup_from_csv` or `read_csv_to_dataframe`. Column dtype information already exists in in the `MODEL_SPEC`, so it would just be a matter of reformatting that information and passing it to the utils function.

`build_lookup_from_csv` required another piece of information, which is the `key_field`, aka the index column (often the LULC column). We don't currently document which column is the index in a structured way, but I think it would be a logical addition to the `MODEL_SPEC`. Then we could pass that to `build_lookup_from_csv`.

In summary, the proposed changes are:
- Add a new optional key to CSV specs called `index_col`. The value of `spec['index_col']` will be a string, which must equal one of the column names found in `spec['columns']`.
    - The name `index_col` is chosen to be consistent with `pandas.read_csv`
    - It will be optional because of the few complicated tables that don't have a single index column, but we should use it wherever possible.
- Modify `utils.build_lookup_from_csv` and `utils.read_csv_to_dataframe` to use information from a CSV spec:
    - build the lookup dictionary using `spec['index_col']` as the index
    - select only the columns from the table that are in `spec['columns']` (or match a pattern in `spec['columns']`)
    - apply a converter function to each column (see the `converters` argument to `pandas.read_csv`) depending on `spec[column]['type']`, for instance:
        - expand relative paths, for path types (`raster, vector, csv, directory, file`)
        - parse boolean values from the table into python `bool`s
        - for string types, strip whitespace and convert to lowercase
        - make integer values `int`s and float values `float`s

emlys:
See also #262.

PR #1334, opened 2023-06-15
Remove `build_lookup_from_csv` and consolidate into `read_csv_to_dataframe`

### LINKED ISSUE #1327: Proposal: Improve division of functionality between `build_lookup_from_csv`, `read_csv_to_dataframe`, `pandas.read_csv`
`read_csv_to_dataframe` is a wrapper around `pandas.read_csv` that adds this functionality:
- Strip whitespace from column names
- Convert column names to lowercase
- Strip whitespace from table values
- Expand relative paths

`build_lookup_from_csv` reformats the table as a dictionary indexed by a particular column, and also adds this functionality:
- Convert table values to lowercase
- Only use certain columns
- Drop empty rows
- Replace NA values with an empty string

It would make more sense if the additional functionality of `build_lookup_from_csv` was moved to `read_csv_to_dataframe`. Particularly the lowercase conversion options are confusing: the `to_lower` arg of `build_lookup_from_csv` affects column names AND values. the `to_lower` arg of `read_csv_to_dataframe` affects column names only. 

Pseudocode outlining the proposed division of functionality:
```
def build_lookup_from_csv(path, index_col, **kwargs):
    df = read_csv_to_dataframe(path, **kwargs)
    return df reformatted as a dictionary indexed by index_col

def read_csv_to_dataframe(path, cols_to_lower=True, vals_to_lower=True, 
        expand_path_cols=[], column_list=None, sep=None, engine='python', encoding='utf-8-sig', **kwargs):
    df = pandas.read_csv(path, **kwargs)
    use only the columns in column_list
    drop empty rows
    strip whitespace from column names and values
    expand relative paths in expand_path_cols
    replace NA values with an empty string
    if cols_to_lower:
        convert column names to lowercase
    if vals_to_lower:
        convert values to lowercase
    return df
```
The data processing that can be done with arguments to `pandas.read_csv` (such as selecting columns, dropping empty rows, and applying converters to column values) should be moved there to avoid duplication.

emlys:
## Description
Fixes #1327 
Consolidated functionality into `read_csv_to_dataframe` and replaced uses of `build_lookup_from_csv(...)` with `read_csv_to_dataframe(...).to_dict(orient='index')`.

Split the `to_lower` functionality into two separate args, `cols_to_lower` and `vals_to_lower`. Some of this might be obsoleted by #1328, but I wanted to take it one step at a time.

## Checklist

### DIFF
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -598,8 +598,8 @@ def expand_path(path, base_path):
 
 
 def read_csv_to_dataframe(
-        path, index_col=None, usecols=None, cols_to_lower=True,
-        vals_to_lower=True, expand_path_cols=[], sep=None, engine='python',
+        path, index_col=False, usecols=None, cols_to_lower=True,
+        vals_to_lower=True, expand_path_cols=None, sep=None, engine='python',
         encoding='utf-8-sig', **kwargs):
     """Return a dataframe representation of the CSV.
 
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -646,7 +646,7 @@ def read_csv_to_dataframe(
             path, index_col=False, sep=sep, engine=engine, encoding=encoding, **kwargs)
     except UnicodeDecodeError as error:
         LOGGER.error(
-            f'{path} must be encoded as UTF-8 or ASCII')
+            f'The file {path} must be encoded as UTF-8 or ASCII')
         raise error
 
     # strip whitespace from column names
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -674,7 +674,8 @@ def read_csv_to_dataframe(
     # Set 'index_col' as the index of the dataframe
     if index_col:
         try:
-            dataframe = dataframe.set_index(index_col, drop=False)
+            dataframe = dataframe.set_index(
+                index_col, drop=False, verify_integrity=True)
         except KeyError:
             # If 'index_col' is not a column then KeyError is raised for using
             # it as the index column
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -713,7 +713,7 @@ def read_csv_to_dataframe(
         # the columns are stripped of leading/trailing whitespace in
         # ``read_csv_to_dataframe``, and also lowercased if ``to_lower`` so we only
         # need to convert the rest of the table.
-        if index_col:
+        if index_col and isinstance(index_col, str):
             index_col = index_col.lower()
         # lowercase column names
         if usecols:
--- a/src/natcap/invest/utils.py
+++ b/src/natcap/invest/utils.py
@@ -618,59 +618,10 @@ def build_lookup_from_csv(
         KeyError
             If ``key_field`` is not present during ``set_index`` call.
     """
-    # Reassign to avoid mutation
-    col_list = column_list
-    # if a list of columns are provided to use and return, make sure
-    # 'key_field' is one of them.
-    if col_list and key_field not in col_list:
-        col_list.append(key_field)
 
-    table = read_csv_to_dataframe(
-        table_path, to_lower=to_lower, index_col=False, **kwargs)
+    table = read_csv_to_dataframe(table_path, cols_to_lower=to_lower,
+        vals_to_lower=to_lower, index_col=key_field, **kwargs)
 
-    # if 'to_lower`, case handling is done before trying to access the data.
-    # the columns are stripped of leading/trailing whitespace in
-    # ``read_csv_to_dataframe``, and also lowercased if ``to_lower`` so we only
-    # need to convert the rest of the table.
-    if to_lower:
-        key_field = key_field.lower()
-        # lowercase column names
-        if col_list:
-            col_list = [col.lower() for col in col_list]
-        # lowercase values
-        table = table.applymap(
-            lambda x: x.lower() if isinstance(x, str) else x)
-
-    # Set 'key_field' as the index of the dataframe
-    try:
-        table.set_index(key_field, drop=False, inplace=True)
-    except KeyError:
-        # If 'key_field' is not a column then KeyError is raised for using
-        # it as the index column
-        LOGGER.error(f"'key_field' : '{key_field}' could not be found as a"
-                     f" column in the table. Table path: {table_path}.")
-        raise
-
-    # Subset dataframe by columns if desired
-    if col_list:
-        table = table.loc[:, col_list]
-
-    # look for NaN values and warn if any are found.
-    table_na = table.isna()
-    if table_na.values.any():
-        LOGGER.warning(
-            f"Empty or NaN values were found in the table: {table_path}.")
-    # look to see if an entire row is NA values
-    table_na_rows = table_na.all(axis=1)
-    na_rows = table_na_rows.index[table_na_rows].tolist()
-    # if a completely empty row, drop it
-    if na_rows:
-        LOGGER.warning(
-            "Encountered an entirely blank row on line(s)"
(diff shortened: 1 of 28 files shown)


tests/test_coastal_blue_carbon.py
```diff
@@ -358,7 +358,6 @@ def test_read_transition_matrix(self):
             transition_csv.write('b,,NCC,accum\n')
             transition_csv.write('c,accum,,NCC\n')
             transition_csv.write(',,,\n')
-            transition_csv.write(',legend,,')  # simulate legend
```
emlys, line 361: Here's one minor functional change. This was testing that extra stuff in the table would be ignored. I don't think we allow that in other tables in invest, so it seemed easier to drop it from the test than figure out how to accomplish that with `read_csv_to_dataframe`. But let me know if there's a reason to keep it!

phargogh: Yep, you're right that this table legend is unique among the InVEST models!  It's also weird in that it breaks from the normal table structure only to provide some extra metadata that's also described in the UG chapter.  Personally I'm OK with removing it, but because this is a notable functional change we should also be sure to 

- [ ] remove the legend-generating code from the preprocessor
- [ ] Note the functional change in the model's HISTORY
- [ ] remove the legend from the `transitions_template.csv` in the sample data repo

What do you think?

tests/test_utils.py
```diff
@@ -610,25 +610,25 @@ def tearDown(self):
...
-                '_': 'apple'
-            },
-        }
-        self.assertDictEqual(result, expected_dict)
+
+        csv_file = os.path.join(self.workspace_dir, 'csv.csv')
+
+        with open(csv_file, 'w') as file_obj:
+            file_obj.write(textwrap.dedent(
+                """
```
phargogh, line 621: I see you have a `.strip()` there at the end (which is nice!) so this isn't a request to change anything, just an FYI!  I recently learned that you can also do this to avoid the leading newline:

```
"""\
HEADER,
A,
b
"""

File: src/natcap/invest/utils.py [hunk 3]
```diff
@@ -714,7 +613,15 @@ def read_csv_to_dataframe(
...
     Args:
         path (str): path to a CSV file
-        to_lower (bool): if True, convert all column names to lowercase
+        index_col (str): name of column to use as the dataframe index. If
+            ``cols_to_lower``, this column name and the dataframe column names
+            will be lowercased before they are compared. If ``usecols``
+            is defined, this must be included in ``usecols``.
+        usecols (list(str)): list of column names to subset from the dataframe.
+            If ``cols_to_lower``, these names and the dataframe column names
+            will be lowercased before they are compared.
```
phargogh, line 622: Since `index_col` is required to be in `usecols`, would it make sense to simply add it into `usecols` if it isn't already there?  It seems like this might be a minor bit of de-duplication when writing out a function call, but I'm curious what you think!

phargogh (changes requested):
Thanks @emlys !  I left a few questions, comments and suggestions.  And thanks for having this feature implementation as a nice, atomic PR in the bigger picture :)

src/natcap/invest/utils.py
emlys: That certainly could make sense! I was going for consistency with `pandas.read_csv`, which will raise an error if you don't include the value of `index_col` in your `usecols` list. I think it also parallels nicely with the structure of the `MODEL_SPEC` - when we implement #1328, we might pipe in the spec `columns`, which already include the index column.

tests/test_utils.py
emlys: That is much nicer, I made this change everywhere in `test_utils.py`

src/natcap/invest/utils.py
```diff
@@ -732,23 +639,56 @@ def read_csv_to_dataframe(
...
     try:
+        # set index_col=False to force pandas not to index by any column
+        # this is useful in case of trailing separators
+        # we'll explicitly set the index column later on
         dataframe = pandas.read_csv(
-            path, sep=sep, engine=engine, encoding=encoding, **kwargs)
+            path, index_col=False, sep=sep, engine=engine, encoding=encoding, **kwargs)
     except UnicodeDecodeError as error:
         LOGGER.error(
             f'{path} must be encoded as UTF-8 or ASCII')
```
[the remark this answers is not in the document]
emlys: I believe this is about the contents of the table. I updated this to say "The file {path} must be..."

src/natcap/invest/utils.py
```diff
@@ -732,23 +639,56 @@ def read_csv_to_dataframe(
...
+            LOGGER.error(f"The column '{index_col}' could not be found "
+                         f"in the table {path}")
+            raise
+
+    # convert table values to lowercase
+    if vals_to_lower:
+        dataframe = dataframe.applymap(
+            lambda x: x.lower() if isinstance(x, str) else x)
+
+    # expand paths
```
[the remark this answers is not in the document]
emlys: I think there definitely are cases where a table contains a path column that we _don't_ want to lowercase, but other columns that we _do_ want to lowercase. The `read_csv_to_dataframe` approach has not given us the option to distinguish between columns, so we've had to set `to_lower=False` on any table that contains paths. 

When #1328 is implemented, that will give us the chance to apply different processing to each column. For now I don't think this is an issue, because `vals_to_lower` should always be `False` if `expand_paths` are defined.

src/natcap/invest/utils.py
```diff
@@ -714,7 +613,15 @@ def read_csv_to_dataframe(
...
-        to_lower (bool): if True, convert all column names to lowercase
+        index_col (str): name of column to use as the dataframe index. If
+            ``cols_to_lower``, this column name and the dataframe column names
+            will be lowercased before they are compared. If ``usecols``
+            is defined, this must be included in ``usecols``.
+        usecols (list(str)): list of column names to subset from the dataframe.
+            If ``cols_to_lower``, these names and the dataframe column names
+            will be lowercased before they are compared.
+        cols_to_lower (bool): if True, convert all column names to lowercase
+        vals_to_lower (bool): if True, convert all table values to lowercase
```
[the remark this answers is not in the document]
emlys: Renamed these to `convert_cols_to_lower` and `convert_vals_to_lower`, hopefully that's clearer.

tests/test_coastal_blue_carbon.py
emlys: Ah, I didn't realize that the preprocessor and sample data relied on this. In the interest of self-contained PRs, I'd rather not have to make all those changes here.

How about this for now? I made a small change in `coastal_blue_carbon.py` so that it will skip over rows that have nothing in the `lulc-class` column (which should be all the legend rows). That way the functionality is the same, but `read_csv_to_dataframe` doesn't have to handle it. And we can remove the legend in a separate issue if it's worth it?

phargogh: Sounds like a fine solution to me!  I took the liberty of describing this in https://github.com/natcap/invest/issues/1336

src/natcap/invest/utils.py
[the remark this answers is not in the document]
phargogh: Yes, much clearer.  Thanks!

src/natcap/invest/utils.py
phargogh: That all makes sense to me!

phargogh (approved):
Thanks @emlys !  I don't have any further comments or changes to request, so I'm adding my approval.

One thing I am seeing though is that the mac binary build is failing when it was passing at the beginning of the PR.  Could you just take a peek at that and make sure it isn't failing because of something in the PR here?  If it isn't, please feel free to merge away.  Thanks!

emlys:
Thanks @phargogh! I'm seeing the same error in the latest build on `main` (https://github.com/natcap/invest/actions/runs/5317455763/jobs/9628004338), so I think it's unrelated.

PR #1338, opened 2023-06-23
Workbench: upgrade react & testing-library

### LINKED ISSUE #1325: Workbench: upgrade to React18

davemfish:
We're upgrading to React18, which was pretty straightforward. More troublesome was upgrading `@testing-library/userEvent` and `@testing-library/react`.

Fixes #1325 

## Checklist


workbench/src/main/main.js
```diff
@@ -177,15 +179,6 @@ export function main() {
...
   app.on('ready', async () => {
-    if (ELECTRON_DEV_MODE) {
-      const {
-        default: installExtension,
-        REACT_DEVELOPER_TOOLS,
-      } = require('electron-devtools-installer');
-      await installExtension(REACT_DEVELOPER_TOOLS, {
-        loadExtensionOptions: { allowFileAccess: true },
-      });
-    }
```
davemfish, line 188: See #862 , react devtools don't work at the moment.

workbench/tests/binary_tests/puppet.test.js
```diff
@@ -10,8 +10,6 @@ import { spawn, spawnSync } from 'child_process';
 
 import rimraf from 'rimraf';
 import puppeteer from 'puppeteer-core';
-import { getDocument, queries, waitFor } from 'pptr-testing-library';
```
davemfish, line 13: We had to stop using `pptr-testing-library` because it seems to be no longer maintained. And the latest release depends on a different version of `testing-library/dom` than does `testing-library/react`. And having two different versions of that dependency caused problems.

workbench/tests/renderer/app.test.js
```diff
@@ -537,345 +526,8 @@ describe('InVEST global settings: dialog interactions', () => {
   });
 });
 
-describe('InVEST subprocess testing', () => {
```
davemfish, line 540: I moved this block to its own file. This file was getting a bit lengthy.

workbench/tests/renderer/app.test.js
```diff
@@ -887,14 +539,21 @@ describe('Translation', () => {
     });
   });
 
+  afterAll(() => {
+    Object.defineProperty(global.window, 'location', {
+      configurable: true,
+      value: location,
+    });
+  });
```
davemfish, line 547: We were missing this manual "unmocking" of `window.location

workbench/tests/renderer/invest_subprocess.test.js
```diff
@@ -0,0 +1,405 @@
...
+    await userEvent.click(carbon);
+    const workspaceInput = await findByLabelText(
+      `${spec.args.workspace_dir.name}`
+    );
+    await userEvent.type(workspaceInput, fakeWorkspace);
+    const execute = await findByRole('button', { name: /Run/ });
+    await userEvent.click(execute);
+    await waitFor(() => expect(execute).toBeDisabled());
+
+    act(() => {
```
davemfish, line 155: rare instance of needing to use `act` because the actions we're wrapping here are not managed by `react-testing-library`

workbench/tests/renderer/invest_subprocess.test.js
```diff
@@ -0,0 +1,405 @@
+import fs from 'fs';
```
davemfish, line 1: Not much is new here, everything was just moved from `app.test.js`. Except the updates to `await userEvent` and to use `act` in a few places.

workbench/tests/renderer/investtab.test.js
```diff
@@ -266,39 +266,6 @@ describe('Sidebar Buttons', () => {
     expect(archiveDatastack).toHaveBeenCalledTimes(1);
   });
 
-  test('Multiple Save Clicks: each triggers a unique alert', async () => {
```
davemfish, line 269: This test seemed extraneous. I think the original intent was to test behavior of someone rapidly clicking all the save buttons. But that's much harder to do now that they are nested inside a `Save as...` dialog. We have assertions for the alerts in other tests.

workbench/src/renderer/components/LogTab/index.jsx
```diff
@@ -85,9 +86,15 @@ export default class LogTab extends React.Component {
...
-    }));
+    // flushSync will override react18 batched updates
+    // and force state updates to happen now. We're managing
+    // the rate of updates ourselves in this component via
+    // debouncedLogUpdate.
+    ReactDom.flushSync(() => {
+      this.setState((state) => ({
+        logdata: state.logdata.concat(this.cache)
+      }));
+    });
```
davemfish, line 97: I tested removing the `cache` and `debounced` updates, to just let react manage state updates entirely on its own. That seemed to work fine now, as of React18. But still seemed like there were excessive re-renders during high volume stdout coming from invest.

So I think our approach here is likely safer on machines with lower performance. There's no reason to update the logfile more rapidly than once every 10ms.

dcdenu4 (approved):
Thanks @davemfish for updating our packages and handling all the broken workflows it caused. That video you linked was playing in the back of my mind as I looked over this!

Issue #1339, opened 2023-06-23
Binary build failures on GHA macos-latest


davemfish:
It seems like this must have been triggered by an OS update? There was an macOS upgrade for the runners a few days ago. https://github.com/actions/runner-images/releases

This doesn't seem like our bug, but I haven't found mention of it anywhere else yet. The exception is from,
```
2023-06-23T18:07:31.5238040Z subprocess.CalledProcessError: Command '['/Users/runner/work/invest/invest/dist/invest/invest', '-sS', '-c', 'import platform; print(platform.mac_ver()[0])']' returned non-zero exit status 2.
```

https://github.com/natcap/invest/actions/runs/5358675028/jobs/9722366040
```
2023-06-23T18:06:36.8749610Z dist/invest/invest list
2023-06-23T18:07:30.8626000Z usage: invest [-h] [--version] [-v | --debug]
2023-06-23T18:07:30.8730470Z               [--taskgraph-log-level {DEBUG,INFO,WARNING,ERROR}]
2023-06-23T18:07:30.8769960Z               [-L {en,es,zh}]
2023-06-23T18:07:30.8809550Z               {list,run,validate,getspec,serve,export-py} ...
2023-06-23T18:07:30.8869260Z invest: error: argument subcommand: invalid choice: 'import platform; print(platform.mac_ver()[0])' (choose from 'list', 'run', 'validate', 'getspec', 'serve', 'export-py')
2023-06-23T18:07:31.4447430Z Traceback (most recent call last):
2023-06-23T18:07:31.4565640Z   File "invest/cli.py", line 16, in <module>
2023-06-23T18:07:31.4667460Z   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
2023-06-23T18:07:31.4754900Z   File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
2023-06-23T18:07:31.4856940Z   File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
2023-06-23T18:07:31.4958690Z   File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
2023-06-23T18:07:31.4970270Z   File "natcap/invest/datastack.py", line 37, in <module>
2023-06-23T18:07:31.5071900Z   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
2023-06-23T18:07:31.5173800Z   File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
[...]

PR #1341, opened 2023-06-26
REC bugfix - overlapping predictor polygons inflate area coverage metrics

### LINKED ISSUE #1310: REC: polygon area metrics overcount when polygons overlap
For predictor methods `polygon_area_coverage` and `polygon_percent_coverage`, the predictor polygons are intersected with the AOI polygons and the area of the predictor intersection is summed. If the predictor layer includes overlapping polygons, then the area of the overlap is counted once for each polygon. This results in incorrect computed areas, which only really become noticeable when the value is greater than the area of the AOI cell (or percentages over 100).

I can't think of any reason why overlapping polygons should be double-counted like this. I think both of these metrics imply that it is the _union_ of the predictor polygons that matters in terms of assessing its spatial coverage. So we should probably do that union before computing areas.

davemfish:
`unary_union` polygons instead of calculating the cumulative area of the individual polygons.

Fixes #1310

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -96,6 +96,10 @@ Unreleased Changes
       set to 0. The old behavior was not well documented and caused some
       confusion when nodata pixels did not line up. It's safer not to fill in
       unknown data. (`#1317 <https://github.com/natcap/invest/issues/1317>`_)
+* Visitation: Recreation and Tourism
+    * Fixed a bug where overlapping predictor polygons would be double-counted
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Looks good to me!  Thanks for the fix and the helpful comments in the test as well.

For future us, the mac binary failure is related to https://github.com/natcap/invest/issues/1339 and not this PR.

PR #1342, opened 2023-06-27
UFRM bugfix - all 0s in biophysical table are okay

### LINKED ISSUE #1123: UFRM: incorrect ValueError raised if biophysical table includes a row of all 0s
This came up on this [forum post](https://community.naturalcapitalproject.org/t/valueerror-the-biophysical-table-is-missing-a-row-for-lucode-s-255/3229/6).

As is somewhat common, a LULC raster without a nodata value was used. But the raster was filled with 0s that probably represent nodata (or maybe water), and so the biophysical table was given a row for `lucode 0`, filled with 0s for `CN_A, CN_B, CN_C, CN_D`.

I don't know the details of this model, and whether setting those parameters to 0 is equivalent to setting that pixel value to nodata, but I don't think it matters.

### Here's the problem
This model does fancy things using a sparse matrix as a lookup table for these CN values. Such a matrix can appear to have a row full of 0s for two different reasons:
1. Because the biophysical table had a row full of 0s (this should not raise an error, but does)
2. Because a landuse code was missing from the biophysical table (this should raise an error, and does)

davemfish:
Added a test and fixed the logic by which we check for non-existent values in a sparse matrix, which are different from explicit 0s .

Fixes #1123 

## Checklist

### DIFF
--- a/src/natcap/invest/urban_flood_risk_mitigation.py
+++ b/src/natcap/invest/urban_flood_risk_mitigation.py
@@ -871,14 +871,14 @@ def _lu_to_cn_op(
     # this is an array where each column represents a valid landcover
     # pixel and the rows are the curve number index for the landcover
     # type under that pixel (0..3 are CN_A..CN_D and 4 is "unknown")
-    valid_lucodes = lucode_array[valid_mask].astype(int)
+    valid_lucode_array = lucode_array[valid_mask].astype(int)
 
     try:
-        cn_matrix = lucode_to_cn_table[valid_lucodes]
+        cn_matrix = lucode_to_cn_table[valid_lucode_array]
     except IndexError:
         # Find the code that raised the IndexError, and possibly
         # any others that also would have.
-        lucodes = numpy.unique(valid_lucodes)
+        lucodes = numpy.unique(valid_lucode_array)
         missing_codes = lucodes[lucodes >= lucode_to_cn_table.shape[0]]
         raise ValueError(
             f'The biophysical table is missing a row for lucode(s) '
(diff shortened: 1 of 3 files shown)


src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -871,25 +871,26 @@ def _lu_to_cn_op(
...
 
     # Even without an IndexError, still must guard against
     # lucodes that can index into the sparse matrix but were
-    # missing from the biophysical table. They have rows of all 0.
-    if not cn_matrix.sum(1).all():
-        empty_rows = numpy.where(lucode_to_cn_table.sum(1) == 0)
-        missing_codes = numpy.intersect1d(valid_lucodes, empty_rows)
+    # missing from the biophysical table. Do this by intersecting
+    # rows with no stored values with the lulc array
+    empty_rows = cn_matrix.getnnz(1) == 0
```
dcdenu4, line 891: Nice

tests/test_ufrm.py
```diff
@@ -198,6 +198,26 @@ def test_ufrm_value_error_on_bad_lucode(self):
...
+
+        # a user may define a row with all 0s
+        cn_table.iloc[0] = [0] * cn_table.shape[1]
+        cn_table.to_csv(good_cn_table_path, index=False)
+        args['curve_number_table_path'] = good_cn_table_path
+
+        try:
+            urban_flood_risk_mitigation.execute(args)
+        except ValueError:
+            self.fail('unexpected ValueError')
```
dcdenu4, line 219: Should this fail message be more informative?
```suggestion
            self.fail('unexpected ValueError when testing curve number row with all zeros')

```

dcdenu4 (changes requested):
Thanks @davemfish ! Just one minor suggestion.

dcdenu4 (approved):
Thanks!

PR #1355, opened 2023-07-28
Code sign the electron app

### LINKED ISSUE #727: Sign the entire macos app bundle
As of #697, we're correctly signing the mac binary DMG with the correct type of Apple-issued certificate 🎉 This is important for security because it ensures that the entire DMG (not just the app bundle in the DMG) is what we expect it to be. So security-wise I think we're fine. It also (I'm pretty sure) prevents any Gatekeeper warnings when the user opens the DMG.

The next step in our Apple security journey will be to sign the app bundle itself. This is one of the things needed in order to prevent Gatekeeper warnings when opening the app. The command to do so is
```
codesign --deep --sign <identifier> <path/to/bundle>`
```
This recursively signs everything in the bundle from the most-deeply-nested up to the top. It complains about the following directories:
```
Contents/MacOS/invest_dist/documentation/userguide/.doctrees
Contents/MacOS/invest_dist/cryptography-35.0.0.dist-info
Contents/MacOS/invest_dist/pygeoprocessing-2.3.2.dist-info
Contents/MacOS/invest_dist/wheel-0.37.0-py3.9.egg-info
Contents/MacOS/invest_dist/natcap.invest-3.9.1.post72+g6af0b498.dist-info
Contents/MacOS/invest_dist/setuptools-59.1.1-py3.8.egg-info
Contents/MacOS/invest_dist/charset_normalizer-2.0.0.dist-info
Contents/MacOS/invest_dist/taskgraph-0.11.0.dist-info
```
If you delete all these, it works! This is a good sign that our app bundle structure is not severely wrong. So we just need to remove or reorganize these folders somehow.

emlys:
## Description
Fixes #727 

It seems like electron-builder has improved their code signing automation since I last tried to use it. The only change I had to make was moving the license file into the Resources directory of the app bundle. Note that the code signing steps don't run in PRs, so check out the actions run on my branch instead.

Mac users will now see a slightly-less-scary warning when they open the app for the first time. I'm not sure if it has any effect on Windows.

Before:
image
After:
image

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -356,6 +356,7 @@ jobs:
         uses: google-github-actions/setup-gcloud@v0
 
       - name: Build Workbench (macOS)
+        if: github.event_name != 'pull_request' && matrix.os == 'macos-latest' # secrets not available in PR
         working-directory: workbench
         env:
           GH_TOKEN: env.GITHUB_TOKEN
@@ -369,6 +370,7 @@ jobs:
           yarn run dist
 
       - name: Build Workbench (Windows)
+        if: github.event_name != 'pull_request' && matrix.os == 'windows-latest' # secrets not available in PR
         working-directory: workbench
         env:
           GH_TOKEN: env.GITHUB_TOKEN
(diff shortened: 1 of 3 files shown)


davemfish:
Thanks, Apple! For no longer encouraging our users to trash our software 🙄

dcdenu4 (approved):
Thanks @emlys!

PR #1364, opened 2023-08-07
Workbench: cleanup the process shutdown routines

### LINKED ISSUE #1356: Workbench: fragility in the shutdown routine
When the Workbench exits, it kills the python subprocess running `ui_server.py`, and then it polls an endpoint until it gets no response, indicating the server did shut down. Only after that is the electron process allowed to exit. (I have a note about not allowing electron to exit too quickly because the signal for killing the python process never goes through.)

This system is fragile because it's possible for a `ui_server` to be running that is not associated with the currently exiting Workbench process. So the Workbench won't be able to kill that server. In that case the Workbench will continuously poll the server waiting for an `ECONNREFUSED` that it never receives.

This is most likely to come up in development; we haven't heard about it in production. But there is probably a better way to manage the shutdown routine.

### LINKED ISSUE #1362: Workbench: MacOS pptr test does not cleanly exit electron after each test
I noticed this test failing on the last couple GHA runs. And I think I was able to reproduce it locally on Mac. I observed that after the first test, the app remained open whereas it should fully shutdown during the `afterEach` and then restart in `beforeEach`.

davemfish:
There was a race condition in the shutdown routine of the workbench. It would mainly come up during the Windows puppeteer test, where the python subprocess would not be given a chance to exit before electron & jest exited. Using `execSync` instead of `exec` to call `taskkill` seems reliable. 

Fixed #1356 by removing the potential for an infinite loop while shutting down the python process. Polling the server to see if it was still live was actually papering over the real problem of the race condition mentioned above.

Fixes #1362 by making sure to close instances of pptr `pages` before trying to close the `browser`. 

Unrelated to that, tests were sometimes unreliable because the script was trying to click on buttons before their click handlers were ready. I added some pauses in the execution to accommodate that and things appear reliable now.

## Checklist

### DIFF
--- a/workbench/tests/binary_tests/puppet.test.js
+++ b/workbench/tests/binary_tests/puppet.test.js
@@ -167,6 +167,7 @@ test('Run a real invest model', async () => {
   const downloadModal = await page.waitForSelector('.modal-dialog');
   const downloadModalCancel = await downloadModal.waitForSelector(
     'aria/[name="Cancel"][role="button"]');
+  await page.waitForTimeout(300); // waiting for click handler to be ready
   await downloadModalCancel.click();
   // We need to get the modelButton from w/in this list-group because there
   // are buttons with the same name in the Recent Jobs container.
@@ -235,11 +236,13 @@ test('Check local userguide links', async () => {
   const downloadModal = await page.waitForSelector('.modal-dialog');
   const downloadModalCancel = await downloadModal.waitForSelector(
     'aria/[name="Cancel"][role="button"]');
+  await page.waitForTimeout(500); // waiting for click handler to be ready
   await downloadModalCancel.click();
 
   const investList = await page.waitForSelector('.invest-list-group');
   const modelButtons = await investList.$$('aria/[role="button"]');
 
+  await page.waitForTimeout(300); // first btn click does not register w/o this pause
   for (const btn of modelButtons) {
     await btn.click();
     const link = await page.waitForSelector('text/User\'s Guide');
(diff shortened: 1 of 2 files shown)


workbench/tests/binary_tests/puppet.test.js
```diff
@@ -133,13 +133,15 @@ beforeEach(() => {
 
 afterEach(async () => {
   try {
+    const pages = await BROWSER.pages();
+    await Promise.all(pages.map(page => page.close()));
```
dcdenu4, line 137: Nice.

workbench/tests/binary_tests/puppet.test.js
```diff
@@ -232,11 +235,13 @@ test('Check local userguide links', async () => {
   const downloadModal = await page.waitForSelector('.modal-dialog');
   const downloadModalCancel = await downloadModal.waitForSelector(
     'aria/[name="Cancel"][role="button"]');
+  await page.waitForTimeout(500); // waiting for click handler to be ready
```
dcdenu4, line 238: Any reason this one is longer than the others? I don't think it's necessary if different click events could need different wait times but would it be convenient to have a constant variable defining these wait values?

dcdenu4 (commented):
Thanks as always @davemfish for iterating and making the Workbench more stable. Just had a minor comment.

workbench/tests/binary_tests/puppet.test.js
davemfish: I don't know why it's different so I've made it the same, and I added a variable.

But it's likely each button is different, in terms of when it becomes functional. It seems dependent on whether the javascript containing the click handler's function has loaded yet, or whether the element is visible yet, or otherwise visually blocked somehow. All we know for sure is that the element is present in the DOM (that's what the `waitForSelector` accomplishes). But these other things remain unknown.

dcdenu4 (approved):

PR #1370, opened 2023-08-11
Workbench: validate metadata that is used to open invest models

### LINKED ISSUE #1275: Make sure two different workbench versions can run on the same system

### LINKED ISSUE #1286: Workbench - handle unexpected data in localStorage

davemfish:
This PR adds some validation before rendering data from the recent jobs store, to avoid rendering a card for a model that no longer exists in the current version of invest, and to skip over jobs that are somehow missing critical pieces of data.

It also adds error handling for trying to load a datastack from a deprecated model.

Fixes #1275 and fixes #1286 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/SetupTab/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/index.jsx
@@ -26,6 +26,7 @@ import { argsDictFromObject } from '../../utils';
 import { ipcMainChannels } from '../../../main/ipcMainChannels';
 
 const { ipcRenderer } = window.Workbench.electron;
+const { logger } = window.Workbench;
 
 /** Initialize values of InVEST args based on the model's UI Spec.
  *
@@ -274,8 +275,20 @@ class SetupTab extends React.Component {
   }
 
   async loadParametersFromFile(filepath) {
-    const datastack = await fetchDatastackFromFile(filepath);
     const { pyModuleName, switchTabs, t } = this.props;
+    let datastack;
+    try {
+      datastack = await fetchDatastackFromFile(filepath);
+    } catch (error) {
+      logger.error(error);
+      alert( // eslint-disable-line no-alert
+        t(
+          'No InVEST model data can be parsed from the file:\n {{filepath}}',
+          { filepath: filepath }
+        )
+      );
+      return;
+    }
     if (datastack.module_name === pyModuleName) {
       this.batchUpdateArgs(datastack.args);
       switchTabs('setup');
--- a/workbench/src/renderer/components/SetupTab/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/index.jsx
@@ -55,7 +55,7 @@ function initializeArgValues(argsSpec, uiSpec, argsDict) {
     if (argsSpec[argkey].type === 'boolean') {
       value = argsDict[argkey] || false;
     } else if (argsSpec[argkey].type === 'option_string') {
-      if  (argsDict[argkey]) {
+      if (argsDict[argkey]) {
         value = argsDict[argkey];
       } else { // default to first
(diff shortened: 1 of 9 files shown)


workbench/src/renderer/server_requests.js
```diff
@@ -82,7 +82,6 @@ export function fetchDatastackFromFile(payload) {
       headers: { 'Content-Type': 'application/json' },
     })
       .then((response) => response.json())
-      .catch((error) => logger.error(error.stack))
```
emlys, line 85: Just making sure I understand - this error would now be caught farther up?

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -274,8 +275,20 @@ class SetupTab extends React.Component {
...
-    const datastack = await fetchDatastackFromFile(filepath);
     const { pyModuleName, switchTabs, t } = this.props;
+    let datastack;
+    try {
+      datastack = await fetchDatastackFromFile(filepath);
+    } catch (error) {
+      logger.error(error);
+      alert( // eslint-disable-line no-alert
+        t(
+          'No InVEST model data can be parsed from the file:\n {{filepath}}',
```
emlys, line 286: I wonder if this message could provide any more information to help the user solve the problem. Maybe something like 
> File could not be opened as an InVEST model run. Are you sure this is a valid datastack for a current InVEST model?

emlys (commented):
@davemfish this all makes sense! Just one suggestion about the alert message.

workbench/src/renderer/components/SetupTab/index.jsx
davemfish: I agree it's not perfect but I'm kind of at a loss for how to improve it. This is from the "Load Parameters" button on a model's sidebar, so we're not really opening a "model run" here. If "datastack" had a clear meaning to users then I'd agree we could say, "this is not a valid datastack", but I'm afraid that term is mostly insider-jargon. 

My hope is that anyone who sees this message will just realize they clicked on the wrong file or something. If they really don't know what file they should be looking for, then that's a larger design flaw and we should try to add some guidance before the user reaches this alert. We have a tooltip that tries to do that now.

workbench/src/renderer/server_requests.js
davemfish: That's right. It felt better to add the user-facing alert message to UI component, so that's where the error-handling is now.

workbench/src/renderer/components/SetupTab/index.jsx
emlys: Oh I see what you mean. Yeah, this should be fine for now.

emlys (approved):

PR #1371, opened 2023-08-11
Close tabs on middle click

### LINKED ISSUE #1261: Middle-clicking a workbench tab opens a new blank window
When I ran an InVEST model in the workbench 3.13.0 release candidate, I went to close a model tab like I do in my web browser: by middle-clicking the tab.  Doing so brought up a window:

Screenshot 2023-03-17 at 3 11 38 PM

I think I would expect it to close the tab, but maybe middle-click is supposed to indicate some other type of action?

To reproduce (at least on my mac):
1. run a model in the workbench on the sample data
2. When the model finishes executing, middle-click the model tab
3. Observe new window opening.

dcdenu4:
## Description
Prevent the default behavior of a middle click opening a new window / tab on an InVEST Tab and instead close that tab.

I was unsure of whether this should live on the `Nav.Link` element or up one on the `Nav.Item` element. I opted for `Nav.Link` because the default nature of middle click on links...

I was unsure of whether this deserved a test? I could not find a "close tab" / "switch tab" test in `investtabtest`

Manually verified behavior on Windows with mouse wheel middle click in development build.

Fixes #1261 

## Checklist
- ~[ ] Updated the user's guide (if needed)~

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -59,6 +59,9 @@ Unreleased Changes
     * Closing the main window will now close any user's guide windows that are
       open. Fixed a bug where the app could not be reopened after closing.
       (`#1258 <https://github.com/natcap/invest/issues/1258>`_)
+    * Middle clicking an InVEST model tab was opening a blank window. Now
+      middle clicking will close that tab as expected.
+      (`#1261 <https://github.com/natcap/invest/issues/1261>`_)
 * Forest Carbon
     * The biophysical table is now case-insensitive.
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Looks good to me! I don't think this requires testing. But FYI the tests that open and close tabs are in `app.test.js`. 

And I think it makes sense to place this handler on the `Nav.Link`, not the `Nav.Item`.

PR #1375, opened 2023-08-16
Move validation out of execute methods

### LINKED ISSUE #1373: Don't call `validate` from `execute`
On the team call today, we decided to standardize on keeping `validate` orthogonal to `execute`. 5 models call `validate` at the start of `execute`, and some others duplicate validation in `execute` that's already handled in `validate`. We'll separate these out.

emlys:
## Description
Fixes #1373 

- Removed calls to `validate` from `execute`
- Removed validation from `execute` that's redundant to what `validate` does
- For Wind Energy, Recreation, and NDR, moved custom validation from `execute` into `validate`. I briefly tested the UI for these models to confirm that the new validation messages show up.

I did not do anything about custom validation that happens in functions other than `execute` - trying to keep the scope small.

## Checklist


tests/test_hra.py
```diff
@@ -1476,10 +1476,3 @@ def test_model_stressor_mismatch(self):
                       str(cm.exception))
         self.assertIn("Missing from criteria table: transportation",
                       str(cm.exception))
-
-        args['risk_eq'] = 'some other risk eq'
-        with self.assertRaises(ValueError) as cm:
```
davemfish, line 1481: Depending on my HRA comment, we may want to keep this in.

tests/test_recreation.py
```diff
@@ -697,9 +697,8 @@ def test_data_missing_in_predictors(self):
         table_path = os.path.join(
             SAMPLE_DATA, 'predictors_data_missing.csv')
 
-        with self.assertRaises(ValueError):
-            recmodel_client._validate_same_projection(
-                response_vector_path, table_path)
+        self.assertIsNotNone(recmodel_client._validate_same_projection(
```
davemfish, line 700: The docstring of this test could use updating: no longer raises exception.

tests/test_recreation.py
```diff
@@ -709,9 +708,8 @@ def test_data_different_projection(self):
         table_path = os.path.join(
             SAMPLE_DATA, 'predictors_wrong_projection.csv')
 
-        with self.assertRaises(ValueError):
-            recmodel_client._validate_same_projection(
-                response_vector_path, table_path)
+        self.assertIsNotNone(recmodel_client._validate_same_projection(
```
davemfish, line 711: update docstring please

tests/test_recreation.py
```diff
@@ -721,10 +719,9 @@ def test_different_tables(self):
             SAMPLE_DATA, 'predictors_data_missing.csv')
         scenario_table_path = os.path.join(
             SAMPLE_DATA, 'predictors_wrong_projection.csv')
-
-        with self.assertRaises(ValueError):
+        self.assertIsNotNone(
```
davemfish, line 722: update docstring please

tests/test_recreation.py
```diff
@@ -984,9 +980,8 @@ def test_predictor_id_too_long(self):
             'results_suffix': '',
             'workspace_dir': self.workspace_dir,
         }
-
-        with self.assertRaises(ValueError):
-            recmodel_client.execute(args)
+        msgs = recmodel_client.validate(args)
+        self.assertIn('more than 10 characters long', msgs[0][1])
```
davemfish, line 984: update docstring please

tests/test_recreation.py
```diff
@@ -1125,9 +1120,9 @@ def test_year_order(self):
                 SAMPLE_DATA, 'predictors_scenario.csv'),
             'workspace_dir': self.workspace_dir,
         }
-
-        with self.assertRaises(ValueError):
-            recmodel_client.execute(args)
+        msgs = recmodel_client.validate(args)
+        self.assertEqual(
+            'Start year must be less than or equal to end year.', msgs[0][1])
```
davemfish, line 1125: Depending on my other comment, we may also still want to check for `assertRaises(ValueError)`

tests/test_routedem.py
```diff
@@ -59,23 +59,6 @@ def _make_dem(target_path):
...
-        args = {
-            'workspace_dir': self.workspace_dir,
-            'algorithm': 'invalid',
-            'dem_path': os.path.join(self.workspace_dir, 'dem.tif'),
-            'results_suffix': 'foo',
-            'calculate_flow_direction': True,
-        }
-
-        RouteDEMTests._make_dem(args['dem_path'])
-        with self.assertRaises(RuntimeError) as cm:
```
davemfish, line 74: It looks like this test would still make sense if it checked for `KeyError` instead of `RuntimeError`. I think it's important that a bad algorithm raises some kind of error, so might as well keep this test around in my opinion.

tests/test_scenic_quality.py
```diff
@@ -225,38 +225,6 @@ def test_no_valuation(self):
...
-            'refraction': 0.13,
-            'do_valuation': True,
-            'valuation_function': 'INVALID FUNCTION',
-            'a_coef': 1,
-            'b_coef': 0,
-            'max_valuation_radius': 10.0,
-            'n_workers': -1,
-        }
-
-        with self.assertRaises(ValueError):
```
davemfish, line 257: Similar comment as in `test_routedem`. This would now raise a `NameError`, I think. And it would be bad if it raised no error at all, so it might be nice to just keep the test.

tests/test_urban_nature_access.py
```diff
@@ -807,22 +807,6 @@ def test_polygon_overlap(self):
             [polygon_1, polygon_2, polygon_3], vector_path, wkt, 'GeoJSON')
         self.assertTrue(urban_nature_access._geometries_overlap(vector_path))
 
-    def test_invalid_search_radius_mode(self):
```
davemfish, line 810: Related to my comment in the model, we may want to keep this test.

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -1646,21 +1611,16 @@ def error_handler(err_level, err_no, err_msg):
...
-                f"{path} might have a different projection than the base AOI\n"
-                f"base:{base_ref.ExportToPrettyWkt()}\n"
-                f"current:{ref.ExportToPrettyWkt()}")
             invalid_projections = True
     if invalid_projections:
-        raise ValueError(
-            "One or more of the projections in the table did not match the "
-            "projection of the base vector")
+        return ("One or more of the projections in the table did not match "
+                "the projection of the base vector")
```
davemfish, line 1623: Could we include `base_vector_path` in the message instead of saying "base vector"? I'm worried it's not obvious that "base vector" is the AOI vector. And maybe the same for using `table_path` instead of "table"?

davemfish (approved):
Thank you @emlys ; lots of useful improvements here!

PR #1376, opened 2023-08-21
improve validation message when bounding boxes don't intersect #502

### LINKED ISSUE #502: Improve language in spatial overlap validation messages
@newtpatrol  suggests the current language isn't well-understood by users and we could do better:

>"Bounding box does not intersect at least one other [...]", and as we've seen on the forum, a lot of people don't understand what that means. Could it be changed to something like "Not all of the spatial layers overlap each other. Do they all have the same projected coordinate system?" 

I agree this proposed language is clearer.

emlys:
## Description
Fixes #502 
I took a stab at rewording the validation message based off the conversation in the issue. Let me know what you think!

## Checklist


dcdenu4 (approved):
Thanks @emlys , looks like a better message to me!

PR #1377, opened 2023-08-21
Update PR template

davemfish:
I think our last bullet here is a bit narrow and probably predates the Workbench. Testing the UI -- which a PR author should do manually when it makes sense -- is about testing the Workbench, not really the individual model UI.

## Checklist


dcdenu4 (approved):
Yep, good update!

PR #1381, opened 2023-08-24
HRA: use pygeoprocessing to create rasters

### LINKED ISSUE #1312: HRA: edge case where aligned datasets have different sizes
HRA accepts raster & vector inputs and aligns them all onto a raster stack with a target bounding box of the union of all the inputs.

For the rasters, we use `align_and_resize_raster_stack`.

For the vectors, we create an empty raster to rasterize onto, which has its size & geotransform determined by the same `target_bounding_box` and `pixel_size` passed to `align_and_resize`. Specifically, the size of the target raster is determined by,

```python
n_cols = int(numpy.ceil(
    abs((bbox_maxx - bbox_minx) / target_pixel_size[0])))
n_rows = int(numpy.ceil(
    abs((bbox_maxy - bbox_miny) / target_pixel_size[1])))
```

Rounding up with `numpy.ceil` makes sense...if the original vector includes geometries extending right out to the edge of `target_bounding_box`, then we don't want a sub-pixel gap between the last row/col of pixels and the edge of the `target_bounding_box`.

For the raster inputs, which pass through `align_and_resize` and then `warp_raster`,
`warp_raster` does a similar thing to reconcile a `target_bb` with width & height that may not be a multiple of the `target_pixel_size` , but just slightly different:

```python
    # sometimes bounding boxes are numerically perfect, this checks for that
    x_residual = (
        abs(target_x_size * target_pixel_size[0]) -
        (working_bb[2] - working_bb[0]))
    if not numpy.isclose(x_residual, 0.0):
        target_x_size += 1
    y_residual = (
        abs(target_y_size * target_pixel_size[1]) -
        (working_bb[3] - working_bb[1]))
    if not numpy.isclose(y_residual, 0.0):
        target_y_size += 1

    if target_x_size == 0:
        LOGGER.warning(
            "bounding_box is so small that x dimension rounds to 0; "
            "clamping to 1.")
        target_x_size = 1
    if target_y_size == 0:
        LOGGER.warning(
            "bounding_box is so small that y dimension rounds to 0; "
            "clamping to 1.")
        target_y_size = 1

    # this ensures the bounding boxes perfectly fit a multiple of the target
    # pixel size
    working_bb[2] = working_bb[0] + abs(target_pixel_size[0] * target_x_size)
    working_bb[3] = working_bb[1] + abs(target_pixel_size[1] * target_y_size)
```

This is effectively the same as the HRA method -- expanding the target bounding box as needed to accomadate an extra row or column of pixels. But there's an edge case where `numpy.isclose(x_residual, 0.0)` is `True`, so no extra column is added during `align_and_resize`. But the extra column _is_ added in HRA because it simply takes `numpy.ceil`.

My inclination would be to change the logic in HRA and use the approach found in `warp_raster` instead of using `numpy.ceil`

davemfish:
Fixes #1312 

## Checklist


phargogh (approved):
Easy!

PR #1384, opened 2023-08-25
REC: `point_nearest_distance` metric now measures from centroid

### LINKED ISSUE #1347: REC: `point_nearest_distance` metric does not do what docs say
> point_nearest_distance: Predictor is a point vector. Metric is the Euclidean distance between the center of each AOI grid cell and the nearest point in this layer.

But actually it does a `polygon.distance(point)` where `polygon` and `point` are shapely geometries. So the measurement is not from centroid to point, but from point to nearest edge of polygon. And the most problematic part is if the point is inside the polygon, the distance returns `0`.

This was probably okay when used with a gridded AOI, but it's a pretty bad approximation for non-gridded AOIs. And even for the former, distance from centroid is probably a better metric.

I think it's as easy as `polygon.centroid.distance(point)`. in `def _point_nearest_distance`

davemfish:
The User's Guide (i.e. model spec) suggested it was measuring from "center" of an AOI polygon to the predictor point, whereas it was actually measuring from the edge of the polygon. Now it measures from the centroid.

Fixes #1347 

## Checklist


tests/test_recreation.py
```diff
@@ -581,14 +581,14 @@ def test_all_metrics_local_server(self):
         expected_grid_vector_path = os.path.join(
             REGRESSION_DATA, 'predictor_data_all_metrics.shp')
         utils._assert_vectors_equal(
-            out_grid_vector_path, expected_grid_vector_path, 1e-3)
+            expected_grid_vector_path, out_grid_vector_path, 1e-3)
```
davemfish, line 584: When it fails, the message from this assertion function expects the first arg to be the "expected" value and second to be the "actual".

dcdenu4 (approved):
Looks good to me, thanks @davemfish!

PR #1392, opened 2023-08-29
Remove support for excel files

### LINKED ISSUE #1391: Drop support for excel files
The two HRA tables are the only places in invest where we allow Excel files. 

Supporting Excel files requires some additional maintenance:
- the `excel_ok` attribute in the model spec
- extra code in validation to handle excel files
- require extra dependencies (`xlrd`, `openpyxl`) necessary for pandas to read Excel files
- HRA has its own table reading functions

It seems that the added maintenance costs are no longer worth it, especially as we are expanding CSV validation and type checking. I’m guessing we originally supported Excel because the HRA tables are complicated and it’s nice to work with them in a spreadsheet software, but users can still do that and export to CSV when they’re done. Stacie and Jade confirmed on slack that they are OK with using CSVs only.

emlys:
## Description
Fixes #1391 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -109,6 +109,8 @@ Unreleased Changes
     * Fixed a bug where vector inputs could be rasterized onto a grid that is
       not exactly aligned with other raster inputs.
       (`#1312 <https://github.com/natcap/invest/issues/1312>`_)
+    * Dropped support for Excel (.xlsx) files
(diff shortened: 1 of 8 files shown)


dcdenu4 (commented):
Thanks @emlys , is there a Users Guide counterpart needed for this?

emlys:
@dcdenu4 User's guide update added!

dcdenu4 (approved):

PR #1393, opened 2023-08-31
Add license agreement back to installer

### LINKED ISSUE #1394: Add license agreement step to Workbench installation
After chatting about whether we should be presenting our license during the installation process I found that a few open source and free projects don't bother and some do. I reached out to Stanford OTL and they recommended we do present our license because it never hurts to be upfront and explicit. So add the license "agree" step back to the installers.

dcdenu4:
## Description
After a consult with Stanford OTL we decided that it would be best to be upfront and explicit with the license in which we distribute InVEST and provide that to users during installation.

The NOTICE and LICENSE are concatenated and shown during installation. I altered NOTICE a bit to make this concatenation more readable. I think it's fine for now and if we ever add items to the NOTICE file can rethink how best to handle it. Removed registered TM symbol from NOTICE because it was not playing nice with NSIS.

Fixes #1394 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -80,6 +80,9 @@ Unreleased Changes
       (`#1261 <https://github.com/natcap/invest/issues/1261>`_)
     * Updated InVEST logo to use new version with registered trademark symbol.
       (`https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy`_)
+    * InVEST is licensed using a permissive open source license. But we have
+      decided to add back the license and agreement step to the installer to
+      be upfront and explicit about how InVEST is licensed.
 * Coastal Blue Carbon
     * Added validation for the transition table, raising a validation error if
       unexpected values are encountered.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -80,7 +80,7 @@ Unreleased Changes
       (`#1261 <https://github.com/natcap/invest/issues/1261>`_)
     * Updated InVEST logo to use new version with registered trademark symbol.
       (`InVEST TM and Logo Use Policy 
-       <https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy>`_)
+      <https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy>`_)
     * InVEST is licensed using a permissive open source license. But we have
       decided to add back the license and agreement step to the installer to
       be upfront and explicit about how InVEST is licensed.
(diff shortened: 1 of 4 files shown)


dcdenu4:
@davemfish and @phargogh , I have tested the Windows NSIS version, but have not tested Mac. Mac DMG is supposed to magically find and incorporate the `license_en.txt` from "build resources" into the installation process. Thanks for seeing if this is working!

HISTORY.rst
```diff
@@ -80,6 +80,9 @@ Unreleased Changes
       (`#1261 <https://github.com/natcap/invest/issues/1261>`_)
     * Updated InVEST logo to use new version with registered trademark symbol.
       (`https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy`_)
```
davemfish, line 82: I think the readthedocs build doesn't like something about this URL. I believe the trailing `_` is meant to be placed on the text that should contain the link, not on the url itself? If you just want the URL printed as-is, I don't think it needs backticks or underscore or parens.

davemfish (changes requested):

phargogh (approved):
The license is presented!  When you open the DMG, the license (and NOTICE file at the top) are presented before the image will mount, so that's great.  There's a minor issue with line wrapping in the NOTICE file, but the content is all there.

Screenshot 2023-08-31 at 9 19 13 AM

dcdenu4:
Thanks @davemfish and @phargogh! I just pushed changes for History and added some spaces in NOTICE to see if that'd help with line wrapping. It looks like it's treating each line as a line as opposed to a paragraph and not wrapping at all unless an individual line needs wrapping.

HISTORY.rst
```diff
@@ -79,7 +79,11 @@ Unreleased Changes
       middle clicking will close that tab as expected.
       (`#1261 <https://github.com/natcap/invest/issues/1261>`_)
     * Updated InVEST logo to use new version with registered trademark symbol.
-      (https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy)
+      (`InVEST TM and Logo Use Policy 
+       <https://naturalcapitalproject.stanford.edu/invest-trademark-and-logo-use-policy>`_)
```
dcdenu4, line 83: @davemfish this look right to me and I'm not sure why it's not rendering correctly when viewing in the rich diff.

davemfish: @dcdenu4 it looks like the new RST-linter (awesome!) is flagging these lines

dcdenu4: Yep, it was an indent problem! Fixed!

dcdenu4:
@davemfish and @phargogh , once the tests pass this should be good to go. Punting on the line wrapping issue for Mac DMG install. Unless someone has a strong feeling about it.

davemfish (approved):

PR #1406, opened 2023-09-07
Add actions job to report translation stats

### LINKED ISSUE #1403: Develop a way to report on translations needing updates
Having a report summarizing how much translated text has changed since the last translation would be helpful to decide when to request new translations.

emlys:
## Description
Fixes #1403 

This PR adds a job to our Github Actions workflow that prints out stats about missing translations.

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -65,6 +65,79 @@ jobs:
             # Skip line-too-long errors (D001)
             python -m doc8 --ignore D001 HISTORY.rst
 
+  report-translation-status:
+    name: Report status of translations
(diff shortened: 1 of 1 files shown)


dcdenu4 (changes requested):
Hey @emlys, I think we decided to put a pin in this for now or otherwise pull the logic of this out into a script to be able to run from the command line as well? Happy to defer to you in what you think would be the best next step here.

emlys:
It seems that most of the interest in this feature was related to the users guide, but we don't have a great way to measure that, because git diff can count how many lines changed, but not how many words. There hasn't been any more demand for this, so closing for now.

PR #1407, opened 2023-09-08
Revert license location in electron build for code signing

dcdenu4:
## Description
We attempted to move the License file to the root directory by using `extraFiles` in the electron build config. This had a side effect that made the Mac code signing step fail. So reverting back to having the License and Notice files be in the `extraResources` location.

## Checklist


davemfish (approved):
Looks good, ty!

PR #1418, opened 2023-09-25
Correcting how conda-forge channel is defined

phargogh:
My builds started failing until I realized that setup-micromamba changed how it defined channels.


emlys (approved):
It's strange that this stopped working given that we have setup-micromamba pinned @ v1. But, this looks correct according to the docs!

Issue #327, opened 2020-10-06
Allow CSV validation to validate rasters/vectors referenced in a CSV column


phargogh:
We now have several models that use CSV files to define paths to other spatial files, including HRA, HQ and CBC (once #209 is complete), and I think Recreation also does something similar.

If we want to validate a spatial filepath that is referenced in one of these inputs, the validation performed is a manual step outlined in the module's `validate` function.  We should allow for an `ARGS_SPEC`-defined configuration option that then will allow for the files in a given column to be validated as a spatial file (with all of the optional validation parameters supported by the check functions).

PR #1419, opened 2023-09-26
Use `read_csv_to_dataframe` in validation

### LINKED ISSUE #1379: Use `read_csv_to_dataframe` in CSV validation
Validation has its own logic for reading and checking CSVs. This is now redundant to `utils.read_csv_to_dataframe` and less comprehensive. It would make sense to read CSVs for validation in the same way that we're going to read them in `execute`. 

Currently, CSV validation check that:
- the csv exists
- the file has the necessary permissions
- exact column names exist in the csv

With `utils.read_csv_to_dataframe`, we can also easily check that:
- at least one column name matching a pattern exists in the csv
- values can be cast to the appropriate data type for each column

### related issues

This makes sense to do before #327 because `read_csv_to_dataframe` handles path expansion. 

I'd also like to revisit whether we can drop support for Excel files (which is only allowed in one table in HRA) rather than writing additional code to support them.

emlys:
## Description
Fixes #1379 

This sets us up for #327.

- Split out some of the functionality of `utils.read_csv_to_dataframe` in a new function, `validation.get_validated_dataframe`. This new function enforces a contract: given a CSV path and a spec, `get_validated_dataframe` will either return a dataframe that matches the spec, or raise an error if that's not possible. `utils.read_csv_to_dataframe` now handles the lower-level CSV parsing details.
- `validation.get_validated_dataframe` is used _both_ in validation and in execute. If it raises an error during validation, that error message will be returned as a validation message. If it raises an error in execute, that error will be raised. This is convenient because we get the same message in either context without duplicating.
- All tables are now read in via `utils.read_csv_to_dataframe` (indirectly through `validation.get_validated_dataframe` in most cases, or directly in the few places (HRA criteria table, Wave Energy machine performance table)). This lets us apply the same basic rules to all CSVs, such as encoding, dropping empty rows/columns, and lower-casing column names.

## Checklist


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -462,6 +462,13 @@
                                     "type": "integer",
                                     "about": "Shore point ID"
                                 },
+                                "R_hab": {
```
emlys, line 465: Moving this up so that the `r_hab` column is matched before we get to the `[HABITAT]` pattern, which matches everything.

src/natcap/invest/wave_energy.py
```diff
@@ -287,26 +287,20 @@
...
         "machine_param_path": {
             "type": "csv",
-            "rows": {
-                "capmax": {
-                    "about": gettext("Maximum capacity for device."),
-                    "type": "number",
-                    "units": u.kilowatt
-                },
-                "hsmax": {
+            "columns": {
```
emlys, line 290: I switched this to columns so that we don't have to validate the individual rows, as discussed on the last team call.

phargogh: Yep, makes sense!  Could you add an inline comment about this so we are sure to remember?

tests/test_habitat_quality.py
```diff
@@ -2056,6 +2056,7 @@ def test_habitat_quality_argspec_missing_threat_header(self):
                 '0.07,1.0,threat_2,threat_2_c.tif,threat_2_f.tif\n')
 
         validate_result = habitat_quality.validate(args, limit_to=None)
+        print(validate_result)
```
phargogh, line 2059: Left over from debugging?
```suggestion
```

phargogh (changes requested):
Thanks @emlys !  I really like the new contract being enforced here in the validated dataframe ... really cleans things up conceptually.

I just had a few minor comments and suggestions, and then would you object to adding a note to `HISTORY` about this?  I think the main thing that might be worth mentioning there is something along the lines of improving consistency in validation of tables, which should improve the readability of validation errors.  Just an idea ... I defer to you on it!

src/natcap/invest/utils.py
```diff
@@ -597,7 +597,7 @@ def expand_path(path, base_path):
     return os.path.abspath(os.path.join(os.path.dirname(base_path), path))
 
 
-def read_csv_to_dataframe(path, spec, **kwargs):
+def read_csv_to_dataframe(path, **kwargs):
```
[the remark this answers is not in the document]
emlys: Oh true!

src/natcap/invest/validation.py
```diff
@@ -572,19 +572,108 @@ def check_boolean(value, **kwargs):
...
+                elif col_spec['type'] in {'freestyle_string', 'option_string'}:
+                    df[col] = df[col].apply(
+                        lambda s: s if pandas.isna(s) else str(s).strip().lower())
+                    df[col] = df[col].astype(pandas.StringDtype())
+                elif col_spec['type'] in {'number', 'percent', 'ratio'}:
+                    df[col] = df[col].astype(float)
+                elif col_spec['type'] == 'integer':
+                    df[col] = df[col].astype(pandas.Int64Dtype())
+                elif col_spec['type'] == 'boolean':
+                    df[col] = df[col].astype('boolean')
```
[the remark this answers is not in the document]
emlys: Sure! That should never happen with our code because the model specs are tested, but it would be good to have in case of plugins.

emlys:
@phargogh I think I addressed all your comments

phargogh (approved):
Looks great, thank you!

PR #1420, opened 2023-09-27
Wind energy: handle the case where the grid points path is empty or None

### LINKED ISSUE #1417: Wind Energy: model expects a grid points vector when it shouldn't
Bug report from Rob on slack:
> Hi folks, I'm seeing an issue with the offshore wind model I wanted to bring to your attention. First read is that it could be related to not providing a grid points shapefile, but instead relying on the "Average Shore-To-Grid Distance" field, yet the model still expecting a grid points shapefile. [Here's the datastack](https://drive.google.com/file/d/1KQOfR3HRcw3ydVIDoOmO6mvYnka3xnWG/view?usp=sharing) - it's all sample data except the AOI and the field parameters. Log for model run [here](https://drive.google.com/file/d/1j5hzrepIDM4amWu4szWDv0lcDVVFdhGj/view?usp=sharing). Ran into issue on 3.11 and updated to 3.14 and still seeing it.

emlys:
## Description
Fixes #1417 
Rob ran into this issue where the model tried to read an empty string as a CSV path.

Also fixing a PropTypes issue that came up when I loaded the datastack to reproduce this issue - numbers should be allowed. This did not affect the functionality.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -44,6 +44,9 @@ Unreleased Changes
 * SDR
     * RKLS, USLE, avoided erosion, and avoided export rasters will now have
       nodata in streams (`#1415 <https://github.com/natcap/invest/issues/1415>`_)
+* Wind Energy
+    * Fixed a bug where model would error when the grid points path was empty
+      (`#1417 <https://github.com/natcap/invest/issues/1417>`_)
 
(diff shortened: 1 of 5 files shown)


dcdenu4 (approved):
Thanks Emily!

PR #1428, opened 2023-10-16
Prompt user to extract datastack archive to permanent location

### LINKED ISSUE #1308: ``get_datastack_info`` from tgz returns an unusable args dictionary
As Stacie pointed out, 

>when I drag and drop the .tgz or "Load parameters from file" into Workbench, the entries that it fills the UI in with give "file not found" errors. But, when I drag in the parameter .json file within the archive, it populates correctly.

This amounts to the fact that `datastack.get_datastack_info('some.tgz')` seems to extract only the `.json` file to a temp directory, but not the rest of the data from the archive. So it defaults to returning an `args` dict with the paths from the `.json`, which are relative paths. They would only be valid paths if the rest of the archive was extracted.

As far as I can tell, we don't have a use-case for needing an `args` dict from an archive unless we are also extracting all the data from that archive.

It might be a bad idea to have `datastack.get_datastack_info` return something different depending on the type of datastack. So maybe we leave that as-is, but follow it up with a call to `extract_datastack_archive`, which also returns the `args` dict. And I believe the only place where we use this is in `ui_server.py`

```python
@app.route(f'/{PREFIX}/post_datastack_file', methods=['POST'])
def post_datastack_file():
    """Extracts InVEST model args from json, logfiles, or datastacks.

    Body (JSON string): path to file

    Returns:
        A JSON string.
    """
    filepath = request.get_json()
    stack_type, stack_info = datastack.get_datastack_info(
        filepath)
    model_name = PYNAME_TO_MODEL_NAME_MAP[stack_info.model_name]
    result_dict = {
        'type': stack_type,
        'args': stack_info.args,
        'module_name': stack_info.model_name,
        'model_run_name': model_name,
        'model_human_name': MODEL_METADATA[model_name].model_title,
        'invest_version': stack_info.invest_version
    }
    return json.dumps(result_dict)
```

Implementing this for use by the Workbench will also require asking the user to choose a location to extract to, so might necessitate a new UI component for that.

emlys:
## Description
Fixes #1308
Using the Save As dialog to prompt users to choose a location to extract their datastack archive. Now the entire archive is extracted and the file paths loaded into the input form are valid. This does introduce a slight delay while extraction happens, but I didn't think it was bad enough to need a loading indicator. 

The dialog looks like this on mac:
image

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -42,6 +42,8 @@ Unreleased Changes
       have been replaced with ``numpy.prod``.
       https://github.com/natcap/invest/issues/1410
     * Add support for python 3.11 (`#1103 <https://github.com/natcap/invest/issues/1103>`_)
+    * Datastack archives will now be correctly extracted
+      (`#1308 <https://github.com/natcap/invest/issues/1308>`_)
 * SDR
     * RKLS, USLE, avoided erosion, and avoided export rasters will now have
       nodata in streams (`#1415 <https://github.com/natcap/invest/issues/1415>`_)
--- a/src/natcap/invest/datastack.py
+++ b/src/natcap/invest/datastack.py
@@ -177,7 +177,8 @@ def get_datastack_info(filepath, extract_path=None):
         filepath (string): The path to a file on disk that can be extracted as
             a datastack, parameter set, or logfile.
         extract_path (str): Path to a directory to extract the datastack, if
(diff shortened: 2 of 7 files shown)


davemfish (approved):
Thanks so much @emlys , this looks great to me

PR #1430, opened 2023-10-18
Validate vector geometry types

### LINKED ISSUE #1374: InVEST models don't validate input vector geometry
There was a user who was getting odd results on the [forums](https://community.naturalcapitalproject.org/t/why-does-the-sdr-model-complete-and-produce-a-tif-file-but-fails-to-show-the-content-in-qgis/3993/7) and it turns out they were passing in a watershed vector with a line geometry instead of a polygon.

Can we validate vector geometry type through the MODEL_SPEC?

emlys:
## Description
Fixes #1374 

## Checklist

### DIFF
--- a/src/natcap/invest/coastal_vulnerability.py
+++ b/src/natcap/invest/coastal_vulnerability.py
@@ -3507,34 +3507,6 @@ def validate(args, limit_to=None):
     invalid_keys = validation.get_invalid_keys(validation_warnings)
     sufficient_keys = validation.get_sufficient_keys(args)
 
-    if 'shelf_contour_vector_path' not in invalid_keys:
-        # We can assume that this vector can already be opened because it
-        # passed validation so far.
-        vector = gdal.OpenEx(args['shelf_contour_vector_path'], gdal.OF_VECTOR)
-        layer = vector.GetLayer()
-        # linestring codes:
-        # https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogr_core.h
-        if layer.GetGeomType() not in [
-                ogr.wkbLineString, ogr.wkbMultiLineString,
-                ogr.wkbLineStringM, ogr.wkbMultiLineStringM,
-                ogr.wkbLineStringZM, ogr.wkbMultiLineStringZM,
-                ogr.wkbLineString25D, ogr.wkbMultiLineString25D]:
-            validation_warnings.append(
-                (['shelf_contour_vector_path'], POLYLINE_VECTOR_MSG))
-
-    if ('slr_vector_path' in sufficient_keys and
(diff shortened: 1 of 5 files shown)


src/natcap/invest/coastal_vulnerability.py
```diff
@@ -3507,34 +3507,6 @@ def validate(args, limit_to=None):
     invalid_keys = validation.get_invalid_keys(validation_warnings)
     sufficient_keys = validation.get_sufficient_keys(args)
 
-    if 'shelf_contour_vector_path' not in invalid_keys:
```
emlys, line 3510: This is now redundant to `validation.validate`.

tests/test_wave_energy.py
```diff
@@ -678,34 +678,3 @@ def test_validate_keys_missing_values(self):
         expected_error = (
             ['dem_path', 'wave_base_data_path'], validation.MESSAGES['MISSING_VALUE'])
         self.assertTrue(expected_error in validation_error_list)
-
-    def test_validate_bad_aoi_incorrect_proj_units(self):
-        """WaveEnergy: test validating AOI vector with incorrect units."""
-        from natcap.invest import wave_energy, validation
-
```
emlys, line 685: These tests aren't specific to wave energy, and the same thing is covered in the validation tests. I removed them because the sample data vectors they use have the wrong geometry type, which now breaks.

src/natcap/invest/validation.py
```diff
@@ -335,8 +337,8 @@ def load_fields_from_vector(filepath, layer_id=0):
     return fieldnames
 
 
-def check_vector(filepath, fields=None, projected=False, projection_units=None,
-                 **kwargs):
+def check_vector(filepath, geometries, fields=None, projected=False,
```
dcdenu4, line 340: Could you add `geometries` to the Args definition?

HISTORY.rst
```diff
@@ -42,6 +42,8 @@ Unreleased Changes
       have been replaced with ``numpy.prod``.
       https://github.com/natcap/invest/issues/1410
     * Add support for python 3.11 (`#1103 <https://github.com/natcap/invest/issues/1103>`_)
+    * Vector geometry types will now be validated for all models
```
dcdenu4, line 45: Is this true? I see the update to `validation.py` but I don't see where each model would be taking advantage of the new param in `check_vector`.

dcdenu4 (changes requested):
Thanks @emlys, one suggestion and one question. I also see there are some merge conflicts in History.

src/natcap/invest/validation.py
emlys: good catch! Added it to the docstring.

PR #1437, opened 2023-10-25
Update to pygeoprocessing 2.4.2

### LINKED ISSUE #1421: Migrate to pygeoprocessing kernel library
The decay kernel functions in `utils` are now redundant to `pygeoprocessing.kernels`. These can be deprecated, along with any model-specific kernel functions that are now available from pygeoprocessing.

emlys:
## Description
Fixes #1421

### enhancements

Replaced `raster_calculator` with `raster_map` in places where it simplified things.
There are still many places where `raster_calculator` is needed, mostly related to nodata masking. I didn't write full docstrings for the simple one-line functions that are passed to `raster_map`, hoping that they can be replaced with lambdas in the near future.

Replaced kernel creation functions with `pygeoprocessing.kernels` where possible.

Replaced `utils.array_equals_nodata` with `pygeoprocessing.array_equals_nodata`.

### changes to tests
A few expected values in the UCM tests had to be updated because of changes to `zonal_statistics`. Out of almost 1500 building polygons in the test data, zonal stats results were different (and more correct) for 7 of them. I am guessing because we used to reproject the raster to match the vector; now we reproject the vector to match the raster, which should be more accurate. This probably only showed up in UCM because it uses so many polygons and they are small relative to the pixel size.

In one habitat quality test, we were incorrectly creating a raster from an int8 array and then inserting the value 255 into it. Previously this worked because pygeoprocessing incorrectly created a uint8 raster from the int8 array. Now pygeoprocessing copies over the type correctly, and so the 255 became a -1. Fixed by updating the original array dtype.

## Checklist


src/natcap/invest/habitat_quality.py
```diff
@@ -298,17 +298,6 @@
                 "filtered_[THREAT]_aligned.tif": {
                     "about": "Filtered threat raster",
                     "bands": {1: {"type": "ratio"}},
-                },
-                "kernels": {
```
emlys, line 302: This was not actually used anywhere.

phargogh: Ah, this looks like it's left over from the old convolution-based workflow we recently reverted.  Thanks for catching this!

tests/test_ucm.py
```diff
@@ -69,7 +69,7 @@ def test_ucm_regression_factors(self):
             'avg_cc': 0.222150472947109,
             'avg_tmp_v': 37.325275675470998,
             'avg_tmp_an': 2.325275675470998,
-            'avd_eng_cn': 3520213.280928277,
+            'avd_eng_cn': 3520217.313878,
```
phargogh, line 72: Are these values changing because of a change in how we are masking?  If the behavior is changing, I wonder if we should note that change in HISTORY

tests/test_pollination.py
```diff
@@ -198,7 +198,7 @@ def test_pollination_no_farm_regression(self):
             result_sum += numpy.sum(data_block)
         # the number below is just what the sum rounded to two decimal places
         # when I manually inspected a run that appeared to be correct.
-        self.assertAlmostEqual(result_sum, 58.669518, places=2)
+        self.assertAlmostEqual(result_sum, 58.407155, places=2)
```
phargogh, line 201: Are these changing because of the change in how we're building kernels?  Or is it more about masking?  Either way, it might be useful to note in the HISTORY why values are changing slightly.

phargogh (commented):
Wow, thank you for this big update, @emlys !  It is _so_ nice to see all of the array masking logic abstracted away to `raster_map`.

I just had two questions for you about how the model's behavior was changing in UCM and in Pollination that were resulting in updated regression values.  It might save us some headaches (and forums questions!) down the road if we could identify and document in HISTORY what's changing that results in different outputs.

tests/test_ucm.py
emlys: These had to be updated because of changes to zonal_statistics. Out of almost 1500 building polygons in the test data, zonal stats results were different (and more correct) for 7 of them. I am guessing because we used to reproject the raster to match the vector; now we reproject the vector to match the raster, which should be more accurate. This probably only showed up in UCM because it uses so many polygons and they are small relative to the pixel size.

tests/test_pollination.py
emlys: This is because of the switch to `pygeoprocessing.kernels`, which is more correct in some cases. The kernel implementation in pollination didn't always produce symmetrical kernels:
image
With `pygeoprocessing.kernels`:
image

tests/test_ucm.py
emlys: I added a general note in history.

tests/test_pollination.py
emlys: Added a note to history.

phargogh (approved):
Thank you!

PR #1441, opened 2023-11-02
CV: bugfix for passing numpy float type to gdal

### LINKED ISSUE #992: CV bathymetry masking does not handle unset nodata

davemfish:
Sometime since this issue was created, we implemented `pygeoprocessing.choose_nodata`, which returns exactly the type that GDAL expects to receive in `SetNoDataValue()`. So that's the fix.

## Description
Fixes #992 

## Checklist


davemfish:
This PR can stay on hold until #1437

phargogh (approved):
Makes sense to me!  Thanks @davemfish

PR #1461, opened 2023-12-05
Handle BOMs when loading HRA criteria tables

### LINKED ISSUE #1460: HRA criteria table loading fails cryptically when CSV has a BOM
A user on the forums reported this exception:

```python
2/04/2023 13:21:47 natcap.invest.utils ERROR Exception while executing natcap.invest.hra
File “natcap\invest\hra.py”, line 537, in execute
File “natcap\invest\hra.py”, line 1948, in _parse_criteria_table
AssertionError: The criteria table is missing these section headers: HABITAT NAME
12/04/2023 13:21:47 natcap.invest.utils INFO Elapsed time: 0.04s
12/04/2023 13:21:47 natcap.invest.utils INFO Execution finished.
```

After looking into the error, it turns out that this is due to HRA not opening the CSV with the UTF-8-SIG codec.

phargogh:
This PR corrects (and tests for) loading the HRA criteria table using the UTF-8-SIG encoding.

Fixes #1460 

## Checklist

### DIFF
--- a/tests/test_hra.py
+++ b/tests/test_hra.py
@@ -353,7 +353,8 @@ class HRAUnitTests(unittest.TestCase):
 
         # Sanity check: make sure the file has the expected BOM
         bom_char = "\uFEFF"  # byte-order marker in 16-bit hex value
-        assert open(criteria_table_path).read().startswith(bom_char)
+        with open(criteria_table_path) as criteria_table:
+            assert criteria_table.read().startswith(bom_char)
 
         target_composite_csv_path = os.path.join(self.workspace_dir,
                                                  'composite.csv')
--- a/tests/test_hra.py
+++ b/tests/test_hra.py
@@ -326,6 +326,36 @@ class HRAUnitTests(unittest.TestCase):
         pandas.testing.assert_frame_equal(
             expected_composite_dataframe, composite_dataframe)
 
+    def test_criteria_table_parsing_with_bom(self):
+        """HRA: criteria table - parse a BOM."""
+        from natcap.invest import hra
+
+        criteria_table_path = os.path.join(self.workspace_dir, 'criteria.csv')
+        with open(criteria_table_path, 'w') as criteria_table:
+            bom_char = "\uFEFF"  # byte-order marker in 16-bit hex value
+            criteria_table.write(
+                textwrap.dedent(
+                    f"""\
+                    {bom_char}HABITAT NAME,eelgrass,,,hardbottom,,,CRITERIA TYPE
+                    HABITAT RESILIENCE ATTRIBUTES,RATING,DQ,WEIGHT,RATING,DQ,WEIGHT,E/C
+                    recruitment rate,2,2,2,2,2,2,C
+                    connectivity rate,2,2,2,2,2,2,C
+                    ,,,,,,,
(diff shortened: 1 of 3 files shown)


phargogh:
With the update to use `utils.read_csv_to_dataframe`, the only value-add here is the test to make sure we can handle the BOM when parsing.

dcdenu4 (approved):
BOMs away!

dcdenu4 (changes requested):
Oops, I was a little premature with my approval. Looks like the Windows tests are failing on this.

tests/test_hra.py
```diff
@@ -326,6 +326,41 @@ def test_criteria_table_parsing(self):
...
+                    fishing,RATING,DQ,WEIGHT,RATING,DQ,WEIGHT,E/C
+                    frequency of disturbance,2,2,3,2,2,3,C
+                    management effectiveness,2,2,1,2,2,1,E
+                    """
+                ))
+
+        # Sanity check: make sure the file has the expected BOM
+        bom_char = "\uFEFF"  # byte-order marker in 16-bit hex value
+        with open(criteria_table_path) as criteria_table:
+            assert criteria_table.read().startswith(bom_char)
```
dcdenu4, line 357: This bit is still failing on Windows and I think it's because of the following from Python's codec docs (https://docs.python.org/3/library/codecs.html#encodings-and-unicode)

>On encoding the utf-8-sig codec will write 0xef, 0xbb, 0xbf as the first three bytes to the file. On decoding utf-8-sig will skip those three bytes if they appear as the first three bytes in the file. In UTF-8, the use of the BOM is discouraged and should generally be avoided.

So maybe it's being stripped out when opening and decoding? Although this is contradictory to what I'm seeing in my terminal:

```python
>>> with open("test-bom.csv", 'w', encoding='utf-8-sig') as my_table:
...     my_table.write("HABITAT NAME,eelgrass,,,hardbottom,,,CRITERIA TYPE")
>>>
>>> with open("test-bom.csv") as table:
...     print(table.read())
...
ï»¿HABITAT NAME,eelgrass,,,hardbottom,,,CRITERIA TYPE
```
Which makes sense because:
>To increase the reliability with which a UTF-8 encoding can be detected, Microsoft invented a variant of UTF-8 (that Python calls "utf-8-sig") for its Notepad program: Before any of the Unicode characters is written to the file, a UTF-8 encoded BOM (which looks like this as a byte sequence: 0xef, 0xbb, 0xbf) is written. As it’s rather improbable that any charmap encoded file starts with these byte values (which would e.g. map to

>LATIN SMALL LETTER I WITH DIAERESIS
>RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK
>INVERTED QUESTION MARK

So, I'm not sure why the test assertion error looks like the BOM is being decoded and stripped, where as when I open and read it I get the funky character mapping...

phargogh:
Thanks @dcdenu4 !  It looks like the missing part was to open it up in binary mode ... that's going to bypass any interpretation or OS-specific assumptions of whether to strip those leading bytes.

dcdenu4 (approved):
Thanks James. There are some Workbench related dependencies failing, but they're unrelated.

PR #1473, opened 2023-12-12
add a datastack arg to the export-py CLI command.

### LINKED ISSUE #1472: CLI: export python script that includes args from a datastack
`invest export-py` does not currently accept a datastack arg, though the function that builds the python script can take an `args` dict. So this should be a trivial change to add a command-line arg to the parser.

I'm wanting this feature as it is the quickest way I can think of to run a model on a given datastack and drop into a python debugger on exceptions.

davemfish:
`invest export-py routedem -d data\invest-sample-data\RouteDEM\routedem_gura.invs.json -f run_routedem.py`

Now populates the `args` dict in `run_routedem.py` with values from the given parameter set. 

We already have tests that assert the args are included in the script correctly. I added one to test the new command-line arg.

Fixes #1472 

## Checklist


tests/test_cli.py
```diff
@@ -370,6 +370,35 @@ def test_export_python_default_filepath(self):
...
+            'args': {}
+        }
+        new_parameter_set_path = os.path.join(
+            self.workspace_dir, 'paramset.invs.json')
+        with open(new_parameter_set_path, 'w') as parameter_set_file:
+            parameter_set_file.write(
+                json.dumps(datastack_dict, indent=4, sort_keys=True))
+
+        target_filepath = os.path.join(self.workspace_dir, 'foo.py')
+        with redirect_stdout() as stdout_stream:
```
emlys, line 389: My understanding is this just prevents the stdout from showing up in the test output? If `stdout_stream` is not used, can this be simplified?
```suggestion
        with redirect_stdout():
```
```

emlys (changes requested):
Thanks @davemfish! My only other comment is that it's a little redundant to have to provide the model name as an argument, when it's also known from the datastack. I see why that's necessary using the `export_py` subparser though. Another option would be to make a separate command for converting a datastack to a script - but that would be extra work for a pretty minor issue.

davemfish:
> Thanks @davemfish! My only other comment is that it's a little redundant to have to provide the model name as an argument, when it's also known from the datastack. I see why that's necessary using the `export_py` subparser though. Another option would be to make a separate command for converting a datastack to a script - but that would be extra work for a pretty minor issue.

Yeah that's a great point @emlys . Okay let's not be hasty and design this properly then. I'm going to put this PR in draft mode and we can discuss further in #1472

davemfish:
Closing this since I'm not working on it. We can always re-open if we work on #1472

PR #1476, opened 2023-12-13
force autotest script to exit with code 1 if any models failed. Part 2

### LINKED ISSUE #1472: CLI: export python script that includes args from a datastack

davemfish:
Moved the autotest step to near the end of the job so that it can fail and everything else still runs anyway.

I tested this on my fork by removing the conditional about when this step will run:
https://github.com/davemfish/invest/actions/runs/7196631748/job/19601935165

Fixes #1472 

## Checklist


dcdenu4 (approved):
Waiting for tests

PR #1496, opened 2024-01-04
enable manual trigger for autorelease part 2 #1493

### LINKED ISSUE #1493: AUTORELEASE: error in part 2 checking for number of artifacts
`release-part-2.yml` checks for a specific number of artifacts in the release. This number can vary depending on number python versions supported, so we should be more resilient to this in the workflow, maybe not checking for a specific number at all.

### LINKED ISSUE #1494: autorelease workflow: add manual trigger for part 2
On our call today we discussed the issues with the autorelease process on 3.14.1. We agreed it makes sense to add an option to manually trigger part 2 of the autorelease workflow. This would've saved some time during the 3.14.1 release.

Most likely, any problem in part 2 will be due to a bug in the workflow itself, not in the invest code being released. These kind of problems are to be expected because we run the workflow infrequently and don't test it, so things become outdated. When that happens, it's convenient to be able to fix and re-run the workflow.

emlys:
## Description
Fixes #1493, fixes #1494 

Unfortunately there's no simple way to test these changes. I did confirm that the trigger works on my fork.

## Checklist

### DIFF
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -7,6 +7,12 @@ on:
     branches:
       - main
 
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        type: string
+
 env:
   GITHUB_TOKEN: ${{ secrets.AUTORELEASE_BOT_PAT }}
 
@@ -29,7 +35,7 @@ jobs:
 
   publish_release:
     # run this job if a PR was merged from an autorelease branch into main
-    if: startsWith(github.head_ref, 'autorelease') && github.event.pull_request.merged == true
+    if: ${{ ! github.head_ref || (startsWith(github.head_ref, 'autorelease') && github.event.pull_request.merged == true) }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/release-part-2.yml
```diff
@@ -58,8 +70,6 @@ jobs:
           # move the artifacts out of the folders
           mv artifacts/*/* artifacts
           rm -rf artifacts/Wheel*
-          # verify that all 6 wheels are there
-          if [ `find artifacts/* | wc -l` -ne 6 ]; then exit 1; fi
```
emlys, line 62: Without this check, if we change the name of the wheel artifacts, and forget to update it here, then the release will proceed without the necessary wheels. It seems like you can upload files to a release after it's published, so this shouldn't be a critical issue.

davemfish: Okay, that sounds good.

Actually it has me thinking that the more dangerous part is the artifacts for PyPI, since we can't modify that release after it's published. Like what if we changed the artifact names from "Wheel for *". We might not have any wheels downloaded, but we might still call `twine upload artifacts/natcap.invest*` with the source distribution `natcap.invest....tgz` available.

davemfish (commented):
This looks good, thanks, @emlys . In the event we accidentally trigger this without doing part 1, things should fail when it tries to find the non-existent Actions run for the tag.

Your comment about the potential for release artifact name-changes made me think about that case for the PyPI artifacts. There might be a fail-case to prevent there. But we don't necessarily need to make any more changes to this workflow right now. What do you think?

emlys:
@davemfish my understanding is you can upload additional files to a PyPI release too. On testpypi I am able to add another file after the first upload to a particular release.

davemfish:
> @davemfish my understanding is you can upload additional files to a PyPI release too. On testpypi I am able to add another file after the first upload to a particular release.

Ahh okay, great! Thanks for correcting me, I was blindly assuming it was not possible because of the general strictness of things on PyPI.

davemfish (approved):

PR #1499, opened 2024-01-05
Bugfix/1498 wind energy cast numbers

### LINKED ISSUE #1498: Wind Energy not casting number arguments
Rob G. reported in Slack that he and a colleague were experiencing issues with the wind energy model when trying to edit the `foundation_cost` in the Workbench UI. They were getting:

```python
2024-01-03 21:09:01,050 (natcap.invest.wind_energy) wind_energy._calculate_npv_levelized_rasters(1357) INFO Creating output NPV and levelized rasters.
2024-01-03 21:09:01,063 (taskgraph.Task) Task.add_task(706) ERROR Something went wrong when adding task calculate_npv_levelized_rasters (18), terminating taskgraph.
  File "taskgraph\Task.py", line 674, in add_task
  File "taskgraph\Task.py", line 1093, in _call
  File "natcap\invest\wind_energy.py", line 1415, in _calculate_npv_levelized_rasters
TypeError: can only concatenate str (not "float") to str
```

It turns out that the wind energy model passes all of `args` into `_calculate_npv_levelized_rasters` and does not do any casting on the number fields it uses from `args`.

dcdenu4:
## Description
Fixes an issue where certain number inputs were not being properly cast to float or int. So if the values were of string type the model would have unintended side effects like trying to concatenate instead of add in an operation. TaskGraph would also treat these values different when trying to 'clean' the arguments, interpreting strings as filepaths.

Fixes #1498 

## Checklist


davemfish (commented):
Thanks, @dcdenu4 , this fix looks good. My only suggestion would be to consider modifying one of the existing regression tests to raise this bug, by passing in a `str` for one of those numeric args. What do you think?

dcdenu4:
Hey @davemfish, I modified a few tests to pass in some strings, to test this check. Also did a HISTORY conflict merge.

davemfish (approved):
Thank you!

PR #1510, opened 2024-01-23
UNA: Handle the case where an admin geometry doesn't overlap any people

### LINKED ISSUE #1503: Urban Nature Access: handle the case where an admin boundary doesn't overlap any people
A [user on the forums](https://community.naturalcapitalproject.org/t/something-went-wrong-when-adding-task-aggregate-supply-demand-to-admin-units-single-rasters-17/4405) encountered a `ZeroDivisionError` that appears to be caused by an admin boundary polygon that doesn't overlap any people in the population raster.  The workaround suggested to the user was to locate that polygon and remove it from the vector until a fix can be implemented.

phargogh:
This PR addresses a case where an admin geometry doesn't overlap any people in the population raster, as might happen with a rural park.  Per the science team, `NaN` (which OGR saves as NULL, it appears) is used as a value in this case.

Fixes #1503

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -42,6 +42,9 @@ Unreleased Changes
     * Fixed a ``NameError`` that occurred when running the model using
       search radii defined per population group with an exponential search
       kernel. https://github.com/natcap/invest/issues/1502
+    * Fixed an issue where Urban Nature Access would crash if an administrative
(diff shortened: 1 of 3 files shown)


davemfish (approved):
Thanks, James, looks all good to me!

PR #1511, opened 2024-01-24
Proposal (and bugfix) to implement value-based conditional requirement

### LINKED ISSUE #1165: Implement conditional requirement for `required` keys anywhere in the args spec
When writing out the MODEL_SPEC for UNA, I inadvertently wrote `{"required": <expression>}` for vector and CSV columns, forgetting that validation of conditional requirement wasn’t yet implemented.  Here’s what I’m thinking for a solution:

1. Create a copy of args where all the required keys (regardless of where it is in the nested dict) are checked for expressions.  If an expression is found, evaluate it like we evaluate required keys at the top level.  The result is bool values where the expressions were before.
2. This new dict could be validated as we validate it now since we already check for required: True when validating individual inputs.

~~Related to #722~~ 
~~Needed for #1150~~

# Estimated time remaining

To be safe (and to make sure this is tested), I'd guess this is several days of effort, maybe 1 week of development time.

### LINKED ISSUE #1509: Catch missing uniform search radius in UI
This issue was just [posted to the user forum](https://community.naturalcapitalproject.org/t/urban-nature-access-valueerror-encountered-nan-in-calculation-but-default-nan-is-none/4379/8), and I replicated it with Workbench 3.14.1. 

In Workbench, with Urban Nature Access, If the user chooses a "Search Radius Mode" of "Uniform radius", the model requires that the user also provide "Uniform Search Radius" in the user interface. But the Workbench UI does not show an error when a value for "Uniform Search Radius" is not provided when it is required (and instead, throws a not useful error at runtime). Based on the behavior of other models, I think this should be caught in the UI, or at the very least caught while running with a user-friendly error message.

Currently, this error gets thrown while running the model:

> 2024-01-22 19:32:39,229 (natcap.invest.utils) utils.prepare_workspace(166) ERROR Exception while executing natcap.invest.urban_nature_access
> Traceback (most recent call last):
>   File "natcap\invest\utils.py", line 164, in prepare_workspace
>   File "invest\cli.py", line 470, in main
>   File "natcap\invest\urban_nature_access.py", line 934, in execute
> ValueError: could not convert string to float: ''

phargogh:
This proposal implements value-based conditional requirement expressions in validation.

The use case for this is that sometimes (like in UNA, see #1509 ), validation is conditional not only on whether a parameter is provided and has a value (is "sufficient"), but sometimes validation depends on the _value_ of another key.  In the case of UNA, several different inputs are required depending on the value of `search_radius_mode`.  If `search_radius_mode` represents uniform search radii, then `args['search_radius']` is required.  If `search_radius_mode` represents search radii by population group, then `args['population_gorup_radii_table']` is required.

An alternate approach to this could have been to just write out the required logic in UNA's `validate` function, but this feature also seemed useful and perhaps worth expanding.  I would be happy to replace this implementation with such a model-specific implementation if that is preferred.

Implementation-wise, I thought it might be convenient to be able to use a dot-notation in the MODEL_SPEC expression to represent the argument's value.  So for the above example, one might use `search_radius_mode.value == "{RADIUS_OPT_UNIFORM}"`.  Behind the scenes, validation (as proposed) will tokenize the expression and rewrite `search_radius_mode.value` into a new local variable name.  This allows backwards compatibility with existing expressions throughout InVEST, while supporting the proposed dot-notation.

Critical feedback is very welcome!

Fixes #1509
Fixes #1165 

## Checklist


src/natcap/invest/urban_nature_access.py
```diff
@@ -273,7 +273,7 @@
                     'type': 'number',
                     'units': u.meter,
                     'required':
-                        f'search_radius_mode == "{RADIUS_OPT_POP_GROUP}"',
+                        f'search_radius_mode.value == "{RADIUS_OPT_POP_GROUP}"',
```
emlys, line 276: I think it's not necessary to include the conditional requirement on the column here, since the table as a whole has the same requirement. And it might not have any effect because we don't yet evaluate conditional requirement on columns within a table ( that would be great to add, though!)

src/natcap/invest/urban_nature_access.py
```diff
@@ -125,7 +125,7 @@
                 "pop_[POP_GROUP]": {
                     "type": "ratio",
                     "required": (
-                        f"(search_radius_mode == '{RADIUS_OPT_POP_GROUP}') "
+                        f"(search_radius_mode.value == '{RADIUS_OPT_POP_GROUP}') "
```
emlys, line 128: I suspect this won't have the intended effect if validation doesn't yet evaluate conditonal requirement in vector fields: https://github.com/natcap/invest/blob/9e50106f64bd0525163e347030c4d439bb4b910f/src/natcap/invest/validation.py#L847-L849

src/natcap/invest/urban_nature_access.py
```diff
@@ -92,7 +92,7 @@
                     'type': 'number',
                     'units': u.meter,
                     'required':
-                        f'search_radius_mode == "{RADIUS_OPT_URBAN_NATURE}"',
+                        f'search_radius_mode.value == "{RADIUS_OPT_URBAN_NATURE}"',
```
emlys, line 95: Similarly, this might not have the desired effect since I think CSV validation doesn't evaluate conditional requirement for columns: https://github.com/natcap/invest/blob/9e50106f64bd0525163e347030c4d439bb4b910f/src/natcap/invest/validation.py#L622-L623

emlys (commented):
@phargogh I really like where this is going! Our existing validation doesn't yet evaluate conditional requirement of CSV columns or vector fields, so I think that would need to be added. I think we just left it out of the original `ARGS_SPEC` design to keep things simple, but it would be a great improvement to have.

I found two other models where I left in a placeholder conditional requrirement for CSV columns. It would be great to have those working too:
https://github.com/natcap/invest/blob/9e50106f64bd0525163e347030c4d439bb4b910f/src/natcap/invest/forest_carbon_edge_effect.py#L83-L106
https://github.com/natcap/invest/blob/9e50106f64bd0525163e347030c4d439bb4b910f/src/natcap/invest/urban_cooling_model.py#L72-L96

It's interesting that we've done this before in a limited way with boolean inputs by lumping in their true/false value with "input sufficiency". Which has worked well, but would it make more sense to explicitly check the _value_ of the input like you propose here? For instance,
- `"required": "do_valuation.value == True"` vs
- `"required": "do_valuation.value"` vs
- `"required": "do_valuation"`?

phargogh:
Thanks so much for taking a look at this @emlys !  I made a few changes that ultimately resulted in a larger refactor that really simplified `validate`.  Here's a summary:

## Expressions Now Use Arg Values

Symbols in expressions that represent args keys will now contain the user-provided value of the arg.  If the key is not present in `args`, a value of `False` is assumed in these expressions.  The symbol map passed to the expression is therefore `{key: args.get(key, False) for key in spec}`.

Even though we have been assuming so far that keys in expressions represent sufficiency (whether a key has a non-empty value in args), sufficiency is really just the truthy evaluation of the arg value ... so why not just use the arg value itself?  There was only one place in all of InVEST where the expression needed to be changed (in UCM) because of this change in meaning.

This also gets around your question about the value or sufficiency of checkboxes.  By switching to using the values of the checkboxes, we're being more explicit about just using the value of the input (whatever type that is), and allowing the programmer to decide how to handle its type in an expression that evaluates to a `bool`.

## Expression Results are Cast to `bool`

Now that expressions use arg values, the result of the `eval`'d expression is now explicitly cast to `bool`.  This allows us to ensure that we have only 2 possible states for conditional validity: valid and invalid.  This works because a string with nonzero length is interpreted as `True` in a truthy context.

## Switching to Truthiness = No Need to `tokenize`

Switching to truthiness means that there's no need to do any special parsing or mangling of variable names - we've simplified the meaning of the expression to be closer to how python handles expressions.

src/natcap/invest/urban_nature_access.py
phargogh: Good point about not needing this conditional requirement in this parameter because the whole table is required!  I have removed it, and I also implemented conditional requirement for vector fields and CSV rows/columns.

src/natcap/invest/urban_nature_access.py
phargogh: I implemented conditional validation by rewriting the `required` attribute for vectors and CSVs to a boolean, which meant that this function doesn't have to change at all.  The rewriting happens within `validate()`, where we have easy access to the args themselves.  I also removed this comment about conditional requirement in cb70a12c7.

src/natcap/invest/urban_nature_access.py
phargogh: It works now!  As with vectors, at this point in the validity evaluation (at least when calling `validate`) the `required` attribute will be either True or False.

Since conditional requirement could be based on the value of other args, I didn't implement anything here in `get_validated_dataframe` because I believe we can now assume that conditional requirement has already been evaluated by a call to `validate`.  Maybe that's incorrect though ... what do you think?

src/natcap/invest/validation.py
```diff
@@ -930,65 +929,48 @@ def validate(args, spec, spatial_overlap_opts=None):
...
-    # it has a value.  A sufficient input need not be valid.  Sufficiency is
-    # used by the conditional requirement phase (step 3 in this function) to
-    # determine whether a conditionally required input is required.
-    # The only special case about sufficiency is with boolean values.
-    # A boolean value absent from args is insufficient.  A boolean input that
-    # is present in args but False is in sufficient.  A boolean input that is
-    # present in args and True is sufficient.
+    # has a truthy value.  If the input is missing from args or is falsy, it is
+    # insufficient.
     insufficient_keys = missing_keys.union(keys_with_no_value)
```
emlys, line 943: Maybe no longer needed? This is overwritten two lines down

src/natcap/invest/validation.py
```diff
@@ -930,65 +929,48 @@ def validate(args, spec, spatial_overlap_opts=None):
...
-
-        # Any other input type must be sufficient because it is in args and
-        # has a value.
-        else:
-            sufficient_inputs[key] = True
 
-    # step 3: evaluate required status of conditionally required keys
-    # keep track of keys that are explicity not required due to
-    # their condition being false
+    # Step 3: evaluate whether conditionally required keys are required.
```
emlys, line 956: Would it simplify things to combine the non-conditional requirement check (what's currently labeled "step 1" above) into this? Since they produce the same validation warnings

src/natcap/invest/validation.py
```diff
@@ -1005,11 +987,30 @@ def validate(args, spec, spatial_overlap_opts=None):
...
             LOGGER.debug(f'Provided key {key} does not exist in MODEL_SPEC')
             continue
 
+        # rewrite parameter_spec for any nested, conditional validity
+        axis_keys = None
+        if parameter_spec['type'] == 'csv':
+            axis_keys = ['columns', 'rows']
+        elif parameter_spec['type'] == 'vector':
+            axis_keys = ['fields']
+
```
emlys, line 1002: Could we also include `'directory'` here with its key `'contents'`? This isn't used much, but would be nice to have a comprehensive solution

src/natcap/invest/validation.py
```diff
@@ -1005,11 +987,30 @@ def validate(args, spec, spatial_overlap_opts=None):
...
+        if axis_keys:
+            for axis_key in axis_keys:
+                if axis_key not in parameter_spec:
+                    continue
+                for nested_key, nested_spec in parameter_spec[axis_key].items():
+                    if ('required' in nested_spec
+                            and isinstance(nested_spec['required'], str)):
+                        parameter_spec[axis_key][nested_key]['required'] = (
```
emlys, line 1010: Neat!

src/natcap/invest/validation.py
```diff
@@ -930,65 +929,48 @@ def validate(args, spec, spatial_overlap_opts=None):
...
-    # used by the conditional requirement phase (step 3 in this function) to
-    # determine whether a conditionally required input is required.
-    # The only special case about sufficiency is with boolean values.
-    # A boolean value absent from args is insufficient.  A boolean input that
-    # is present in args but False is in sufficient.  A boolean input that is
-    # present in args and True is sufficient.
+    # has a truthy value.  If the input is missing from args or is falsy, it is
+    # insufficient.
     insufficient_keys = missing_keys.union(keys_with_no_value)
     sufficient_inputs = {}
```
emlys, line 944: It looks like `sufficient_inputs` isn't used anywhere after it's created. Would it be possible to rewrite this code block without it?

src/natcap/invest/validation.py
```diff
@@ -844,9 +846,6 @@ def get_headers_to_validate(spec):
     """
     headers = []
     for key, val in spec.items():
-        # for now only check headers that are always required
-        # assume that any conditionally-required headers are validated by the
-        # model's validate function
         # if 'required' isn't a key, it defaults to True
         if ('required' not in val) or (val['required'] is True):
```
emlys, line 850: At this point, will `required` always be a bool? If so, could simplify to
```suggestion
        if val['required']:
```

src/natcap/invest/validation.py
```diff
@@ -930,65 +929,48 @@ def validate(args, spec, spatial_overlap_opts=None):
...
+    expression_values = {
+        input_key: args.get(input_key, False) for input_key in spec.keys()}
     excluded_keys = set()
     for key in conditionally_required_keys:
-        # An input is conditionally required when the expression given
-        # evaluates to True.
         is_conditionally_required = _evaluate_expression(
-            expression=spec[key]['required'],
-            variable_map=sufficient_inputs)
+            expression=f'bool({spec[key]["required"]})',
```
emlys, line 967: This is a really minor stylistic thing, but I wonder would it make sense to cast the result of `_evaluate_expression` to `bool`, rather than putting `bool(...)` in the expression? Just to minimize the amount of code that we have to `eval`

src/natcap/invest/validation.py
```diff
@@ -930,65 +929,48 @@ def validate(args, spec, spatial_overlap_opts=None):
...
     sufficient_inputs = {}
-    for key, parameter_spec in spec.items():
-        # If the key isn't present, no need to validate.
-        # If it's required and isn't present, we wouldn't have gotten to this
-        # point in the function.
-        if key not in args:
+    insufficient_keys = set()
+    for key in spec.keys():
+        try:
+            sufficient_inputs[key] = bool(args[key])
```
emlys, line 948: Would an arg value of `0` be labeled insufficient here?

emlys (changes requested):
Thanks @phargogh! I really like how this simplifies validation. I had some questions but nothing major.

src/natcap/invest/validation.py
```diff
@@ -1,23 +1,27 @@
...
 import inspect
+import io
 import logging
 import os
 import pprint
 import queue
 import re
 import threading
+import token
+import tokenize
```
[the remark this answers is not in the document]
phargogh: True!  I removed these and did some more light linting in 482b32762

src/natcap/invest/validation.py
phargogh: Yep! Patched in b32fcbf96

src/natcap/invest/validation.py
phargogh: Dramatically!  `validate` has now been cut down to 3 distinct "phases" from the original 5.  Conditional requirement is now just a special case of requirement, as it should be.  1deee3e55

src/natcap/invest/validation.py
phargogh: Happy to!  Although it looking more closely at `check_directory`, we don't currently do anything with the `contents` key.  The main `validation` loop will probably need a bit of a refactor to support this kind of recursion.  880e31fbd

src/natcap/invest/validation.py
phargogh: Definitely, and it was addressed in one of the other changes mentioned in this latest round of revisions.

src/natcap/invest/validation.py
phargogh: Unfortunately no, we haven't fully cleaned the SPEC by this point in `validate`, so `required` could still be absent.  We have so far only handled the case where `required` is present and is a `str`, which we have rewritten as a bool.

src/natcap/invest/validation.py
phargogh: Makes sense to me!  It'll be a little less to re-`eval` on each call, so there's probably a negligible speedup this way too.  Patched (and added a separate missing cast to `bool` in d70b15dae)

src/natcap/invest/validation.py
phargogh: This specific line has been removed in a prior edit, but this is a case that the authors of expressions will need to be aware of.  One cost of this proposed simplification is that it is possible for there to be a falsey value (like `0`) which was previously acknowledged as "sufficient" to now be effectively insufficient.  This can be mitigated by better, context-aware expressions in cases where this may be an issue, so I think we'll largely be OK.  In practice, I don't think there are many places where we are likely to encounter a value of `0`.

src/natcap/invest/validation.py
```diff
@@ -905,111 +913,73 @@ def validate(args, spec, spatial_overlap_opts=None):
...
-    # We want to do validation on optional keys, like `n_workers`,
-    # but not on conditionally excluded keys, like fields that are greyed out
-    # because a checkbox is unchecked.
+    # Phase 2: Check whether any input with a value validates with its
+    # type-specific check function.
     invalid_keys = set()
-    sufficient_keys = set(args.keys()).difference(insufficient_keys)
-    for key in sufficient_keys.difference(excluded_keys):
+    insufficient_keys = (missing_keys | keys_with_no_value)
+    for key in set(args.keys()) - insufficient_keys:
```
emlys, line 953: If I'm understanding, there is a subtle change here: before, we removed `excluded_keys` from the set to validate. Now, any keys that have a value will be validated, even if they're not required. I'm not sure what the correct answer is, but it seems worth noting the change.

src/natcap/invest/urban_nature_access.py
emlys: I think that's fine for now... `get_validated_dataframe` is used in some models outside of `validate`, but then this validation should've already happened.

src/natcap/invest/validation.py
phargogh: Yes, this exclusion was definitely something I needed to address, and the fix now includes a better variable name for `excluded_keys` to clarify how they were being used.

phargogh:
Thanks @emlys !  Yes, the test failures were definitely related and I believe I have addressed everything.  Let me know what you think!

src/natcap/invest/validation.py
emlys: @phargogh might there still be a difference if a non-required key has a real value? Suppose you have a `valuation_table` that's required only if `do_valuation` is True. Then if we provide args where `do_valuation` is False and `valuation_table` is a valid path, should we still validate `valuation_table`?

phargogh: Great question, and thank you for thinking critically about this!

This specific snippet we're discussing here isn't updating for some reason, but there is now logic to cover the case where a non-required parameter has a truthy value.  So yes, in the latest revision (at least as of rev https://github.com/natcap/invest/pull/1511/commits/617e76d9bbaaa2ecc66cf041b967d7fd2748a945) I would expect the validation you describe to take place.

emlys: Ah okay, thanks for clarifying. I think that validation would _not_ take place before, because `excluded_keys` were removed from the set to validate: https://github.com/natcap/invest/blob/9a706487a1f8a8a9334435842f3c75f50fa7c590/src/natcap/invest/validation.py#L1003-L1004
Is that a change worth noting in history?

phargogh: Oh!  I see what you mean.  Sorry for my confusion about this!  Yes, this is indeed a change in behavior that I _think_ is for the best, but I will share a dev build with Stacie so that she can try it out.  In the meantime, I have noted the changes in `HISTORY`.

emlys (approved):
Awesome, thanks @phargogh!

PR #1513, opened 2024-01-25
Fix issue with over- and under-supply values missing from UNA administrative vector

### LINKED ISSUE #1512: UNA: `Pund_adm` and `Povr_adm` are both `0` when radii defined by population group
Stacie noticed the following:

> While running the sample data (3.14.0) to get results from the different options that are available, I noticed that if I run with population group radii, then only data for those groups appears in the output administrative unit table, it does not include the total population values. Which might be ok, but the UG indicates that both should be present, saying "If the user has selected to aggregate results by population group or has elected to run the model with search radii defined per population group, these additional fields will be created". Is this a bug, or expected behavior that just needs some UG rewording?

Looking at how the model is written, it appears that the implementation (and the user's guide definition) are intended to sum across all population groups when population groups are provided.  So this is a bug.

phargogh:
This PR addresses an issue with the sample data when running UNA with search radii defined per population group.  The fields `Pund_adm` and `Povr_adm`, per the User's guide, are supposed to be the sum of over- and under-supplied populations across all population groups, but these values just weren't being summed in the output vector.  This has been corrected and test values updated.

Fixes #1512

## Checklist


dcdenu4 (approved):
Looks fairly straightforward to me, thanks @phargogh !

PR #1520, opened 2024-02-06
Fix nodata value masking issue in UNA's urban nature balance

### LINKED ISSUE #1519: `inf` values ending up in UNA outputs
Stacie reported that some infinity values are showing up in the urban nature balance outputs.

Digging into it, this appears to be an issue where a specific nodata value is expected, but another is being provided.

Stacie has shared her inputs with me and I am working on it.

phargogh:
UNA was incorrectly assuming that by the time we were calculating urban nature supply, all of the nodata values in the raster would be `FLOAT32_NODATA`, which was not actually happening with Stacie's data.  This PR updates how we calculate the balance to handle any nodata values.

Fixes #1519

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -52,6 +52,9 @@ Unreleased Changes
       administrative units vector now has the correct values for these fields,
       consistent with the user's guide chapter.
       https://github.com/natcap/invest/issues/1512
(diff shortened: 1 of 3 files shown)


emlys (approved):
Makes sense!

PR #1521, opened 2024-02-06
Two fixes for SDR - convert processing stack to `long` and reduce rasterizations.

### LINKED ISSUE #1431: SDR and probably other routed models have int32 limitations
In https://github.com/natcap/pygeoprocessing/pull/351, we fixed an issue in MFD routing where rasters that had over 2^32 pixels, there would be an indexing issue and a related nodata comparison issue.

These problems with indexing also apply to all of the InVEST models that use MFD, which include SDR, NDR and SWY.  @jagoldstein encountered this issue in SDR specifically (see logfile below), but we should fix the issue for all of the routed models as well.

This is important for the GEF work.

[InVEST-natcap.invest.sdr.sdr-log-2023-10-17--14_22_25.txt](https://github.com/natcap/invest/files/13045836/InVEST-natcap.invest.sdr.sdr-log-2023-10-17--14_22_25.txt)

phargogh:
This PR comes out of work done while helping Jesse to run SDR on a 30m DEM of all of Colombia.  There are 2 main fixes:

* The processing stack is now of type `long`.  The raster has more than 2^32 pixels, so `int` was overflowing and resulting in the negative dimensions error encountered.
* I reworked the zonal statistics step in SDR to reduce the number of times rasterization happens.  `pygeoprocessing.zonal_statistics` already supports this, so all I did was reformat the function call.

Fixes #1431

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -52,6 +52,12 @@ Unreleased Changes
       administrative units vector now has the correct values for these fields,
       consistent with the user's guide chapter.
       https://github.com/natcap/invest/issues/1512
+* SDR
+    * Fixed an issue encountered in the sediment deposition function where
+      rasters with more than 2^32 pixels would raise a cryptic error relating
+      to negative dimensions. https://github.com/natcap/invest/issues/1431
+    * Optimized the creation of the summary vector by minimizing the number of
+      times the target vector needs to be rasterized.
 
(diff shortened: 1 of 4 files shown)


dcdenu4 (approved):
Thanks @phargogh , I just see a HISTORY conflict. Otherwise looks good to me!

PR #1522, opened 2024-02-07
`ManagedFlowDirRaster` cython class

emlys:
## Description
This PR adds a Cython extension class called `ManagedFlowDirRaster`, a subclass of `_ManagedRaster`. `ManagedFlowDirRaster` implements new helper functions `is_local_high_point`, `yield_upslope_neighbors`, and `yield_downslope_neighbors`. These helper functions replace some common code in SDR, NDR, and SWY where we iterate through the 8 neighbors of each pixel.

In a future PR, we'll extend `ManagedFlowDirRaster` to support D8 as well as MFD rasters.

## Checklist


davemfish (commented):
So nice to have these class methods! I left some misc comments for consideration.

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +308,98 @@ cdef class _ManagedRaster:
...
+            xi (int): x coordinate of the pixel i to consider
+            yi (int): y coordinate of the pixel i to consider
+            skip_oob (bool): if True, do not yield neighbors that fall
+                outside the raster bounds.
+
+        Yields:
+            (j, xj, yj, p_ij) tuples representing neighboring pixels, where
+            j is the index 0-7 of the neighbor,
+            xj is the x coordinate of pixel j in raster space,
+            yj is the y coordinate of pixel j in raster space, and
```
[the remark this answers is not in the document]
emlys: Sure, "pixel space" does sound better!

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +308,98 @@ cdef class _ManagedRaster:
...
+            return False
+        return True
+
+    def yield_upslope_neighbors(self, long xi, long yi):
+        """Yield upslope neighbors of a given pixel.
+
+        Args:
+            xi (int): x coordinate of the pixel i to consider
+            yi (int): y coordinate of the pixel i to consider
+            skip_oob (bool): if True, do not yield neighbors that fall
```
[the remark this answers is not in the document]
emlys: Removing this arg from the docstring

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -183,26 +166,23 @@ def ndr_eff_calculation(
...
-                        continue
-                    ds_row = row_offsets[i] + global_row
-                    if ds_row < 0 or ds_row >= n_rows:
+                has_outflow = False
+                for i, ds_col, ds_row, p_ij in (
+                        mfd_flow_direction_raster.yield_downslope_neighbors(
+                            global_col, global_row, skip_oob=False)):
+                    has_outflow = True
+                    if (ds_col < 0 or ds_col >= n_cols or
+                        ds_row < 0 or ds_row >= n_rows):
```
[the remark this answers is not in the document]
emlys: The difference is that this code is checking whether the pixel has _any_ outflow, not just outflow to an in-bounds pixel. If `has_outflow` is False, an error is raised below. 

I'm not totally sure if that's necessary - maybe that check should happen in `yield_downslope_neighbors` or somewhere else. But I was trying to preserve the original code structure as much as possible since I don't fully understand it yet

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +308,98 @@ cdef class _ManagedRaster:
...
+        cdef float flow_ij, flow_sum, p_ij
+
+        flow_dir = <int>self.get(xi, yi)
+        flow_sum = 0
+        downslope_neighbor_tuples = []
+        for n_dir in xrange(8):
+            # flows in this direction
+            xj = xi + COL_OFFSETS[n_dir]
+            yj = yi + ROW_OFFSETS[n_dir]
+            if skip_oob and (xj < 0 or xj >= self.raster_x_size or
```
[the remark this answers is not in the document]
emlys: It seems to be expected on the edge of a raster. If a pixel on the right edge of a raster has flow in the '0' direction (to the right), then it's flowing out of bounds.

phargogh (commented):
Thanks for this @emlys !  Conceptually, I think this is a great step forwards towards a unified D8/MFD routing implementation.  However, because these functions will be called many, many times when iterating over a raster, we need to be super careful about efficiency and `cython -a` is highlighting a bunch of places that can be further optimized to remove unnecessary computation.  How would you like to address those?  Would you prefer a separate PR to optimize those or to include specific suggestions in this PR?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -433,41 +350,33 @@ def route_baseflow_sum(
...
 
     current_pixel = 0
     for offset_dict in pygeoprocessing.iterblocks(
             (flow_dir_mfd_path, 1), offset_only=True, largest_block=0):
-        win_xsize = offset_dict['win_xsize']
-        win_ysize = offset_dict['win_ysize']
-        xoff = offset_dict['xoff']
-        yoff = offset_dict['yoff']
 
         # search block for a peak pixel where no other pixel drains to it.
```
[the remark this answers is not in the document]
emlys: Agreed, this must have been here for a long time. Removed it

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -136,51 +125,22 @@ def calculate_sediment_deposition(
...
+            for col_index in range(offset_dict['win_xsize']):
+                xs = offset_dict['xoff'] + col_index
 
                 # if this can be a seed pixel and hasn't already been
                 # calculated, put it on the stack
-                if seed_pixel and sediment_deposition_raster.get(
-                        seed_col, seed_row) == target_nodata:
-                    processing_stack.push(seed_row * n_cols + seed_col)
+                if (mfd_flow_direction_raster.is_local_high_point(xs, ys) and
+                        sediment_deposition_raster.get(xs, ys) == target_nodata):
```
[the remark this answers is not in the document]
emlys: Sure, I changed all of these to `numpy.isclose`

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -257,48 +212,21 @@ cpdef calculate_local_recharge(
...
                     # mfd values yet
-                    j_neighbor_end_index = 0
-                    mfd_dir_sum = 0
-                    for n_dir in xrange(8):
-                        if not upslope_defined:
+                    l_sum_avail_i = 0
+                    for _, xj, yj, p_ij in flow_raster.yield_upslope_neighbors(xi, yi):
+                        # pixel flows inward, check upslope
+                        l_sum_avail_j = target_l_sum_avail_raster.get(xj, yj)
+                        if is_close(l_sum_avail_j, target_nodata):
```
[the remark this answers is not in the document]
emlys: I removed the local `is_close` and replaced it with `numpy.isclose`

src/natcap/invest/sdr/sdr_core.pyx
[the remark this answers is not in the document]
phargogh: Uffda, that `numpy.isclose` is a python function, so it will be _super_ slow to call!  I would strongly discourage calling `numpy.isclose()` unless there's a C or C++ equivalent available. Instead, if there are different combinations of types we should be handling in our equality checks, the more direct solution might just be to adapt `_is_close()` to handle the types we're targeting.

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +310,118 @@ cdef class _ManagedRaster:
...
+            p_ij is the proportion of flow from pixel j that flows into pixel i
+        """
+        cdef int n_dir, flow_dir_j, idx
+        cdef long xj, yj
+        cdef float flow_ji, flow_dir_j_sum
+
+        cdef NeighborTuple n
+        cdef vector[NeighborTuple] upslope_neighbor_tuples
+
+        for n_dir in xrange(8):
```
dcdenu4, line 356: Is `xrange` valid in Cython still? I know it was deprecated in Python 3.

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +310,118 @@ cdef class _ManagedRaster:
...
+            ManagedFlowDirRaster self, long xi, long yi, bint skip_oob=True):
+        """Yield downslope neighbors of a given pixel.
+
+        Args:
+            xi (int): x coordinate of the pixel i to consider
+            yi (int): y coordinate of the pixel i to consider
+            skip_oob (bool): if True, do not yield neighbors that fall
+                outside the raster bounds.
+
+        Yields:
```
dcdenu4, line 388: Can we update this to `Returns`? It's a very nitpicky thing but I have been wondering whether it's worth changing all the `yield` terminology used to `get` or `return`, since we're no longer yielding anything. In the Python sense anyway.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -183,36 +164,33 @@ def ndr_eff_calculation(
...
-                    outflow_weight = (flow_dir >> (i*4)) & 0xF
-                    if outflow_weight == 0:
-                        continue
-                    outflow_weight_sum += outflow_weight
-                    ds_col = col_offsets[i] + global_col
-                    if ds_col < 0 or ds_col >= n_cols:
+                has_outflow = False
+                for neighbor in mfd_flow_direction_raster.yield_downslope_neighbors(
+                        global_col, global_row, skip_oob=False):
+                    has_outflow = True
```
dcdenu4, line 170: We know this to be `True` right away because `yield_downslope_neighbors` will return empty and not enter here otherwise?

dcdenu4 (changes requested):
Hey @emlys, I made a pass through and just had a couple questions and comments.

A slightly annoying nitpick which maybe isn't necessary would be to update the `yield` terminology since the functions are no longer `yielding` in the Python sense. At the very least updating the docstrings is probably appropriate. Let me know what you think!

src/natcap/invest/sdr/sdr_core.pyx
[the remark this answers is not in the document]
emlys: Good point @phargogh. When I used `numpy.isclose`, the test suite ran noticeably slower. I added a `is_close` cython function to `managed_raster.pxd` that can be used in all the other cython modules.

src/natcap/invest/ndr/ndr_core.pyx
emlys: Correct, if a pixel has no downslope neighbors (i.e. has a MFD value of 0), then an error is raised below

src/natcap/invest/managed_raster/managed_raster.pyx
emlys: Updated to `Returns`

src/natcap/invest/managed_raster/managed_raster.pyx
emlys: Cython still supports it, we use it a few times farther up in `ManagedRaster`. I updated these to `range`.

emlys:
@phargogh I fully cythonized the new functions so there is almost no more highlighting (the function signatures are still highlighted - not sure why). I timed the SDR base regression test to compare performance with these changes vs. `main` and saw no significant difference.

emlys:
@dcdenu4 I agree about getting rid of the "yield" terminology, I updated the docstrings and function names.

src/natcap/invest/managed_raster/managed_raster.pxd
```diff
@@ -37,3 +46,16 @@ cdef class _ManagedRaster:
...
+
+    cdef bint is_local_high_point(ManagedFlowDirRaster self, long xi, long yi)
+
+    cdef vector[NeighborTuple] get_upslope_neighbors(ManagedFlowDirRaster self, long xi, long yi)
+
+    cdef vector[NeighborTuple] get_downslope_neighbors(ManagedFlowDirRaster self, long xi, long yi, bint skip_oob=*)
+
+
+cdef inline int is_close(double x, double y):
+    return abs(x-y) <= (1e-8+1e-05*abs(y))
```
phargogh, line 61: Oh!  I just remembered a recent case in `pygeoprocessing` where we added in a check for whether `x` and/or `y` were `nan`s and returning an appropriate check accordingly.  Maybe we could save ourselves some future headaches by including such a check here?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -199,46 +177,22 @@ cpdef calculate_local_recharge(
 
     for offset_dict in pygeoprocessing.iterblocks(
             (flow_dir_mfd_path, 1), offset_only=True, largest_block=0):
-        win_xsize = offset_dict['win_xsize']
-        win_ysize = offset_dict['win_ysize']
-        xoff = offset_dict['xoff']
-        yoff = offset_dict['yoff']
```
phargogh, line 205: Same comment here about explicitly typing these values

src/natcap/invest/managed_raster/managed_raster.pxd
```diff
@@ -4,6 +4,15 @@
 from libcpp.list cimport list as clist
 from libcpp.pair cimport pair
 from libcpp.set cimport set as cset
+from libcpp.vector cimport vector
+
+cdef struct s_neighborTuple:
+    int direction
+    int x
+    int y
```
phargogh, line 12: I was just about to ask if we should consider making these `long`s, but for a raster to exceed 2^31 pixels in the x dimension, that would be a truly massive raster!  For a raster covering the whole globe, 2^31 pixels would mean a pixel width of 1.87cm along the equator.  So we'll probably never need to make this a `long`, and when we do, we'll have other problems of computability anyways.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -224,28 +202,23 @@ def ndr_eff_calculation(
...
-                neighbor_col = col_offsets[i] + global_col
-                if neighbor_col < 0 or neighbor_col >= n_cols:
-                        continue
-                neighbor_row = row_offsets[i] + global_row
-                if neighbor_row < 0 or neighbor_row >= n_rows:
-                    continue
-                neighbor_outflow_dir = inflow_offsets[i]
+            # for i in range(8):
+            for neighbor in mfd_flow_direction_raster.get_upslope_neighbors(
+                    global_col, global_row):
```
phargogh, line 216: Will this approach cause a memory leak?  Specifically: `get_upslope_neighbors` creates a new `vector` on each function call, so are we likely to get into a case where we're leaking memory on very large rasters?

From what I can tell, this is a kind of interesting case that's relatively new to the language, where the compiler should create a copy of our populated vector and return it, so the memory should be managed as stack memory in the calling function and therefore freed when the caller exits.  But this would be good to verify on a big-ish raster to verify that our memory doesn't keep steadily climbing.

phargogh (changes requested):
This is looking good!  I had a few questions and a few suggestions for some further tuning, but this is really looking like a solid solution.  Thanks!

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -300,3 +309,119 @@ cdef class _ManagedRaster:
...
+                    yj < 0 or yj >= self.raster_y_size):
+                continue
+            flow_ij = (flow_dir >> (n_dir * 4)) & 0xF
+            flow_sum += flow_ij
+            if flow_ij:
+                n.direction = n_dir
+                n.x = xj
+                n.y = yj
+                n.flow_proportion = flow_ij
+                downslope_neighbor_tuples.push_back(n)
```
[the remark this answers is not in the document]
emlys: Oh interesting. It must be creating copies somewhere, because it's returning the expected results. You're right it makes more sense to explicitly allocate the new object.

src/natcap/invest/managed_raster/managed_raster.pyx
```diff
@@ -25,6 +26,14 @@ ctypedef pair[int, double*] BlockBufferPair
...
 cdef int MANAGED_RASTER_N_BLOCKS = 2**6
 
+# These offsets are for the neighbor rows and columns according to the
+# ordering: 3 2 1
+#           4 x 0
+#           5 6 7
+cdef int *ROW_OFFSETS = [0, -1, -1, -1,  0,  1, 1, 1]
+cdef int *COL_OFFSETS = [1,  1,  0, -1, -1, -1, 0, 1]
+cdef int* FLOW_DIR_REVERSE_DIRECTION = [4, 5, 6, 7, 0, 1, 2, 3]
+
```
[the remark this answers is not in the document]
emlys: Sure! I don't think they're needed elsewhere at the moment, but could be in the future.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -115,49 +115,30 @@ def ndr_eff_calculation(
...
-        xoff = offset_dict['xoff']
-        yoff = offset_dict['yoff']
-        for row_index in range(win_ysize):
-            global_row = yoff + row_index
-            for col_index in range(win_xsize):
-                global_col = xoff + col_index
+        for row_index in range(offset_dict['win_ysize']):
+            global_row = offset_dict['yoff'] + row_index
+            for col_index in range(offset_dict['win_xsize']):
+                global_col = offset_dict['xoff'] + col_index
```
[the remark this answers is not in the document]
emlys: Sure, I put them back and added a comment explaining it, since usually we avoid creating new variables for dictionary values like that. Unfortunately, casting to int like `<int>offset_dict['win_xsize']` doesn't seem to remove the python overhead.

src/natcap/invest/managed_raster/managed_raster.pxd
emlys: That would be conveninent, but do you think the additional overhead would be a problem? I assume that calling `math.isnan` or `numpy.isnan` would cause a slowdown

phargogh: You're right, they would!  But there should also be a `cimport`able `math` module with an `isnan` function, which should make the addition negligibly slower.  Looks like we added this over in pygeoprocessing's routing, so I bet we might be able to just use the same solution here: https://github.com/natcap/pygeoprocessing/blob/254c7c8fbbf5220e3d28d69e8cdf945fc8a6c3a9/src/pygeoprocessing/routing/routing.pyx#L197

src/natcap/invest/ndr/ndr_core.pyx
emlys: I think you are right that a copy is made. The docs [say](https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html) that C++ objects can be stack allocated by using the `cdef` syntax, and then [exist only for the duration of the function](https://cython.readthedocs.io/en/latest/src/userguide/glossary.html#term-Stack-allocation). I tried to test this by comparing the memory address of the vector before and after it's returned (`print(<unsigned long>&upslope_neighbor_tuples)`) and they are indeed different addresses. But interestingly, the same address is reused in subsequent calls of `get_upslope_neighbors` - which I assume means it's being freed after each call, and then just happens to be re-allocated on the next call.

I don't have a very large MFD raster on hand to test with, but could run RouteDEM on the global DEM if you think it's worth double checking.

src/natcap/invest/managed_raster/managed_raster.pxd
emlys: Oh neat! I added that check to the `is_close` in `managed_raster.pxd` for now, and eventually when we consolidate things into pygeoprocessing, those functions will probably be combined.

src/natcap/invest/ndr/ndr_core.pyx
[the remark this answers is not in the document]
phargogh: >Unfortunately, casting to int like `<int>offset_dict['win_xsize']` doesn't seem to remove the python overhead.

Yeah, that makes sense ... I am not aware of a way to statically type dictionary accesses, since the values could be any type.

src/natcap/invest/managed_raster/managed_raster.pxd
phargogh: Awesome, thanks!

src/natcap/invest/ndr/ndr_core.pyx
phargogh: I'm satisfied with this as-is!  Looking into this a little bit more, I think we will be OK.

Although the reference to the `vector` is on the stack, the vector's contents are allocated on the heap since it manages its own contents and how much memory has been allocated for it.  This means that when we return a `vector` from a function, its contents aren't necessarily erased until they are explicitly `free`d.  And then it also seems like the compiler is using some sort of [copy elision](https://en.wikipedia.org/wiki/Copy_elision) to avoid actually re-copying the memory, so my educated guess at what's causing the reference behavior you mention is that the compiler is just re-configuring the vector itself to use the same contents that are on the heap.  The fact that we're re-using the same reference makes me feel confident that the compiler is helping us out rather than creating a new object on each iteration.

phargogh (approved):
Thanks @emlys !

davemfish (approved):

PR #1524, opened 2024-02-13
restrict scipy version to one that does not use `packaging.tags`

### LINKED ISSUE #1508: Error in "build binaries" job on macOS
"Build binaries" job is failing with this error when it tries to run `invest list` on the built binaries.
```
subprocess.CalledProcessError: Command '['/Users/runner/work/invest/invest/dist/invest/invest', '-sS', '-c', 'import platform; print(platform.mac_ver()[0])']' returned non-zero exit status 2.
```

davemfish:
I'm not sure if we want to do this, but it does seem to allow our mac binary builds to work again. The change in scipy is probably this one in `_testutils.py`:

https://github.com/scipy/scipy/commit/96081aa10e929e807b199f8e441dd1693e304f03

Fixes #1508 

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-GDAL>=3.4.2,<3.6.0
+GDAL>=3.4.2
 Pyro4==4.77  # pip-only
 pandas>=1.2.1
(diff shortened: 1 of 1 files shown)


dcdenu4 (approved):
I think this is okay to do and can't see it having a huge impact on anyone. Where as, it fixes a 3rd party bug for us that allows us to have reliable builds.

PR #1525, opened 2024-02-19
AWY: add results suffix to a few files

### LINKED ISSUE #1517: AWY model file suffix is missing from a few intermediate files
`kc_raster.tif`, `root_depth.tif` and `veg.tif` in the `intermediate` dir do not get the results suffix.

In addition, the `MODEL_SPEC` lists the `intermediate` dir as nested without an `output` dir, when they are actually at the same level.

davemfish:
A few intermediate file outputs were missing the suffix. Based on the variable names, perhaps these files used to be temporary and that's why they didn't get the suffix. But they are no longer temporary. 

I also corrected a couple minor bugs in the output spec for this model (`intermediate` folder not nested within `output`) and in the carbon model.

Fixes #1517 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -38,6 +38,9 @@
 
 Unreleased Changes
 ------------------
+* Annual Water Yield
+    * Added the results_suffix to a few intermediate files where it was
+      missing. https://github.com/natcap/invest/issues/1517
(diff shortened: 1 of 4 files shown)


emlys (approved):
Great, thanks @davemfish!

PR #1530, opened 2024-02-20
Bugfix CV: mask out nan nodata

### LINKED ISSUE #1528: CV does not handle `nan` in bathymetry
A raster with `nan` as `nodata` will yield `nan` values from `extract_bathymetry_along_ray` if any of the fetch ray overlaps nodata.

This looks like a case where we're using `math.isclose` where we should be using `numpy.isclose(x, y, equal_nan=True)`.

davemfish:
We were not properly masking out `nan` values from the bathymetry raster.

I updated an `isclose` check and updated a test to confirm that `nan` values can be properly masked like any other nodata value. Other raster ops in this model were already doing the proper checks via `pygeoprocessing.array_equals_nodata`.

Fixes #1528 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -41,6 +41,11 @@ Unreleased Changes
 * Annual Water Yield
     * Added the results_suffix to a few intermediate files where it was
       missing. https://github.com/natcap/invest/issues/1517
+* Coastal Vulnerability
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Makes sense to me, thanks @davemfish !

PR #1537, opened 2024-03-11
upgraded eslint packages to latest available

davemfish:
One of these packages was not compatible with `node v20`, now it is. I also ran the checks on node 18 before upgrading to 20. So I think we should be compatible with both now.

Some of the cascading version changes seemed to impact jest. Mocking behavior changed somehow and one test was failing, so I updated our jest config to resolve that. I assume we were somehow depending on a bug that got fixed.

## Checklist


workbench/tests/renderer/investtab.test.js
```diff
@@ -227,7 +227,7 @@ describe('Sidebar Buttons', () => {
     const response = 'saved';
     archiveDatastack.mockImplementation(() => new Promise(
       (resolve) => {
-        setTimeout(() => resolve(response), 1000);
+        setTimeout(() => resolve(response), 500);
```
davemfish, line 230: This test failed once intermittently. I think the `waitFor` at line 245 times out after 1000ms by default, and it was trying to wait for this to resolve, so it was hit/miss.

emlys (approved):
great, thanks @davemfish!

PR #1545, opened 2024-03-26
Translations update

### LINKED ISSUE #1623: Spanish and Chinese translation update
We received Spanish translations back in May / June and have gotten most translation materials from CAS.

emlys:
## Description
Fixes #1623

## Checklist


workbench/readme.md
```diff
@@ -121,18 +121,24 @@ These instructions assume you have defined the two-letter locale code in an envi
...
-   If the translator works with PO files, we can convert them to JSON using this tool: https://github.com/i18next/i18next-gettext-converter
+5. Receive the updated PO files from the translator, then convert PO to JSON using [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter). Replace `src/main/i18n/$LL.json` and `src/renderer/i18n/$LL.json` with the updated versions:
+   ```
+   i18next-conv -l $LL -s new_main_translations.po -t src/main/i18n/$LL.json
+   i18next-conv -l $LL -s new_renderer_translations.po -t src/renderer/i18n/$LL.json
 
-6. Replace `src/main/i18n/$LL.[json,po]` and `src/renderer/i18n/$LL.json` with the updated versions received from the translator
+   ```
 
 7. Commit the changes:
```
dcdenu4, line 140: It looks like step 6 was removed and that this should be the new step 6.

requirements-docs.txt
```diff
@@ -2,3 +2,4 @@ Sphinx>=1.3.1,!=1.7.1
 sphinx-rtd-theme
 sphinx-intl
 sphinx-reredirects
+pyyaml
```
dcdenu4, line 5: What relies on the `pyyaml` dependency?

workbench/src/renderer/components/DataDownloadModal/index.jsx
```diff
@@ -196,11 +228,11 @@ class DataDownloadModal extends React.Component {
                   )}
                 />
                 <Form.Check.Label>
-                  {modelName}
+                  {displayNames[modelName]}
```
dcdenu4, line 231: Nice catch to include translations for these. I'm thinking about having to remember to update this list manually if model names change or are added / removed. Since we're using `modelName` here, is it possible to somehow programmatically wrap `modelName` with the `t` designation?

dcdenu4: \`${t(modelName)}\`

dcdenu4 (commented):
Thanks @emlys , just a few quick questions and comments.

requirements-docs.txt
emlys: See https://github.com/natcap/invest.users-guide/commit/df9acf867db2882aae4c4f7c7103243160d5e4b1

workbench/src/renderer/components/DataDownloadModal/index.jsx
emlys: No, `t(...)` should always go around a string literal because it's used to extract strings for translation. If it goes around a variable, there's no way to know what strings need to be translated.

dcdenu4 (approved):

PR #1546, opened 2024-03-28
Move GDAL constraint into constraints file and update to GDAL 3.8

### LINKED ISSUE #916: [Proposal, RFC] Separate requirements from constraints
# The Problem

This came up while working on #880. 

We have long struggled with how best to define the packages that are required for InVEST's execution under various circumstances, and we have been simply defining everything in `requirements.txt`.  The result of this is that `natcap.invest` now has much stricter-than-actually-necessary requirements, which could limit someone's ability to install, say, on `debian:stable`, due to a restriction in place for a `Windows` build.

## Use Case: Installing `natcap.invest` as a library

When someone wants to install `natcap.invest` in a development environment, we should assume that the user has a deep technical understanding of their system and are able to meet the functional requirements needed in order to run `natcap.invest`.  So long as we are clear about the package versions needed to use `natcap.invest`, we should let the user make sure that they are installed correctly.

The packages required for a user on `debian:stable` should not be determined by a build issue on Windows.

## Use Case: Building InVEST binaries

When building InVEST binaries, we are at the whim of package maintainers elsewhere and their CI systems that may release buggy, ill-compiled, or incomplete binary builds.  When we discover those issues for a particular package and platform, we should be able to exclude that package for that platform, or else exclude that package version altogether, without inhibiting a technical user from being able to install whatever dependency versions they want.

## Use Case: `conda install -c conda-forge natcap.invest`

Our [`conda-forge` package's requirements definition](https://github.com/conda-forge/natcap.invest-feedstock/blob/main/recipe/meta.yaml) is a mildly-updated version of our package requirements.  Like with the `natcap.invest` library use case, we should assume that `conda-forge` is catching build issues in the CI process and not uploading faulty packages, so this requirements list should be _functional_ requirements as in `requirements.txt`. 

# Proposed Solution: add a `pip` constraints file

`pip` has 2 forms of requirement specifications:

* `requirements.txt`, which lists functional package requirements.  Packages in this list must be installed.
* `constraints.txt`, which lists any constraints on packages.  Constraints listed apply to packages being installed through `requirements.txt`.

So, InVEST's `requirements.txt` would look like:

```diff
-GDAL>=3.1.2,!=3.3.0  # 3.3.0 had a bug that broke our windows builds: https://github.com/OSGeo/gdal/issues/3898
+GDAL>=3.1.2
 Pyro4==4.77  # pip-only
 pandas>=1.2.1
-numpy>=1.11.0,!=1.16.0
-Rtree>=0.8.2,!=0.9.1
+numpy>=1.11.0
+Rtree>=0.8.2
 Shapely>=1.7.1,<2.0.0
 scipy>=1.6.0,<1.8.0  # pip-only
 pygeoprocessing>=2.3.2  # pip-only
taskgraph[niced_processes]>=0.11.0  # pip-only
psutil>=5.6.6
chardet>=3.0.4
openpyxl
pint
Babel
```

And we would add a file, `constraints-binaries.txt` that contains constraints needed for the binaries:

```diff
+# constraints-binaries.txt
+#
+# If we have any version constraints needed for our binary builds (any OS),
+# this is where they go.
+
+# 3.3.0 had a bug that broke our windows builds: https://github.com/OSGeo/gdal/issues/3898
+GDAL!=3.3.0;platform_system=='Windows'
+
+# Numpy 1.16.0 had some release bugs that were patched in 1.16.1.
+# https://numpy.org/doc/stable/release/1.16.1-notes.html#numpy-1-16-1-release-notes
+numpy!=1.16.0
+
+# We've had problems with rtree builds being broken.  This is a list of broken
+# builds that we should not build InVEST against.
+Rtree!=0.9.1
```

If one wanted to install packages through `pip`, this could be done via `pip install -r requirements.txt -c constraints-binaries.txt`.

# Implications:

## Tests

To inspire greater confidence in our binary builds, I think our build matrix should use the version constraints placed on our binary builds.  If it becomes useful later on to test other combinations of packages, we can make that call later on.

## `scripts/convert-requirements-to-conda-yml.py`

Because we use `conda` in our CI environments for building and testing, we will need to update this script to resolve our requirements and constraints as `pip` does for us if we were just using `pip`.  I have not yet found a reliable way to resolve these constraints with 3rd-party tools.  If none exist, it will fall on us to implement, test and maintain this functionality.

## `.github/actions/setup_env/action.yml`

Our custom action will need to be updated to accommodate constraint files, which are then passed to `scripts/convert-requirements-to-conda-yml.py`.

## `.github/workflows/build-and-test.yml`

The appropriate constraint files will need to be provided through this configuration.  One benefit of this constraint file approach, however, is that we have the option to define different sets of constraints for different use cases as they arise.

emlys:
## Description
Fixes #916 
Create a new file `constraints_tests.txt` that stores dependency constraints needed to run our tests, but not necessarily to use invest in general.

Exclude gdal versions 3.6.* and 3.7.* which had the bug.

Fortunately, conda seems to have no problem with multiple requirements for the same package in one environment yaml file. This makes it easy to include the constraints file in our workflow - we can pass it in to the `convert-requirements-to-conda-yml.py` script just like another requirements file. The resulting yml will contain this:
```
  - GDAL>=3.4.2
  - GDAL!=3.6.*,!=3.7.*
```

Also includes updates needed for the latest versions of GDAL:

- A validation test was broken with python 3.8 and gdal>=3.6.0. This was due to an error in the test.
- An HRA test broke with gdal>=3.6.4 (due to the underlying GEOS update). This appeared to be an error in HRA that probably didn't affect results most of the time.
- Some NDR tests broke with gdal >=3.7.0, probably related to the new `GDT_Int8` type and deprecation of the `PIXELTYPE=SIGNEDBYTE` flag (hooray!)

## Checklist

### DIFF
--- a/constraints_tests.txt
+++ b/constraints_tests.txt
@@ -1,4 +1,6 @@
-# This file contains package constraints needed to run the invest test suite
+# This file contains package constraints needed to run the invest test suite.
+# It follows the pip constraints file format:
+# https://pip.pypa.io/en/stable/user_guide/#constraints-files
 
 # A gdal bug caused our test suite to fail, but this issue is unlikely to
 # occur with regular use of invest. https://github.com/OSGeo/gdal/issues/8497
--- a/constraints_tests.txt
+++ b/constraints_tests.txt
@@ -1,0 +1,5 @@
+# This file contains package constraints needed to run the invest test suite
+
+# A gdal bug caused our test suite to fail, but this issue is unlikely to
+# occur with regular use of invest. https://github.com/OSGeo/gdal/issues/8497
+GDAL!=3.6.*,!=3.7.*
--- a/src/natcap/invest/hra.py
+++ b/src/natcap/invest/hra.py
@@ -633,10 +633,6 @@ def execute(args):
         # If the input is a vector, reproject to the AOI SRS and simplify.
         # Rasterization happens in the alignment step.
         elif gis_type == pygeoprocessing.VECTOR_TYPE:
-            # Habitats and stressors are rasterized with ALL_TOUCHED=TRUE
-            if name in habitats_info or name in stressors_info:
-                habitat_stressor_vectors.add(source_filepath)
-
             # Using Shapefile here because its driver appears to not raise a
             # warning if a MultiPolygon geometry is inserted into a Polygon
             # layer, which was happening on a real-world sample dataset while
@@ -679,6 +675,10 @@ def execute(args):
                 target_path_list=[target_simplified_vector],
                 dependent_task_list=[reprojected_vector_task]
             ))
+
+            # Habitats and stressors are rasterized with ALL_TOUCHED=TRUE
+            if name in habitats_info or name in stressors_info:
+                habitat_stressor_vectors.add(target_simplified_vector)
 
         # Later operations make use of the habitats rasters or the stressors
         # rasters, so it's useful to collect those here now.
@@ -1648,7 +1648,6 @@ def _simplify(source_vector_path, tolerance, target_vector_path,
     for source_feature in source_layer:
         target_feature = ogr.Feature(target_layer_defn)
         source_geom = source_feature.GetGeometryRef()
-
         simplified_geom = source_geom.SimplifyPreserveTopology(tolerance)
         if simplified_geom is not None:
             target_geom = simplified_geom
--- a/src/natcap/invest/ndr/ndr_core.pyx
+++ b/src/natcap/invest/ndr/ndr_core.pyx
@@ -415,9 +415,9 @@ def ndr_eff_calculation(
 
     # create direction raster in bytes
     def _mfd_to_flow_dir_op(mfd_array):
-        result = numpy.zeros(mfd_array.shape, dtype=numpy.int8)
+        result = numpy.zeros(mfd_array.shape, dtype=numpy.uint8)
         for i in range(8):
-            result[:] |= (((mfd_array >> (i*4)) & 0xF) > 0) << i
+            result[:] |= ((((mfd_array >> (i*4)) & 0xF) > 0) << i).astype(numpy.uint8)
         return result
 
     pygeoprocessing.raster_calculator(
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -2002,6 +2002,7 @@ class TestValidationFromSpec(unittest.TestCase):
         layer = vector.CreateLayer('layer', vector_srs, ogr.wkbPoint)
         new_feature = ogr.Feature(layer.GetLayerDefn())
         new_feature.SetGeometry(ogr.CreateGeometryFromWkt('POINT (1 1)'))
+        layer.CreateFeature(new_feature)
 
         new_feature = None
         layer = None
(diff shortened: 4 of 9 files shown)


src/natcap/invest/hra.py
```diff
@@ -680,6 +676,10 @@ def execute(args):
                 dependent_task_list=[reprojected_vector_task]
             ))
 
+            # Habitats and stressors are rasterized with ALL_TOUCHED=TRUE
+            if name in habitats_info or name in stressors_info:
+                habitat_stressor_vectors.add(target_simplified_vector)
```
emlys, line 681: It appears that the intent was to apply `ALL_TOUCHED=TRUE` when rasterizing habitats and stressors, but that was not happening because `target_simplified_vector`, not `source_filepath`, is used.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -415,9 +415,9 @@ def ndr_eff_calculation(
 
     # create direction raster in bytes
     def _mfd_to_flow_dir_op(mfd_array):
-        result = numpy.zeros(mfd_array.shape, dtype=numpy.int8)
+        result = numpy.zeros(mfd_array.shape, dtype=numpy.uint8)
```
emlys, line 418: I believe this always should've been `uint8`. I think `result` is supposed to be an 8-digit binary number indicating which of the 8 directions around a pixel have flow. (i.e. `10000001` means there is flow to neighbors 0 and 7). The max value `11111111` = 255, so uint8 is needed. 

I'm not sure exactly why it worked before, but it probably broke because of changes in GDAL's handling of signed 8-bit integers in 3.7.0.

tests/test_validation.py
```diff
@@ -2002,6 +2002,7 @@ def test_spatial_overlap_error(self):
         layer = vector.CreateLayer('layer', vector_srs, ogr.wkbPoint)
         new_feature = ogr.Feature(layer.GetLayerDefn())
         new_feature.SetGeometry(ogr.CreateGeometryFromWkt('POINT (1 1)'))
+        layer.CreateFeature(new_feature)
```
emlys, line 2005: The feature wasn't being added to the layer, so the layer was actually empty. I'm not sure what determines the extent of an empty layer, but I'm guessing it's undefined behavior. Some change in GDAL (or GEOS or PROJ?) caused the extent to include a `nan`.

By adding a feature to the layer, we get the expected behavior.

constraints_tests.txt
```diff
@@ -0,0 +1,5 @@
+# This file contains package constraints needed to run the invest test suite
```
phargogh, line 1: Since the concept of a constraints file is pip-specific (I think?) but we're using conda/mamba to manage our environments, would it be worth adding a note here about the syntax that is allowed in this file?

I'm guessing [environment markers](https://peps.python.org/pep-0496/) might be a big one that wouldn't be supported without further work on our requirements conversion script, for example!

phargogh (changes requested):
I just had one comment about helping future us remember what belongs in the constraints file since we are regularly using it for conda/mamba applications.  I'd be happy with a comment in the constraints file, but it could also be interesting to update the conda/mamba conversion script to handle environment markers if you thought that might be a better solution.  Totally up to you :)

Aside from that, it looks great to me, thanks @emlys!

emlys:
Thanks @phargogh, I added a note about the constraints format!

phargogh (approved):
The test failures appear to be specific to the recmodel's connection to the server, so I'm happy approving and merging this.

Issue #1549, opened 2024-04-03
SWY tests do not test raster outputs


dcdenu4:
When working on #1541 , I thought I would add a test for `QF.tif`, but realized that there was no real regression testing happening for rasters anywhere. It instead does some vector aggregate testing for regression tests. Since a lot of functions have been moved to `pgp.raster_map` or the like for simplification, this means unit tests aren't really picking up on these either. So we end up with limited ability to know when introduced changes cause a bug in the raster outputs.

SWY should have better regression tests for raster outputs.

davemfish:
Here's a few things I found while reviewing the test suite. @dcdenu4 would you mind reviewing this briefly since you have more context for how this model works and where we can benefit from more tests?

* There are a number of unit tests for the monthly quickflow calculation
* `test_local_recharge_undefined_nodata` appears to be a test without an assertion
* regression tests on the vector outputs effectively cover the `Vri.tif` and `L.tif` output rasters
* other recharge outputs of `calculate_local_recharge` are not under test (`L_avail.tif`, `L_sum.tif`).
    * Perhaps it's worth adding some assertions on existing regression tests 
    * Or add a unit test for `calculate_local_recharge`, but that requires synthesizing more input data
* baseflow output of `route_baseflow_sum` is not under test (`B.tif`, `B_sum.tif`).
    * Again, could assertions to existing regression tests, or create a unit test with synthetic data.
* `_calculate_curve_number_raster` is not unit-tested

PR #1550, opened 2024-04-03
Fix a quickflow summation bug in SWY

### LINKED ISSUE #1541: SWY QFi sum is producing coarse results with unexpected value ranges
@jagoldstein and @nadinetrahan both reported seeing this bug. Jesse believes that a bug might have been introduced in InVEST 3.8.9, as earlier InVEST results seemed okay.

Jesse:
>I noticed mine looks buggy in my latest run with v3.14.1. It doesn't match my ArcPro raster math result (summing the 12 monthly QFs... which looks good).
QF outputs from the previous times I used this model look good with earlier versions of InVEST, so I'm wondering if a bug was introduced. Otherwise, something must be off with my inputs.

>the range of values is way too high too: 5.4e7 - 4.2e5 versus the range of sums I calculated  0 - 4156

[ QFi summed via ArcGIS | QFi summed from InVEST | QFi summed from InVEST]

qfi-sum-arc-good
qfi-bug
qfi-bug-coarse

dcdenu4:
## Description
Fix a pixelwise sum op for QF. It was summing up whole blocks as opposed to doing a pixelwise sum. Remove unneeded task dependency.

I didn't add a test for this because SWY doesn't really have any regression testing for the raster outputs. It instead tests against the aggregated vector results. This seems like a larger issue which I created an issue for in #1549 .

Fixes #1541 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -68,6 +68,11 @@ Unreleased Changes
       raster. ``nan`` pixels will now be propertly ignored before calculating
       mean depths along fetch rays.
       https://github.com/natcap/invest/issues/1528
+* Seasonal Water Yield
+    * Fixed an issue where the monthly quickflow values were being summed over
+      a block area and not summed pixelwise. This caused the quickflow 
+      output ``QF.tif`` to have malformed values.
+      https://github.com/natcap/invest/issues/1541
 * Urban Nature Access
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks @dcdenu4! I must have introduced that bug when I added `raster_map`. Sorry about that!

PR #1567, opened 2024-05-07
update to Cython 3

### LINKED ISSUE #556: Update to Cython 3.0
Cython 3.0 [still has pre-release status on PyPI](https://pypi.org/project/Cython/#history), so I think this can be on hold for now. Updating will have some benefits, including:
* Use python 3 syntax by default
* Some minor bug fixes
* No more deprecated numpy API warnings in the compile logs! This will make it easier to see any actual problems.

I believe Cython 3.0 is not compatible with numpy versions pre-1.7, but that's okay because we're already using `numpy>=1.11.0`. Helpful docs: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html

emlys:
## Description
Fixes #556

## Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,7 +10,7 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-Cython<3.0.0
+Cython>=3.0.0
 virtualenv>=12.0.1
 pytest
 pytest-subtests
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,7 +10,6 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-Cython>=3.0.0
 virtualenv>=12.0.1
 pytest
 pytest-subtests
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 import platform
 import subprocess
 
-import Cython.Build
+from Cython.Build import cythonize
 import numpy
 from setuptools import setup
 from setuptools.command.build_py import build_py as _build_py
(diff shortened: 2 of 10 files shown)


setup.py
```diff
@@ -46,14 +46,14 @@ def run(self):
 
 setup(
     install_requires=_REQUIREMENTS,
-    ext_modules=[
+    ext_modules=cythonize([
```
emlys, line 49: Using `cythonize` rather than cython's `build_ext` seems to be more standard nowadays, and lets us set the language level directive.

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -129,9 +127,9 @@ cdef class _ManagedRaster:
         self.block_xbits = numpy.log2(self.block_xsize)
         self.block_ybits = numpy.log2(self.block_ysize)
         self.block_nx = (
-            self.raster_x_size + (self.block_xsize) - 1) / self.block_xsize
+            self.raster_x_size + (self.block_xsize) - 1) // self.block_xsize
```
emlys, line 130: Was relying on python 2 style integer division. Now we need to explicitly use floor division.

requirements-dev.txt
```diff
@@ -10,7 +10,6 @@
 # scripts/convert-requirements-to-conda-yml.py as though it can only be found
 # on pip.
 
-Cython<3.0.0
```
emlys, line 13: Cython doesn't really belong in requirements because it's only needed to build invest, and because it's listed under `build-system.requires` in `pyproject.toml`, pip installs it for us in a separate virtual environment at build time.

setup.py
```diff
@@ -46,14 +46,14 @@ def run(self):
...
         Extension(
             name=f'natcap.invest.{package}.{module}',
             sources=[f'src/natcap/invest/{package}/{module}.pyx'],
-            include_dirs=[numpy.get_include()],
             extra_compile_args=compiler_and_linker_args,
             extra_link_args=compiler_and_linker_args,
-            language='c++'
+            language='c++',
+            define_macros=[("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
```
emlys, line 56: This takes care of the deprecated numpy API warnings we've had forever.

dcdenu4 (approved):
Thanks @emlys , this all looks good to me.

PR #1568, opened 2024-05-08
update sample data commit hash

emlys:
## Description
For the last CBC PR, I increased the analysis year in the sample data, but I forgot to also extend the price table through that new year, which caused the binary tests to fail on main. Didn't catch it earlier because we don't run invest-autotest on PRs. Here I'm updating the sample data again to correct that.

## Checklist


phargogh (approved):
Sounds great, thanks for catching that!

PR #1573, opened 2024-05-10
Include pip packages in package_versions.txt

### LINKED ISSUE #1527: package_versions.txt is missing some included packages
This is the file we include with our binary distributions that list all the python dependencies and their versions that were included in the invest binaries. It seems to be missing some packages.

The 3.14.0 dist is missing listings for taskgraph and pygeoprocessing
The 3.14.1 dist includes taskgraph but  not pygeoprocessing.

davemfish:
`package_versions.txt` is meant to be a list of all packages & versions included in the binary builds. It's a convenient way to see which version of package was bundled with the workbench, after the workbench is installed.

We were building it with `micromamba list`, but that excludes pip installed packages (pygeoprocessing, taskgraph, among others). I did not find any way for `micromamba` to include these, so I appended the `pip list` to the `micromamba list`. There is some overlap, which I don't understand, but I think that's okay given the purpose of this file.

Fixes #1527 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,6 +251,7 @@ $(INVEST_BINARIES_DIR): | $(DIST_DIR) $(BUILD_DIR)
 	-$(RMDIR) $(INVEST_BINARIES_DIR)
 	$(PYTHON) -m PyInstaller --workpath $(BUILD_DIR)/pyi-build --clean --distpath $(DIST_DIR) exe/invest.spec
 	$(CONDA) list > $(INVEST_BINARIES_DIR)/package_versions.txt
+	$(PYTHON) -m pip list >> $(INVEST_BINARIES_DIR)/package_versions.txt
 	$(INVEST_BINARIES_DIR)/invest list
 
 # Documentation.


emlys (commented):
@davemfish this makes sense, I just had a couple of questions. Is `package_versions.txt` a standard thing? I assumed it was but I couldn't find anything about it online.

And, if the intent is to list all the package versions used in the binary builds, would it make sense to get that info from Pyinstaller somehow? Since Pyinstaller should know what is actually included in the package.

davemfish:
> @davemfish this makes sense, I just had a couple of questions. Is `package_versions.txt` a standard thing? I assumed it was but I couldn't find anything about it online.
> 
> And, if the intent is to list all the package versions used in the binary builds, would it make sense to get that info from Pyinstaller somehow? Since Pyinstaller should know what is actually included in the package.

@emlys the Pyinstaller idea is a good one, I hadn't thought of that. As far as I know `package_versions.txt` is just something we do for our own info, but I don't know the origin of it.

davemfish:
> > @davemfish this makes sense, I just had a couple of questions. Is `package_versions.txt` a standard thing? I assumed it was but I couldn't find anything about it online.
> > And, if the intent is to list all the package versions used in the binary builds, would it make sense to get that info from Pyinstaller somehow? Since Pyinstaller should know what is actually included in the package.
> 
> @emlys the Pyinstaller idea is a good one, I hadn't thought of that. As far as I know `package_versions.txt` is just something we do for our own info, but I don't know the origin of it.

It looks like we would have to add some custom code to our pyinstaller script to do this. I kinda feel like it's not worth it given how little we actually refer to `package_versions.txt`.

emlys (approved):
Ok, thanks for looking into it!

PR #1564, opened 2024-04-30
Addressing viewshed numerical error issue on M1 Macs

phargogh:
This (strangely) fixes the viewshed issue on M1 macs.  See the inline comment for a rundown.

RE:#1562

## Checklist
- [x] Updated HISTORY.rst and link to any relevant issue (if these changes are user-facing)
~~- [ ] Updated the user's guide (if needed)~~
~~- [ ] Tested the Workbench UI (if relevant)~~

Issue #1562, opened 2024-04-26
Tests failing on main


emlys:
Tests are failing on macOS, all python versions. Probably due to some dependency update.
```
FAILED tests/test_coastal_vulnerability.py::CoastalVulnerabilityTests::test_aoi_invalid_geometry - RuntimeWarning: invalid value encountered in intersection
FAILED tests/test_ndr.py::NDRTests::test_masking_invalid_geometry - AssertionError: 
Arrays are not equal

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 1
Max relative difference: inf
 x: array([[1]])
 y: array([[0]], dtype=uint8)
FAILED tests/test_scenic_quality.py::ViewshedTests::test_primitive_peak - AssertionError: 
Arrays are not equal

Mismatched elements: 4 / 64 (6.25%)
Max absolute difference: 1.
Max relative difference: 1.
 x: array([[1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1],...
 y: array([[1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1.],...
```

emlys:
This appears to be related to a change in the github actions runner. `macos-latest` is now `macos-14-arm64`. Since the difference has to do with AMR64 and/or AMR64 versions of packages, I can't reproduce it locally.

emlys:
## Scenic Quality test failure
@phargogh was able to reproduce this one locally on his M1 mac. It comes down to a difference in floating point error between Intel and M1. The relevant code is:
https://github.com/natcap/invest/blob/55804f9fff59053ed0241213c87fb811faed59fb/src/natcap/invest/scenic_quality/viewshed.pyx#L834-L841
We have a situation where `adjusted_dem_height` and `z` should be equal, but on M1, `z` has floating point error that makes it than `adjusted_dem_height`, and so the conditional evaluates differently. This is like how need `isclose` for equality comparison, we need a similar thing for greater/lesser-than-or-equal comparison. Probably worth scanning the rest of our code base for this situation.

## NDR test failure
[...]

PR #1578, opened 2024-05-13
Correcting viewshed behavior on M1 with a compiler flag

phargogh:
It turns out that the incorrect viewshed behavior patched in #1564 was because of a change in the default value of a parameter in clang-14.  Deliberately restoring the parameter value to the clang-13 default value corrects the issue.

See https://github.com/llvm/llvm-project/issues/91824 for the thread with the llvm devs.  I'm still hoping to find out why this is an issue on ARM64 in the first place and not on x86_64.

RE:#1562

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,8 @@ import os
 import platform
 import subprocess
 
-from Cython.Build import cythonize
 import numpy
+from Cython.Build import cythonize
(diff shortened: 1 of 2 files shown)


emlys (approved):
Neat!

Issue #1581, opened 2024-05-23
3.14.2 release


phargogh:
Placeholder issue for the 3.14.2 release.

There are currently issues with the conda-forge build from yesterday's pygeoprocessing release (https://github.com/conda-forge/pygeoprocessing-feedstock/pull/29), so we will need to be sure to make pygeoprocessing installed through pypi instead.

PR #1584, opened 2024-05-28
Fix sdist version information in github actions sdist builds

phargogh:
Fixes an issue with our sdist builds where the version metadata were not included, leading to an incorrect version string (e.g. `natcap_invest-0.0.post1+g6bcb9e5`.  I also modified the sdist build process to upgrade python warnings to errors, so this shouldn't happen in the future without a very loud failure.

RE: #1581

## Checklist


emlys (approved):
Thanks @phargogh!

PR #1597, opened 2024-07-17
Reimplement managed raster class in C++

emlys:
## Description
This PR reimplements the `ManagedRaster` class in C++ for better performance. Tested with `sdr.calculate_sediment_deposition` and the SDR sample data, it's now just slightly faster than the original implementation on `main` (0.59 vs 0.61 seconds), whereas the python implementation on `feature/routing-refactor` takes nearly twice as long.

I'm a beginner at C++ so I'm sure I'm not always using the right conventions or design patterns, but this seems like an okay starting point. I took some guidance from the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html). Ignoring some guidelines that don't seem to make sense in this context. For instance, I have all the code in a header file because splitting it out into `ManagedRaster.h` and `ManagedRaster.cpp` significantly slowed it down.

Happy to discuss and walk through this PR on a zoom call as well.

## Checklist


src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -68,19 +73,21 @@ def ndr_eff_calculation(
     # cell sizes must be square, so no reason to test at this point.
     cdef float cell_size = abs(stream_info['pixel_size'][0])
 
-    cdef _ManagedRaster stream_raster = _ManagedRaster(stream_path, 1, False)
-    cdef _ManagedRaster crit_len_raster = _ManagedRaster(
-        crit_len_path, 1, False)
+    cdef ManagedRaster stream_raster = ManagedRaster(
+        stream_path.encode('utf-8'), 1, False)
```
phargogh, line 77: I imagine once we have the python C interface for logging figured out that we might be able to do this string encoding as well!

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -206,26 +227,32 @@ def ndr_eff_calculation(
...
                 else:
                     raise Exception("got to a cell that has no outflow!")
             # search upslope to see if we need to push a cell on the stack
             # for i in range(8):
-            for neighbor in mfd_flow_direction_raster.get_upslope_neighbors(
-                    global_col, global_row):
+            up_iterator = UpslopeNeighborIterator(
+                mfd_flow_direction_raster, global_col, global_row)
+            neighbor = up_iterator.next()
+            while neighbor.direction < 8:
```
phargogh, line 243: I remember one of the things we talked about last week was the need to be careful to cover all of the iteration cases (e.g. `continue`, end of loop block) with calling `up_iterator.next()`.  What if we just handled that up here at the top of the loop? I suppose there's a slight cost to holding 2 neighbor objects in memory, but maybe it would be easier to maintain?  I'm thinking of something like the below (github wouldn't allow me to suggest something on the whole loop block)

```suggestion
            neighbor = up_iterator.next()
            next_neighbor = neighbor
            while neighbor.direction < 8:
                neighbor = next_neighbor
                next_neighbor = up_iterator.next()
```

src/natcap/invest/sdr/sdr_core.pyx
```diff
@@ -154,28 +159,35 @@ def calculate_sediment_deposition(
                     # the weighted sum of flux flowing onto this pixel from
                     # all neighbors
                     f_j_weighted_sum = 0
-                    for neighbor in (
-                            mfd_flow_direction_raster.get_upslope_neighbors(
-                                global_col, global_row)):
+                    up_iterator = UpslopeNeighborIterator(
+                        mfd_flow_direction_raster, global_col, global_row)
+                    neighbor = up_iterator.next()
+                    while neighbor.direction < 8:
```
phargogh, line 165: Same question here about whether it'd make sense to track a neighbor/next neighbor object in the loop to make it easier on us to maintain and avoid infinite loops.

src/natcap/invest/sdr/sdr_core_main.pyx
phargogh, line 1: I remember we talked about this file, but I forget if it's intended to continue to be in the PR.  Should this file be removed as a part of the PR or should it be retained for now?

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
```diff
@@ -101,76 +105,77 @@ cpdef calculate_local_recharge(
...
     # set to -1 if not defined
     # precipitation and evapotranspiration data should
     # always be non-negative
-    et0_m_raster_list = []
+    cdef vector[ManagedRaster] et0_m_rasters
     et0_m_nodata_list = []
-    for et0_path in et0_path_list:
-        et0_m_raster_list.append(_ManagedRaster(et0_path, 1, 0))
+    for i, et0_path in enumerate(et0_path_list):
+        et0_m_rasters.push_back(ManagedRaster(et0_path.encode('utf-8'), 1, 0))
```
phargogh, line 120: Is the enumeration still needed?  It looks like there are a few more enumerations in the next 30 lines or so

phargogh (commented):
This is looking really great @emlys !  I had a few questions, but I think they're all pretty minor.  Happy to talk through anything that would be helpful!

src/natcap/invest/sdr/sdr_core_main.pyx
emlys: Oops, it's meant to be removed!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield_core.pyx
emlys: Not needed!

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,607 @@
...
+                    band->GetActualBlockSize(block_xi, block_yi, &actual_x, &actual_y);
+                    actualBlockWidths[block_yi * block_nx + block_xi] = actual_x;
+                }
+            }
+
+            this->lru_cache = new LRUCache<int, double*>(MANAGED_RASTER_N_BLOCKS);
+            closed = 0;
+        }
+
+        void set(long xi, long yi, double value) {
```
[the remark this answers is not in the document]
emlys: I'm not seeing a measurable speedup on the SDR sample data. As a guideline, the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Inline_Functions) recommends against inlining functions over about 10 lines.

-> Caveat that I'm not at all confident that my speed testing will detect relatively small changes.. I get weird results sometimes.

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,607 @@
...
+            if (write_mode) {
+
+                std::set<int>::iterator dirty_itr = dirty_blocks.find(block_index);
+                if (dirty_itr == dirty_blocks.end()) {
+                    dirty_blocks.insert(block_index);
+                }
+            }
+        }
+
+        double get(long xi, long yi) {
```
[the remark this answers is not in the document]
emlys: See comment above

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,607 @@
...
+                    if (xoff + win_xsize > raster_x_size) {
+                        win_xsize = win_xsize - (xoff + win_xsize - raster_x_size);
+                    }
+                    if (yoff + win_ysize > raster_y_size) {
+                        win_ysize = win_ysize - (yoff + win_ysize - raster_y_size);
+                    }
+                    CPLErr err = band->RasterIO( GF_Write, xoff, yoff, win_xsize, win_ysize,
+                        double_buffer, win_xsize, win_ysize, GDT_Float64, 0, 0 );
+                    if (err != CE_None) {
+                        std::cout << "Error writing block\n";
```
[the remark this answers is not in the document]
emlys: Good point, the error stream makes more sense!

setup.py
```diff
@@ -20,8 +20,15 @@
...
+include_dirs = [numpy.get_include(), 'src/natcap/invest/managed_raster']
 if platform.system() == 'Darwin':
-    compiler_and_linker_args = ['-stdlib=libc++']
+    compiler_args = []
+    compiler_and_linker_args = ['-stdlib=libc++', '-std=c++20']
+    library_dirs = [f'{os.environ["CONDA_PREFIX"]}/lib']
+else:
+    compiler_args = ['/std:c++20']
+    library_dirs = [f'{os.environ["CONDA_PREFIX"]}/Library/lib']
+    include_dirs.append(f'{os.environ["CONDA_PREFIX"]}/Library/include')
```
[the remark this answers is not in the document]
emlys: Have we been supporting Linux compilation previously? I think this would not work, but if it didn't work before, maybe that's a separate question.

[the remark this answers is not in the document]
phargogh: Yeah, it's never been a priority in this repository, but we do offer linux builds via conda-forge so I don't think we should _completely_ ignore the use case.  Linux support has historically been kind of informal in this repo.

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,607 @@
...
+        this->x = x;
+        this->y = y;
+        this->flow_proportion = flow_proportion;
+    }
+
+    ~NeighborTuple () {}
+};
+
+
+class ManagedRaster {
```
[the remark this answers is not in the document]
emlys: I ran into issues with implementing a destructor because it seems to get called before I'd expect (such as when the cython object is created and assigned to a variable, it's actually copied and the original destroyed), possibly due to Cython weirdness? So I've put all the clean-up into `close` instead, so we can control when it's called. The default destructor should take care of everything else. (I might be missing something here, though!)

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,607 @@
...
+            n_dir += 1;
+            return n;
+        } else {
+            n_dir += 1;
+            return next_skip(skip);
+        }
+    }
+};
+
+bool is_close(double x, double y) {
```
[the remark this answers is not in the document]
emlys: Inlined it!

setup.py
```diff
@@ -50,13 +57,14 @@ def run(self):
...
             name=f'natcap.invest.{package}.{module}',
             sources=[f'src/natcap/invest/{package}/{module}.pyx'],
-            include_dirs=[numpy.get_include()] + ['src/natcap/invest/managed_raster'],
-            extra_compile_args=compiler_args + compiler_and_linker_args,
+            include_dirs=include_dirs,
+            extra_compile_args=compiler_args + package_compiler_args + compiler_and_linker_args,
             extra_link_args=compiler_and_linker_args,
             language='c++',
+            libraries=['gdal'],
+            library_dirs=library_dirs,
```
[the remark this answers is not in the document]
emlys: Oh neat! Thanks for pointing that out. I'm not sure what the solution would be for Windows, to avoid relying on conda - possibly we'd need to make the GDAL lib location configurable as an env variable

src/natcap/invest/ndr/ndr_core.pyx
emlys: @phargogh I think I've found an even better solution which is to properly implement the iterators to meet the C++ iterator specification. That way we don't need to manually call `next()` at all. I was stumped on that before but I think I've worked it out.

phargogh: Woohoo!  That is super exciting.  Well done and I can't wait to see the iterator!

emlys:
@phargogh this is finally ready for re-review! The biggest change is I figured out how to implement the iterators properly, so they can be iterated with the usual syntax, solving the issues with tracking/incrementing values properly.

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -0,0 +1,652 @@
...
+
+public:
+
+    ManagedFlowDirRaster() {}
+
+    ManagedFlowDirRaster(char* raster_path, int band_id, bool write_mode)
+        : ManagedRaster(raster_path, band_id, write_mode)   // Call the superclass constructor in the subclass' initialization list.
+        {
+            // do something with bar
+        }
```
phargogh, line 350: Is the plan to still do something in this constructor?

phargogh (commented):
Thanks @emlys , this looks good to me!  I had a couple of minor comments, and then I also wanted to confirm: what we have here in this PR looks to be intended for MFD routing, specifically, so is D8 routing to be added in a future PR?

src/natcap/invest/managed_raster/ManagedRaster.h
emlys: Fixed in 25c954af

setup.py
```diff
@@ -20,9 +20,18 @@
...
-
+include_dirs = [numpy.get_include(), 'src/natcap/invest/managed_raster']
+if platform.system() == 'Windows':
+    compiler_args = ['/std:c++20']
+    library_dirs = [f'{os.environ["CONDA_PREFIX"]}/Library/lib']
+    include_dirs.append(f'{os.environ["CONDA_PREFIX"]}/Library/include')
+else:
+    library_dirs = []
+    compiler_args = []
+    compiler_and_linker_args = ['-stdlib=libc++', '-std=c++20']
```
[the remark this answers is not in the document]
emlys: Good question, I think I used some C++20 features during development, but now only need C++17. (It still compiles with 11, but I get warnings that the inline variables are a C++17 feature.) Updated in d50a2c9ee

setup.py
```diff
@@ -20,9 +20,18 @@
...
+    library_dirs = [subprocess.run(
+        ['gdal-config', '--libs'], capture_output=True, text=True
+    ).stdout.split()[0][2:]]  # get the first argument which is the library path
```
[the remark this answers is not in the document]
emlys: Hmmm... the libraries are a dependency of the build, but the `gdal` python bindings themselves aren't necessary. I'm not sure there's any standardized way to document that kind of dependency?

emlys:
@phargogh yes, the D8 feature is over in #1633. I also got the RTD build working here, and added an env variable to configure with GDAL lib path for windows builds.

.github/workflows/build-and-test.yml
```diff
@@ -119,7 +119,7 @@ jobs:
         run: |
           # uninstall InVEST if it was already in the restored cache
           python -m pip uninstall -y natcap.invest
-          python -m build --wheel
+          NATCAP_INVEST_GDAL_LIB_PATH="$CONDA_PREFIX/Library" python -m build --wheel
```
phargogh, line 122: I see we're repeating this new environment variable in a couple places, so would it interfere with anything to move it to the `env` block so we only need to define it once?  I'm pretty sure we can to a top-level `env:` mapping to allow the env var to propagate across all of the different sections.

setup.py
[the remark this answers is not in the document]
phargogh: Sounds good!  Thanks for looking into it.

If I had to guess, maybe the lack of an `--ffp-contract=off` flag (needed for addressing a compilation bug in viewshed) is causing the rest of the errors?  If (big if) that's the only thing standing in the way, then the viewshed tests would fail, which isn't worth it if all we need is a recent enough version of the C++ standard.

.github/workflows/build-and-test.yml
phargogh: Nope, I see now that we won't know what `$CONDA_PREFIX` is going to be until after we're in the activated environment, so this seems like the most direct way to define it.  👍

phargogh (approved):
Thanks so much @emlys it's looking good!

PR #1604, opened 2024-07-31
Distribute mamba with workbench

emlys:
PR to merge the mamba distribution work into the plugins feature branch.

## Description
Fixes #

## Checklist


workbench/src/main/findBinaries.js
```diff
@@ -51,3 +53,37 @@ export default function findInvestBinaries(isDevMode) {
...
+
+
+/**
+ * Return the available mamba executable.
+ *
+ * @param {boolean} isDevMode - a boolean designating dev mode or not.
+ * @returns {string} mamba executable.
+ */
+export function findMambaExecutable(isDevMode) {
+  // Binding to the invest server binary:
```
dcdenu4, line 65: Copied comment?

workbench/src/main/findBinaries.js
```diff
@@ -51,3 +53,37 @@ export default function findInvestBinaries(isDevMode) {
...
+  let mambaExe;
+  const ext = (process.platform === 'win32') ? '.exe' : '';
+  const filename = `invest${ext}`;
```
dcdenu4, line 68: It looks like these are leftover and not used.

workbench/src/main/main.js
```diff
@@ -80,6 +80,7 @@ export const createWindow = async () => {
 
   const investExe = findInvestBinaries(ELECTRON_DEV_MODE);
   settingsStore.set('investExe', investExe);
+  settingsStore.set('mamba', findMambaExecutable(ELECTRON_DEV_MODE));
```
dcdenu4, line 83: This is nitpicky but do we want to be consistent with whether we create a variable for a returned binary path or just call it from within the function signature?

dcdenu4 (changes requested):
Thanks @emlys and sorry for being so slow. I had a few comments and questions but nothing major.

You mentioned commenting out the puppeteer test, I don't see that in here. Was that testing pulled out of this PR or handled elsewhere?

emlys:
@dcdenu4 I fixed those small issues. I had found that the puppeteer test failure was not specific to this branch, but was also happening on `feature/plugins`, so I directly commented it out on that branch. So that change isn't in this PR.

dcdenu4 (approved):
Thanks @emlys , I think this is good to merge!

PR #1614, opened 2024-08-15
Workbench Form Validation UI Updates

### LINKED ISSUE #912: minor improvements to arg form validation UI
Just some thoughts on improving how this looks:

image

- The first message should say "write access", not "w access"
- The message for a directory should say "access to this directory", not "this file"
- The input type should not be repeated in the validation message. Instead of "directory: Input is required but has no value", just say "Input is required but has no value"
- The message should be left-justified in line with the left edge of the input box
- The input label should be vertically centered relative to the input box, not including the validation message. As it currently is, when a validation message appears, the input label moves down to re-center itself relative to the row. That's a little visually distracting.
- It would look more consistent if the dropdown box was outlined in green, indicating that whatever selected option is valid.
- The dropdown arrow V is a little too close to the right edge of the box (not consistent with the other margins)
- It could be nice to use [switches](https://react-bootstrap.github.io/forms/checks-radios/#switches) instead of yes/no radio buttons for boolean inputs.

emilyanndavis:
## Description
Fixes #912.

Specifically:
- Updates permission error messages to distinguish between a directory vs. a file and include verbose permission type (e.g., "You must have write access to this directory" instead of "You must have w access to this file")
- Removes input type (directory, raster, csv, etc.) from error message UI and adds it to the corresponding label instead, to ensure this helpful info persists in the UI regardless of interaction state
- Adds aria-describedby to input fields to associate them with their error messages (for screen reader support)
- Adds custom styling to `<select>` (dropdown) form fields—including green border when valid—for more consistency with appearance of other form fields
- Top-aligns labels to prevent vertical shift when an error message appears
- Aligns start of error message with the left edge of its form field (instead of with the info button)

## Checklist
- [n/a] Updated the user's guide (if needed)

## Representative screenshots
workbench-form-validation-ui_2024-08-15

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -37,7 +37,7 @@
 
 Unreleased Changes
 ------------------
-* Workbench  
+* Workbench
   * Several small updates to the model input form UI to improve usability and visual consistency (https://github.com/natcap/invest/issues/912)
 
 3.14.2 (2024-05-29)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -35,10 +35,11 @@
 
 .. :changelog:
 
+Unreleased Changes
+------------------
+* Workbench  
+  * Several small updates to the model input form UI to improve usability and visual consistency (https://github.com/natcap/invest/issues/912)
 
-..
-  Unreleased Changes
-  ------------------
 
 3.14.2 (2024-05-29)
 -------------------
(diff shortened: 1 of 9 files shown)


davemfish (approved):
Thanks, @emilyanndavis , this is all very thorough and clear! These will be nice enhancements to the UI.

Issue #1599, opened 2024-07-18
Validation is allowing read-only directories as workspaces


davemfish:
I just tried this directory: `C:\\Program Files\\InVEST 3.14.1 Workbench\\runs` via the workbench. There was no validation warning, so this is what happened when I clicked run:

```python
[1]   File "C:\Users\dmf\projects\invest\env\lib\site-packages\natcap\invest\utils.py", line 145, in prepare_workspace
[1]     os.makedirs(workspace)
[1]   File "C:\Users\dmf\projects\invest\env\lib\os.py", line 215, in makedirs
[1]
[1] [09:25:33.665] [undefined]     makedirs(head, exist_ok=exist_ok)
[1]   File "C:\Users\dmf\projects\invest\env\lib\os.py", line 225, in makedirs
[1]
[1] [09:25:33.666] [undefined]     mkdir(name, mode)
[1] PermissionError: [WinError 5] Access is denied: 'C:\\Program Files\\InVEST 3.14.1 Workbench\\runs'
```

This is especially important because a relative path entered in the Workbench UI will be relative to the installation path (`C:\\Program Files\\InVEST 3.14.1 Workbench\\`) where the user may not have write permission.

emlys:
It looks like the necessary validation code is already there to perform this check. Might just need to find where we call `check_directory` on the workspace and set `permissions='rwx'`.

emilyanndavis:
This seems to be a Windows-specific issue, since workspace permissions validation works as expected on macOS.

The cause appears to be a limitation of Python's `os.access` method. From the docs: 

> Note: I/O operations may fail even when [access()](https://docs.python.org/3/library/os.html#os.access) indicates that they would succeed, particularly for operations on network filesystems which may have permissions semantics beyond the usual POSIX permission-bit model.

A more reliable approach may well be to attempt the operation (e.g., write to the workspace directory) and throw an error if it fails, rather than try to determine in advance whether the operation is allowed.

PR #1618, opened 2024-08-21
Allow 'Open Workspace' to open workspace whenever possible, even if logfile does not exist

### LINKED ISSUE #1598: Open Workspace button is enabled even when there is no workspace
If the invest run crashes before creating the workspace, then clicking the "Open Workspace" button will crash the Workbench.

This is a rare edge case, but can be reproduced by an error during python imports, for example a `ModuleNotFoundError` . Or by the case described in #1599

emilyanndavis:
## Description
Fixes #1598, with a solution that is more forgiving (and perhaps more user-friendly) than the one in [PR #1617](https://github.com/natcap/invest/pull/1617).

## Checklist
- [n/a] Updated the user's guide (if needed)

### Open workspace with valid logfile (logfile is no longer selected, for the sake of simplicity)
open-workspace_valid-logfile_not-highlighted_2024-08-29

### Open workspace with missing logfile (in this case, it was manually deleted after a successful run)
open-workspace_missing-logfile_2024-08-23

### Open empty workspace (this might happen if, for example, the workspace is readonly—see #1599)
open-workspace_empty-workspace_2024-08-23

### Error modal when workspace cannot be opened (for example, if it does not exist, as would be the case if the workspace path specified a not-yet-extant subdirectory of a readonly directory—see #1599)
open-workspace_nonexistent-workspace_2024-08-23

### DIFF
--- a/workbench/src/renderer/components/InvestTab/index.jsx
+++ b/workbench/src/renderer/components/InvestTab/index.jsx
@@ -50,10 +50,12 @@ function handleOpenWorkspace(logfile, workspace_dir) {
     logger.info(`Attempting to show logfile in workspace directory: ${logfile}`);
     ipcRenderer.send(ipcMainChannels.SHOW_ITEM_IN_FOLDER, logfile);
   }
-  // Always call shell.openPath.
-  // If shell.showItemInFolder failed, it will have failed silently;
-  // if it succeeded, the subsequent call to shell.openPath will not "undo" the result of shell.showItemInFolder.
-  openWorkspaceDir(workspace_dir);
+  if (workspace_dir) {
+    // Always call shell.openPath.
+    // If shell.showItemInFolder failed, it will have failed silently;
+    // if it succeeded, the subsequent call to shell.openPath will not "undo" the result of shell.showItemInFolder.
+    openWorkspaceDir(workspace_dir);
+  }
 }
 
 async function openWorkspaceDir(workspace_dir) {
@@ -220,7 +222,6 @@ class InvestTab extends React.Component {
       argsValues,
       logfile,
     } = this.props.job;
-    const {workspace_dir} = argsValues;
 
     const { tabID, t } = this.props;
 
@@ -274,7 +275,7 @@ class InvestTab extends React.Component {
                   ? (
                     <ModelStatusAlert
                       status={status}
-                      handleOpenWorkspace={() => handleOpenWorkspace(logfile, workspace_dir)}
+                      handleOpenWorkspace={() => handleOpenWorkspace(logfile, argsValues?.workspace_dir)}
                       terminateInvestProcess={this.terminateInvestProcess}
                     />
                   )
--- a/workbench/src/renderer/components/InvestTab/index.jsx
+++ b/workbench/src/renderer/components/InvestTab/index.jsx
@@ -71,7 +71,6 @@ class InvestTab extends React.Component {
     this.investLogfileCallback = this.investLogfileCallback.bind(this);
     this.investExitCallback = this.investExitCallback.bind(this);
     this.handleOpenWorkspace = this.handleOpenWorkspace.bind(this);
-    this.openWorkspaceDir = this.openWorkspaceDir.bind(this);
     this.showErrorModal = this.showErrorModal.bind(this);
   }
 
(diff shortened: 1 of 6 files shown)


workbench/tests/renderer/investtab.test.js
```diff
@@ -117,10 +117,101 @@ describe('Run status Alert renders with status from a recent run', () => {
...
+    const { findByRole } = renderInvestTab(job);
+    const openWorkspaceBtn = await findByRole('button', { name: 'Open Workspace' });
+    openWorkspaceBtn.click();
+
+    expect(shell.showItemInFolder).not.toHaveBeenCalled();
+    expect(ipcRenderer.invoke).toHaveBeenCalledTimes(1);
+    expect(ipcRenderer.invoke).toHaveBeenCalledWith(ipcMainChannels.OPEN_PATH, job.argsValues.workspace_dir);
+
+    const errorModal = await findByRole('dialog', { name: 'Error opening workspace'});
+    expect(errorModal).toBeTruthy();
```
davemfish, line 214: Nice test!

davemfish (commented):
Hi @emilyanndavis , this looks great to me. I like the new error modal. It's nice that `openPath` returns an error when it doesn't work. 

I appreciate you trying to preserve the behavior of opening the workspace with the logfile selected, when it exists. I don't think we were very intentional about that behavior when we first designed this. I'm sure the main objective was just to open to the workspace directory. 

So I'm curious what you think about dropping the `showItemInFolder` pathway altogether and always using `openPath`? Mainly for the sake of simplicity.

On the other hand if you like having the file selected then I think your solution in this PR is great!

workbench/tests/renderer/investtab.test.js
emilyanndavis: Thanks!

I guess I assumed that highlighting the logfile was intentional, maybe to help users find the logfile when they needed to reach out for support? 

I can imagine that, under normal circumstances, a user might find it strange to see a particular file selected in their workspace, especially since it's the file they're probably least interested in. It's a detail that's simple enough for a user to ignore, but still—if there's no real need for it, I can certainly drop it. I'll defer to your recommendation here.

davemfish: @emilyanndavis I guess I feel like the theoretical benefit of highlighting the file doesn't quite justify the extra logic in implementing it. I'm in favor of simplifying to always use `openPath` if that's okay with you.

workbench/src/renderer/components/InvestTab/index.jsx
```diff
@@ -186,13 +189,34 @@ class InvestTab extends React.Component {
     );
   }
 
+  handleOpenWorkspace(workspace_dir) {
+    if (workspace_dir) {
+      this.openWorkspaceDir(workspace_dir);
+    }
+  }
```
davemfish, line 196: Is there a benefit to defining both `handleOpenWorkspace` and `openWorkspaceDir`? As opposed to doing all the logic in one function? It's fairly inconsequential so if you prefer it this way feel free to leave it as-is.

workbench/src/renderer/components/InvestTab/index.jsx
```diff
@@ -213,88 +237,99 @@ class InvestTab extends React.Component {
...
-            </TabContent>
-          </Col>
-        </Row>
-      </TabContainer>
+                {
+                  status
+                    ? (
+                      <ModelStatusAlert
+                        status={status}
+                        handleOpenWorkspace={() => this.handleOpenWorkspace(argsValues?.workspace_dir)}
```
davemfish, line 281: woah I just learned something new!
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

davemfish (changes requested):
@emilyanndavis this looks good except for some syntax in HISTORY that's not passing our checks. Maybe the indentation?

And I'm not sure why all the other checks are being cancelled. I didn't think the failed HISTORY check would cause that, but maybe it does?

I had one other code-comment, but it doesn't necessarily require any changes. Thanks!

emilyanndavis:
> @emilyanndavis this looks good except for some syntax in HISTORY that's not passing our checks. Maybe the indentation?
> 
> And I'm not sure why all the other checks are being cancelled. I didn't think the failed HISTORY check would cause that, but maybe it does?

Yes, it was the indentation. Should be fixed now. And the other checks were cancelled because I cancelled them once I saw the HISTORY check had failed. :)

workbench/src/renderer/components/InvestTab/index.jsx
emilyanndavis: Probably not. I think this is a holdover from when `handleOpenWorkspace` also called `shell.showItemInFolder` when the logfile was valid. With that complexity gone, it does look a little silly now.

davemfish (approved):
All good now!

PR #1625, opened 2024-09-10
Add twine check and add README_PYTHON.rst to doc8 linting

### LINKED ISSUE #1582: Add a github action to check `python setup.py --long-description`
It might be nice to add a GHA check to lint the text of the long-description, which is visible on our package PyPI page.  This isn't super important because the content of the page so rarely changes, but it still might be nice for future us to have that automated confirmation of RST correctness.

emilyanndavis:
## Description
Resolves #1582.

1. Added `twine check` task, which validates the source distribution and raises an error if there are problems in the long description that will prevent proper rendering on PyPI. (See [twine check example output](https://github.com/emilyanndavis/invest/actions/runs/10782853636/job/29903637756#step:7:22) from an intentionally failed run.)
2. Also updated the `check-history-rst-syntax` so that it also lints README_PYTHON.rst. While `twine check` seems to bail after the first syntax error it encounters, doc8 reports all issues, which would be helpful in the (albeit unlikely) scenario in which README_PYTHON.rst is updated and ends up with multiple syntax issues. (See [doc8 example output](https://github.com/emilyanndavis/invest/actions/runs/10782853636/job/29903628924) from an intentionally failed run.)

I realize there's a lot of overlap between these two tasks, so I'm open to other options if folks think it's not worthwhile to include both.

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -46,8 +46,8 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           python -m flake8 src --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
-  check-history-rst-syntax:
-    name: Check HISTORY RST syntax
+  check-rst-syntax:
+    name: Check RST syntax
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -63,7 +63,7 @@ jobs:
       - name: Lint with doc8
         run: |
             # Skip line-too-long errors (D001)
-            python -m doc8 --ignore D001 HISTORY.rst
+            python -m doc8 --ignore D001 HISTORY.rst README_PYTHON.rst
 
   run-model-tests:
     name: Run model tests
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -193,6 +193,12 @@ jobs:
           # natcap.invest from source and that it imports.
           python -c "from natcap.invest import *"
 
+      - name: Set up environment
+        run: pip install twine
```
phargogh, line 197: Would it be worth including `twine` under the `requirements` list up above, when we're using the actions `setup_env` step?  I don't think it'd be strictly necessary, but I think it might avoid the need to re-resolve the dependencies, which could potentially save some small amount of time.  I think the bigger benefit might be just having all of the python dependencies in once place.

So I'm thinking something like this up above:
```
      - uses: ./.github/actions/setup_env
        with:
          python-version: ${{ matrix.python-version }}
          requirements-files: requirements.txt
          requirements: ${{ env.CONDA_DEFAULT_DEPENDENCIES }} twine
```

phargogh (changes requested):
Makes sense to me!  I just had one very minor suggestion about moving the twine requirement to the environment setup step, but please do let me know if this doesn't make sense or is better the way you have it!

.github/workflows/build-and-test.yml
emilyanndavis: Great idea, thanks!

phargogh (approved):
Thanks @emilyanndavis !

PR #1627, opened 2024-09-11
Fix autotest failures on `main`

phargogh:
## Description
Fixes #

## Checklist

### DIFF
--- a/constraints_tests.txt
+++ b/constraints_tests.txt
@@ -12,4 +12,4 @@ GDAL<3.8.5
 # Pyinstaller 6.10 breaks our windows builds.  Until we can figure out the
 # root cause, let's cap the versions to those that work.
 # https://github.com/natcap/invest/issues/1622
-pyinstaller<6.10
+#pyinstaller<6.10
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -196,7 +196,7 @@ jobs:
       - name: Check long description with twine
         run: twine check $(find dist -name "natcap[._-]invest*")
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: Source distribution
```
davemfish, line 201: It looks like there's a breaking change here: https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

I guess we'll have to rename the artifact. Using `matrix.os` and `matrix.python-version` should do it.

davemfish (changes requested):
Thanks @phargogh . I had one question in `__main__.py` but doesn't require any changes. 

And there's a breaking change in one of the `actions` that we need to address. Thanks for updating all of those for us!

davemfish (approved):

PR #1631, opened 2024-09-19
UNA: allow the LULC raster to not have a nodata value defined

### LINKED ISSUE #1293: Urban Nature Access assumes the LULC has a defined nodata value

phargogh:
UNA has a masking step that assumes we have a nodata value set on both the population and lulc rasters.  If the lulc does not have a nodata value, the masking would raise an error.  This PR fixes this behavior by updating the LULC warping step to set a nodata value that is outside of the range of the dtype used if the raster does not have a nodata value defined.

Fixes #1293

## Checklist

### DIFF
--- a/src/natcap/invest/urban_nature_access.py
+++ b/src/natcap/invest/urban_nature_access.py
@@ -755,13 +755,12 @@ def execute(args):
     squared_lulc_pixel_size = _square_off_pixels(args['lulc_raster_path'])
 
     lulc_alignment_task = graph.add_task(
-        pygeoprocessing.warp_raster,
+        _warp_lulc,
         kwargs={
-            'base_raster_path': args['lulc_raster_path'],
-            'target_pixel_size': squared_lulc_pixel_size,
-            'target_bb': target_bounding_box,
-            'target_raster_path': file_registry['aligned_lulc'],
-            'resample_method': 'near',
+            "source_lulc_path": args['lulc_raster_path'],
+            "target_lulc_path": file_registry['aligned_lulc'],
+            "target_pixel_size": squared_lulc_pixel_size,
+            "target_bounding_box": target_bounding_box,
         },
         target_path_list=[file_registry['aligned_lulc']],
         task_name='Resample LULC to have square pixels'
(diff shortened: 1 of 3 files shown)


src/natcap/invest/urban_nature_access.py
```diff
@@ -2528,6 +2527,40 @@ def _create_mask(*raster_arrays):
...
+    Returns:
+        ``None``.
+    """
+    source_raster_info = pygeoprocessing.get_raster_info(source_lulc_path)
+    target_nodata = source_raster_info['nodata'][0]
+    if target_nodata is None:
+        # Guarantee that our nodata cannot be represented by the datatype -
+        # select a nodata value that's out of range.
+        target_nodata = pygeoprocessing.choose_nodata(
+            source_raster_info['numpy_type']) + 1
```
emilyanndavis, line 2549: I'm curious about the need to add 1 here. I mean, I understand from the comment why you've done this, but I wonder if you think `choose_nodata` should be updated to return the max + 1 instead of just the max? Would that make sense for other use cases, or could it cause problems?

src/natcap/invest/urban_nature_access.py
```diff
@@ -2528,6 +2527,40 @@ def _create_mask(*raster_arrays):
...
+
+    pygeoprocessing.warp_raster(
+        source_lulc_path, target_pixel_size, target_lulc_path,
+        'near', target_bb=target_bounding_box,
+        target_projection_wkt=source_raster_info['projection_wkt'])
+
+    # if there is no defined nodata, set a default value
+    raster = gdal.OpenEx(target_lulc_path, gdal.GA_Update)
+    band = raster.GetRasterBand(1)
+    band.SetNoDataValue(target_nodata)
```
emilyanndavis, line 2559: Forgive me if I'm just misreading something, but could this `SetNoDataValue` unintentionally overwrite an existing nodata value that was defined already?

emilyanndavis (commented):
Looks good, @phargogh; thanks for giving me the chance to review. Just had a couple of questions!

src/natcap/invest/urban_nature_access.py
phargogh: Great question!  Since we can assume that every pixel in the dataset is valid (because `target_nodata is None`), the safest possible nodata value is one that isn't even in the range of values possible with the given datatype.  For a `float32` raster, it's pretty unlikely that we would have a dataset that would have every discrete value that is allowed by the float32 spec, but it could still happen.  Since the nodata value is stored as a piece of metadata separate from the pixel values themselves, we can avoid this possibility completely by using a nodata value that is completely outside of the range of possibilities of the datatype of the pixel values themselves.

If we were to change the datatype of `choose_nodata`, you are right that that would unfortunately break things.  In `raster_map`, the nodata values of the incoming raster are transformed to the nodata values of the target raster, where the target nodata has been selected by `choose_nodata`.  If `choose_nodata` were to return a value of the (dtype's max value)+1, then we'd get an overflow in the pixel values, or we bump up the output raster's datatype to accommodate the extra needed precision, neither of which we want!

So this solution as implemented gets around the situation by only handling this one case, where the lulc doesn't have a nodata value defined, and we just set one (to avoid breaking other things later on in the model) that can never happen in this datatype.

src/natcap/invest/urban_nature_access.py
phargogh: You know, it _shouldn't_ be an issue because `target_nodata` would just be set to the same value it was at before, but this isn't really a clean approach.  I just updated the function in 94f7f006e to only set the nodata value if one isn't already defined.

emilyanndavis (approved):

emilyanndavis (approved):

phargogh:
All checks passed and @emilyanndavis has approved, so I'll go ahead and merge.  Thanks!

PR #1633, opened 2024-09-25
D8 routing

emlys:
## Description
This PR implements D8 equivalents of the routing functions that previously only supported MFD. 
A "routing algorithm" option input (MFD or D8) is added to each model. Based on that option, the appropriate pygeoprocessing routing functions are called. For the model-specific routing functions, I used [C++ templates](https://learn.microsoft.com/en-us/cpp/cpp/templates-cpp) to implement switching between MFD and D8 "modes" on the iterator classes in ManagedRaster.h. To make it possible to call the templated functions, I rewrote the bulk of the SDR, NDR, and SWY routing functions from Cython to C++. I'm happy to discuss pros and cons of this.

Future work: add logging to python from C++ extensions, add templating for `ManagedRaster` to read and return values of different data types

## Checklist


src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -337,15 +340,15 @@ class ManagedRaster {
         }
 };
 
+template<class T>
```
emlys, line 343: This is a class template. `ManagedFlowDirRaster` is _not_ a class, but `ManagedFlowDirRasater<MFD>` and `ManagedFlowDirRasater<D8>` are classes that can be created from this template.

src/natcap/invest/managed_raster/ManagedRaster.h
```diff
@@ -337,15 +340,15 @@ class ManagedRaster {
...
 class ManagedFlowDirRaster: public ManagedRaster {
-
 public:
-
     ManagedFlowDirRaster() {}
 
     ManagedFlowDirRaster(char* raster_path, int band_id, bool write_mode)
         : ManagedRaster(raster_path, band_id, write_mode) {}
 
+    template<typename T_ = T, std::enable_if_t<std::is_same<T_, MFD>::value>* = nullptr>
```
emlys, line 351: This tells the compiler to use this implementation of `is_local_high_point` if the template argument is `MFD`. The same pattern is used in all the iterator classes below.

phargogh: Interesting!  So then if we were to extend these classes to accommodate some additional routing method, we would need to add new classmethods that have a conditional like this?

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -60,211 +48,46 @@ def ndr_eff_calculation(
...
+        stream_path.encode('utf-8'),
+        retention_eff_lulc_path.encode('utf-8'),
+        crit_len_path.encode('utf-8'),
+        to_process_flow_directions_path.encode('utf-8'),
+        effective_retention_path.encode('utf-8')]
+
+    if algorithm == 'MFD':
+        run_effective_retention[MFD](*args)
+    else: # D8
+        run_effective_retention[D8](*args)
```
phargogh, line 93: I get it now!  That is really pretty neat.  So the implication is that the compiler is, effectively, rendering MFD and D8 versions of the underlying functions to allow us to call either one at runtime.  Awesome.

phargogh (approved):
I really don't have anything to suggest at this time ... the solution's architecture makes sense to me given the constraints of a compiled language, I'm not seeing _any_ compiler warnings, tests pass, and the outputs all make sense to me, as do the tests.  This looks great!

PR #1637, opened 2024-10-04
HQ: Use reprojected access vector in rasterization task

### LINKED ISSUE #1615: HQ doesn't use reprojected access vector
I noticed that in HQ, the input vector is reprojected, but the reprojected version is never used. The original version is used in the model:
https://github.com/natcap/invest/blob/e81966093b92bad6b4c28ca2bcb6f04a4e96a513/src/natcap/invest/habitat_quality.py#L542-L566

If it does need to be reprojected, we should be using the reprojected version. If reprojection is unnecesssary, we can remove this step of the model.

emilyanndavis:
## Description
Fixes #1615 by passing the reprojected access vector to the rasterize_access task

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -48,6 +48,9 @@ Unreleased Changes
     * Fixed a bug that, in certain scenarios, caused a datastack to be saved
       with relative paths when the Relative Paths checkbox was left unchecked
       (https://github.com/natcap/invest/issues/1609)
+* Habitat Quality
+    * Access raster is now generated from the reprojected access vector.
+      (https://github.com/natcap/invest/issues/1615)
 
 3.14.2 (2024-05-29)
 -------------------
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -555,7 +555,7 @@ def execute(args):
 
         rasterize_access_task = task_graph.add_task(
             func=pygeoprocessing.rasterize,
-            args=(args['access_vector_path'], access_raster_path),
+            args=(reprojected_access_path, access_raster_path),
             kwargs={
                 'option_list': ['ATTRIBUTE=ACCESS'],
                 'burn_values': None


emilyanndavis:
@dcdenu4 I think this is the only change needed, but I noticed that `create_access_raster_task` and `rasterize_access_task` are being added to `access_task_list`, which allows them to later be passed to `total_degradation_task` as dependencies.

Do I need to add `reproject_access_task` to `access_task_list`?

Or is it sufficient that `reproject_access_task` is already a dependency of `rasterize_access_task` (and therefore implicitly understood as a dependency of `total_degradation_task`)?

Also wondering if you have any suggestions as to how to go about testing this, or if there might be a scenario in another test suite I can use as an example?

dcdenu4:
Hey @emilyanndavis

>Or is it sufficient that reproject_access_task is already a dependency of rasterize_access_task (and therefore implicitly understood as a dependency of total_degradation_task)?

Yes, that's right. Since `rasterize_access_task` can't execute until `reproject_access_task` is completed, we can just add `rasterize_access_task` to `access_task_list` and rely on that dependency chain. Since taskgraph is just queuing up work (tasks) to be run in any order (if `n_workers>1`) we want to make sure `total_degradation_task` doesn't start until all the possible inputs needed have been calculated first.

>any suggestions as to how to go about testing this

Great question, this is an interesting one for testing. In the `Model_Spec`, `access_vector_path` is included in the `args_with_spatial_overlap`. This means that our validation step is going to check that the bounding boxes of all the included spatial inputs overlap. If they don't, an error will be returned. So, we have some upfront check that makes sure the access vector doesn't have a coordinate system that is somewhere else in the world compared to the other inputs. If the `access_vector` did have a different projection from the land use land cover input, there's a good chance the rasterize operation would work just fine, though it could be slightly different than what we'd expect. This would be the minute detail we'd want to test, but one that might be a pain to create and one that might already be tested (more on that in a sec). In `tests/test_habitat_quality.py` we have a `def make_access_shp` function, which uses a default projected coordinate system. We might expect that if we used a different coordinate system that still represented the same area and vector features, the pixels that get set during rasterization might be different. So we could test that the rasterized layer is what we expect and that we can confirm a difference... But, if we did have a different output for that rasterization step, I'd ultimately expect the final outputs of the model to be different too. So in that way we already might be testing against this with our regression tests. All to say, I think it would be interesting to see if an access vector projected in a different space would lead to a different rasterized output, I don't think we need to worry about explicitly testing this.

Happy to talk about this more as I found it hard to clearly communicate here!

emilyanndavis:
Makes sense, @dcdenu4; thank you for the detailed explanation!

dcdenu4 (approved):
Thanks @emilyanndavis ! Looks good to me. Ignoring the RTD failure as unrelated.

PR #1638, opened 2024-10-07
update SDR variable names and correct typo in docstring #1629

### LINKED ISSUE #1629: Small typos/mis-match with user's guide in sdr_core.pyx comments?
I think there are a few small typos/updates to notation needed in [sdr_core.pyx](https://github.com/natcap/invest/blob/main/src/natcap/invest/sdr/sdr_core.pyx) in the comment block starting on line 355.

Specifically:
line 359 currently reads:  `t_i  =      dr_i  * (sum over j ∈ J of f_j * p(i,j)) + E'_i `
I think it should read: `t_i  =      dt_i  * (sum over j ∈ J of f_j * p(j,i))` 
This is because `dr_i` was changed to `dt_i `when deposition (R) was changed to trapping (T) to avoid confusion with the R factor in RUSLE. And  `E'_i`  is not part of the equation for `t_i `(see [eq. 81 in the user's guide](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/sdr.html#equation-ti)). It looks like E'_i is  correctly not included in the actual calculation implemented in the code below (line 653). `p(i,j) `should be changed to `p(j,i)` to denote the flow from pixels j to i.

line 361 and 365 currently include `dr_i`, which should be changed to `dt_i`, for the reason noted above. 

line 361 currently also includes `p(i,j)`, which should be changed to `p(j,i)` for the reason noted above.

It looks like the old variable name `dr_i `is retained in the code in lines 644-662. Perhaps this should be changed to` tr_i `to be consistence with the user's guide notation? But maybe that's more complicated than I am imagining.

And I'd welcome feedback on how to write an issue that's easy to follow! Thanks!

emlys:
## Description
Fixes #1629 
Making the changes suggested by @lmandle 

## Checklist

### DIFF
--- a/src/natcap/invest/sdr/sdr_core.pyx
+++ b/src/natcap/invest/sdr/sdr_core.pyx
@@ -356,13 +356,13 @@ def calculate_sediment_deposition(
 
     This algorithm outputs both sediment deposition (t_i) and flux (f_i)::
 
-        t_i  =      dr_i  * (sum over j ∈ J of f_j * p(i,j)) + E'_i
+        t_i  =      dt_i  * (sum over j ∈ J of f_j * p(j,i))
 
-        f_i  = (1 - dr_i) * (sum over j ∈ J of f_j * p(i,j)) + E'_i
+        f_i  = (1 - dt_i) * (sum over j ∈ J of f_j * p(j,i)) + E'_i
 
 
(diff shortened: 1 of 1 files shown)


dcdenu4 (approved):
Looks good, thanks @emlys

PR #1644, opened 2024-10-09
Redesign plugin interface and support removing plugins

emlys:
## Description
image

- Create an interface to remove a plugin
- Move this, and the "add plugin" button, into a "Manage plugins" modal

Very little thought went into the design of the original "add plugin" button, it was just to get something working. This is a more intentional design.

I didn't add any tests yet, since we talked a while ago about possibly needing less UI tests, but I'm open to writing some if you think it would be helpful.

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -89,6 +89,13 @@ jobs:
           path: data/invest-test-data
           key: git-lfs-testdata-${{ hashfiles('Makefile') }}
 
+      - name: debug
+        run: |
+          ls data/invest-test-data
+          ls data/invest-test-data/wind_energy/smoke
+          ls /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke
+          gdalinfo /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke/landpoly_smoke.shp
+
       - name: Set up python environment
         uses: ./.github/actions/setup_env
         with:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -89,13 +89,6 @@ jobs:
           path: data/invest-test-data
           key: git-lfs-testdata-${{ hashfiles('Makefile') }}
 
-      - name: debug
-        run: |
-          ls data/invest-test-data
-          ls data/invest-test-data/wind_energy/smoke
-          ls /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke
-          gdalinfo /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke/landpoly_smoke.shp
-
       - name: Set up python environment
         uses: ./.github/actions/setup_env
         with:
@@ -129,6 +122,13 @@ jobs:
           python -m build --wheel
           ls -la dist
           pip install $(find dist -name "natcap.invest*.whl")
+
+      - name: debug
+        run: |
+          ls data/invest-test-data
+          ls data/invest-test-data/wind_energy/smoke
+          ls /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke
+          gdalinfo /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke/landpoly_smoke.shp
 
       - name: Run model tests
         run: make test
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -123,13 +123,6 @@ jobs:
           ls -la dist
           pip install $(find dist -name "natcap.invest*.whl")
 
-      - name: debug
-        run: |
-          ls data/invest-test-data
-          ls data/invest-test-data/wind_energy/smoke
-          ls /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke
-          gdalinfo /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke/landpoly_smoke.shp
-
       - name: Run model tests
         run: make test
 
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -26,7 +26,7 @@ export default function PluginModal(props) {
     setShowPluginModal(false);
   };
   const handleModalOpen = () => setShowPluginModal(true);
-  const handleSubmit = () => {
+  const addPlugin = () => {
     setLoading(true);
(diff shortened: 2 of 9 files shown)


workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -29,36 +34,88 @@ export default function PluginModal(props) {
...
+          </Button>
+        </Form.Group>
+        <hr />
+        <Form.Group className="mb-3">
+          <Form.Label htmlFor="url">Remove a plugin</Form.Label>
+          <Form.Control
+            id="plugin-select"
+            as="select"
+            name="plugin"
+            type="text"
```
emilyanndavis, line 93: This might be another leftover from copy-paste? You can remove the `type` attribute here, since it's only relevant for an `input` (and, since you're using `as="select"`, this form field is rendered as a `select`, not an `input`).

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -29,36 +34,88 @@ export default function PluginModal(props) {
...
             type="text"
             placeholder={t('Enter Git URL')}
-            onChange={handleChange}
+            onChange={(event) => setURL(event.currentTarget.value)}
           />
+          <Form.Text className="text-muted">
+            {t('This may take several minutes')}
+          </Form.Text>
+          <Button
+            name="submit"
```
emilyanndavis, line 78: Can you remove the `name` attributes from the `button`s? Since we're not submitting raw form data to a server (instead, we're using an explicit `onClick` handler to pre-process data on the front end), there's no need for the `button`s to have `name`s.

emilyanndavis (changes requested):
Looks good, @emlys! Just a few comments on small cleanup sorts of things.

As for tests, it looks like you have a pretty good baseline already. I might consider adding a test to make sure the list of options in the "Remove a plugin" dropdown is updated as expected after removing a plugin.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -14,13 +13,19 @@ const { ipcRenderer } = window.Workbench.electron;
...
 
-  const handleModalClose = () => setShowAddPluginModal(false);
-  const handleModalOpen = () => setShowAddPluginModal(true);
+  const handleModalClose = () => {
+    setURL(undefined);
+    setErr(false);
+    setShowPluginModal(false);
+  };
+  const handleModalOpen = () => setShowPluginModal(true);
   const handleSubmit = () => {
```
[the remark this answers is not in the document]
emlys: Good idea!

workbench/src/renderer/components/PluginModal/index.jsx
emlys: Good catch!

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -63,34 +63,29 @@ export default function PluginModal(props) {
     <Modal.Body>
       <Form>
         <Form.Group className="mb-3">
-          <Form.Label htmlFor="url">Add a plugin</Form.Label>
+          <Form.Label htmlFor="url">{t('Add a plugin')}</Form.Label>
           <Form.Control
             id="url"
-            name="url"
-            type="text"
```
emilyanndavis, line 50: Would you mind adding `type="text"` back to this form control? ([When `type` is omitted, it does default to `"text"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#type), but it's a best practice to always include a `type` attribute on an `<input>`.)

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -63,34 +63,29 @@ export default function PluginModal(props) {
...
         </Form.Group>
         <hr />
         <Form.Group className="mb-3">
-          <Form.Label htmlFor="url">Remove a plugin</Form.Label>
+          <Form.Label htmlFor="plugin-select">{t('Remove a plugin')}</Form.Label>
           <Form.Control
             id="plugin-select"
             as="select"
-            name="plugin"
-            type="text"
```
emilyanndavis: (Just to be clear, removing `type="text"` from this form control was correct, because this one is rendered as a `<select>`, not an `<input>`.)

workbench/tests/renderer/plugin.test.js
```diff
@@ -123,4 +123,41 @@ describe('Add plugin modal', () => {
...
+
+    const pluginDropdown = await findByLabelText('Remove a plugin');
+    await userEvent.selectOptions(pluginDropdown, [getByRole('option', { name: 'Foo' })]);
+
+    const submitButton = await findByText('Remove');
+    userEvent.click(submitButton);
+    await waitFor(() => {
+      expect(spy.mock.calls.map((call) => call[0])).toContain(ipcMainChannels.REMOVE_PLUGIN);
+    });
+    // expect the plugin to have disappeared from the model list and the dropdown
```
emilyanndavis, line 159: 👍🏻

emilyanndavis (approved):

dcdenu4 (changes requested):
Hey @emlys, I don't have anything to add on top of Emily Davis's review. So, I think it's just the test failures to debug.

emlys:
@dcdenu4 the workbench tests are passing now. The other failures are the same ones I'm seeing on main

.github/workflows/build-and-test.yml
```diff
@@ -123,6 +123,13 @@ jobs:
           ls -la dist
           pip install $(find dist -name "natcap.invest*.whl")
 
+      - name: debug
+        run: |
+          ls data/invest-test-data
+          ls data/invest-test-data/wind_energy/smoke
+          ls /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke
+          gdalinfo /Users/runner/work/invest/invest/tests/../data/invest-test-data/wind_energy/smoke/landpoly_smoke.shp
+
```
dcdenu4, line 132: Hey @emlys, should we remove this debug portion now?

dcdenu4 (changes requested):
Hey @emlys, thanks for looking into those tests. It looks like some of the GHA debugging code in `build-and-test.yml` was left in though.

Actually, now that I look at the GHA failures, it seems like they are all failing on that debug step in the workflow and not getting to actually running the tests.

.github/workflows/build-and-test.yml
emlys: Oops! Thanks for catching that

dcdenu4 (approved):
@emlys thanks for the updates. I'm still not sure about the test failures, but I'm going to push this through. I think one thing we haven't been doing is keeping up with merging `main` in, so things don't get stale.

PR #1646, opened 2024-10-11
Handle GDAL-land exceptions raised after upgrading pygeoprocessing

### LINKED ISSUE #1645: Using `gdal.UseExceptions` raising exceptions in tests
A bunch of our tests are failing in validation because of Pygeoprocessing 2.4.4's addition of using `gdal.UseExceptions` (within a context manager) to have different exception objects raised.

This is happening in a couple of places:
* In `validation`
* In Forest Carbon Edge Effects (due to needing to enable partial reprojection)
* In Forest Carbon Edge Effects (due to points being outside of the reprojection domain, see https://github.com/natcap/pygeoprocessing/issues/409)
* In Seasonal Water Yield (due to an incorrect filepath)

phargogh:
This just adds support for GDAL's `RuntimeError`s in addition to pygeoprocessing's `ValueError`s .  I left both in, in case we have a case where we are supporting earlier versions of pygeoprocessing since it was trivial to support both versions of the exception.

I also addressed an exception occurring in one of the tests because of a file name we changed recently.

Fixes #1645 

## Checklist


src/natcap/invest/utils.py
```diff
@@ -121,6 +121,25 @@ def capture_gdal_logging():
         gdal.PopErrorHandler()
 
 
+@contextlib.contextmanager
+def _set_gdal_configuration(opt, value):
```
davemfish, line 125: This looks handy

davemfish (approved):

PR #1649, opened 2024-10-15
HQ: generate rarity output in CSV format (in addition to raster)

### LINKED ISSUE #721: Output table alongside raster for Habitat Quality rarity
The HQ rarity map is not all that useful and would be more informative as an output CSV table. A listing of LULC classes for current / future and their rarity values would be easier to consume. This came up when discussing #720 . 

- [ ] Update Users Guide

emilyanndavis:
## Description
Fixes #721 

## Checklist
- [ ] Updated the user's guide (Will update HQ outputs once the updates here have been finalized)

### DIFF
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -862,7 +862,6 @@ def _compute_rarity_operation(
         base_lulc_path_band[0])
     base_pixel_size = base_raster_info['pixel_size']
     base_area = float(abs(base_pixel_size[0]) * abs(base_pixel_size[1]))
-    base_nodata = base_raster_info['nodata'][0]
 
     lulc_code_count_b = _raster_pixel_count(base_lulc_path_band)
 
@@ -870,7 +869,6 @@ def _compute_rarity_operation(
     lulc_raster_info = pygeoprocessing.get_raster_info(lulc_path_band[0])
     lulc_pixel_size = lulc_raster_info['pixel_size']
     lulc_area = float(abs(lulc_pixel_size[0]) * abs(lulc_pixel_size[1]))
-    lulc_nodata = lulc_raster_info['nodata'][0]
 
     # Trim cover_x to the mask of base.
     pygeoprocessing.raster_map(
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -1,6 +1,7 @@
 # coding=UTF-8
 """InVEST Habitat Quality model."""
 import collections
+import csv
 import logging
 import os
 
@@ -983,12 +984,11 @@ def _generate_rarity_csv(rarity_dict, target_csv_path):
         None
     """
     lulc_codes = sorted(rarity_dict)
-    cols = ['lulc_code', 'rarity_value']
-    with open(target_csv_path, 'w') as csvfile:
-        csvfile.write(str(','.join(cols) + '\n'))
+    with open(target_csv_path, 'w', newline='') as csvfile:
+        writer = csv.writer(csvfile, delimiter=',')
+        writer.writerow(['lulc_code', 'rarity_value'])
         for lulc_code in lulc_codes:
-            row = [str(lulc_code), str(rarity_dict[lulc_code])]
-            csvfile.write(str(','.join(row) + '\n'))
+            writer.writerow([lulc_code, rarity_dict[lulc_code]])
 
 
 def _raster_pixel_count(raster_path_band):
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -908,6 +908,16 @@ def _compute_rarity_operation(
 
 
 def _generate_rarity_csv(rarity_dict, target_csv_path):
+    """Generate CSV containing rarity values by LULC code.
+
+    Args:
+        rarity_dict (dict): dictionary containing LULC codes (as keys)
+            and their associated rarity values (as values).
+        target_csv_path (string): path to output CSV.
+
+    Returns:
+        None
+    """
     lulc_codes = sorted(rarity_dict)
     cols = ['LULC_code', 'rarity_value']
     with open(target_csv_path, 'w') as csvfile:
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -280,10 +280,10 @@ MODEL_SPEC = {
                 },
                 "rarity_c.csv": {
                     "about": ("Table of rarity values by LULC code for the "
-                              "current landscape"),
-                    "index_col": "LULC_code",
+                              "current landscape."),
+                    "index_col": "lulc_code",
                     "columns": {
-                        "LULC_code": {
+                        "lulc_code": {
                             "type": "number",
                             "units": u.none,
                             "about": "LULC class",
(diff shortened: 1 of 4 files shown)


src/natcap/invest/habitat_quality.py
```diff
@@ -213,36 +213,36 @@
         "output": {
             "type": "directory",
             "contents": {
-                "deg_sum_out_c.tif": {
+                "deg_sum_c.tif": {
```
dcdenu4, line 216: I agree that the "out" in these file names is pretty unnecessary. However, in practice we don't like to update intermediate / output names too often because it is something that could affect a users workflow. It's a light backwards incompatible API change essentially. I think it's alright to leave these changes in, but just wanted to mention that!

src/natcap/invest/habitat_quality.py
```diff
@@ -278,6 +278,70 @@
                     "created_if": "lulc_bas_path and lulc_fut_path",
                     "bands": {1: {"type": "ratio"}}
                 },
+                "rarity_c.csv": {
+                    "about": ("Table of rarity values by LULC code for the "
+                              "current landscape"),
```
dcdenu4, line 283:
```suggestion
                              "current landscape."),
```

src/natcap/invest/habitat_quality.py
```diff
@@ -278,6 +278,70 @@
...
+                    "index_col": "LULC_code",
```
dcdenu4, line 284: I think we have started to standardize using all lowercase for table outputs. Could we have this be `lulc_code`?

src/natcap/invest/habitat_quality.py
```diff
@@ -278,6 +278,70 @@
...
+                                "the likelihood that the preservation of that habitat "
+                                "type on the current or future landscape is important "
+                                "to biodiversity conservation."),
+                        },
+                    },
+                    "created_if": "lulc_bas_path",
+                },
+                "rarity_f.csv": {
+                    "about": ("Table of rarity values by LULC code for the "
+                              "future landscape"),
```
dcdenu4, line 315:
```suggestion
                              "future landscape."),
```

src/natcap/invest/habitat_quality.py
```diff
@@ -829,7 +896,8 @@ def total_degradation(*arrays):
 
 
 def _compute_rarity_operation(
-        base_lulc_path_band, lulc_path_band, new_cover_path, rarity_path):
+        base_lulc_path_band, lulc_path_band, new_cover_path,
+        rarity_raster_path, rarity_csv_path):
```
dcdenu4, line 900: Might be worth adding to the docstring that this function generates a raster and csv output.

src/natcap/invest/habitat_quality.py
```diff
@@ -895,13 +962,35 @@ def _compute_rarity_operation(
...
+    Args:
+        rarity_dict (dict): dictionary containing LULC codes (as keys)
+            and their associated rarity values (as values).
+        target_csv_path (string): path to output CSV.
+
+    Returns:
+        None
+    """
+    lulc_codes = sorted(rarity_dict)
+    cols = ['LULC_code', 'rarity_value']
```
dcdenu4, line 986: As mentioned above, I believe we've started to standardize lowercase for table column headers. Even though LULC is an acronym, I think it should still follow that standard.

src/natcap/invest/habitat_quality.py
```diff
@@ -895,13 +962,35 @@ def _compute_rarity_operation(
...
+    with open(target_csv_path, 'w') as csvfile:
+        csvfile.write(str(','.join(cols) + '\n'))
+        for lulc_code in lulc_codes:
+            row = [str(lulc_code), str(rarity_dict[lulc_code])]
+            csvfile.write(str(','.join(row) + '\n'))
```
dcdenu4, line 991: This looks like a very reasonable writing solution. But I did want to mention the `csv` module and `writer` object, https://docs.python.org/3/library/csv.html#csv.writer. One nice thing about using this is we don't have to worry about that newline character, which I think is best to leave up to the operating system when we can. Different OS's handle line endings differently. Not that this is the case here, but if we did have a list of rows we could also just write those out with `writerows(list_of_rows)`.

tests/test_habitat_quality.py
```diff
@@ -245,6 +246,14 @@ def test_habitat_quality_presence_absence_regression(self):
             # so we should exclude those new nodata pixel values.
             assert_array_sum(raster_path, assert_value, include_nodata=False)
 
+        for csv_filename in ['rarity_c_regression.csv',
+                             'rarity_f_regression.csv']:
+            csv_path = os.path.join(args['workspace_dir'], csv_filename)
+            rarity_table = pandas.read_csv(csv_path)
+            assert 'LULC_code' in rarity_table.columns
+            self.assertAlmostEqual(rarity_table['rarity_value'].sum(),
```
dcdenu4, line 254: Could you leave a comment of where the 0.6667 values comes from?

tests/test_habitat_quality.py
```diff
@@ -2105,3 +2114,5 @@ def test_habitat_quality_validate_missing_fut_column(self):
                 header='column', header_name='fut_path')
         )]
         self.assertEqual(validate_result, expected)
+
+# @TODO: ¿add rarity CSV test?
```
dcdenu4, line 2118: Thinking of adding an additional test!?

dcdenu4 (changes requested):
Thanks @emilyanndavis! I had a few comments and suggestions.

One thing I went back and forth on was whether these should be two CSV outputs or one. I think the spirit of the issue was to combine `rarity_c` and `rarity_f` into one table, such that comparing rarity values across current and future scenarios was easier. The way you've set it up makes total sense, especially from how the code is broken up. It'd take a little bit of reworking to combine them and I'm not myself convinced it's really worth it. A user can quickly get at the desired effect themselves too. Curious your thoughts on this.

src/natcap/invest/habitat_quality.py
emilyanndavis: Totally makes sense. The only reason I updated these is because I noticed the output files don't include "out" in their names. So I'm not truly changing the output names, just updating the spec to reflect what the model actually does.

tests/test_habitat_quality.py
emilyanndavis: Oops! I was, but then I realized it'd be simpler to update an existing test instead.

src/natcap/invest/habitat_quality.py
emilyanndavis: I think I based this on an example I found in another model, but I will try switching over to `csv.writer`; thanks for the suggestion!

tests/test_habitat_quality.py
emilyanndavis: Yes, and I rewrote this part of the test; hopefully it's a bit more meaningful now. (I also switched from pandas to `csv.reader`.)

emilyanndavis:
> One thing I went back and forth on was whether these should be two CSV outputs or one. I think the spirit of the issue was to combine `rarity_c` and `rarity_f` into one table, such that comparing rarity values across current and future scenarios was easier. The way you've set it up makes total sense, especially from how the code is broken up. It'd take a little bit of reworking to combine them and I'm not myself convinced it's really worth it. A user can quickly get at the desired effect themselves too. Curious your thoughts on this.

Oh, that hadn't even occurred to me, but it makes sense! I agree, though, the way the code is currently set up isn't immediately conducive to combining outputs into a single CSV, and my gut says it's not worth it right now. Maybe in a future iteration, if/when we find that's what users need.

dcdenu4 (approved):

PR #1664, opened 2024-10-24
Update validation.py to avoid SyntaxError

### LINKED ISSUE #1663: "Validate sample data" task—and model tests—fail on a SyntaxError in validation.py
After a fresh clone of the sample data (i.e., after deleting the `data/invest-sample-data` folder), `make validate_sampledata` is crashing on the following error:

```
File "invest/src/natcap/invest/validation.py", line 645
   patterns.append(column.replace('(', '\(').replace(')', '\)'))
                                       ^^^^
SyntaxError: invalid escape sequence '\('
```

emilyanndavis:
## Description
Fixes #1663 with Python raw string notation, as recommended by the [Python `re` module docs](https://docs.python.org/3.12/library/re.html):

> Also, please note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a `SyntaxWarning` and in the future this will become a `SyntaxError`. This behaviour will happen even if it is a valid escape sequence for a regular expression.
> 
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'.

I still don't know what triggered the switch to `SyntaxError`, since even the Python 3.13 docs still describe this as "future" behavior, but in any case the raw string notation appears to solve the problem.

## Checklist

### DIFF
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -642,7 +642,7 @@ def get_validated_dataframe(
             patterns.append(f'{groups[0]}(.+){groups[2]}')
         else:
             # for regular column names, use the exact name as the pattern
-            patterns.append(column.replace('(', '\\(').replace(')', '\\)'))
+            patterns.append(column.replace('(', r'\(').replace(')', r'\)'))
 
     # select only the columns that match a pattern
     df = df[[col for col in df.columns if any(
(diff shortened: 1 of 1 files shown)


phargogh (approved):
Thanks for looking into this @emilyanndavis !

Looking at the docs, it appears that in python 3.11, a `DeprecationWarning` was raised, and in python 3.12+ this warning was changed to `SyntaxWarning`.  This change matters because at the bottom of `pyproject.toml`, we tell `pytest` to elevate all warnings to exceptions _except_ for `DeprecationWarning`, `FutureWarning`, and one very specific `RuntimeWarning` related to an issue on M1 chips.  So that's why this behavior changed suddenly.

emilyanndavis:
Ooh, mystery solved! 🕵🏻 Thanks, @phargogh!

PR #1676, opened 2024-11-04
Wrap mamba commands in quotes

### LINKED ISSUE #1665: Plugins: removing plugin on Windows brings up cmd shell
Hopefully we're just missing a hide windows shell flag.

### LINKED ISSUE #1675: Plugins: error when there's a space in the mamba command path, Windows
It looks like we should wrap the command in quotes to get around this issue for Windows.

dcdenu4:
## Description
During installation on Windows there is a space in the path of where InVEST is installed. The space then breaks the path for running the mamba command. Wrapping that mamba command in quotes.

Also, adding `windowsHide:true` so a windows cmd shell doesn't pop open during plugin removal.

Fixes #1675
Fixes #1665

## Checklist


dcdenu4:
Closing because I was misunderstanding how the mamba commands were being handled.

PR #1680, opened 2024-11-05
Also handling sanitized package names. RE:#1679

### LINKED ISSUE #1679: Update source distribution filename in twine upload
Reading through the python docs more carefully, it appears the [distribution names are supposed to be sanitized slightly, including converting `.` characters to `_`](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode).  This has only recently been implemented in `python -m build` (although I can't find it in the [changelog](https://build.pypa.io/en/stable/changelog.html)), but apparently only for source distributions and not yet for wheels.

In any case, our build automation should support uploading distributions with these sanitized package names.

phargogh:
This should be a simple fix, just adding the sanitized distribution names to the `twine upload` command.  Further context is described in #1679 

Fixes #1679 

## Checklist

### DIFF
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -99,7 +99,7 @@ jobs:
           twine upload \
             --username="__token__" \
             --password=${{ secrets.PYPI_NATCAP_INVEST_TOKEN }} \
-            artifacts/natcap.invest*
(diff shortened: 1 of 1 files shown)


emlys (approved):
Thanks!

PR #1681, opened 2024-11-06
Feature/1368 display changelog in workbench

### LINKED ISSUE #1368: Workbench: display invest changelog the first time a new install is launched

emilyanndavis:
## Description
Fixes #1368.

On first launch of a new version, the workbench will display a modal containing the relevant section of the changelog (i.e., only the items that are new in this version).

On the first launch (of any version) ever, the workbench will display the "download sample data" modal as usual. Then, only once the download modal has been closed, the changelog modal will open. Once the changelog modal has been closed, it will not open again.

Note this also supports multiple installed versions. That is, a user who is actively using multiple versions of the workbench will not see the changelog every time they close one version and open another.

## Checklist

## Example
Note the changelog for 3.14.2 is rendered because 3.14.2 is the most recent version. Once a new version is released, the app will automatically render the relevant section of the changelog.

workbench-changelog_2024-11-07


workbench/src/main/main.js
```diff
@@ -9,35 +9,37 @@ import {
...
-import setupGetElectronPaths from './setupGetElectronPaths';
 import setupRendererLogger from './setupRendererLogger';
-import { ipcMainChannels } from './ipcMainChannels';
-import menuTemplate from './menubar';
-import ELECTRON_DEV_MODE from './isDevMode';
-import BASE_URL from './baseUrl';
-import { getLogger } from './logger';
-import i18n from './i18n/i18n';
-import pkg from '../../package.json';
+
```
emilyanndavis, line 42: The only imports added to this file are `setupBaseUrl` and `setupIsNewVersion`. All the other changes are just reorganization to make the file a bit easier to read.

workbench/src/main/main.js
```diff
@@ -1,5 +1,5 @@
 import path from 'path';
-
+import i18n from './i18n/i18n';
```
emilyanndavis, line 2: The only imports added to this file are `setupBaseUrl` and `setupIsNewVersion`. All the other changes are just reorganization to make the file a bit easier to read.

workbench/src/renderer/components/Changelog/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+                    `<section.*?>[\\s]*?<h1>${escapedVersionStr}\\b[\\s\\S]*?</h1>[\\s\\S]*?</section>`
+                );
+                const sectionMatches = htmlString.match(sectionRegex);
+                if (sectionMatches && sectionMatches.length) {
+                    let latestVersionSection = sectionMatches[0];
+                    const linkRegex = /<a\shref/g;
+                    // Ensure all links open in a new window and are styled with a relevant icon.
+                    latestVersionSection = latestVersionSection.replaceAll(
+                        linkRegex,
+                        '<a target="_blank" class="link-external" href'
```
davemfish, line 37: Do you think it's possible to have these open in the default browser, instead of an electron window? That usually requires calling `electron.shell.openExternal` via a click event (we have `ipcMainChannels.OPEN_EXTERNAL_URL`). But given the static html, I'm not sure if that's possible. Curious if you have any ideas, but it's not critical to change.

davemfish (commented):
Thanks, @emilyanndavis , this works great! Looks like you got the puppeteer tests working. And there's probably something unrelated going on the the GHA Mac runners.

I have mostly minor comments; happy to discuss any of them.

workbench/tests/renderer/changelog.test.js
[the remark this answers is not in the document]
davemfish: Same with `Changelog/index.jsx` I suppose

workbench/tests/renderer/changelog.test.js
```diff
@@ -0,0 +1,81 @@
...
+        const changelogModal = await findByRole('dialog', { name: 'New in this version' });
+        expect(changelogModal).toBeInTheDocument();
+    });
+
+    test('Changelog modal does not open when current version has been run before', async () => {
+        const { findByRole } = render(<App isNewVersion={false} />);
+        let changelogModalFound = true;
+        try {
+            await findByRole('dialog', { name: 'New in this version' });
+        } catch {
```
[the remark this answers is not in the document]
emilyanndavis: Yes, this was exactly my thought process. It's a bit of an awkward approach, but I needed an async query, and `findBy` throws an error (instead of returning `null`) when no match is found. (And yes, I double-checked, and `queryBy` does produce a false negative when testing for the presence of this modal).

workbench/tests/renderer/changelog.test.js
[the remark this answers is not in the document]
emilyanndavis: Yes, and I think I figured out how to wrangle my editor settings so this _should_ happen automagically in the future.

workbench/src/renderer/components/Changelog/index.jsx
emilyanndavis: Good question. I did come across this pattern in our code, but I didn't follow it for two reasons:
1. I have a knee-jerk reaction whenever I see a click handler added to a link. Links are supposed to "just work" the way they are, and forcing them to behave in a non-default manner is often—though not always—a sign that there might be a simpler solution.
2. I misunderstood the purpose of `OPEN_EXTERNAL_URL` (even though it's clearly documented, so that's on me!) and thought it only opened links in a new (Electron) window. So when these links opened in a new window with nothing more than `target="_blank"`, I figured that was all I needed.

Looking around on the web just now, I came across some references to [`setWindowOpenHandler`](https://www.electronjs.org/docs/latest/api/web-contents#contentssetwindowopenhandlerhandler) in the Electron API. I'm not 100% sure it does what you're suggesting, or that it'd be "better" than our existing solution (heck, for all I know, it just adds click handlers to all the links behind the scenes), but I was curious to know if you had ever seen it or even tried it.

davemfish: > 1. I have a knee-jerk reaction whenever I see a click handler added to a link. Links are supposed to "just work" the way they are, and forcing them to behave in a non-default manner is often—though not always—a sign that there might be a simpler solution.

This is a good point!

>Looking around on the web just now, I came across some references to [setWindowOpenHandler](https://www.electronjs.org/docs/latest/api/web-contents#contentssetwindowopenhandlerhandler) in the Electron API. I'm not 100% sure it does what you're suggesting, or that it'd be "better" than our existing solution (heck, for all I know, it just adds click handlers to all the links behind the scenes), but I was curious to know if you had ever seen it or even tried it.

I never noticed this before. It looks potentially useful for standardizing some options we use when we want to open a new electron window. And I guess it can also "deny" the creation of the new electron window and do something else instead. The challenge might be that sometimes we want to open a new electron window, and sometimes we might prefer a default-browser window.

Cases where we use an electron window:
* About menu pages
* User's Guide link -- though this one could be changed to a non-electron window if we want to, it links to local html files, not the web version.

There are a couple reasons why I think it's a nicer experience to use the user's default browser for links on the web:
1. I might already be logged into github in my browser, so opening a github link there is more convenient than opening it in an electron window.
2. Our electron windows don't have any of the useful browser UI features, such as a Back button, address bar field, etc. So if I wanted to start on the linked page and then continue browsing around github from there, I'm better off in my own browser than in electron. 

I think these are the two main reasons why we setup that `OPEN_EXTERNAL_URL` handler and use it for the FAQ link to the Forum. But given the static html of the changelog, it would be more complicated to add a click handler.

emilyanndavis: These are all great points. I'm going to see if I can make it work.

workbench/src/renderer/components/Changelog/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+const { ipcRenderer } = window.Workbench.electron;
+
+export default function Changelog(props) {
+    const { t } = useTranslation();
+    const [htmlContent, setHtmlContent] = useState('');
+
+    // Load HTML from external file (which is generated by Python build process).
+    useEffect(() => {
+        async function loadHtml() {
```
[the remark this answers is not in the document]
emilyanndavis: Updated!

workbench/src/renderer/components/Changelog/index.jsx
emilyanndavis: Updated! All the links in the changelog modal now open in the browser.

workbench/src/renderer/components/Changelog/index.jsx
```diff
@@ -9,74 +9,95 @@ import pkg from '../../../../package.json';
...
+  useEffect(() => {
+    const openLinkInBrowser = (event) => {
+      event.preventDefault();
+      ipcRenderer.send(
+        ipcMainChannels.OPEN_EXTERNAL_URL, event.currentTarget.href
+      );
+    };
+    document.querySelectorAll('.link-external').forEach(link => {
+      link.addEventListener('click', openLinkInBrowser);
+    });
```
davemfish, line 66: This looks like a great solution!

davemfish (approved):

PR #1684, opened 2024-11-11
Wrap plugin env in quotes

dcdenu4:
## Description
When testing a recent installation the plugin env path was being stored in the settings config without escaped quotes for path spacing. When running the plugin model I was seeing:
```sh

[11:57:18.970] [undefined] about to run model with command: "C:/Users/ddenu/AppData/Local/Programs/InVEST 3.14.2.post397+g9c664c177 Workbench/resources/miniforge3/condabin/mamba.bat" run,--prefix C:\Users\ddenu\AppData\Local\Programs\InVEST 3.14.2.post397+g9c664c177 Workbench\resources\Miniforge3\envs\invest_plugin_schistosomiasis,--live-stream,invest,-vv,--taskgraph-log-level=INFO,--language "en",run,schistosomiasis,-d "C:\Users\ddenu\AppData\Roaming\invest-workbench\tmp\data-Da2SMk\datastack.json"
[11:57:20.346] [undefined] Not a conda environment: C:\Users\ddenu\AppData\Local\Programs\InVEST
```

Wrapping the env path follows similar practices to how we store the invest and mamba paths.

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -60,7 +60,7 @@ export function setupAddPlugin() {
             pyname: pluginPyName,
             type: 'plugin',
             source: pluginURL,
-            env: envPath,
+            env: `\"{envPath}\"`,
           }
         );
         logger.info('successfully added plugin');
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -60,7 +60,7 @@ export function setupAddPlugin() {
             pyname: pluginPyName,
             type: 'plugin',
             source: pluginURL,
-            env: `\"{envPath}\"`,
+            env: envPath,
           }
         );
         logger.info('successfully added plugin');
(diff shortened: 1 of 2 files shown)


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -60,7 +60,7 @@ export function setupAddPlugin() {
             pyname: pluginPyName,
             type: 'plugin',
             source: pluginURL,
-            env: envPath,
+            env: `\"{envPath}\"`,
```
emlys, line 63: Why do the quotes need to be escaped?

emlys: Oh I see, because they're a special character in the config JSON.

dcdenu4:
> IMO, it might be even better to wrap the path in quotes where it is used, instead of where it's stored, because that's where the quotes are needed.

I initially went that route, but noticed in the config file we had stored investExe and mamba similarly.
```json
{
	"nWorkers": -1,
	"taskgraphLoggingLevel": "INFO",
	"loggingLevel": "INFO",
	"language": "en",
	"investExe": "\"C:/Users/ddenu/AppData/Local/Programs/InVEST 3.14.2.post397+g9c664c177 Workbench/resources/invest/invest.exe\"",
	"mamba": "\"C:/Users/ddenu/AppData/Local/Programs/InVEST 3.14.2.post397+g9c664c177 Workbench/resources/miniforge3/condabin/mamba.bat\"",
	"core": {
		"port": 63911,
		"pid": 11296
	},
```
I like handling it on the code side as well though. I'll make that change and re-request.

PR #1685, opened 2024-11-13
Updated Readme

### LINKED ISSUE #670: InVEST Readme could use a refresh
From this discussion, https://github.com/natcap/cookbook/pull/1#issuecomment-929720320, it seems like there is consensus that the InVEST Readme could use a bit of a review and overhaul.

davemfish:
Added notes on Python environment and `natcap.invest` installation.
Added notes on how we link to external repos, sample-data, test-data, and User's Guide.

Some other small sections were removed because they were out of date.

Fixes #670 

It looks like all the tests are failing for an unrelated reason (see #1686 ) so I'm not going to address that in this PR.

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,15 +57,16 @@ Or on Windows, use the following instead from a CMD prompt::
     > .\env\bin\activate
 
 This makefile target is included for convenience. It uses ``conda`` and installs packages from ``conda-forge``.
+It also uses the `-p` flag with `conda create`, creating a `./env` folder containing the environment.
 
-Using a different environment name
-""""""""""""""""""""""""""""""""""
-If you prefer a different name for your environment, you may pass the environment name as
+Using a different environment folder name
+"""""""""""""""""""""""""""""""""""""""""
+If you prefer a different path for your environment, you may pass the environment path as
 a parameter to make::
 
     $ make ENV=myEnv env
 
-You could then activate the environment created at ``myEnv``.
+You could then activate the environment created at ``./myEnv``.
 
 
 Using a different environment management tool
--- a/README.rst
+++ b/README.rst
@@ -153,6 +153,13 @@ To build the user's guide::
 This will build HTML and PDF documentation, writing them to ``dist/userguide``
 and ``dist/InVEST_*_Documentation.pdf``, respectively.
 
+The User's Guide is maintained in a separate git reporsitory. InVEST will build
+the User's Guide with the commit defined in the ``Makefile``::
+
+   GIT_UG_REPO                 := https://github.com/natcap/invest.users-guide
+   GIT_UG_REPO_PATH            := doc/users-guide
+   GIT_UG_REPO_REV             := f203ec069f9f03560c9a85b268e67ebb6b994953
+
 
 API Documentation
 +++++++++++++++++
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -74,13 +70,25 @@ You could then activate the environment created at ``myEnv``.
...
+Installing ``natcap.invest`` from local source code
+"""""""""""""""""""""""""""""""""""""""""""""""""""
+From an activated virtual environment::
 
+    $ make install
+
+In practice, it can be convenient to use an "editable install" instead to avoid needing
+to re-install after making changes to source code::
+
+   $ pip install -e .
```
emilyanndavis, line 89: Yes! Took me weeks to figure this out 🤣

README.rst
```diff
@@ -74,13 +70,25 @@ You could then activate the environment created at ``myEnv``.
...
+
+``requirements.txt``, ``requirements-dev.txt`` and ``requirements-docs.txt`` list the python
+dependencies needed.
+
+Installing ``natcap.invest`` from local source code
+"""""""""""""""""""""""""""""""""""""""""""""""""""
+From an activated virtual environment::
 
+    $ make install
+
```
emilyanndavis, line 85: I would add a note here about the need to uninstall before reinstalling. Something along these lines:

```
To update your local installation after making changes to source code::
    $ pip uninstall natcap-invest
    $ make install
```
This'll get you unstuck if you (intentionally or unintentionally) use `make install`, and it underscores why an editable install (described next) is so compelling.

README.rst
```diff
@@ -74,13 +70,25 @@ You could then activate the environment created at ``myEnv``.
...
+In practice, it can be convenient to use an "editable install" instead to avoid needing
+to re-install after making changes to source code::
+
+   $ pip install -e .
+
+But note that any changes to non-Python files will still require compilation.
```
emilyanndavis, line 91: I think it would be worth explaining what this means in practice, i.e., if I've changed non-Python files, do I need to uninstall and reinstall `natcap.invest`?

emilyanndavis, line 91: I would also add a parenthetical: `non-Python files (excluding the workbench)`, just to avoid confusion.

README.rst
```diff
@@ -145,6 +153,13 @@ To build the user's guide::
 This will build HTML and PDF documentation, writing them to ``dist/userguide``
 and ``dist/InVEST_*_Documentation.pdf``, respectively.
 
+The User's Guide is maintained in a separate git reporsitory. InVEST will build
```
emilyanndavis, line 156: It is admittedly fun to say `reporsitory`, but we should probably fix the typo anyway. 😃

emilyanndavis (changes requested):
Thanks @davemfish, this is a vast improvement! Just had a few small suggestions.

README.rst
davemfish: It took me years!

README.rst
davemfish: I'm adding a note to explain that the workbench is not part of the `natcap.invest` python library.

davemfish:
> Thanks @davemfish, this is a vast improvement! Just had a few small suggestions.

Thanks for the suggestions, @emilyanndavis. I pushed some changes.

emilyanndavis (approved):
Looks good, thanks!

claire-simpson (approved):
Looks good. I left a small comment tangentially related to your changes but tbd if its relevant

README.rst
```diff
@@ -57,10 +56,7 @@ Or on Windows, use the following instead from a CMD prompt::
...
     > .\env\bin\activate
 
-This makefile target is included for convenience ... you may of course choose to
-manage your own virtual environment.  ``requirements.txt``,
-``requirements-dev.txt`` and ``requirements-docs.txt`` list the python
-dependencies needed.
+This makefile target is included for convenience. It uses ``conda`` and installs packages from ``conda-forge``.
 
 Using a different environment name
 """"""""""""""""""""""""""""""""""
```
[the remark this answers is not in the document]
davemfish: Yes, I think you're right. Since the `env` target of the Makefile calls conda like this:

`(CONDA) create -p $(ENV)`

It's using the `-p` flag, so `ENV` will always be interpreted as a path to a directory. 

>Maybe this is a me problem but the only way I could get a named environment was to specify ENV=path/to/miniconda/base /myEnv

That makes sense; not a you problem. I guess `make env` was designed with the opinion that prefixed environments are a good idea. But creating our own environments in whatever way we prefer as individuals is also a good idea!

claire-simpson (approved):

PR #1688, opened 2024-11-15
Spawn call was missing shell true option

dcdenu4:
A spawn call was missing the `shell: true` option, and is set to `false` by default. This was causing an error on Windows.

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -15,7 +15,8 @@ const logger = getLogger(__filename.split('/').slice(-1)[0]);
  * Spawn a child process and log its stdout, stderr, and any error in spawning.
  *
  * child_process.spawn is called with the provided cmd, args, and options,
- * and the windowsHide option set to true.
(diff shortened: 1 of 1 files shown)


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -27,7 +27,8 @@ const logger = getLogger(__filename.split('/').slice(-1)[0]);
  */
 function spawnWithLogging(cmd, args, options) {
   logger.info(cmd, args);
-  const cmdProcess = spawn(cmd, args, { ...options, windowsHide: true });
+  const cmdProcess = spawn(
+    cmd, args, { ...options, shell: true, windowsHide: true });
```
emlys, line 31: Suggest mentioning this in the docstring

dcdenu4: Added to docstring in https://github.com/natcap/invest/pull/1688/commits/276632db06803c29b00d469f65ac2df0b95f99e7

emlys (approved):
👍 , the same tests are failing upstream

PR #1690, opened 2024-11-19
enable gdal.UseExceptions in entrypoints and tests

emlys:
Enable gdal exceptions in the CLI. Update all models to expect and handle gdal exceptions where they are now raised.

I added a FutureWarning saying that we will eventually enable `gdal.UseExceptions` on import. I think that would be reasonable to do in the next minor release. And for anyone using invest with other packages that require `gdal.DontUseExceptions`, they can handle that themselves with a context manager.

## Description
Fixes #

## Checklist

### DIFF
--- a/src/natcap/invest/cli.py
+++ b/src/natcap/invest/cli.py
@@ -19,6 +19,7 @@ from natcap.invest import set_locale
 from natcap.invest import spec_utils
 from natcap.invest import ui_server
 from natcap.invest import utils
+from pygeoprocessing.geoprocessing_core import GDALUseExceptions
 
 
 DEFAULT_EXIT_CODE = 1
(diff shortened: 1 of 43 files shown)


src/natcap/invest/wind_energy.py
```diff
@@ -2590,67 +2594,25 @@ def _calculate_distances_land_grid(base_point_vector_path, base_raster_path,
...
-        output_feature = ogr.Feature(feature_def=target_layer.GetLayerDefn())
-
-        # Since the original feature is of interest add its fields and
-        # Values to the new feature from the intersecting geometries
-        # The False in SetFrom() signifies that the fields must match
-        # exactly
-        output_feature.SetFrom(point_feature, False)
-        target_layer.CreateFeature(output_feature)
-        target_vector.SyncToDisk()
-        target_layer.DeleteFeature(point_feature.GetFID())
```
emlys, line 2641: This raised an exception from GDAL for some reason I'm not totally clear on. It seemed simpler to refactor this to avoid the situation - rather than creating a new vector for each individual point, we can filter.

src/natcap/invest/wave_energy.py
```diff
@@ -1148,23 +1148,18 @@ def _copy_vector_or_raster(base_file_path, target_file_path):
...
         source_dataset = gdal.OpenEx(base_file_path, gdal.OF_VECTOR)
         target_driver_name = _VECTOR_DRIVER_NAME
-
-        # Raise an exception if the file can't be opened by GDAL
-        if source_dataset is None:
-            raise ValueError(
-                'File %s is neither a GDAL-compatible raster nor vector.'
-                % base_file_path)
-
+    else:
```
emlys, line 1158: `pygeoprocessing.get_gis_type` can return a bitwise AND of `RASTER_TYPE` and `VECTOR_TYPE`, indicating it can be opened as both, but I'm not sure if/how we'd want to handle that case.

src/natcap/invest/cli.py
```diff
@@ -218,267 +219,268 @@ def main(user_args=None):
...
+        # logging.getLogger().setLevel(logging.DEBUG)
+        # Also FYI: using logging.DEBUG means that the logger will defer to
+        # the setting of the parent logger.
+        logging.getLogger('natcap').setLevel(logging.DEBUG)
+
+        if args.subcommand == 'list':
+            # reevaluate the model names in the new language
+            importlib.reload(model_metadata)
             if args.json:
-                message = json.dumps(
```
emlys, line 388: The diff looks weird here, but it is really just indented one level.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -1737,15 +1737,16 @@ def extract_bathymetry_along_ray(
...
+                xoff=ix, yoff=iy,
+                win_xsize=win_size, win_ysize=win_size)
+        except RuntimeError as ex:
             location = {'xoff': ix, 'yoff': iy, 'win_xsize': win_size,
                              'win_ysize': win_size}
             raise ValueError(
-                f'got a {value} when trying to read bathymetry at {location}. '
-                'Does the bathymetry input fully cover the fetch ray area?')
+                f'Failed to read bathymetry at {location}. Does the bathymetry '
+                'input fully cover the fetch ray area?') from ex
```
phargogh, line 1749: Neat!  I was not aware of this `raise x from y` syntax.

src/natcap/invest/wave_energy.py
phargogh: Yep, it'd be ambiguous how to handle it, so I agree, better to raise an error.

phargogh (approved):
Looks great, thank you @emlys !

PR #1691, opened 2024-11-20
Plugins: update how investUsageLogger gets port information

dcdenu4:
## Description
Since we're now managing multiple invest servers on different ports, we're storing related port information in the `settingsStore` config. `investUsageLogger` was still using `process.env.PORT`. This PR attempts to transition that to the new setup.

I'd love some feedback and input on how to manage this. The solution I have here works, but feels a little tacked on. I couldn't think of a clean way for `investUsageLogger` to locally know if a plugin or core model was being run to get the port information there. So, I end up passing the port in to the `start` and `exit` functions from `setupInvestHandlers`.

### DIFF
--- a/workbench/tests/main/main.test.js
+++ b/workbench/tests/main/main.test.js
@@ -225,7 +225,8 @@ describe('createWindow', () => {
 });
 
 describe('investUsageLogger', () => {
-  const expectedURL = `http://127.0.0.1:${process.env.PORT}/api/log_model_start`;
+  const PORT = 3000;
+  const expectedURL = `http://127.0.0.1:${PORT}/api/log_model_start`;
   beforeEach(() => {
     // the expected response
     const response = {
@@ -244,12 +245,12 @@ describe('investUsageLogger', () => {
     const investStdErr = '';
     const usageLogger = investUsageLogger();
 
-    usageLogger.start(modelPyname, args);
+    usageLogger.start(modelPyname, args, PORT);
     expect(fetch.mock.calls).toHaveLength(1);
     expect(fetch.mock.calls[0][0]).toBe(expectedURL);
     const startPayload = JSON.parse(fetch.mock.calls[0][1].body);
 
-    usageLogger.exit(investStdErr);
+    usageLogger.exit(investStdErr, PORT);
     expect(fetch.mock.calls).toHaveLength(2);
     const exitPayload = JSON.parse(fetch.mock.calls[1][1].body);
 
--- a/workbench/src/main/investUsageLogger.js
+++ b/workbench/src/main/investUsageLogger.js
@@ -13,9 +13,9 @@ const PREFIX = 'api';
 export default function investUsageLogger() {
   const sessionId = crypto.randomUUID();
 
-  function start(modelPyName, args) {
+  function start(modelPyName, args, port) {
     logger.debug('logging model start');
-    fetch(`${HOSTNAME}:${process.env.PORT}/${PREFIX}/log_model_start`, {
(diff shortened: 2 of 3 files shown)


workbench/tests/main/main.test.js
```diff
@@ -225,7 +225,8 @@ describe('createWindow', () => {
 });
 
 describe('investUsageLogger', () => {
-  const expectedURL = `http://127.0.0.1:${process.env.PORT}/api/log_model_start`;
+  const PORT = 3000;
```
davemfish, line 228: This constant could be anything, right? We're mocking the actual fetch and only confirming that the URL is being constructed in the intended way. Should we add a comment here so it's clear we're not actually using a port in this test?

workbench/tests/main/main.test.js
```diff
@@ -225,7 +225,8 @@ describe('createWindow', () => {
 });
 
 describe('investUsageLogger', () => {
-  const expectedURL = `http://127.0.0.1:${process.env.PORT}/api/log_model_start`;
```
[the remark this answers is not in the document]
dcdenu4: Done in https://github.com/natcap/invest/pull/1691/commits/eeb6bfb382a192f0e04ae819bc3d871b1d6aa6b8

workbench/tests/main/main.test.js
dcdenu4: Added a comment in https://github.com/natcap/invest/pull/1691/commits/8331349f42fa6799ac42cb81ff165e42285542cb

workbench/package.json
```diff
@@ -8,10 +8,10 @@
     "start": "yarn build-main && yarn build:preload && concurrently --kill-others \"yarn serve\" \"electron .\"",
     "serve": "cross-env MODE=development node scripts/watch.js",
     "lint": "eslint --cache --color --ext .jsx,.js src",
-    "test": "cross-env PORT=56788 jest --runInBand --testPathIgnorePatterns /tests/binary_tests/ /tests/sampledata_linkcheck/",
+    "test": "cross-env jest --runInBand --testPathIgnorePatterns /tests/binary_tests/ /tests/sampledata_linkcheck/",
```
davemfish, line 11: `cross-env` is the program for setting environment variables in a cross-platform way, so we don't need that anymore.

emlys (commented):
Nothing to add to Dave's comments, I think this is a good solution. It could be a little cleaner to look up the port and `pyModuleName` in the settings store using the `modelRunName` - if `modelRunName` is in `settingsStore.get('plugins')`, then it's a plugin. There's an example of this on line 93 of setupInvestHandlers.js in the diff.

dcdenu4:
> Nothing to add to Dave's comments, I think this is a good solution. It could be a little cleaner to look up the port and `pyModuleName` in the settings store using the `modelRunName` - if `modelRunName` is in `settingsStore.get('plugins')`, then it's a plugin. There's an example of this on line 93 of setupInvestHandlers.js in the diff.

Thanks @emlys , I was looking into this for a solution because I thought it'd be cleaner too. But, for the `exit` function call we don't have access to `pyModuleName`, so it'd need to be passed in like it is for `start`. Or maybe we could have something like:
```js
let port;
function start() {
  port = "logic to get port from pyModuleName";
}
function end(){
  "use port, which should be initialized by the time this call occurs";
}
```

davemfish (approved):

dcdenu4:
Thanks Dave, I see some tests failing related to GDAL 3.10, which we fixed in `main`. So, I'll go ahead and merge this and we can merge `main` in the future.

PR #1697, opened 2024-11-25
#1166 Add support for zero-padded months in SWY

### LINKED ISSUE #1166: Support zero-padded month numbers in SWY?
A [user on the forums](https://community.naturalcapitalproject.org/t/precipitation-data-not-red/3385/3) wanted to have their SWY month numbers zero-padded, which the model can't currently handle.

So instead of a file named `et0-1.tif`, we could also support `et0-01.tif`.

This is a really interesting idea, though, since padding single-digit months would help to keep things in lexicographic order when listing files in a directory and wouldn't be difficult to implement as an option since we're using a regular expression to find these files anyways.

claire-simpson:
## Description
Updated the regex to allow users to have zero-padded month numbers in the monthly precipitation and evapotranspiration files. Now, filenames like et0_01.tif and precip01.tif are permitted. 

Fixes #1166

## Checklist

### DIFF
--- a/src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
+++ b/src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
@@ -659,15 +659,8 @@ def execute(args):
     output_align_list = [
         file_registry['lulc_aligned_path'], file_registry['dem_aligned_path']]
     if not args['user_defined_local_recharge']:
-        et0_dir_list = [
-            os.path.join(args['et0_dir'], f) for f in os.listdir(
-                args['et0_dir'])]
-        precip_dir_list = [
-            os.path.join(args['precip_dir'], f) for f in os.listdir(
-                args['precip_dir'])]
-
-        et0_path_list, precip_path_list = _get_monthly_file_lists(
-            N_MONTHS, et0_dir_list, precip_dir_list)
+        precip_path_list = _get_monthly_file_lists(N_MONTHS, args['precip_dir'])
+        et0_path_list = _get_monthly_file_lists(N_MONTHS, args['et0_dir'])
 
         input_align_list = (
             precip_path_list + [args['soil_group_path']] + et0_path_list +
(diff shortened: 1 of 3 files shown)


tests/test_seasonal_water_yield_regression.py
```diff
@@ -318,6 +318,30 @@ def tearDown(self):
...
+
+        n_months = 12
+
+        # Make fake paths with file names with zero-padded months
+        padded_et0_dir_list = [
+            os.path.join("/fake/path", "et0_" + str(mo).zfill(2) + ".tif")
+            for mo in range(1, n_months+1)]
+        padded_precip_dir_list = [
+            os.path.join("/fake/path", "Precip" + str(mo).zfill(2) + ".tif")
+            for mo in range(1, n_months + 1)]
```
davemfish, line 336: This is a very nice self-contained unit test!

Do you think it's worth constructing one of these lists with zero-padded numbers and one without? To demonstrate that `_get_monthly_file_lists` should accommodate both?

We could also consider a test that asserts a `ValueError` is raised in certain cases, since the function is designed with those cases in mind too.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -659,33 +659,15 @@ def execute(args):
     output_align_list = [
         file_registry['lulc_aligned_path'], file_registry['dem_aligned_path']]
     if not args['user_defined_local_recharge']:
-        precip_path_list = []
-        et0_path_list = []
-
         et0_dir_list = [
             os.path.join(args['et0_dir'], f) for f in os.listdir(
                 args['et0_dir'])]
         precip_dir_list = [
```
davemfish, line 665: Do you think it makes sense to do these list comprehensions within the `_get_monthly_file_lists` function? It seems like logically they are closely related to the rest of the code in the function, so it might be nice to encapsulate it there. And then the function would take `args['et0_dir']` as a direct input.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -1380,6 +1362,46 @@ def _aggregate_recharge(
...
+    """
+    et0_path_list = []
+    precip_path_list = []
+    
+    for month_index in range(1, n_months + 1):
+        month_file_pattern = re.compile(r'.*[^\d]0?%d\.[^.]+$' % month_index)
+
+        for data_type, dir_list, path_list in [
+                ('et0', et0_dir_list, et0_path_list),
+                ('Precip', precip_dir_list, precip_path_list)]:
```
davemfish, line 1389: Now that we have this convenient function to use, what do you think about having the function operate on only one directory, and return one list of filepaths? In other words, instead of using this `for` loop here, we make two separate calls to `_get_monthly_file_list` from up in `execute`.

davemfish (commented):
Hi @claire-simpson , this looks like a great solution! A nice convenient little function + unit test. I had a few fairly minor comments that perhaps we can discuss, but this looks great overall.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
claire-simpson: Yep good call!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
claire-simpson: Yes I like this, will update

tests/test_seasonal_water_yield_regression.py
claire-simpson: I believe the other tests all use non-zero-padded file names so this concept does get tested - but maybe its still good to have an explicit test for this with the new regex? Happy to add this. 

As for the fail cases - I'm thinking double zeros  (i.e. raising ValueError on precip001.tif). Did you have anything else in mind?

davemfish (approved):
Looks great, thank you!

PR #1706, opened 2024-12-11
SWY: check for `numpy.nan` pixels in routing functions

### LINKED ISSUE #1705: SWY cython `is_close` does not handle `numpy.nan`
We use an `is_close` function to check if pixel values are nodata. It's possible for a pixel value to be `numpy.nan`, so this function should allow for that.

davemfish:
We have updated our `is_close` and `numpy.isclose` calls in most places to check for `numpy.nan`, but this is one place where we were not doing that. It's plausible for the rasters being processed by this module to have `numpy.nan` pixels.

Fixes #1705 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -90,6 +90,9 @@ Unreleased Changes
     * Added support for zero padding in month numbers in ET and precipitation
       file names (i.e., users can now name their file Precip_01.tif).
       (https://github.com/natcap/invest/issues/1166)
+    * Fixed a bug where ``numpy.nan`` pixel values would not be correctly
+      detected as nodata in local recharge and baseflow routing functions.
+      (https://github.com/natcap/invest/issues/1705)
 * Urban Flood Risk
     * Fields present on the input AOI vector are now retained in the output.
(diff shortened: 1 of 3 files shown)


emlys (commented):
Thanks @davemfish! Just wondering if this has an effect of slowing down the model significantly?

davemfish:
> Thanks @davemfish! Just wondering if this has an effect of slowing down the model significantly?

@emlys I did some tests and found no noticeable slowdown of the model until I used a datastack that took 56 minutes to run on my machine. In that case, a version with the `isnan` checks took 57 minutes.

At a recent meeting we discussed this and agreed this is acceptable for now since doing the `isnan` checks fixes a bug where `nans` would otherwise pollute some math.

I also created #1714 and #1715 so that we can revisit a broader invest-wide solution and revert this change later if we want to.

emlys (approved):
@davemfish thanks, I was going to wait for the checks to pass, but it seems that the macos checks have been stalled all day (on this and other branches too), so feel free to merge given that all seemingly relevant checks have passed.

PR #1707, opened 2024-12-11
Distribute micromamba executable for plugins

emlys:
## Description
Fixes #

## Checklist


emlys:
@dcdenu4 I think the ReadTheDocs failure is unrelated

.github/workflows/build-and-test.yml
```diff
@@ -367,22 +367,21 @@ jobs:
...
-        if: matrix.os == 'macos-13'
-        run: bash Miniforge3-MacOSX-x86_64.sh -b -p dist/miniforge3
-
-      - name: Download mamba installer for distribution (Windows)
-        if: matrix.os == 'windows-latest'
-        run: curl -fsSLo Miniforge3-Windows-x86_64.exe "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Windows-x86_64.exe"
+        run: |
+          curl -Ls https://micro.mamba.pm/api/micromamba/osx-64/latest | tar -xvj bin/micromamba
+          mv bin/micromamba dist/
+          ./dist/micromamba --help
```
dcdenu4, line 375: Is this just to make sure the command works, and if it doesn't we'll fail / exit here? If so, it might be nice to have a short comment here.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -81,19 +81,19 @@ export function setupAddPlugin() {
...
+          micromamba,
+          ['run', '--name', envName, 'pip', 'install', `git+${pluginURL}`]
         );
         logger.info('installed plugin into its env');
         // Write plugin metadata to the workbench's config.json
-        const envInfo = execSync(`${mamba} info --envs`, { windowsHide: true }).toString();
+        const envInfo = execSync(`${micromamba} info --name ${envName}`, { windowsHide: true }).toString();
         logger.info(`env info:\n${envInfo}`);
-        const regex = new RegExp(String.raw`^${envName} +(.+)$`, 'm');
+        const regex = /env location : (.+)/;
```
dcdenu4, line 96: I don't think I've seen this syntax before. Is the `/.../` syntax for a regex, without needing it to be wrapped in a string?

workbench/electron-builder-config.js
```diff
@@ -36,8 +36,12 @@ const config = {
...
     },
     {
-      from: '../dist/Miniforge3',
-      to: 'Miniforge3',
+      from: '../dist/micromamba', // mac
+      to: 'micromamba',
+    },
+    {
+      from: '../dist/micromamba.exe', // windows
+      to: 'micromamba',
```
dcdenu4, line 44:
```suggestion
      to: 'micromamba.exe',
```

dcdenu4 (changes requested):
Hey @emlys, this will be great. The install size and time it takes to install InVEST is drastically better with this approach. I had a few questions, but mostly the Windows micromamba executable path still isn't quite right. I left a suggestion for what I think should do it. In the electron config when moving that file, making sure the `exe` extension remains in the filename.

.github/workflows/build-and-test.yml
emlys: Yes, just testing the executable. Which might not be necessary any more, but I think it's fine to leave in. Added a comment

workbench/src/main/setupAddRemovePlugin.js
emlys: Yes, it's a [regex literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#creating_a_regular_expression). My linter recommended this over a string.

workbench/electron-builder-config.js
emlys: Added this in another commit before I saw this

dcdenu4 (approved):
Hey @emlys , this works for me now on Windows. I was able to install plugin "foo" and run it. I see one Mac test failed, but it looks like maybe it timed out.

PR #1711, opened 2024-12-12
Upgrade `setup-micromamba` action version

### LINKED ISSUE #1709: GHA caches: `[object Promise]` appearing in cache keys
https://github.com/natcap/invest/actions/caches

Some cache keys contain the string `[object Promise]`, which probably means something's wrong with how we're building the key, and the caches might not be updating like we expect.

emlys:
## Description
Fixes #1709 
Not sure why exactly, but updating to the latest major version of `setup-micromamba` appears to fix the cache key issue. You can see in the `Setup environment` step, the cache key no longer has the extra string containing `[object Promise]`.

## Checklist


phargogh (approved):
Thanks @emlys !

PR #1716, opened 2024-12-17
Corrections to args spec regarding evapotranspiration and preciptation

### LINKED ISSUE #1693: investspec "evapotranspiration" wording needs to be changed
The Data Needs section of the User Guide for Urban Cooling and Annual Water Yield currently says this:

> Evapotranspiration ([raster](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/input_types.html#raster), units: mm, required): Map of evapotranspiration values.

Just calling it "evapotranspiration" is not precise enough, because there is a difference between _potential_ ET and _reference_ ET and _actual_ ET (all of which users might encounter in an ET data search), and both of these models' equations specify _reference_ ET (then they calculate potential and/or actual ET using Kc provided in the biophysical table). So the Data Needs section should specify "reference evapotranspiration", like it does for Seasonal Water Yield. 

So please update this in investspec - thanks!

### LINKED ISSUE #1694: SWY Data Needs updates to :investspec:
While reviewing SWY recently, I noticed two things that should be improved in the Data Needs section of the User Guide, but seem to be part of :investspec:, so need to be updated by the software team.

- Update the descriptions for _Precipitation directory_ and _ET0 directory_ to explain the “conditionally required” designation. I thought this was an error, until I looked at the UI and saw that they turn off if you enter a local recharge map. Please add something like "conditionally required, if not supplying a User-defined Recharge Layer".

- The _Monthly Alpha Table_ description is missing info on required fields.

Thanks!

davemfish:
Fixes #1693 
Fixes #1694  

## Checklist

### DIFF
--- a/src/natcap/invest/spec_utils.py
+++ b/src/natcap/invest/spec_utils.py
@@ -93,7 +93,7 @@ PRECIP = {
     "name": gettext("precipitation")
 }
 ET0 = {
-    "name": gettext("evapotranspiration"),
+    "name": gettext("reference evapotranspiration"),
     "type": "raster",
     "bands": {
         1: {
@@ -101,7 +101,7 @@ ET0 = {
             "units": u.millimeter
         }
     },
-    "about": gettext("Map of evapotranspiration values.")
+    "about": gettext("Map of reference evapotranspiration values.")
 }
 SOIL_GROUP = {
     "type": "raster",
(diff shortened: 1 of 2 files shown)


src/natcap/invest/spec_utils.py
```diff
@@ -93,15 +93,15 @@
...
+    "name": gettext("reference evapotranspiration"),
     "type": "raster",
     "bands": {
         1: {
             "type": "number",
             "units": u.millimeter
         }
     },
-    "about": gettext("Map of evapotranspiration values.")
+    "about": gettext("Map of reference evapotranspiration values.")
```
davemfish, line 104: I confirmed that the 3 models that use `spec_utils.ET0` all expect "reference" evapotranspiration. They are SWY, Urban Cooling, and AWY.

emlys (approved):
Thanks @davemfish!

PR #1728, opened 2025-01-06
Fixing an artifact name in release automation

### LINKED ISSUE #1718: Do the InVEST 3.14.3 release

phargogh:
I think this was the one issue encountered with the release automation, and this did trigger a rollback which I could manually reconstruct, so it all worked out!

Fixes #1718 

## Checklist

### DIFF
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -76,7 +76,7 @@ jobs:
                           Workbench-macOS-binary \
                           InVEST-sample-data \
                           InVEST-user-guide \
-                          "Source distribution"
+                          "Source distribution macos-13 3.12"
           do
             gh run download $RUN_ID --dir artifacts --name "$artifact"
           done
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -75,11 +75,17 @@ jobs:
           for artifact in Workbench-Windows-binary \
                           Workbench-macOS-binary \
                           InVEST-sample-data \
-                          InVEST-user-guide \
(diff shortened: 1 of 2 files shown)


.github/workflows/release-part-2.yml
```diff
@@ -76,7 +76,7 @@ jobs:
                           Workbench-macOS-binary \
                           InVEST-sample-data \
                           InVEST-user-guide \
-                          "Source distribution"
+                          "Source distribution macos-13 3.12"
```
dcdenu4, line 79: So we're just choosing an arbitrary source distribution to check, but choosing not to check both Windows and Mac source distributions? Just want to make sure I understand and then wondering if we should have a note about this being hard coded and that eventually we'll need to update this manually.

dcdenu4 (approved):
Hey @phargogh , just jumping in since I saw this. I had a quick question but otherwise happy to approve.

.github/workflows/release-part-2.yml
phargogh: I'm not 100% comfortable with hardcoding this, but yes, we need to pick one of the source distributions and python versions to upload here.  We're not checking anything, though, just downloading one of the `.tar.gz` artifacts and uploading it to the github release object.

Having said that, I dislike having a statically-defined string here that will need to be updated in the future, so I've fixed it in e5a3065e6

phargogh:
Thanks @dcdenu4! I think I've addressed your concern; the steps affected by this PR aren't actually testing anything, they're just downloading artifacts so that they can be uploaded to the github release object and pypi.

PR #1737, opened 2025-01-14
Align NDR rasters to DEM grid; Urban Cooling rasters to LULC grid

### LINKED ISSUE #1488: Align output NDR rasters to input DEM
The output rasters from NDR should align with the input DEM coordinates. They currently do not. This, it seems, is because the model aligns to [an intersection of the input raster and watershed bounding boxes](https://community.naturalcapitalproject.org/t/ndr-raster-output-alignment/4130/2).

The online documentation already states:
> Raster inputs may have different cell sizes, and they will be resampled to match the cell size of the DEM. Therefore, all model results will have the same cell size as the DEM.

I propose, further, that the most consistent, desirable intended behaviour is that rasters are resampled onto the subset of DEM coordinates within whatever the final bounding box is. We're already treating (some) DEM spatial properties as somehow special.

Another strong reason to motivate this change is a use case wherein a user is processing two (or more) non-conterminous regions. It is entirely plausible, and more computationally efficient, to process the two regions separately. As the current resampling stands, this will all but guarantee the two sets of results lie on non-commensurable coordinates. That is, the multiple sets of coordinates will differ by an offset that is not a multiple of the pixel spacing.

Another justification is the case when the user has already done the work of aligning their input rasters. The current approach all but guarantees their outputs will lie on some offset coordinates. This seems ... "rude" ;-)

Further motivation centres on the subsequent use and analysis of the model outputs. The user may have other data, similarly aligned, they want to merge with the NDR output. They've now got additional work to do to cast the output back onto their preferred coordinates. It is true that the model provides `aligned_` versions of the input rasters, but those inputs (required for driving the NDR model) are very likely not the only data the user has.

The user must ensure all input rasters are already in the same projected CRS. So, as with any software, there are already some expectations for data preparation on the user.

In short, the current approach, whilst attempting to resolve a "many to one" mapping of coordinates, actually inadvertently creates a many to many mapping when combining outputs across a number of runs (watersheds), and does this in a hard to predict and reconcile way, and also breaks what should be a "one to one" mapping of coordinates (in the case where all input rasters are aligned). Respecting the input DEM coordinates for the outputs will preserve the special case of "one to one" mapping for already aligned rasters, preserve the mapping across multiple runs (assuming the user has supplied consistent DEMs), and preserve the mapping a user might reasonably expect between their other raster data (that they will very likely have already aligned to the DEM) and the NDR model outputs.

emlys:
## Description
Fixes #1488 - align NDR rasters to the grid of the DEM using the `raster_align_index` parameter.

Also changing the `raster_align_index` in Urban Cooling to align to the LULC instead of ET0. I think that's what was intended based on the comment and that the LULC projection is used.

I took a look at other `align_and_resize` calls in invest, and these were the only ones that I thought need a change. Several others already align to the DEM or LULC grid, and others don't have an obvious target to align to.

This does change the model results slightly, but I think the differences are small enough it's not concerning

## Checklist


davemfish (approved):
This is great, thanks! The small changes in the results look completely reasonable to me.

PR #1738, opened 2025-01-14
Halt auto-scrolling of log output on user-initiated scrolling

### LINKED ISSUE #1533: Workbench log scroll bar skips back to bottom
When scrolling back up the log in a long-running model, the scroll bar behaves weirdly. It seems like it skips back to the bottom each time a new message is logged, even when my cursor is on the scroll bar and I'm dragging it upward. This makes it very difficult to look at earlier log messages or go back to the top.

emilyanndavis:
## Description
Fixes #1533. When a user-initiated scroll event is detected (based on scroll direction), auto-scrolling is halted.

Note this behavior will persist in a model tab as long as it is open. Once auto-scrolling has been halted in a given tab, it will not resume in that tab, even on subsequent model runs. This shouldn't cause any problems, but if it becomes annoying, we can implement a mechanism for resuming auto-scrolling.

## Checklist

### DIFF
--- a/workbench/src/renderer/components/LogTab/index.jsx
+++ b/workbench/src/renderer/components/LogTab/index.jsx
@@ -15,22 +15,46 @@ function LogDisplay(props) {
 
   const [autoScroll, setAutoScroll] = useState(true);
   const [prevScrollTop, setPrevScrollTop] = useState(0);
+  const [userInitiatedScroll, setUserInitiatedScroll] = useState(true);
+  let scrollHandlerTimer;
+
+  // `scrollOffsetThreshold` is used to determine when user has scrolled to bottom of window.
+  // It includes a buffer to account for imprecision stemming from rounding errors and/or
+  // scroll event throttling. 24px (the height of one line of log output) seems to be
+  // large enough to detect scroll events we want, and small enough to avoid false positives.
+  const scrollOffsetThreshold = 24;
 
   useEffect(() => {
     if (autoScroll) {
+      setUserInitiatedScroll(false);
       ref.current.scrollTop = ref.current.scrollHeight;
     }
   }, [props.logdata]);
 
-  // Listen for scroll events. If scroll direction is up,
-  // assume scrolling was user-initiated, and halt auto-scrolling.
-  const checkScrollDirection = () => {
-    const currentScrollTop = ref.current.scrollTop;
-    const scrollingUp = (currentScrollTop < prevScrollTop);
-    if (scrollingUp) {
-      setAutoScroll(false);
+  const handleScroll = () => {
+    if (scrollHandlerTimer) {
+      clearTimeout(scrollHandlerTimer);
     }
-    setPrevScrollTop(currentScrollTop);
+    scrollHandlerTimer = setTimeout(() => {
+      const currentScrollTop = ref.current.scrollTop;
+      if (userInitiatedScroll) {
+        if (autoScroll) {
+          // If user has scrolled up, halt auto-scrolling.
+          const scrollingUp = (currentScrollTop < prevScrollTop);
+          if (scrollingUp) {
+            setAutoScroll(false);
+          }
+        } else {
+          // If user has scrolled back to the bottom, resume auto-scrolling.
+          const currentScrollOffset = ref.current.scrollHeight - currentScrollTop;
+          if (Math.abs(ref.current.offsetHeight - currentScrollOffset) <= scrollOffsetThreshold) {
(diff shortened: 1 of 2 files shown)


emilyanndavis:
> Would it make sense to also resume auto-scrolling if the user scrolls all the way to the bottom? Similar to how terminal output scrolling works (at least in my terminal)

That sounds reasonable to me. I've come up with something that should provide the behavior you're looking for.

I should mention that pausing auto-scrolling only gives the user control over the relative position of the content in the log tab. It doesn't pause the process that adds content to the log tab. So, even when auto-scrolling is paused, you'll continue to see the size and relative position of the scroll bar handle changing as the height of the scrollable content continues to increase. If that becomes frustrating—or if what you were hoping for all along was a pause in the flow of log content—then I'll have to make some changes here. Please try it out and let me know what you think.

workbench/src/renderer/components/LogTab/index.jsx
```diff
@@ -13,15 +13,56 @@ const { ipcRenderer } = window.Workbench.electron;
...
+    }
   }, [props.logdata]);
 
+  const handleScroll = () => {
+    if (scrollHandlerTimer) {
+      clearTimeout(scrollHandlerTimer);
+    }
+    scrollHandlerTimer = setTimeout(() => {
+      const currentScrollTop = ref.current.scrollTop;
+      if (userInitiatedScroll) {
```
emlys, line 40: Since this is called after an `onScroll` event, should `userInitiatedScroll` always be true here?

workbench/src/renderer/components/LogTab/index.jsx
```diff
@@ -13,15 +13,56 @@ const { ipcRenderer } = window.Workbench.electron;
 function LogDisplay(props) {
   const ref = useRef();
 
+  const [autoScroll, setAutoScroll] = useState(true);
+  const [prevScrollTop, setPrevScrollTop] = useState(0);
+  const [userInitiatedScroll, setUserInitiatedScroll] = useState(true);
```
emlys, line 18: Should `userInitiatedScroll` start out `false`, since no scroll events have happened yet? And I think the `useEffect` will set it to false anyway when the first data appears

emlys (commented):
Thanks @emilyanndavis, I tried this out and it worked very smoothly! I just had a couple of questions about how the scrolling logic works.

workbench/src/renderer/components/LogTab/index.jsx
emilyanndavis: Not always. A `scroll` event is fired whenever the user scrolls, _and_ whenever the content is scrolled programmatically, such as when our code directly updates `ref.current.scrollTop` every time there's an update to the `logdata`. Unfortunately, there's no built-in data on a `scroll` event that can tell you whether it was triggered by a user interaction or by code.

So, what I've done here to differentiate scroll events is the following:
1.  Assume all scroll events are user-initiated. This is why `userInitiatedScroll` is initialized with `true` and is always reset to `true` just before `handleScroll` returns.
2. Filter out code-initiated scroll events. Immediately before updating `ref.current.scrollTop`, set `userInitiatedScroll` to `false`. We do this because, as soon as `ref.current.scrollTop` is updated, a `scroll` event is fired, which gets picked up by the `onScroll` listener, which calls `handleScroll`. When `handleScroll` is invoked, `userInitiatedScroll` is now `false`, which allows us to skip the special handling (calculations and comparisons) we need to determine whether to pause/resume auto-scrolling based on a user interaction. 

If there were a simple way to detect only user-initiated scroll events, the code would be simpler, too. 😃  Happy to add a comment if you think that would be helpful!

workbench/src/renderer/components/LogTab/index.jsx
emilyanndavis: You're right, it does get set to `false` as soon as the first log data comes in, so, in a way, it doesn't matter much what the initial value is. I initialized it to `true` to indicate that the assumption is that all scroll events are user-initiated unless otherwise specified, since we can't determine whether a given scroll event is user-initiated, but we _can_ flag programmatic scroll events as _not_ user-initiated.

workbench/src/renderer/components/LogTab/index.jsx
emlys: Ah okay, thanks for explaining! Yes, I would suggest adding a couple of comments describing this to make the code more maintainable for us in the future.

emlys (approved):
Thank you!

PR #1743, opened 2025-01-21
Support installing plugins from specific ref; and from local path

### LINKED ISSUE #1651: Plugins: should be able to install from different kinds of paths
Currently you can only install a plugin from the default branch of a git repo. We should support specifying a different branch or revision; and a local repo.

emlys:
## Description
Fixes #1651

## Checklist


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -52,32 +52,56 @@ function spawnWithLogging(cmd, args, options) {
...
 
+        const micromamba = settingsStore.get('micromamba');
         // Access plugin metadata from the pyproject.toml
         const pluginID = pyprojectTOML.tool.natcap.invest.model_id;
         const pluginName = pyprojectTOML.tool.natcap.invest.model_name;
         const pluginPyName = pyprojectTOML.tool.natcap.invest.pyname;
-
         // Create a conda env containing the plugin and its dependencies
         const envName = `invest_plugin_${pluginID}`;
         await spawnWithLogging(
```
emilyanndavis, line 107: Forgive me: this isn't related to this PR, and you might even have created an issue for this already, in which case, please ignore!

It looks like this call to create the micromamba env is referencing specific version restrictions for Python & GDAL. Seems like it would be a good idea to store these args someplace that's easier to keep up to date rather than hard-coding them here.

emilyanndavis (changes requested):
Looks good, @emlys! Just a handful of small suggestions.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -61,31 +72,85 @@ export default function PluginModal(props) {
...
+  if (installFrom === 'url') {
+    pluginFields = (
+      <>
+        <Form.Row>
+          <Form.Group as={Col} xs={7} className="mb-1">
+            <Form.Label>Git URL</Form.Label>
+            <Form.Control
+              id="url"
+              type="text"
+              placeholder="https://github.com/foo/bar.git"
```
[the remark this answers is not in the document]
emlys: Good suggestion!

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -52,32 +52,56 @@ function spawnWithLogging(cmd, args, options) {
...
+          );
+          // Read in the plugin's pyproject.toml, then delete it
+          pyprojectTOML = toml.parse(fs.readFileSync(
+            upath.join(tmpPluginDir, 'pyproject.toml')
+          ).toString());
+          fs.rmSync(tmpPluginDir, { recursive: true, force: true });
+        } else { // install from local path
+          logger.info(`adding plugin from ${path}`);
+          installString = path;
+          // Read in the plugin's pyproject.toml, then delete it
```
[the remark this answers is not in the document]
emlys: Ooh yep, thanks

workbench/src/main/setupAddRemovePlugin.js
emlys: You're absolutely right, we hard-coded those just to get something working, but as of https://github.com/natcap/invest/pull/1733 we'll read them from the `pyproject.toml`.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -61,31 +72,85 @@ export default function PluginModal(props) {
...
-          <Form.Control
-            id="url"
-            type="text"
-            placeholder={t('Enter Git URL')}
-            onChange={(event) => setURL(event.currentTarget.value)}
-          />
-          <Form.Text className="text-muted">
-            {t('This may take several minutes')}
-          </Form.Text>
+          <Form.Label className="col-form-label-lg" htmlFor="url">{t('Add a plugin')}</Form.Label>
```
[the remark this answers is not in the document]
emlys: I agree it makes more sense as a heading. I changed "Add a plugin" and "Remove a plugin" to headings, and made a label "Install from" for the install options dropdown. I made the dropdown in line with the "Install from" label, but I couldn't get it perfectly aligned - which do you think looks better?
image
image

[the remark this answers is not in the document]
emilyanndavis: Awesome! I would vote for the second option, with the label stacked above the form field, since that's consistent with the other fields in this form. If the alignment isn't quite working like you expect, you might try mimicking the structure you've used elsewhere in the form, i.e., adding `as={Col}` to the `Form.Group` and nesting that inside a `Form.Row` ... and/or seeing if you are using any utility classes (such as `mx-1`) you no longer need.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -61,31 +72,85 @@ export default function PluginModal(props) {
...
             {t('Add')}
           </Button>
+          <Form.Text className="text-muted">
+            {t('This may take several minutes')}
+          </Form.Text>
         </Form.Group>
         <hr />
         <Form.Group className="mb-3">
-          <Form.Label htmlFor="plugin-select">{t('Remove a plugin')}</Form.Label>
+          <h5 className="mb-3">{t('Remove a plugin')}</h5>
```
emilyanndavis, line 153: I think it's great that you converted this (and "Add a plugin") to a heading, but now the plugin dropdown is technically missing a label. A visible label (marked up as a `<label>`) is usually best, but we do have a few options in this case:
1. Add a `<Form.Label>` with `htmlFor="plugin-select"`. What that label says is up to you, as long as it describes the purpose of the form field ("Select a plugin", "Plugin to remove", etc.).
2. OR add an `aria-label` to the `<Form.Control>`, with its value set to something that describes the purpose of the form field (i.e., whatever you would use for a visible label, if you chose option 1).
3. OR add an `id` to the "Remove a plugin" heading, and add `aria-labelledby` to the `<Form.Control>`, with its value set to the heading's `id`. This tells assistive technologies (screen readers, voice input control, etc.) to associate the form field with the heading, essentially allowing the "Remove a plugin" heading to act both as a heading _and_ as a label for the form field.

emilyanndavis (changes requested):
Thanks for making those changes, @emlys! Just one more request!

workbench/src/renderer/components/PluginModal/index.jsx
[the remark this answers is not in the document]
emlys: Okay, I changed it to the second option, removed some unneeded styling, and it does look better now!

workbench/src/renderer/components/PluginModal/index.jsx
emlys: I went with option 1 (and changed the `id` to be a little more specific)

emilyanndavis: Great, thank you!

emilyanndavis:
Looks like all those label changes broke some of the workbench tests. As soon as those are fixed, I'll happily approve!

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -52,32 +52,56 @@ function spawnWithLogging(cmd, args, options) {
...
+            installString = `git+${url}@${revision}`;
+            logger.info(`adding plugin from ${installString}`);
+          } else {
+            installString = `git+${url}`;
+            logger.info(`adding plugin from ${installString} at default branch`);
+          }
+
+          // Create a temporary directory and check out the plugin's pyproject.toml,
+          // without downloading any extra files or git history
+          const tmpPluginDir = fs.mkdtempSync(upath.join(tmpdir(), 'natcap-invest-'));
```
dcdenu4, line 70: Is the trailing dash in `natcap-invest-` something common for a temp directory? Just curious!

dcdenu4 (changes requested):
Thanks @emlys ! I had one quick curiosity question. And just noting that I merged the other `git` PR which has introduced some conflicts!

workbench/src/main/setupAddRemovePlugin.js
emlys: It's a prefix for the temp dir name, so the actual name will be something like `natcap-invest-xkdfdjlksdad`

emilyanndavis (approved):

dcdenu4 (approved):
Thanks @emlys and @emilyanndavis !

PR #1744, opened 2025-01-22
Enhanced testing for Seasonal Water Yield (SWY) to test intermediate rasters

### LINKED ISSUE #1549: SWY tests do not test raster outputs

claire-simpson:
## Description
Improved testing for SWY:
- Added three new unit tests to validate functions that create intermediate raster outputs (`route_baseflow_sum`, `calculate_local_recharge`, and `_calculate_curve_number_raster`)
- Added assertion to existing `test_local_recharge_undefined_nodata`

Fixes #1549

## Checklist


tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+                0.5, stream_path,
+                os.path.join(self.workspace_dir, 'target_li_path.tif'),
+                os.path.join(self.workspace_dir, 'target_li_avail_path.tif'),
+                os.path.join(self.workspace_dir, 'target_l_sum_avail_path.tif'),
+                os.path.join(self.workspace_dir, 'target_aet_path.tif'),
+                os.path.join(self.workspace_dir, 'target_precip_path.tif'))
+        except Exception as e:
+            assert False, f"calculate_local_recharge raised an exception: {e}"
+
+    def test_calculate_local_recharge(self):
```
davemfish, line 1277: Since the setup in this test is almost identical to `test_local_recharge_undefined_nodata` do you think it would make sense to combine them? Basically taking the assertions you added in this test and applying them to the pre-existing test that had no assertions?

tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+        # note: obtained these arrays by running `route_baseflow_sum`
+        expected_b = numpy.array([[10.5, 0.9999998, 0],
+                                  [0.1422222, 2.2666667, 0]])
+        expected_b_sum = numpy.array([[16.916666, 1.8666663, 0],
+                                      [0.1422222, 2.5333333, 0]])
+
+        assert numpy.allclose(b, expected_b, equal_nan=True), \
+            f"Baseflow raster values do not match. Expected: {expected_b}, Got: {b}"
+        assert numpy.allclose(b_sum, expected_b_sum, equal_nan=True), \
+            f"b_sum raster values do not match. Expected: {expected_b_sum}, Got: {b_sum}"
```
davemfish, line 1434: Instead of the builtin `assert`, `unittest` and `numpy.testing` both provide lots of really convenient assertion methods. These are nice because their names tend to be very descriptive in what they're asserting, and they handle the string formatting when the assertion fails. For example, in this case we could use `numpy.testing.assert_allclose`.

And in other, simpler cases, we tend to use one of these: https://docs.python.org/3/library/unittest.html#assert-methods

tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+
+    def test_calculate_curve_number_raster(self):
+        """test `_calculate_curve_number_raster`"""
+        from natcap.invest.seasonal_water_yield import seasonal_water_yield
+        import pandas
```
davemfish, line 1439: Could we move the `pandas` import to the top of the module? I think it's only important to import the modules we are testing inside the scope of the test.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+        lulc_array[0, 0] = 2
+        make_raster_from_array(lulc_array, lulc_raster_path)
+
+        # make small soil raster
+        soil_group_path = os.path.join(self.workspace_dir, "soil_group.tif")
+        soil_groups = 4
+        soil_array = numpy.zeros((3, 3), dtype=numpy.int32)
+        for i, row in enumerate(soil_array):
+            row[:] = i % soil_groups + 1
+        make_raster_from_array(soil_array, soil_group_path)
```
davemfish, line 1454: I noticed this module has `make_soil_raster` and `make_lulc_raster` and `make_biophysical_csv` already defined. Does it make sense to use those functions to create the input data we need here? It may not, but I'm curious what you think.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+             "cn_b": 82, "cn_c": 0, "cn_d": 0},
+            {"lucode": 2, "Description": "lulc 3", "cn_a": 65,
+             "cn_b": 22, "cn_c": 1, "cn_d": 0}])
+
+        cn_path = os.path.join(self.workspace_dir, "cn.tif")
+
+        seasonal_water_yield._calculate_curve_number_raster(
+            lulc_raster_path, soil_group_path, biophysical_df, cn_path)
+
+        cn = pygeoprocessing.raster_to_numpy_array(cn_path)
```
davemfish, line 1470: Sometimes it can be nice to call this variable `actual_cn` so it's easy to see that it is the data we are testing.

davemfish (changes requested):
Thanks, @claire-simpson , I had a few minor suggestions. We can talk them over of course.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -1260,17 +1260,221 @@ def test_local_recharge_undefined_nodata(self):
...
+                [quickflow_path for i in range(12)], flow_dir_path,
+                [kc_path for i in range(12)], {i: 0.5 for i in range(12)}, 0.5,
+                0.5, stream_path,
+                os.path.join(self.workspace_dir, 'target_li_path.tif'),
+                os.path.join(self.workspace_dir, 'target_li_avail_path.tif'),
+                os.path.join(self.workspace_dir, 'target_l_sum_avail_path.tif'),
+                os.path.join(self.workspace_dir, 'target_aet_path.tif'),
+                os.path.join(self.workspace_dir, 'target_precip_path.tif'))
+        except Exception as e:
+            assert False, f"calculate_local_recharge raised an exception: {e}"
```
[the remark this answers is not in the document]
claire-simpson: Yep I was struggling with how to assert that there were no errors and landed on this unsatisfying solution.. but I think your note below makes the most sense to improve this test and remove redundancy!

tests/test_seasonal_water_yield_regression.py
claire-simpson: Thank you! This is super helpful. I have been seeing different types of assertions and wasn't sure which was best

tests/test_seasonal_water_yield_regression.py
claire-simpson: Yes definitely, I noticed another test importing pandas in this way so I followed suit, but I can change both

tests/test_seasonal_water_yield_regression.py
claire-simpson: Yes I noticed these as well! I chose not to use them as the `make_soil_raster` and `make_lulc_raster` both create 100x100 pixel rasters, which seemed a bit unnecessarily large to me as then I'd also be hard-coding a 100x100 pixel raster target. 

The issue I had with `make_biophysical_csv` is that the column names are uppercase vs. the `_calculate_curve_number_raster` function expected lowercase. I could alternatively change the `make_biophysical_csv` function output to be lowercase, however the function as a whole _does_ accept uppercase column names (as shown in the sample data)  so I elected not to change the `make_biophysical_csv` function

davemfish: Cool, that all makes sense to me, thanks!

davemfish (approved):
This looks great, thanks!

PR #1748, opened 2025-01-23
Schedule release workflow to run first tuesday of the month

### LINKED ISSUE #1736: Run the bugfix release workflow on a monthly schedule
We decided to try scheduling the bugfix release workflow to run automatically on a monthly basis. We can opt to skip or delay, but going to try out having a default release schedule. 

Should be possible to automate this by setting a [`schedule`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) for the action.

emlys:
## Description
Fixes #1736 
I tested the bash if/else expression, but not the whole workflow since it can't be run on a branch other than `main`. There's a good chance that some part of this won't work perfectly, but it might be easy enough to fix on the first scheduled release day, rather than trying to set up a test.

## Checklist

### DIFF
--- a/.github/workflows/release-part-1.yml
+++ b/.github/workflows/release-part-1.yml
@@ -1,6 +1,12 @@
 name: Release (Part 1 of 2)
 
 on:
+  schedule:
+    # every tuesday at 10:20am PST (6:20pm UTC)
+    # cron doesn't support "first tuesday of the month", so we will use github
+    # actions syntax below to skip tuesdays that aren't in the first week.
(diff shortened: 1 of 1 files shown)


phargogh (approved):
Nice!  This is a really neat solution.  I particularly like the bash conditional in there, and I agree it'll be easy enough to just resolve any issues when they arise with our upcoming release.  But it looks good, so I don't imagine there will be any issues :)

PR #1749, opened 2025-01-24
Wind Energy: prevent excessive logging of wind data

### LINKED ISSUE #1497: Workbench cannot handle wind energy's debug logging from Taskgraph
Rob gave us a datastack that includes a large `wind_data_path` CSV input. It's about 5MB, which is 2-3x larger than our sample data CSV. When `Taskgraph` logging level is set to `DEBUG`, a single line that is more than 18 million characters long is logged. This causes the Workbench renderer to consume an unreasonable amount of memory and ultimately crash. 

The line begins,

```
2024-01-04 14:56:17,380 (taskgraph.Task) Task.is_precalculated(1166) DEBUG other_arguments: [[{0: {'LONG': -124.255, 'LATI': 46.9703, 'REF_LAM': 8.42, 'K': 1.92, 
```

That `other_argument` is the `wind_data` argument in,

```python
compute_density_harvested_task = task_graph.add_task(
    func=_compute_density_harvested_fields,
    args=(wind_data, parameters_dict, number_of_turbines,
          wind_data_pickle_path),
    target_path_list=[wind_data_pickle_path],
    task_name='compute_density_harvested_fields')
```

For some reason, instead of passing the CSV as an argument to `_compute_density_harvested_fields`, we first read the CSV into a dict and pass that, so `wind_data` here is a dict. I don't see any reason why we need to do this, so passing the CSV into the function would avoid this logging problem.

Should the workbench more gracefully handle logging lines that are this long? Perhaps.

emilyanndavis:
## Description
Fixes #1497. Specifically, we now pass a file path (instead of a dict) to `_compute_density_harvested_fields` to prevent Taskgraph from logging the entire wind data dictionary (which can use enough memory to crash the Workbench).

## Checklist

### DIFF
--- a/src/natcap/invest/wind_energy.py
+++ b/src/natcap/invest/wind_energy.py
@@ -694,14 +694,6 @@ def execute(args):
 
     number_of_turbines = int(args['number_of_turbines'])
 
-    # Create a list of the biophysical parameters we are looking for from the
-    # input csv files
-    biophysical_params = [
-        'cut_in_wspd', 'cut_out_wspd', 'rated_wspd', 'hub_height',
-        'turbine_rated_pwr', 'air_density', 'exponent_power_curve',
-        'air_density_coefficient', 'loss_parameter'
-    ]
-
     # Read the biophysical turbine parameters into a dictionary
     turbine_dict = validation.get_validated_dataframe(
         args['turbine_parameters_path'],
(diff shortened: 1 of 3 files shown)


src/natcap/invest/wind_energy.py
```diff
@@ -694,14 +694,6 @@ def execute(args):
...
     number_of_turbines = int(args['number_of_turbines'])
 
-    # Create a list of the biophysical parameters we are looking for from the
-    # input csv files
-    biophysical_params = [
-        'cut_in_wspd', 'cut_out_wspd', 'rated_wspd', 'hub_height',
-        'turbine_rated_pwr', 'air_density', 'exponent_power_curve',
-        'air_density_coefficient', 'loss_parameter'
-    ]
-
```
emilyanndavis, line 704: This change is unrelated; I just noticed `biophysical_params` wasn't being referenced anywhere, so I removed it.

tests/test_wind_energy.py
```diff
@@ -348,20 +348,18 @@ def test_calculate_npv_levelized_rasters(self):
         price_list = [0.10, 0.10, 0.10, 0.10, 0.10]
 
         srs = osr.SpatialReference()
-        srs.ImportFromEPSG(3157) #UTM Zone 10N
+        srs.ImportFromEPSG(3157)  # UTM Zone 10N
         projection_wkt = srs.ExportToWkt()
         origin = (443723.127327877911739, 4956546.905980412848294)
-        pos_x = origin[0]
-        pos_y = origin[1]
```
emilyanndavis, line 355: No substantive changes to the tests, just some PEP8 formatting cleanup and removal of unused variables.

davemfish (approved):
Looks like a great solution to me!

PR #1763, opened 2025-01-31
Automate Code Signing of Windows Binary

### LINKED ISSUE #1580: Automate code-signing as part of our release process
In #1523 we were able to set up a yubikey to sign our binaries.  The process for signing binaries was pretty manual: download binary, sign binary, upload binary.  It would be fantastic to be able to do this in the cloud as a part of our usual build automation.

phargogh:
This PR restores automated codesigning as a part of our GHA-based build process.  Configuration (minus secrets) has been provided for a queue implemented as a GCS-backed cloud function, and a worker running as a debian:bookworm `systemd` service.  Our github actions configuration and secrets have been updated to use the new codesigning.

I would also like to add that there are several enhancements that I think we should look into long term, including:
* Using identity federation to manage access to the function instead of the primitive token-based approach used
* Triggering the `enqueue` operation based on GCS events instead of an HTTP request

I have removed the old windows codesigning stuff that is no longer relevant.  I think `make codesign_windows` should still work on a windows computer that has a yubikey (and corresponding minidriver) available, but we aren't using it for our build automation any longer.  I left the recipe in the Makefile in case we need to sign locally.

This has been tested and is working on my fork ([example signed dev build binary](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/phargogh/3.14.2.post367+g7ff35670c/workbench/invest_3.14.2.post367+g7ff35670c_workbench_win32_x64.exe))

Fixes #1580 

## Checklist


codesigning/gcp-cloudfunc/main.py
```diff
@@ -0,0 +1,175 @@
...
+        return jsonify('Invalid token'), 403
+
+    if request.method != 'POST':
+        logging.info('Rejecting request due to invalid HTTP method')
+        return jsonify('Invalid request method'), 405
+
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(CODESIGN_DATA_BUCKET)
+
+    logging.debug('Data POSTed: %s', data)
```
emlys, line 76: I'd suggest the modern f-string style
```suggestion
    logging.debug(f'Data POSTed: {data}')
```

codesigning/gcp-cloudfunc/main.py
```diff
@@ -0,0 +1,175 @@
...
+            'basename': os.path.basename(next_file_url),
+            'gs-uri': unquote(next_file_url.replace(
+                f'{GOOGLE_PREFIX}/', 'gs://')),
+        }
+        logging.info(f'Dequeued {next_file_url}')
+        return jsonify(data)
+
+    elif data['action'] == 'enqueue':
+        url = data['url']
+        logging.info('Attempting to enqueue url" %s', url)
```
emlys, line 102:
```suggestion
        logging.info(f'Attempting to enqueue url {url}')
```

codesigning/gcp-cloudfunc/main.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if not url.endswith('.exe'):
+            logging.info("Rejecting URL because it doesn't end in .exe")
+            return jsonify('Invalid URL to sign'), 400
+
+        if not url.startswith(GOOGLE_PREFIX):
+            logging.info('Rejecting URL because it does not start with %s',
+                         GOOGLE_PREFIX)
```
emlys, line 110:
```suggestion
            logging.info(f'Rejecting URL because it does not start with {GOOGLE_PREFIX}')
```

codesigning/gcp-cloudfunc/main.py
```diff
@@ -0,0 +1,175 @@
...
+        mday, mmonth, myear = response.headers['Last-Modified'].split(' ')[1:4]
+        modified_time = datetime.datetime.strptime(
+            ' '.join((mday, mmonth, myear)), '%d %b %Y')
+        if modified_time < datetime.datetime(year=2024, month=6, day=1):
+            logging.info('Rejecting URL because it is too old')
+            return jsonify('File is too old'), 400
+
+        with get_lock():
+            # first, check to see if the file has already been signed.
+            signed_files_list = codesign_bucket.blob('signed_files.json')
```
emlys, line 143: Does this track all files that have ever been signed?

codesigning/gcp-cloudfunc/main.py
```diff
@@ -0,0 +1,175 @@
...
+    """Handle requests to this GCP Cloud Function.
+
+    All requests must be POST requests and have a JSON body with the following
+    attributes:
+
+        * token: a secret token that matches the ACCESS_TOKEN environment
+            variable that is defined in the cloud function configuration.
+        * action: either 'enqueue' or 'dequeue'
+
+    If the action is 'enqueue', the request must also have a 'url' attribute.
```
emlys, line 62: It might be helpful to clarify in the docstring that the `url` is a url to the .exe file that's being enqueued for signing

codesigning/signing-worker/natcap-codesign.service
```diff
@@ -0,0 +1,33 @@
...
+Description=NatCap Code Signing for Windows EXE Binaries
+User=natcap-codesign
+Group=natcap-codesign
+WorkingDirectory=/tmp
+
+
+[Service]
+# Run in the foreground
+Type=simple
+Restart=no  # change back to always
```
emlys, line 32: Does this need to be changed back to always?

codesigning/signing-worker/natcap-codesign.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Args:
+        filename (str): The local file to upload.
+        path_on_bucket (str): The path to the file on the GCS bucket, including
+            the ``gs://`` prefix.
+
+    Returns:
+        ``None``
+    """
+    subprocess.run(['gsutil', 'cp', filename, path_on_bucket], check=True)
```
emlys, line 77: How does the signing worker authenticate to `gsutil`?

codesigning/signing-worker/playbook.yml
```diff
@@ -0,0 +1,116 @@
...
+        repo: "deb http://deb.debian.org/debian {{ ansible_distribution_release }}-backports main"
+        filename: bookworm-backports.list
+
+    - name: Install osslsigncode from backports
+      ansible.builtin.apt:
+        update_cache: true
+        default_release: "{{ ansible_distribution_release }}-backports"
+        pkg:
+          # The backports version is needed because the version in bookworm has
+          # a critical bug in it that prevents it from working with our
```
emlys, line 57: Is this bug something that's likely to be fixed soon/ever? It could be helpful to include a link to the bug, if possible, so we can check on it in the future

codesigning/signing-worker/.gitignore
```diff
@@ -0,0 +1 @@
+natcap-servers-1732552f0202.json
```
emlys, line 1: Does this file get created as part of the signing process?

codesigning/enqueue-binary.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+import sys
+
+import requests
+
+DATA = {
+    'token': os.environ['ACCESS_TOKEN'],
+    'action': 'enqueue',
+    'url': sys.argv[1].replace(
+        'gs://', 'https://storage.googleapis.com/'),
```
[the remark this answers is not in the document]
phargogh: Well, they're two different ways of identifying where files live.  Within the google ecosystem, the `gs://` path is required, like when working with the `gsutil` command line, but web browsers of course need `https://`.  In our case, it just so happens that the location within the bucket is the same when accessed by both schemes.  I'm not sure this helps ... please do let me know if it doesn't!

codesigning/gcp-cloudfunc/main.py
phargogh: No, just the ones that have been signed by the service.  This seemed OK since if we request to sign an already-signed file, signing should just overwrite the existing signature with the new one.  In the future, it would be a nice enhancement to use `osslsigncode` to check and see if a file has already been signed and then add it to this file if so.

codesigning/gcp-cloudfunc/main.py
phargogh: Updated the docstring in 0e643894c14d50a1e1f3673fcbd0060c5acab1e5

codesigning/signing-worker/natcap-codesign.service
phargogh: Ha!  Yes, I think so.  Patched in 33a60e539

codesigning/signing-worker/natcap-codesign.py
phargogh: There's a line in the Ansible playbook that does the authentication, so the service script pretty much "just works"

codesigning/signing-worker/playbook.yml
phargogh: Yes!  It was a bug introduced in osslsigncode 2.5 (which is the version in `debian:bookworm`) and was fixed in 2.6.  The version in `debian:bookworm-backports` is version 2.9, which is the latest released.  I've updated the comments here to reflect this in dfc275ae1

codesigning/signing-worker/.gitignore
phargogh: No, it's a private key from google cloud. Part of the ansible playbook copies the key into a known place, so I just wanted to prevent git from tracking it.  I added a comment to the gitignore about this in 3801a351f

dcdenu4 (changes requested):
Thanks @phargogh for getting this workflow set up so quickly. It looks good to me, just a few grammatical suggestions.

I had a few general questions for my understanding though:
- The signed windows binary will end up on GCS, but the github artifact will be unsigned?
- Does this play nicely with the automated release workflows, thinking about the above?

phargogh:
@dcdenu4 great point about using the signed binary in the release!  I just updated this in 782dbd141 so that the binary is fetched from GCS at release time.  Because we have a human review step, it's pretty likely that the binary will be signed before the release part 2 automation is run, but it isn't a guarantee if we are super quick on the draw for merging the release PR or if there are a lot of binaries in the queue to sign.

FYI @emlys

codesigning/gcp-cloudfunc/main.py
emlys: Ah okay! If this is only used for releases, and not signing regular builds on main, this file shouldn't get prohibitively large

dcdenu4 (approved):
Thanks @phargogh , looks good to me!

emlys (approved):

phargogh:
OK!  @dcdenu4 I'm re-requesting your review because there have been a few nontrivial changes in response to our discussion during the team meeting.

@emlys The only changes since your approval have been bugfix-related, so pretty minor :)

dcdenu4 (changes requested):
Thanks @phargogh ! This looks good to me, just a conflict with History :)

dcdenu4 (approved):

PR #1765, opened 2025-02-04
Ignore compiled InVEST models.

dcdenu4:
For development of the plugins feature branch, plugins use pip to install said feature branch. For Windows machines this means we expect a C++ compiler, which isn't always guaranteed. For the release of plugins, this won't be an issue because we'll point to a conda forge feedstock with everything pre-built.

Obviously, this will break testing. Happy to try and skip testing the compiled models. An alternative to making this change in the feature branch is to create another branch from `feature/plugins` and make this change there. And then have plugins point to that branch instead of this one. Curious what you think @emlys .

Created issue https://github.com/natcap/invest/issues/1766 to remind us to revert this for the release.


emlys (changes requested):
Thanks @dcdenu4! I do think it would be helpful to skip the tests so we're not seeing a bunch of failures, that should be possible by adding to the top of each test module:
```
pytestmark = pytest.mark.skip()
```
https://docs.pytest.org/en/latest/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module

emlys (approved):
Thanks!

PR #1776, opened 2025-02-06
Add unit tests for 5 models

### LINKED ISSUE #1603: Add more unit tests for models that could use them
Some of our older models' test suites only test the final model results, which is effective at catching bugs, but not very helpful in identifying where a bug originated. It would be great if all models had least some unit tests that check intermediate results. Not every single function needs a unit test - some are pretty trivial.

The task:
Identify functions that contain significant logic and/or occur at a critical point in the model. Write unit tests for these functions and/or assert that intermediate results are produced correctly.

Some suggestions to get started with:
- Annual water yield: unit test for `fractp_op`
- Forest carbon: unit test for `_calculate_tropical_forest_edge_carbon_map`

This is complete when all invest models have at least a couple of unit tests that will help us identify issues in the intermediate steps of the models.

claire-simpson:
## Description
I've added unit tests for Annual Water Yield, Carbon, Blue Carbon, Coastal Vulnerability, and Crop Production (regression). These tests assume that the functions we have currently are producing the correct output. 

Fixes #1603 (in part)

## Checklist


emlys (approved):
Wow, thanks @claire-simpson! This must have taken a lot of work and it's going to make these test suites much more comprehensive. My only suggestion would be: if we update test suites for more models and find that the helper functions for creating test data, checking results, etc become redundant, should we abstract those out to a test_utilities module? But I don't think that needs to happen in this PR.

PR #1786, opened 2025-02-15
Log to python from C++ extensions

emlys:
## Description
Reimplement the progess logging that we previously had.
Also add `except +` to cython header definitions to make sure that any C++ exceptions propagate.
I tried it out and confirmed that the logging shows up in the model log file.

## Checklist


phargogh (approved):
Looks good to me!  I have no comments or critiques, and it's fun to know that we might start seeing some C++ exceptions being promoted by Cython to Python exceptions

Issue #1666, opened 2024-10-24
Plugins: show warning to user if some resources were not removed during removal of plugin


dcdenu4:
I'm seeing:
```
 WARNING conda.gateways.disk.delete:unlink_or_rename_to_trash(182): Could not remove or rename C:\Users\ddenu\miniforge3\envs\invest_plugin_schistosomiasis\Library\share\proj\proj.db.  Please remove this file manually (you may need to reboot to free file handles)
[1] WARNING conda.gateways.disk.delete:unlink_or_rename_to_trash(182): Could not remove or rename C:\Users\ddenu\miniforge3\envs\invest_plugin_schistosomiasis\Scripts\invest.exe.  Please remove this file manually (you may need to reboot to free file handles)
[1] [11:24:47.964] [undefined] successfully removed plugin
```

dcdenu4:
It could be because I had the environment activated in another shell.

dcdenu4:
Actually, what was happening was I had a few zombie InVEST processes running that were never properly ended. These must have been using the plugin env, because once I killed these I was able to force remove the plugin env.

emlys:
I don't think we can guarantee that no processes outside the workbench are using the env, but we can at least show this warning to the user

emlys:
@dcdenu4 I could not reproduce this on mac - I tried launching the workbench, opening another shell and running `invest serve` in the plugin environment, and then removing the plugin via the workbench interface. The removal succeeded and didn't show any warnings. The plugin environment directory remained, but all packages were removed from it. 

Maybe this specific failure mode is unlikely enough in regular use that we don't need to handle it?

PR #1790, opened 2025-02-18
Display plugin install status, and any installation errors

### LINKED ISSUE #1653: Plugins: provide step indicator during installation
When installing a plugin, it can take quite a bit of time. And though the spinny wheel is a great indicator it can feel like nothing is happening. I wonder if we could add a "M out of N steps completed" progress callback? I think that updating would help reassure users it's working.

emlys:
## Description
Fixes #1653 

Moved the spinner into the button, and changed the button text to reflect status of the installation. There are 3-4 different steps:
image

The "Remove" button is the same, but it just displays "Removing..." because there's only one step.

If there is an error, it now displays the error message and a button to open the log file:
image

## Checklist


dcdenu4 (approved):
Thanks @emlys , this looks good to me. Some nifty electron communication.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -71,6 +75,7 @@ export function setupAddPlugin() {
           // Create invest_base environment, if it doesn't already exist
           // The purpose of this environment is just to ensure that git is available
           if (!fs.existsSync(baseEnvPrefix)) {
+            mainWindow.webContents.send('plugin-install-status', i18n.t('Creating base environment...'));
```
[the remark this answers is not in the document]
dcdenu4: Is the behavior the same because we only have one window, mainWindow? But otherwise `e.reply` would be communicating with whatever window sent it?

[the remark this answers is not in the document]
davemfish: Yeah that sounds right, Doug. Here's the docs on `event.reply`

>reply Function - A function that will send an IPC message to the renderer frame that sent the original message that you are currently handling. You should use this method to "reply" to the sent message in order to guarantee the reply will go to the correct process and frame.
https://www.electronjs.org/docs/latest/api/structures/ipc-main-event

emlys:
@davemfish, I tried `event.reply` and got an error that `event.reply is not a function`. At the suggestion of [this stackoverflow thread](https://stackoverflow.com/questions/57038073/ipcmain-event-reply-is-not-a-function) I'm using `event.sender.send`, which works just fine. It's not clear why but I'm guessing that event objects in `handle` are different than event objects in `on`. The other places we use `event.reply`, it's in an `on` handler.

davemfish:
> @davemfish, I tried `event.reply` and got an error that `event.reply is not a function`. At the suggestion of [this stackoverflow thread](https://stackoverflow.com/questions/57038073/ipcmain-event-reply-is-not-a-function) I'm using `event.sender.send`, which works just fine. It's not clear why but I'm guessing that event objects in `handle` are different than event objects in `on`. The other places we use `event.reply`, it's in an `on` handler.

Kind of annoying that they are different, and not well-documented. Maybe eventually we will use `handle` for everything, if that's what electron wants us to do.

emlys:
@dcdenu4, I realized that it's very similar to display any errors during uninstall, so I added that to the PR (see also https://github.com/natcap/invest/issues/1666. This doesn't cover warnings, just errors). 

Also changed the uninstall command from `mamba remove` to `mamba env remove` - this removes the environment directory as well as the packages in it.

dcdenu4 (approved):
Thanks @emlys. I'll leave it to @davemfish for the final sign off and merge.

I was curious what your current thinking is for testing plugins? It seems like it'd be good to have tests for the install / uninstall error handling, but I know that getting things working in puppeteer was a blocker early on.

workbench/tests/renderer/plugin.test.js
```diff
@@ -73,13 +73,14 @@ describe('Add plugin modal', () => {
...
+    await findByText('Adding...');
     await waitFor(() => {
+      const calledChannels = spy.mock.calls.map((call) => call[0]);
       expect(calledChannels).toContain(ipcMainChannels.ADD_PLUGIN);
     });
-    // expect the plugin dialog to have disappeared
-    await waitFor(() => expect(queryByRole('dialog')).toBeNull());
+    // close the modal
+    const overlay = await findByRole('dialog');
+    await act(async () => { await userEvent.click(overlay); });
```
davemfish, line 83: From the look of it, this should work just the same as `await userEvent.click(overlay);`

It looks like there are a few other cases in this file where `userEvent` is sometimes `awaited`, sometimes not, and sometimes wrapped in `act`. I think the cleanest and most reliable pattern is to always do,

`await userEvent.*`

workbench/tests/jest.overrides.js
```diff
@@ -23,6 +23,9 @@ if (global.window) {
   global.window._ = (x) => x;
 
   global.window.fetch = fetch;
+
+  global.window.electronAPI = {};
+  global.window.electronAPI.onPluginInstallStatus = () => {};
```
davemfish, line 28: I don't think we're using `window.electronAPI` anymore, right?

davemfish (changes requested):
Hey @emlys , this looks really good. I left a few small comments that might resolve some warnings you may see when running the tests.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -15,52 +15,63 @@ const { ipcRenderer } = window.Workbench.electron;
...
+  const [url, setURL] = useState('');
+  const [revision, setRevision] = useState('');
+  const [path, setPath] = useState('');
+  const [installErr, setInstallErr] = useState('');
+  const [uninstallErr, setUninstallErr] = useState('');
   const [pluginToRemove, setPluginToRemove] = useState(undefined);
-  const [loading, setLoading] = useState(false);
+  const [installLoading, setInstallLoading] = useState(false);
+  const [uninstallLoading, setUninstallLoading] = useState(false);
+  const [statusMessage, setStatusMessage] = useState(undefined);
```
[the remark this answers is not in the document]
emlys: Thanks @davemfish. I intended to use strings for all these but clearly I missed some.

workbench/tests/renderer/plugin.test.js
emlys: I was attempting to resolve the `Warning: An update to Transition inside a test was not wrapped in act(...).` messages, but you're right this doesn't seem to be doing anything here. I added `await` to all the `userEvent` calls in the test suite, and that removes the `act()` warnings... even when the warning is coming from a different line of code.

davemfish: Great! I think that's the right approach. My understanding is that even though React issues those `not wrapped in act` warnings, the correct approach for us, since we're using `testing-library` APIs, is basically never to wrap in `act`, but to look for something async that we forgot to `await`.

davemfish (approved):

PR #1792, opened 2025-02-19
Upgrade Workbench JS dependencies

### LINKED ISSUE #1470: Workbench node package upgrades
Let's first do security & bugifx updates. And then assess the level of breaking changes that would come from minor/major upgrades. We can address those upgrades one at a time, or however it makes sense.

emilyanndavis:
## Description
(Partially) resolves #1470

These upgrades resolve the vast majority of vulnerabilities identified by dependabot & `yarn audit`.

The following have been upgraded to new major versions:
- `electron-builder` to v24 (which is either the latest stable version or possibly one major version behind... the details of their releases are a bit confusing)
- `vite` to v5, which will continue getting patches as long as v6 is the latest (and we'll plan to upgrade to v6 soon)
- `i18next-conv` and `i18next-parser` to v15 and v9, respectively (latest)

The following have been upgraded to the latest patch versions without moving to new major versions:
- `jest`
- `jest-environment-jsdom`
- `@babel/cli`
- `eslint`

I also added `@babel/core` as a dev dependency to remove a bunch of peer dependency warnings.

I tried upgrading `puppeteer-core`, but I haven't been able to do so without also breaking the puppeteer test. I've created #1800 for upgrading `puppeteer-core` alongside `electron`.

I also created #1798 for upgrading `react-boostrap` and `Bootstrap`, which will definitely cause breaking changes.

## Checklist


davemfish (approved):
Thanks!

PR #1795, opened 2025-02-19
Plugin modal: tell user to install MSVC redistributable if needed

### LINKED ISSUE #1794: Plugins: handle VS C++ Windows compiler issue
In later versions of micromamba for Windows, it's expected the install system has the VS C++ compiler on path. There's a few options for us to consider here:
 
- Can we rollback Micromamba to before there was need for a Windows compiler?
- Including the installer as part of the Windows build?
- Install it for Windows when trying to install a plugin?

Some context from James on our previous history with VCRedist

>RE: VCRedist, this is not surprising, but it is frustrating.  [We used to have to do this for InVEST](https://github.com/natcap/invest/blob/12f1d1af9d9158b0af7fcaf1be2a5ce75ad4fb93/installer/windows/invest_installer.nsi#L505) as a part of our NSIS installed (not sure if it was technically legal for us to do so, though), so maybe we will just need to figure out vcredist installation as a part of the micromamba install?
Related [discussion on github](https://github.com/mamba-org/mamba/issues/2928), along with a [scripted solution](https://github.com/mamba-org/mamba/issues/2928#issuecomment-2639284945)

I was able to use the following script silently, except that it prompted a pop up of whether I wanted to allow this program to make changes to my comptuer.
```
Invoke-WebRequest -URI "https://aka.ms/vs/17/release/vc_redist.x64.exe" -OutFile "$env:Temp\vc_redist.x64.exe"; Start-Process "$env:Temp\vc_redist.x64.exe" -ArgumentList "/quiet /norestart" -Wait; Remove-Item "$env:Temp\vc_redist.x64.exe"
```

emlys:
## Description
Fixes #1794 

When a user opens the "Manage plugins" modal, check for the existence of the DLL that micromamba needs. If not found at the expected path, display a message and point the user to download and install the MSVC redistributable, which installs the DLL.

## Checklist


dcdenu4 (changes requested):
Thanks @emlys ! This worked great for me in a bare bones Windows VM. I had one suggestion about add a `/norestart` flag to the exe install call.

Otherwise the only thing that jumps out is that when we finish the 3rd party installer the user is left at the InVEST main page. Would it be possible / reasonable to open the manage plugin window on completion of the MSVCpp install?

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -75,6 +83,16 @@ export default function PluginModal(props) {
     });
   };
 
+  const downloadMSVC = () => {
+    setShowPluginModal(false);
+    ipcRenderer.invoke(ipcMainChannels.DOWNLOAD_MSVC).then(
+      () => {
+        console.log('msvc done');
```
dcdenu4, line 90: Do we want to keep this `console.log`?

dcdenu4 (changes requested):
Thanks @emlys, looks good. There was a `console.log` call that I wasn't sure we wanted to keep.

workbench/src/renderer/components/PluginModal/index.jsx
emlys: Oops, fixed!

dcdenu4 (approved):
Thanks for this @emlys !

PR #1814, opened 2025-02-27
Expose runoff proxy average as a user-specified parameter in NDR RPI calculation

### LINKED ISSUE #1741: NDR runoff proxy index depends on watershed extent
@phawthorne raised this issue:
The runoff proxy index is calculated as $RPI_i = RP_i / RP_{av}$ (dividing runoff proxy by average runoff proxy within the watersheds). This is problematic because average runoff proxy $RP_{av}$ changes depending on which pixels/watersheds are included in the raster. @phawthorne said they have even considered inserting fake pixel values into the RP raster in order to change the $RP_{av}$. We could expose the $RP_{av}$ parameter and allow the user to specify it instead.

claire-simpson:
## Description
Allows a user to specify their own average runoff proxy value. More specifically, this PR modifies the calculation of the runoff proxy index (RPI) in the NDR model by exposing the average runoff proxy as an optional user-defined parameter. Previously, RP average was dynamically computed based on the included watershed pixels, leading to inconsistencies when model inputs changed (e.g., different watershed extents or climate scenarios). By allowing users to specify a runoff proxy average, this change ensures consistent RPI values across scenarios, enabling better comparisons between baseline and future conditions.

Changes:

- Introduces new optional model parameter `Runoff Proxy Average`
- Updates RPI calculation to use the user-defined average runoff proxy (if specified) or the automatically, dynamically calculated average if not specified by the user
- Adds a test and updates documentation

Fixes #1741

UG PR: https://github.com/natcap/invest.users-guide/pull/162

## Checklist


tests/test_ndr.py
```diff
@@ -435,3 +486,102 @@ def test_masking_invalid_geometry(self):
...
             expected_array,
             pygeoprocessing.raster_to_numpy_array(target_raster_path))
+
+    def test_synthetic_runoff_proxy_av(self):
+        """
+        Test that runoff proxy average is calculated correctly if
+        (1) the user specifies a runoff proxy average value,
+        (2) the user does not specify a value so the runoff proxy average
+            is auto-calculated
+        """
```
dcdenu4, line 496: I appreciate the thoroughness of this testing! There is a part of me that wonders if simpler is better here though. I think we're just updating the `_normailze_raster` function and `runoff_proxy_index` is essentially just a wrapper for that function. So could the updated tests for this just be testing the `_normalize_raster` function directly? That way we don't have to run the whole model and worry about setting up other inputs. Curious what you think!

dcdenu4 (changes requested):
Thanks @claire-simpson! I had a few comments and suggestions, mostly around testing.

src/natcap/invest/ndr/ndr.py
```diff
@@ -1198,28 +1222,34 @@ def validate(args, limit_to=None):
...
     """Calculate normalize raster by dividing by the mean value.
 
     Args:
         base_raster_path_band (tuple): raster path/band tuple to calculate
             mean.
         target_normalized_raster_path (string): path to target normalized
             raster from base_raster_path_band.
+        user_provided_mean (float, optional): user-provided runoff proxy
+            average. If provided, this value will be used instead of
+            computing the mean from the raster.
```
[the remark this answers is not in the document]
claire-simpson: Makes sense!

src/natcap/invest/ndr/ndr.py
```diff
@@ -576,6 +593,12 @@ def execute(args):
         args['biophysical_table_path'],
         **MODEL_SPEC['args']['biophysical_table_path'])
 
+    runoff_proxy_av = args.get("runoff_proxy_av")
+    try:
+        runoff_proxy_av = float(runoff_proxy_av) if runoff_proxy_av not in [None, ""] else None
+    except ValueError:
+        raise ValueError(f"Invalid value for runoff_proxy_av: {runoff_proxy_av}. Expected a number.")
```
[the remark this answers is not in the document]
claire-simpson: The issue I was having was if no value was entered for `runoff_proxy_av` in the workbench, it was getting assigned ''. So the reason I added this block was to ensure that `None` would instead be assigned to `runoff_proxy_av` if the user didn't specify anything (as I use `runoff_proxy_av` as an arg in the `_normalize_raster` function call). (And I also wanted to handle the case where a user didn't include `runoff_proxy_av` in the args dict when using the Python API). There is probably a cleaner way to do this though? Thoughts?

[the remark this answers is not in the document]
claire-simpson: I think you are right that this is overkill though and I at least don't have to wrap this in a try/except block!

tests/test_ndr.py
claire-simpson: Works for me! My initial thought was to go for the "safer" option in case something changed with the _normalize_raster RPI task (and this was helpful testing when RP_av = None, '', or a number) but I don't think thats actually necessary and does add extra code to the test!

src/natcap/invest/ndr/ndr.py
```diff
@@ -576,6 +593,10 @@ def execute(args):
         args['biophysical_table_path'],
         **MODEL_SPEC['args']['biophysical_table_path'])
 
+    # Ensure that if user doesn't explicitly assign a value, runoff_proxy_av = None
+    runoff_proxy_av = args.get("runoff_proxy_av")
+    runoff_proxy_av = float(runoff_proxy_av) if runoff_proxy_av not in [None, ""] else None
```
dcdenu4, line 598: That makes sense related to your comment about using this below as a kwarg and expecting it to be `None`. Since the empty string `""` and `None` are "falsey", a subjectively more readable form of this could be:
```python
runoff_proxy_av = args.get("runoff_proxy_av")
runoff_proxy_av = float(runoff_proxy_av) if runoff_prox_av else None
```
- https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false

src/natcap/invest/ndr/ndr.py
```diff
@@ -576,6 +593,10 @@ def execute(args):
         args['biophysical_table_path'],
         **MODEL_SPEC['args']['biophysical_table_path'])
 
+    # Ensure that if user doesn't explicitly assign a value, runoff_proxy_av = None
```
dcdenu4, line 596:
```suggestion
    # Ensure that if user doesn't explicitly assign a value,
    # runoff_proxy_av = None
```

I think this was a few characters over 80 for line length.

tests/test_ndr.py
```diff
@@ -435,3 +488,42 @@ def test_masking_invalid_geometry(self):
...
+        make_simple_raster(runoff_proxy_path, runoff_proxy_array)
+        target_rpi_path = os.path.join(self.workspace_dir, "out_raster.tif")
+
+        # Calculate RPI with user-specified runoff proxy average
+        runoff_proxy_av = 2
+        ndr._normalize_raster((runoff_proxy_path, 1), target_rpi_path,
+                              user_provided_mean=runoff_proxy_av)
+
+        actual_rpi = pygeoprocessing.raster_to_numpy_array(target_rpi_path)
+        expected_rpi = pygeoprocessing.raster_to_numpy_array(
```
dcdenu4, line 515: I think we can avoid the `pgp.raster_to_numpy_array` call here since we already have `runoff_proxy_array` created above!

tests/test_ndr.py
```diff
@@ -435,3 +488,42 @@ def test_masking_invalid_geometry(self):
...
+
+        numpy.testing.assert_allclose(actual_rpi, expected_rpi)
+
+        # Now calculate RPI with auto-calculated RP average
+        ndr._normalize_raster((runoff_proxy_path, 1), target_rpi_path,
+                              user_provided_mean=None)
+
+        actual_rpi = pygeoprocessing.raster_to_numpy_array(target_rpi_path)
+
+        expected_rpi = pygeoprocessing.raster_to_numpy_array(runoff_proxy_path)
```
dcdenu4, line 526: Same as comment above, we already have the array defined that we can use.

tests/test_ndr.py
```diff
@@ -16,6 +17,58 @@
...
+    origin = (461251, 4923445)
+    pixel_size = (30, -30)
+    no_data = -1
+
+    pygeoprocessing.numpy_array_to_raster(
+        array, no_data, pixel_size, origin, projection_wkt,
+        base_raster_path)
+
+
+def make_simple_vector(path_to_shp):
```
dcdenu4, line 42: I don't think we're calling this anymore, so we can go ahead and remove it.

tests/test_ndr.py
```diff
@@ -16,6 +17,58 @@
     os.path.dirname(__file__), '..', 'data', 'invest-test-data', 'ndr')
 
 
+def make_simple_raster(base_raster_path, array):
```
dcdenu4, line 20: Since we're only using this function once now do you think we could move the projection and metadata setup to the test itself, along with the `numpy_array_to_raster` call? We typically have a rule of 3 around whether duplicated code should be placed into its own function / module, etc... That's not a hard rule for everything, but in this case I think it makes sense. If we expanded on these tests and were creating a bunch of rasters using the same set up, this function would definitely be worth having.

dcdenu4 (dismissed):
Thanks @claire-simpson, a few more comments and suggestions!

tests/test_ndr.py
claire-simpson: Ok good rule to know, thanks!

src/natcap/invest/ndr/ndr.py
claire-simpson: One last consideration - what if the user wants to set the runoff proxy average to 0? I think thats the one reason why keeping the "not in ["None, ""]" makes sense.. However I think its unlikely a user would enter a 0 (unless they just didn't want to normalize the runoff_proxy). What do you think?

src/natcap/invest/ndr/ndr.py
```diff
@@ -1198,28 +1222,34 @@ def validate(args, limit_to=None):
...
-        value_mean /= value_count
+    if user_provided_mean is None:
+        value_sum, value_count = pygeoprocessing.raster_reduce(
+            function=lambda sum_count, block: (  # calculate both in one pass
+                sum_count[0] + numpy.sum(block), sum_count[1] + block.size),
+            raster_path_band=base_raster_path_band,
+            initializer=(0, 0))
+        value_mean = value_sum
+        if value_count > 0:
+            value_mean /= value_count
```
davemfish, line 1250: Do you think it would be useful to log the calculated mean? I could imagine a user wanting to let the model calculate the mean for a baseline scenario, but then wanting to re-use that same value for another scenario where the runoff proxy raster has been modified. If I'm understanding the use-case correctly.

davemfish (commented):
@claire-simpson , this looks like a really nice enhancement. I had one suggestion about adding a log message, but other than that this looks good to go.

src/natcap/invest/ndr/ndr.py
claire-simpson: Great idea @davemfish!

davemfish (approved):

PR #1799, opened 2025-02-19
pin sqlite<3.49.1

emlys:
## Description
See #1797 

The issue is resolved by avoiding the latest sqlite version. sqlite is a conda package and not a python package, so there is no change to `requirements.txt`, just to the github actions workflow. If we determine that there is a real issue with this version, we might update the invest conda-forge feedstock to reflect that. For now, just getting the builds working.
## Checklist
- [ ] Updated HISTORY.rst and link to any relevant issue (if these changes are user-facing)
- [ ] Updated the user's guide (if needed)
- [ ] Tested the Workbench UI (if relevant)

PR #1817, opened 2025-02-28
Pinning sqlite for binary builds

dcdenu4:
## Description
Related to #1799. It looks like the dependency wasn't being handled in the binary build environment that is set up. This PR moves the restriction to the top conda dependency environment variable, to avoid missing it in the subsequent set ups. Even though it's not necessary for each environment being set up. Happy to chat about this and whether this is something that would be better maintained in `scripts/convert-requirements-to-conda-yml.py` or elsewhere.

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -328,7 +328,7 @@ jobs:
             requirements-dev.txt
             requirements-docs.txt
             constraints_tests.txt
-          requirements: ${{ env.CONDA_DEFAULT_DEPENDENCIES }} pandoc 
+          requirements: ${{ env.CONDA_DEFAULT_DEPENDENCIES }} pandoc
 
       - name: Make install
         run: make install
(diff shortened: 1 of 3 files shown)


phargogh (changes requested):
`sqlite` is interesting because it is a `conda-forge` package, but it is not a package on PyPI.  Anything we put into `requirements.txt` MUST be a PyPI-installable dependency.

Instead, I think this addition should go in our requirements-to-conda-forge conversion script at https://github.com/natcap/invest/blob/main/scripts/convert-requirements-to-conda-yml.py

emlys:
Sorry to jump in, but wasn't this done in https://github.com/natcap/invest/pull/1799?

dcdenu4:
Ha, thanks for jumping in @emlys . I completely missed this and it explains why the binary build wasn't catching it.

phargogh (approved):

phargogh:
Ah wait shoot, I merged before @emlys could review.  I'm sorry about that!  @dcdenu4 please feel free to revert the merge

emlys:
Looks good to me!

PR #1831, opened 2025-03-07
Carbon model: remove REDD, change scenario names, change output filenames, fix bug in model_spec

### LINKED ISSUE #1087: Carbon model requires same dimensions for current/future/REDD scenarios - missing documentation
A user posted about getting an error: `the raster dimensions of [‘lulc_cur_path’, ‘lulc_fut_path’] are not equivalent.` I can't find this requirement documented anywhere. 

 https://community.naturalcapitalproject.org/t/future-lulc/3084

### LINKED ISSUE #1758: Remove REDD scenario option from Carbon Storage and Sequestration model
We should consider removing the REDD Scenario section. It confuses users and is redundant with the "Calculate Sequestration/Future LULC" option, AFAICT. @newtpatrol do you agree or am I missing some context?

And it might actually be better/more accurate to call the scenarios "Baseline" instead of "Current" and "Alternate" or "Other" or "Scenario" or something instead of "Future". This would signal that users can compare scenarios from the past with the current, or whatever they want (not limited to current and future only).

I created this issue in this repo because @emilyanndavis stated that she "suspect[s] it would affect more than the UI". Also, it looks like the `invest-workbench` repo was archived.

### LINKED ISSUE #1825: Rename Carbon model output files
There was a discussion in InVEST Issue #1758 where @newtpatrol noted that the model output names are a bit inconsistent, specifically:

>If the model is called "Storage and Sequestration", then why aren't the "tot_c" outputs named "storage" and the "delta" outputs named "sequestration"?

Is this a change we'd like to make? @newtpatrol  @jagoldstein @rmgriffin

### LINKED ISSUE #1827: Carbon model valuation bug: discount and price change are entered as a ratio (0 - 1) but are treated as percentage?
In the Carbon Storage and Sequestration model, users are prompted to enter the `discount rate` and `annual price change` as a ratio (value between 0 and 1), however in the `_calculate_valuation_constant` function, both of these values are described as being entered as a percentage and the code treats these like:

`ratio` = (
        1 / ((1 + `discount_rate `/ 100) *
             (1 + `rate_change` / 100)))

which possibly implies they're supposed to be entered as percentage values between 0-100. 

This equation is consistent with the [UG equation](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/carbonstorage.html#equation-carbon-value), however overall the UG is a bit ambiguous as well, with an example value in Data Needs>Annual Market Discount Rate given as a percentage. 

Should the division by 100 occur, or is that meant to convert values the equation is expecting to be in the range of 0-100 into ratios in the range of 0-1? If the equation is correct, we should at least clarify the language around ratio and percentage.

claire-simpson:
## Description
Fixes #1758
- Removes the REDD scenario option
- Renames "Current" to "Baseline" scenario
- Renames "Future" to "Alternate" scenario
- Checks to ensure that the Alternate year is after the Baseline year

Fixes #1825
- Changes output filename prefixes
    - `tot_c` -> `c_storage`
    - `delta` -> `c_change`
    
Fixes #1827
- Fixes bug where `discount_rate` and `rate_change` were incorrectly treated as ratios instead of percentages 

Fixes #1087
- Updates MODEL_SPEC to clarify that alternate LULC raster must be aligned with baseline raster

Sample Data PR: https://bitbucket.org/natcap/invest-sample-data/pull-requests/38
(There is no test data for the carbon model)

UG PR: https://github.com/natcap/invest.users-guide/pull/163

## Checklist


tests/test_carbon.py
```diff
@@ -158,43 +158,32 @@ def test_carbon_full(self):
...
-                ('Total cur', 0.86),
-                ('Total fut', 0.28),
-                ('Total redd', 0.08),
-                ('Change in C for fut', -0.58),
-                ('Change in C for redd', -0.78),
-                ('Net present value from cur to fut', -34.22),
-                ('Net present value from cur to redd', -46.02),
+                ('Baseline Carbon Storage', 0.86),
+                ('Alternate Carbon Storage', 0.28),
+                ('Change in C for Alternate', -0.58),
```
claire-simpson, line 185: Thought this could be a good opportunity to make the html report a bit clearer but thse row names could possibly be improved. For reference, the UG description is:

> So, [Baseline Carbon Storage] is the total amount of carbon within the area of interest for the [baseline] LULC map, summarized from the raster result [c_storage_bas].tif; [Alternate Carbon Storage] is the total amount of carbon within the area of interest for the [alternative] LULC map, summarized from the raster result c_storage_alt.tif; and Change in C for [Alternate] is the total amount of change between baseline and alternate within the area of interest.

src/natcap/invest/carbon.py
```diff
@@ -46,68 +45,43 @@
...
-            "about": gettext(
-                "A map of LULC for the REDD policy scenario. "
-                "All values in this raster must have corresponding entries in "
-                "the Carbon Pools table. Required if REDD Scenario Analysis "
-                "is selected."),
-            "name": gettext("REDD LULC")
+                "A map of LULC for the alternate scenario, which must occur "
+                "after the baseline scenario. All values in this raster must "
+                "have corresponding entries in the Carbon Pools table. "
+                "This raster must align with the Baseline LULC raster. "
```
dcdenu4, line 82: I was surprised to see that the model does not handle this. It looks like the model checks up front that the different rasters all have the same cell size and dimensions. And validation will check that they spatially overlap, but not that they align too.

I'm trying to think through why we would NOT want to `pygeoprocessing.align_and_resize_raster_stack` based on the `baseline` raster. Not an issue necessary to address here, but maybe worth a new issue and input from that same carbon group.

src/natcap/invest/carbon.py
```diff
@@ -263,35 +215,27 @@
...
+    'c_soil_alt': 'c_soil_alt.tif',
+    'c_dead_alt': 'c_dead_alt.tif',
 }
 
 _TMP_BASE_FILES = {
-    'aligned_lulc_cur_path': 'aligned_lulc_cur.tif',
-    'aligned_lulc_fut_path': 'aligned_lulc_fut.tif',
-    'aligned_lulc_redd_path': 'aligned_lulc_redd.tif',
+    'aligned_lulc_bas_path': 'aligned_lulc_bas.tif',
+    'aligned_lulc_alt_path': 'aligned_lulc_alt.tif',
```
dcdenu4, line 238: Interesting artifact of the model previously calling `pgp.align_and_resize_raster_stack`? It doesn't look like these are currently used anywhere. The plot thickens...

src/natcap/invest/carbon.py
```diff
@@ -707,17 +649,14 @@ def _generate_report(raster_file_set, model_args, file_registry):
 
         # value lists are [sort priority, description, statistic, units]
         report = [
-            (file_registry['tot_c_cur'], 'Total cur', carbon_units),
-            (file_registry['tot_c_fut'], 'Total fut', carbon_units),
-            (file_registry['tot_c_redd'], 'Total redd', carbon_units),
-            (file_registry['delta_cur_fut'], 'Change in C for fut',
+            (file_registry['c_storage_bas'], 'Baseline Carbon Storage',
```
dcdenu4, line 652: Much more readable.

tests/test_carbon.py
dcdenu4: They read well to me, but perhaps Stacie or Jesse would have feedback. You could share an example html report with them in Slack to get some quick thoughts!

tests/test_carbon.py
```diff
@@ -373,9 +355,8 @@ def _make_simple_lulc_raster(base_raster_path):
                              out_carbon_stock_path)
 
         # open output carbon stock raster and check values
-        actual_carbon_stock = gdal.Open(out_carbon_stock_path)
-        band = actual_carbon_stock.GetRasterBand(1)
-        actual_carbon_stock = band.ReadAsArray()
+        actual_carbon_stock = pygeoprocessing.raster_to_numpy_array(
```
dcdenu4, line 358: Nice.

tests/test_carbon.py
```diff
@@ -463,6 +430,21 @@ def test_missing_keys_valuation(self):
...
              'rate_change',
-             'lulc_cur_year',
-             'lulc_fut_year'])
+             'lulc_bas_year',
+             'lulc_alt_year'])
         self.assertEqual(invalid_keys, expected_missing_keys)
+
+    def test_invalid_lulc_years(self):
+        """Test that providing an Alternate LULC year < Baseline LULC error
+        raises a ValueError"""
```
dcdenu4, line 439:
```suggestion
        """Test Alternate LULC year < Baseline LULC year raises a ValueError."""
```
I think this get's us to less than 80 characters and the one-liner docstring sentence!

tests/test_cli.py
```diff
@@ -221,7 +221,7 @@ def test_validate_carbon(self):
         validation_output = stdout_stream.getvalue()
         # it's expected that these keys are missing because the only
         # key we included was the workspace_dir
-        expected_warning = [(['carbon_pools_path', 'lulc_cur_path'],
+        expected_warning = [(['carbon_pools_path', 'lulc_bas_path'],
```
dcdenu4, line 224: Good find!

tests/test_translation.py
```diff
@@ -24,7 +24,7 @@
     "InVEST Carbon Model": "ιиνєѕт ςαявσи мσ∂єℓ",
     "Available models:": "αναιℓαвℓє мσ∂єℓѕ:",
     "Carbon Storage and Sequestration": "ςαявσи ѕтσяαgє αи∂ ѕєףυєѕтяαтισи",
-    "current LULC": "ςυяяєит ℓυℓς",
+    "baseline LULC": "ςυяяєит ℓυℓς",
```
dcdenu4, line 27: Is it easy to update that translated `current` to `baseline` ? 😆

dcdenu4 (changes requested):
Thanks @claire-simpson ! I had some minor feedback.

tests/test_translation.py
claire-simpson: I don't know how but I'm gonna do it!

src/natcap/invest/carbon.py
claire-simpson: Yes, very surprising. It looks like in the past (mid 2018) there was a call to `pygeoprocessing_0_3_3.align_dataset_list` ([this](https://github.com/natcap/invest/commit/fc7883d82e6718b11b9ee6de6858c21d96964a5e#diff-91e3c5b0caf23253de581c5c251e477c54497e3db7f63388423351dea3d80787R13-R139) is the most recent commit to the carbon model where I see this, though I did not do a comprehensive investigation). I will create a new issue!

src/natcap/invest/carbon.py
claire-simpson: Good point @dcdenu4 - perhaps I should remove these?

davemfish (approved):
Looks great. These are nice improvements and fixes!

PR #1834, opened 2025-03-11
Habitat Quality: Create degradation raster for each threat

### LINKED ISSUE #1100: Habitat Quality: Create degradation raster for each threat
A user asked about how to calculate degradation for an individual threat ([forum post](https://community.naturalcapitalproject.org/t/how-to-calculate-habitat-degradation-caused-by-each-threat/3147)). This seems like a useful output to have. The model currently calculates and sums the thread degradations all in one step, but I don't see any reason why we shouldn't produce the intermediate degradation rasters too.

megannissel:
## Description
Fixes #1100 

The Habitat Quality model previously calculated and summed total habitat degradation for all threats in a single step. We could instead break this up in order to provide intermediate degradation rasters for each threat, and then use these intermediate rasters to calculate total degradation. 

A note:

Existing tests pass following the change (I still need to dive deeper and see if there are any new test cases we should add!), but I did also compare the total degradation rasters created from our sample data before and after the change. It does seem that the change results in tiny differences, likely due to rounding (summing raster values and then multiplying by access once, vs multiplying by access for each threat and then summing). 

Sample data output values differ, at most, by 2.98e-8. Where do we draw the line in terms of significant enough difference to cause concern?

User's Guide PR: [#167](https://github.com/natcap/invest.users-guide/pull/167)

## Checklist


davemfish (approved):
This looks good, @megannissel , thanks for addressing the changes we talked about.

PR #1838, opened 2025-03-17
Habitat Quality: Improve validation for `MAX_DIST` and `WEIGHT`

### LINKED ISSUE #1052: Habitat Quality handle empty table fields
A user posted an issue on the [forums](https://community.naturalcapitalproject.org/t/habitat-quality-valueerror-could-not-convert-string-to-float/2765) where they were getting the following error:

```python
2022-05-04 16:55:57,649 (natcap.invest.habitat_quality) habitat_quality.execute(259) INFO Starting execute of Habitat Quality model.
2022-05-04 16:55:57,649 (natcap.invest.habitat_quality) habitat_quality.execute(265) INFO Creating workspace
2022-05-04 16:55:57,688 (natcap.invest.habitat_quality) habitat_quality.execute(278) INFO Checking Threat and Sensitivity tables for compliance
2022-05-04 16:55:57,700 (natcap.invest.utils) utils.prepare_workspace(167) ERROR Exception while executing Habitat-Quality
  File "natcap\invest\utils.py", line 165, in prepare_workspace
  File "natcap\invest\ui\model.py", line 1649, in _logged_target
  File "natcap\invest\habitat_quality.py", line 293, in execute
  File "natcap\invest\habitat_quality.py", line 294, in <dictcomp>
ValueError: could not convert string to float: ''
2022-05-04 16:55:57,701 (natcap.invest.utils) utils.prepare_workspace(170) INFO Elapsed time: 0.06s
2022-05-04 16:55:57,701 (natcap.invest.utils) utils.prepare_workspace(173) INFO Execution finished
```

If this is just a case of the threat or sensitivity tables having an empty field, I feel like we could provide a better error message there.

Error reported using InVEST 3.10.2.

megannissel:
## Description

Fixes #1052 

While looking into Issue #1052, I found that, in most cases, our error handling related to empty fields in the Threats Table and Sensitivity Table CSVs has been addressed by improvements to validation since the time the issue was opened. However, in cases where a `NaN` value exists in the `MAX_DIST` or `WEIGHT` columns of the Threats Table, validation passes but the model eventually throws one of the following errors:

- In the case of missing `MAX_DIST`:
```
    if row['max_dist'] <= 0:
       ^^^^^^^^^^^^^^^^^^^^
TypeError: '<=' not supported between instances of 'str' and 'int'
```
- In the case of missing `WEIGHT`:
```
weight_sum = threat_df['weight'].sum()
             ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]
TypeError: can only concatenate str (not "float") to str
```

In order to give the user more helpful feedback, I've added checks to both `execute` (for the benefit of CLI users) and `validate` (for the benefit of Workbench users). 

We _did_ have an existing `if row['max_dist'] <= 0` check in `execute`; in the interest of catching errors as early as possible, I've moved this check up and combined it with the checks for missing `MAX_DIST` and `WEIGHT`. 

## Checklist

### DIFF
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -1266,16 +1266,12 @@ def validate(args, limit_to=None):
                 ['threats_table_path'],
                 MISSING_THREAT_RASTER_MSG.format(threat_list=bad_threat_paths)
             ))
-
-            if 'threats_table_path' not in invalid_keys:
-                invalid_keys.add('threats_table_path')
+            invalid_keys.add('threats_table_path')
 
         if duplicate_paths:
             validation_warnings.append((
                 ['threats_table_path'],
                 DUPLICATE_PATHS_MSG + str(duplicate_paths)))
-
-            if 'threats_table_path' not in invalid_keys:
-                invalid_keys.add('threats_table_path')
+            invalid_keys.add('threats_table_path')
 
     return validation_warnings
--- a/src/natcap/invest/habitat_quality.py
+++ b/src/natcap/invest/habitat_quality.py
@@ -26,9 +26,11 @@ MISSING_THREAT_RASTER_MSG = gettext(
     "A threat raster for threats: {threat_list} was not found or it "
     "could not be opened by GDAL.")
 DUPLICATE_PATHS_MSG = gettext("Threat paths must be unique. Duplicates: ")
+INVALID_MAX_DIST_MSG = gettext(
+    "The maximum distance value for threats: {threat_list} is less than "
+    "or equal to 0. MAX_DIST must be a positive value.")
 MISSING_MAX_DIST_MSG = gettext(
-    "Maximum distance value is missing for threats: {threat_list}. "
-    "MAX_DIST must be a positive value.")
+    "Maximum distance value is missing for threats: {threat_list}.")
 MISSING_WEIGHT_MSG = gettext("Weight value is missing for threats: {threat_list}.")
 
 MODEL_SPEC = {
@@ -480,18 +482,6 @@ def execute(args):
             # for each threat given in the CSV file try opening the associated
             # raster which should be found relative to the Threat CSV
             for threat, row in threat_df.iterrows():
-
-                # Check that weight has been provided
-                if not row['weight']:
-                    raise ValueError(
-                        f"Missing weight ratio for threat: '{threat}'.")
-                # Check that max_dist exists and is greater than 0
-                if not row['max_dist'] or row['max_dist'] <= 0:
-                    raise ValueError(
-                        f"The max distance for threat: '{threat}' is less than"
-                        " or equal to 0, or is missing."
-                        " MAX_DIST should be a positive value.")
-
                 LOGGER.debug(f"Validating path for threat: {threat}")
                 threat_table_path_col = _THREAT_SCENARIO_MAP[lulc_key]
 
@@ -701,6 +691,11 @@ def execute(args):
                     ' calculation for this land cover.')
                 exit_landcover = True
                 break
+            # Check to make sure max_dist is greater than 0
+            if row['max_dist'] <= 0:
+                raise ValueError(
+                    f"The max distance for threat: '{threat}' is less than"
+                    " or equal to 0. MAX_DIST should be a positive value.")
 
             distance_raster_path = os.path.join(
                 intermediate_output_dir,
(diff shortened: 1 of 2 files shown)


src/natcap/invest/habitat_quality.py
```diff
@@ -1220,6 +1233,14 @@ def validate(args, limit_to=None):
                             duplicate_paths.append(
                                 os.path.basename(threat_path))
 
+                    # check that `max_dist` value exists
+                    if not row['max_dist']:
+                        missing_max_dist.append(threat)
+
+                    if not row['weight']:
```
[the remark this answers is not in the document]
emlys: Maybe it would help to keep separate messages for the case where max_dist <= 0, vs. max_dist is undefined?

src/natcap/invest/habitat_quality.py
```diff
@@ -476,6 +480,18 @@ def execute(args):
...
+                if not row['weight']:
+                    raise ValueError(
+                        f"Missing weight ratio for threat: '{threat}'.")
+                # Check that max_dist exists and is greater than 0
+                if not row['max_dist'] or row['max_dist'] <= 0:
+                    raise ValueError(
+                        f"The max distance for threat: '{threat}' is less than"
+                        " or equal to 0, or is missing."
+                        " MAX_DIST should be a positive value.")
+
```
emlys, line 494: I think we don't need these checks for `row['max_dist']` and `row['weight']` here at all, since they're also checked in `validate`. Generally as much validation as possible should happen in `validate`, and should not be duplicated in `execute`. I believe the other validation happening here is because it involves reading rasters, which can be too slow for the instant validation that we expect to see in the workbench.

tests/test_habitat_quality.py
```diff
@@ -802,6 +802,104 @@ def test_habitat_quality_threat_max_dist(self):
...
+            open_table.write(
+                'MAX_DIST,WEIGHT,THREAT,DECAY,BASE_PATH,CUR_PATH,FUT_PATH\n')
+            open_table.write(
+                ',0.7,threat_1,linear,,threat_1_c.tif,threat_1_f.tif\n')
+            open_table.write(
+                '0.07,1.0,threat_2,exponential,,threat_2_c.tif,'
+                'threat_2_f.tif\n')
+
+        with self.assertRaises(ValueError) as cm:
+            habitat_quality.execute(args)
```
emlys, line 849: If we make the changes I suggested above, then these would test `validate`, not `execute`.

tests/test_habitat_quality.py
```diff
@@ -802,6 +802,104 @@ def test_habitat_quality_threat_max_dist(self):
...
+            open_table.write(
+                '0.04,,threat_1,linear,,threat_1_c.tif,threat_1_f.tif\n')
+            open_table.write(
+                '0.07,1.0,threat_2,exponential,,threat_2_c.tif,'
+                'threat_2_f.tif\n')
+
+        with self.assertRaises(ValueError) as cm:
+            habitat_quality.execute(args)
+
+        self.assertIn("Missing weight ratio for threat: 'threat_1'",
```
emlys, line 900: In some other models, we refer to the validation message as a variable, so it's easier if it changes: https://github.com/natcap/invest/blob/36389be8ebbcfe52436530b04340e2395d0b3656/tests/test_coastal_blue_carbon.py#L890-L893

emlys (changes requested):
Thanks @megannissel! I just had some suggestions about the placement of these validation checks.

src/natcap/invest/habitat_quality.py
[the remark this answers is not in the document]
megannissel: The `fillna` is still serving a purpose; the Threats table has several string fields as well, and our checks throughout the model assume a string value will exist for `decay`, `base_path`, `cur_path`, and `fut_path`. We could re-write all of those checks as well, to check for NaNs directly, but I'm not sure the end result would be cleaner or more readable code. 

This would also require importing Pandas, as these fields in the dataframe returned by `validation.get_validated_dataframe` are of `pandas.NAType` rather than `numpy.nan`. Not sure if additional package imports are a concern, but wanted to call it out in case.

src/natcap/invest/habitat_quality.py
```diff
@@ -476,6 +480,18 @@ def execute(args):
             # for each threat given in the CSV file try opening the associated
             # raster which should be found relative to the Threat CSV
             for threat, row in threat_df.iterrows():
+
+                # Check that weight has been provided
+                if not row['weight']:
+                    raise ValueError(
```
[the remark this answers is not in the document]
megannissel: Oh, whoops, thank you; that's a really good catch!

src/natcap/invest/habitat_quality.py
megannissel: Having checks in both `execute` and `validate` was Doug's suggestion, since `validate` only runs in the workbench. Including checks in `execute` would provide more robust error messages to CLI users. Granted, when Doug and I talked about this initially, we were only discussing the issues with `MAX_DIST` error handling, as I hadn't yet realized the same problem existed with `WEIGHT`. I guess it's a question of whether or not we think it's worthwhile to provide more straightforward error messages for CLI users in this case, with the downside of more clutter in the code.

Honestly, I think a better overall solution to problems like this would be to handle missing required CSV values as part of `validation.get_validated_dataframe`, but the scope there is a lot bigger.

src/natcap/invest/habitat_quality.py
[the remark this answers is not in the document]
emlys: Oh true... now I wonder if `validation.get_validated_dataframe` should be responsible for replacing NA with `''` on string columns? Since the `pandas.NA` value isn't particularly useful outside of the dataframe.

src/natcap/invest/habitat_quality.py
megannissel: Thank you for the additional context! Given this precedent, I'm happy to keep the scope here to `validate` functionality and drop the `execute` piece. The errors aren't so obscure that it's hard to determine the root cause (with logging level set to debug, at least). I'll adjust the testing to test `validate` instead.

src/natcap/invest/habitat_quality.py
```diff
@@ -1225,7 +1266,9 @@ def validate(args, limit_to=None):
                 ['threats_table_path'],
                 MISSING_THREAT_RASTER_MSG.format(threat_list=bad_threat_paths)
             ))
-            invalid_keys.add('threats_table_path')
+
+            if 'threats_table_path' not in invalid_keys:
```
emlys, line 1270: There's no downside to re-`add`ing something to a set, so I think this check is unnecessary

megannissel: Ah, so it is. I had (incorrectly) assumed `invalid_keys` was a list rather than a set due to the preexisting check on line 1278. Should have taken a look at `validation.get_invalid_keys`. This being the case, we can drop that check, too.

emlys (approved):
Thanks @megannissel!

PR #1839, opened 2025-03-18
remove `gdal.UseExceptions()` warning on import

### LINKED ISSUE #1702: remove gdal.UseExceptions() FutureWarning
~~Now that invest issues a `FutureWarning` about using `gdal.UseExcpetions()`, scripts that invest generates through the Workbench and CLI should probably include that as part of the boilerplate. It seems like it needs to be called before the `natcap.invest` import to avoid the warning.~~

As a tangent, I wonder if import time is the best place to issue that `FutureWarning`. It kind of forces everyone to see it, which is good, but I also might already intend to call `gdal.UseExceptions()` in my script, I just haven't done it yet because I'm following PEP8 guidelines and doing all my imports first. In that case I don't really want to see the warning everytime.

davemfish:
This warning was too intrusive. We still use `with with GDALUseExceptions():` around all the CLI entry-points, so CLI and workbench users have exceptions enabled.

Python users will no longer get the warning when they import `natcap.invest`, but they will probably see warnings from GDAL during model execution or elsewhere if they haven't called `UseExceptions()`.

Fixes #1702 

## Checklist


emilyanndavis (approved):
Thanks, @davemfish!

PR #1840, opened 2025-03-18
debugging the auto-release workflow

davemfish:
Last time it ran, it ran a step that should have been skipped.
https://github.com/natcap/invest/actions/runs/13930797693/job/38987164220


phargogh (approved):
Makes sense to me!

phargogh:
@davemfish not sure if it makes sense here to wait for the builds to finish, so if it would be better or easier to merge before the builds finish please feel free to do so.  I'll wait until they finish just in keeping with our usual process and then merge.

PR #1851, opened 2025-03-27
Ensure consistent nodata handling in Stormwater volume calculations to prevent output inf values

### LINKED ISSUE #1850: Inf values in Stormwater model volume outputs
On the forum, [two users reported a bug](https://community.naturalcapitalproject.org/t/urban-stormwater-retention-strange-values/5682) in the Stormwater model: the runoff and retention volume rasters and columns in the aggregate.gpkg contain unexpected `inf` values. 

It looks like this is due to a data overflow issue, where values are added to the default `float32` nodata value (`3.40282e+38`) during the volume calculation step. The underlying issue is that, although we do nodata masking, `3.40282e+38` is not recognized as a nodata value. This is because the global variable `FLOAT_NODATA` (which is used during valid pixel masking) is set to `-1`, while the actual nodata value of the input `ratio_array` (to the volume calculation operation) is `3.40282e+38`. 

I propose setting the `target_nodata` value to `FLOAT_NODATA` when creating `[rentention/runoff/percolation]_ratio.tif` rasters to ensure that all nodata pixels are recognized as such when calculating volumes. This appears to be how nodata values were handled in the past, when `pygeoprocessing.raster_calculator` was used to create the `[rentention/runoff/percolation]_ratio.tif` outputs (whereas now we use `pgp.raster_map` which sets a default `target_nodata` value).

claire-simpson:
## Description
Fixes a bug in the Urban Stormwater Retention model that caused `inf` values to appear in the runoff and retention volume outputs. This issue was due to a mismatch between the expected nodata value (`FLOAT_NODATA`, set to `-1`) and the actual nodata value in `_ratio` rasters (set via `pgp.raster_map` to a default value of `3.40282e+38`). The fix ensures that runoff and retention ratio rasters use FLOAT_NODATA as their target_nodata, so nodata pixels are properly masked during volume calculations.

Fixes #1850

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -68,6 +68,10 @@ Unreleased Changes
     * Fixed bug in the calculation of Cooling Capacity (CC) provided by parks,
       where the CC Index was not being properly incorporated.
       https://github.com/natcap/invest/issues/1726
+* Urban Stormwater Retention
+    * Fixed a bug causing ``inf`` values in volume outputs because nodata
+      values were not being set correctly (`InVEST #1850
+      <https://github.com/natcap/invest/issues/1850>`_).
 * Wind Energy
     * Fixed a bug that could cause the Workbench to crash when running the Wind
       Energy model with ``Taskgraph`` logging set to ``DEBUG`` (`InVEST #1497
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks @claire-simpson!

PR #1858, opened 2025-03-28
Handle float nodata comparisons correctly in Scenic Quality

### LINKED ISSUE #1714: Handling `nan` nodata values in Cython modules
This came up originally in #1705 , but is common to several modules.

It's common for our raster processing routines to mask out nodata pixels before applying some calculation. When this happens in the context of `raster_calculator`, we typically create a mask using `numpy.isclose(array, nodata_value, equal_nan=True)`. But in the context of a Cython module, we typically define our own `is_close` function, that looks something like this:
```python
cdef int is_close(double x, double y):
    return abs(x-y) <= (1e-8+1e-05*abs(y))
```
This function does not correctly handle `nan` values. And `nan` is a valid GDAL raster value, and nodata value. `is_close` could be modified to handle `nan` values like this:
```python
cdef int is_close(double x, double y):
    if isnan(x) and isnan(y):
        return 1
    return abs(x-y) <= (1e-8+1e-05*abs(y))
```
That's a solution we implemented in `pygeoprocessing.routing.routing.pyx`

But there's reason to be concerned about the performance of adding this extra check to a function that is called once for every pixel in a raster stack. I'm not sure if we evaluated that concern when implementing the `pygeoprocessing` function.

We could consider an alternative approach: Pre-processing user input data to ensure that it does not include `numpy.nan` values, and to reclassify them to something if it does. I don't know how that would compare performance-wise.

### LINKED ISSUE #1859: Scenic Quality bug in DEM nodata handling
Discovered while working on #1714 

https://github.com/natcap/invest/blob/4908639a4c148f4bbd317f6f9ed1a7bdad756e55/src/natcap/invest/scenic_quality/viewshed.pyx#L961-L976

There is a bug in this conditonal where, if `adjusted_dem_height >= z` AND `target_pixel.distance_to_viewpoint < max_visible_radius` AND `target_dem_height != nodata`, but is within the float tolerance of nodata, then the first condition would apply. But if `target_dem_height` exactly equals nodata, then the second condition would apply. So the visibility raster could be incorrectly set to 1 when the DEM actually has nodata.

This will be fixed as part of #1714.

emlys:
## Description
Fixes #1714 
Fixes #1859 

- Use `is_close` for float nodata comparisons in Scenic Quality
- Fix confusing/buggy conditional
- Add a note about `is_close` performance

## Checklist

### DIFF
--- a/src/natcap/invest/scenic_quality/viewshed.pyx
+++ b/src/natcap/invest/scenic_quality/viewshed.pyx
@@ -40,6 +40,7 @@ from libc cimport math
 cimport numpy
 cimport cython
 from ..managed_raster.managed_raster cimport ManagedRaster
+from ..managed_raster.managed_raster cimport is_close
 
 LOGGER = logging.getLogger(__name__)
 LOGGER.addHandler(logging.NullHandler())
@@ -297,12 +298,13 @@ def viewshed(dem_raster_path_band,
 
     band_to_array_index = dem_raster_path_band[1] - 1
     nodata_value = dem_raster_info['nodata'][band_to_array_index]
-    if viewpoint_elevation == nodata_value:
-        raise LookupError('Viewpoint is over nodata')
 
     # Need to handle the case where the nodata value is not defined.
     if nodata_value is None:
         nodata_value = IMPROBABLE_NODATA
+    else:
+        if is_close(viewpoint_elevation, nodata_value):
+            raise LookupError('Viewpoint is over nodata')
     cdef double nodata = nodata_value
 
     # Verify that pixels are very close to square.  The Wang et al algorithm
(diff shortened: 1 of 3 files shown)


src/natcap/invest/scenic_quality/viewshed.pyx
```diff
@@ -638,22 +639,21 @@ def viewshed(dem_raster_path_band,
...
+        elif (adjusted_dem_height >= z and
+                target_pixel.distance_to_viewpoint < max_visible_radius):
+            visibility_managed_raster.set(n, m, 1)  # the pixel is visible
             aux_managed_raster.set(n, m, adjusted_dem_height)
         else:
-            # If it's close enough to nodata to be interpreted as nodata,
-            # consider it to be nodata.  Nodata implies that visibility is
-            # undefined ... which it is, since there's no defined DEM value for
-            # this pixel.
-            if math.fabs(target_dem_height - nodata) <= 1.0e-7:
```
emlys, line 652: The logic of this conditional was confusing to me and had a potential bug: if `adjusted_dem_height >= z` AND `target_pixel.distance_to_viewpoint < max_visible_radius` AND `target_dem_height != nodata`, but is within the float tolerance of `nodata`, then the first condition would apply. But if `target_dem_height` exactly equaled `nodata`, then the second condition would apply. So this change could produce different results in that case, but I think this is what was intended.

dcdenu4 (approved):
Thanks @emlys, this looks good to me. Thanks for the useful comments for future us.

PR #1861, opened 2025-04-01
Add unit tests for Crop Production, Pollination, Forest Carbon, and HQ models

### LINKED ISSUE #1603: Add more unit tests for models that could use them

claire-simpson:
## Description
This continues PR #1776 and adds more unit tests for Crop Production (percentile), Crop Pollination, Forest Carbon Edge Effects, and Habitat Quality. 

Fixes #1603

## Checklist


tests/test_forest_carbon_edge.py
```diff
@@ -15,13 +19,68 @@
...
+    vector_format = "ESRI Shapefile"
+    fields = {"id": ogr.OFTReal}
+    attribute_list = [{"id": 0}]
+    pygeoprocessing.shapely_geometry_to_vector(shapely_geometry_list,
+                                               path_to_shp, projection_wkt,
+                                               vector_format, fields,
+                                               attribute_list)
+
+
+def make_simple_raster(base_raster_path, array, nodata_val=-1):
```
claire-simpson, line 51: @emlys recommended considering abstracting these functions like `make_simple_raster` out to a test_utils module. I didn't do that as sometimes these rasters or vectors do need to be slightly different, and they do wrap around pgp functions like `numpy_array_to_raster`. Curious to hear other opinions though as sometimes I am using the exact same function.

tests/test_crop_production.py
```diff
@@ -857,6 +884,243 @@ def _create_expected_agg_table():
...
 
+    def test_aggregate_to_polygons(self):
+        """Test `aggregate_to_polygons`"""
+        from natcap.invest.crop_production_percentile import \
+            aggregate_to_polygons
+
+        def _create_expected_pctl_table():
+            """Create expected output results"""
+            # Define the new values manually
+            data = [
```
claire-simpson, line 895: I currently have this huge data frame to use as the `expected_aggregate_pctl_table` which I like as it is comprehensive and clear on what the output is. An alternative could be spot checking values in a few columns, which would result in fewer lines of code and be easier to update if there is a change in the model. Thoughts on which option is preferable?

tests/test_forest_carbon_edge.py
```diff
@@ -243,6 +302,231 @@ def test_carbon_nodata_lulc(self):
...
+            edge_distance_path, spatial_index_pickle_path,
+            n_nearest_model_points, biomass_to_carbon_conversion_factor,
+            tropical_forest_edge_carbon_map_path)
+
+        actual_output = pygeoprocessing.raster_to_numpy_array(
+            tropical_forest_edge_carbon_map_path).sum()
+        expected_output = 3659.7915
+
+        # Comparing sums as 1 value in the output arrays is ~8% different
+        # on windows vs. mac, which seemed too high a relative tolerance to set
```
claire-simpson, line 526: I'm second guessing this approach, but still think this is better than just comparing the whole array with rtol=.08, as that seems like a high enough value that could leave room for errors to slip through. Possibly worth spot checking a few values in the output array though? Or comparing the raw expected and output arrays in addition to the the array sums?

megannissel (commented):
Absolutely love the addition of unit tests for some of our intermediate model functionality!

I did have one clarifying question about the last forest carbon test, but overall I think these look really good, and I appreciate you talking me through your thought process with creating synthetic data. 

Looking forward to the conversation about abstracting out some test helpers; I definitely think it would be worthwhile, both for some of the tests you've already written, and for ones we're all likely to write in the future!

tests/test_forest_carbon_edge.py
claire-simpson: Basically, the test passes on my machine (mac) and on the mac-os runners in GHA, but fails on windows, so my suspicion is there is some difference in how the two operating systems are doing the calculation. I'm not exactly sure where that is getting introduced and `_calculate_tropical_forest_edge_carbon_map` doesn't really produce intermediate products to test. 

My best guesses are that the difference in output values are related to either (1) how the "nearest points to a forest pixel" are determined, or (2) small differences in how the math calculations are done (which we also saw in `numpy.linalg` calculations on different macs). To the first point: the points I'm creating in this test are in a grid (with points spread out more vertically then horizontally) so its possible that with equidistant points to choose from, the mac and windows are choosing different points..? In any case, setting `n_nearest_model_points` to 6 vs 8 reduced the number of errors in the output `tropical_forest_edge_carbon_map` array, but increased the magnitude of the error.

tests/test_crop_production.py
claire-simpson: Would you mind weighing in here @davemfish ?

tests/test_forest_carbon_edge.py
megannissel: Gotcha, thank you for the refresher! Given all of that, I think you're taking a logical approach. And I see you added some spot-checking, too, which looks good to me!

tests/test_forest_carbon_edge.py
claire-simpson: For more on this: issue #1885

tests/test_crop_production.py
megannissel: I personally think I'm still in favor of keeping this test. While we do check the aggregated results in `test_crop_production_percentile`, if that check fails we're still not immediately sure where in the model run things went wrong. If this test fails, though, we immediately know the problem is with the aggregation step. 

But I also respect @davemfish's hesitance to introduce a second complex dataset we'd need to maintain. 

Not in the scope of this PR (and I haven't dug into it to see if it would be practical), but I wonder how realistic it would be to replace the sample data in this test suite entirely; the synthetic data does seem easier to maintain and use for debugging in general.

megannissel (approved):
The question about whether or not to keep `test_aggregate_regression_results_to_polygons` aside (and I'm fine with whichever way you decide to go there), I think this looks great!

tests/test_crop_production.py
davemfish: Thanks for sharing your opinions. I think you should keep the new test if you both agree that it has good benefit, and it sounds like you do.

>Not in the scope of this PR (and I haven't dug into it to see if it would be practical), but I wonder how realistic it would be to replace the sample data in this test suite entirely; the synthetic data does seem easier to maintain and use for debugging in general.

Yeah, it's worth considering this to reduce the maintenance load. Along those lines, it may be reasonable to have an auxiliary python file with a utility function to create the dataframe. That could be a compromise between using the external test-data repo, and coding the data directly in-line in the test. Because in addition to the maintenance load, there's also just an increased cognitive load of reading the tests when they have this large block of in-line data.

megannissel: > Along those lines, it may be reasonable to have an auxiliary python file with a utility function to create the dataframe.

I think this is a great idea

claire-simpson: This sounds good to me too! @davemfish @megannissel Where do you envision this living? Maybe `tests/utils/data_helpers.py`?

davemfish: Maybe something like `tests/crop_production/data_helpers.py` ? There's a little bit of precedent already with `tests/test_datastack_modules/`, though that use-case is a bit different.

megannissel (approved):
Thanks, @claire-simpson! The helper utility looks great, and I appreciate the added docstrings; I think they'll help us keep track of what's being used where.

PR #1873, opened 2025-04-04
Usage logging for plugins

### LINKED ISSUE #1782: Handle usage logging for plugins
## Should we log plugin usage at all?
I think yes, it could be useful and interesting to know which and how many plugins are used

## How do we identify plugins in the context of usage logging?
Currently, usage logging for plugins will happen the same way as core invest models - running a plugin called `foo` will be logged as a run of `natcap.invest.foo`. This probably isn't adequate because there could be any number of different plugins, or versions of plugins, out there called `foo`. While only one at a time can be installed in a given workbench installation, there's no globally unique plugin identifier, since it's just a python package.

One solution could be to log the git url and/or commit hash for plugins installed from a git server. This is nice because we could find the plugin source and see what people are using. But it won't work for plugins installed from a local source, or a private repo if we eventually support that.

Logging the entire source code of a plugin is probably overkill, and might be a privacy concern for someone expecting their local code to be private. Logging a hash of the source doesn't breach privacy but still might surprise someone. I think it would be enough to simply log "Local source" and not uniquely identify plugin runs in that case.

emlys:
## Description
Fixes #1782

This PR adds to the usage information that is sent to the usage logging server on model start. To distinguish between core models and plugins, I added a `type` attribute which can have the value `core` or `plugin`. For logging runs with `type: 'plugin'`, a `source` attribute is also logged, which is the source of the plugin install if installed from github (e.g. `git+https://github.com/foo/bar@main`, or `'local'` if installed from a local path on the user's computer.

The Cloud Run function that receives and stores usage logging will need to be updated to reflect these changes, which should be simple. That function lives on Google Cloud - there is no repo tracking it, so we can't do a PR, but I thought we could walk through it on a team call sometime.
## Checklist

### DIFF
--- a/workbench/src/main/investUsageLogger.js
+++ b/workbench/src/main/investUsageLogger.js
@@ -4,6 +4,7 @@ import fetch from 'node-fetch';
 
 import { getLogger } from './logger';
 import pkg from '../../package.json';
+import { settingsStore } from './settingsStore';
 
 const logger = getLogger(__filename.split('/').slice(-1)[0]);
 const WORKBENCH_VERSION = pkg.version;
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Thanks @emlys, this looks straightforward to me.

PR #1875, opened 2025-04-07
Forest Carbon: Global model vector clipping and reprojection

### LINKED ISSUE #1689: Forest Carbon test data: invalid geometries after reprojection
Enabling `gdal.UseExceptions` made this issue apparent. Two tests (`test_carbon_dup_output` and `test_carbon_full`) started failing because of invalid geometries in the vector resulting from reprojecting the original model data. For #1570 I'm going to just skip over any invalid geometries, since that's effectively what was happening before, but we should check why these are being produced and if there's any way to avoid it.

megannissel:
## Description
Fixes #1689 

### Background:
Enabling `gdal.UseExceptions` led to the discovery that invalid geometries were being created during reprojection of the Forest Carbon "Global Regression Models" vector (`forest_carbon_edge_regression_model_parameters.shp`). This was happening even with partial reprojection enabled. Digging into it, it seems partial reprojection doesn't necessarily remove entire geometries, just points that can't be reprojected; as such, you can end up with geometries that have insufficient points to be valid polygons.

Based on recommendations from @emlys and @phargogh, this PR does away with partial reprojection and instead clips the global vector prior to reprojection, so that only those polygons relevant to the model run need be reprojected. Clipping is based on the LULC raster's bounding box, plus a buffer (explained below). 

While this model does include an AIO input, it's optional and is only used to calculate aggregate statistics at the end of the model run. For consistency, I'm ignoring it for the purposes of clipping. 

### Bounding Box Buffering: 
The Global Regression Model vector data is used to construct a kd-tree. When querying the kd-tree, we set a `distance_upper_bound`. To ensure the clipped vector will contain all of the relevant data to the querying process, I'm buffering the LULC bounding box by the same `DISTANCE_UPPER_BOUND` value (currently set to 500km). 

I'm also keeping all polygons that intersect with the clipping polygon, rather than using `gdal.Layer.Clip()`, in order to preserve the shape of the polygons (this is important because we use the centroid when constructing the kd-tree). 

To summarize my clipping methodology: 
- Get the bounding box of the LULC raster; buffer it by `DISTANCE_UPPER_BOUND`
- Reproject the buffered bounding box to the coordinate system of the vector
- Copy all polygons from the global vector that intersect with the bounding box to a new vector
- Pass the clipped vector along to `_build_spatial_index`, to be reprojected and used to create the kd-tree

### Additional notes:
The User's Guide says, "Note that all spatial inputs must be in the same projected coordinate system and in linear meter units." However, we weren't validating the units within the model. I've updated the model spec to include `"projection_units": u.meter` for validation purposes.

Digging into this model, I also noticed a contradiction. The user's guide says, related to the "Number Of Points To Average," (in "How it works"): 
"The user can designate the number of local models used in the interpolation scheme which is defaulted to 10 but can range anywhere from 1 (only closest point) to 2635 (every regression model on the planet)."

However, within the code itself, we specify the aforementioned `DISTANCE_UPPER_BOUND` of 500km that we use when querying the kd-tree (there's a note that reads, "grid cells are 100km. Becky says 500km is a good upper bound to search"). We may want to update the User's Guide to reflect this, especially now that we're using that `DISTANCE_UPPER_BOUND` as a buffer when clipping the global vector to the LULC raster's bounding box.

## Checklist

### DIFF
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -832,7 +832,7 @@ def _clip_global_regression_models_vector(
                     new_feature.SetField(field_name, field_value)
                 target_layer.CreateFeature(new_feature)
 
-        except (shapely.errors.ReadingError, ValueError):
+        except (shapely.errors.ShapelyError, ValueError):
             invalid_feature_count += 1
             LOGGER.warning(
                 "The geometry at feature %s is invalid and will be "
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -118,7 +118,8 @@ MODEL_SPEC = {
             "about": spec_utils.LULC['about'] + " " + gettext(
                 "All values in this raster must "
                 "have corresponding entries in the Biophysical Table."),
-            "projected": True
+            "projected": True,
+            "projection_units": u.meter
         },
         "pools_to_calculate": {
             "type": "option_string",
@@ -170,6 +171,8 @@ MODEL_SPEC = {
                         "Used only for the asymptotic model.")}
             },
             "geometries": spec_utils.POLYGONS,
+            "projected": True,
+            "projection_units": u.meter,
             "required": "compute_forest_edge_effects",
             "about": gettext(
                 "Map storing the optimal regression model for each tropical "
@@ -254,6 +257,13 @@ MODEL_SPEC = {
                     "bands": {1: {
                         "type": "number", "units": u.metric_ton/u.hectare
                     }}
+                },
+                "regression_model_params_clipped.shp": {
+                    "about": (
+                        "The Global Regression Models shapefile clipped "
+                        "to the study area."),
+                    "geometries": spec_utils.POLYGONS,
+                    "fields": {}
                 }
             }
         },
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -812,36 +812,34 @@ def _clip_global_regression_models_vector(
     target_layer.StartTransaction()
     invalid_feature_count = 0
     for feature in base_layer:
+        invalid = False
         geometry = feature.GetGeometryRef()
-
         try:
             shapely_geom = shapely.wkb.loads(bytes(geometry.ExportToWkb()))
-        # Invalid geometries that cannot be loaded by Shapely;
+        # Catch invalid geometries that cannot be loaded by Shapely;
         # e.g. polygons with too few points for their type
         except shapely.errors.ShapelyError:
-            invalid_feature_count += 1
-            LOGGER.warning(
-                f"The geometry at feature {feature.GetFID()} is invalid "
-                "and will be skipped.")
-            continue
-
-        if not shapely_geom.is_valid:
-            invalid_feature_count += 1
-            LOGGER.warning(
-                f"The geometry at feature {feature.GetFID()} is invalid "
-                "and will be skipped.")
-            continue
-
-        # Check for intersection rather than use gdal.Layer.Clip()
-        # to preserve the shape of the polygons (we use the centroid
-        # when constructing the kd-tree)
-        if shapely_mask.intersects(shapely_geom):
-            new_feature = ogr.Feature(target_layer.GetLayerDefn())
-            new_feature.SetGeometry(ogr.CreateGeometryFromWkb(
-                shapely_geom.wkb))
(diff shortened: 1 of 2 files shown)


src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -733,6 +760,92 @@ def _map_distance_from_tropical_forest_edge(
...
+            # when constructing the kd-tree)
+            if shapely_geom.intersects(shapely_mask):
+                new_feature = ogr.Feature(target_layer.GetLayerDefn())
+                new_feature.SetGeometry(ogr.CreateGeometryFromWkb(
+                    shapely_geom.wkb))
+                for field_name, field_value in feature.items().items():
+                    new_feature.SetField(field_name, field_value)
+                target_layer.CreateFeature(new_feature)
+
+        except (shapely.errors.ShapelyError, ValueError):
```
phargogh, line 835: Could you add a note here of when we should expect each of these exceptions?

src/natcap/invest/forest_carbon_edge_effect.py
```diff
@@ -733,6 +760,92 @@ def _map_distance_from_tropical_forest_edge(
...
+            invalid_feature_count += 1
+            LOGGER.warning(
+                "The geometry at feature %s is invalid and will be "
+                "skipped", feature.GetFID())
+
+    target_layer.CommitTransaction()
```
phargogh, line 841: Good use of transactions and checking of geometry validity!

I know the layer we're running this on should only ever have very simple geometries (and not that many of them, at that), but if you wanted to look at one possible way to speed up the `.intersection` tests, I'd suggest looking at shapely's `prepared` geometries. 
 You'll find we use them all over the place for testing intersections and other things, and it is possible that it might also speed things up a little here as well.  Just a thought!

https://shapely.readthedocs.io/en/stable/reference/shapely.prepare.html

phargogh (changes requested):
Thanks @megannissel !  I just had two minor comments and it looks like there is now a merge conflict on a file, probably because I took too long to review this.

src/natcap/invest/forest_carbon_edge_effect.py
megannissel: Oh, good call. I had started out using `prepared` geometries based on some examples elsewhere in our codebase, but ran into some issues early on (`intersects` failing with `TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.`). So I had scrapped it for the sake of understanding what the code was doing at a basic level before trying to get fancy, but this reminded me that I really needed to go back and figure out what I was doing wrong!

I've now learned that `{geometry}.intersects({prepared_geometry})` throws an error, but flipping it to `{prepared_geometry}.intersects({geometry})` works just fine. (And upon closer reading of the documentation, this makes sense; the prepared geometry is the geom with additional info that helps speed up operations like `intersects` anyway!)

src/natcap/invest/forest_carbon_edge_effect.py
megannissel: It looks like `shapely.errors.ShapelyError` (which replaced a number of prior errors, including `shapely.errors.ReadingError`) is sufficient for catching invalid geoms that shapely can't load. Dropping the `ValueError` and adding a note!

megannissel:
Thanks for your feedback, @phargogh! I've swapped over to using shapely's `prepared` geometries (though please let me know if I'm still not using this feature to its fullest!) and also cleaned up my try / except logic a bit. 

A question for you: should we update the User's Guide to reflect the reality of the model with regards to `DISTANCE_UPPER_BOUND`? Currently there's no mention of this cap; in fact, it (misleadingly) states you could use every regression model by setting "Number of Points to Average" high enough.

And no worries at all about the merge conflict -- Claire and I were both working on tests related to this model simultaneously, so one of us was going to end up with a conflict regardless. Thankfully it was an easy one to resolve AND these model changes pass the tests she added!

phargogh (changes requested):
I just had one additional comment about some duplication I hadn't noticed before, and then

> A question for you: should we update the User's Guide to reflect the reality of the model with regards to DISTANCE_UPPER_BOUND? Currently there's no mention of this cap; in fact, it (misleadingly) states you could use every regression model by setting "Number of Points to Average" high enough.

Yes, I agree that this should be addressed, and updating the UG seems most straightforward to do immediately.  But, looking ahead, I think it would also be worth checking with some folks (Lisa at a minimum, but possibly also Becky) to see what the model _should_ be doing, and then we should make it so, possibly with an ADR.  It's not a huge change either way, but it'd be good to figure out how the model should behave.

phargogh (approved):
Thanks @megannissel !

Issue #322, opened 2020-09-30
Run button doesn't work when workspace from CLI run exists


emlys:
Observed on both main and release/3.9
Steps to reproduce:
1. `python setup.py install` and `invest launch` on main
2. Open any model and run it
3. Download and open the invest-mac-binaries artifact from github from the latest build of main
4. Open the InVEST app and open the same model as you ran before
5. Make sure the workspace is the same as the workspace from the previous run. Click the Run button.

Clicking the Run button has no effect when the workspace already exists from a previous run via the CLI. (No log file is created, no window pops up). If you delete the workspace directory, it works. If the previous run was via the app, it works.

The reverse is not true: if the model is launched via the CLI, it doesn't matter whether the workspace was created previously from the CLI or the app.

phargogh:
I confirm that this is also an issue on my mac, though this is also _only_ an issue when running through the launcher as an application.  Reaching into the application itself and running the binaries seems to work just fine:
* `./InVEST.app/Contents/MacOS/InVEST launch` works great
* `./InVEST.app/Contents/MacOS/invest_dist/invest launch` works great

Only launching the application directly does not work.  Examples include:
* double-clicking on `InVEST.app`
* `open ./InVEST.app`

I was testing this on `release/3.9`, revision `7791b7d8c`.

I tried to reproduce this on Windows as well, but could not reproduce the issue.  Running the launcher through the CLI works fine, running the launcher through the binaries does not work.

dcdenu4:
Should this issue be part of the 3.9 milestone?

phargogh:
@dcdenu4 yes, this would prevent a user from running InVEST, so we should absolutely resolve this before the release.  I'll add it to the milestone.

emlys:
Appears related to this console message:
`Sandbox: invest(13237) System Policy: deny(1) file-read-data /Users/emily/Documents/seasonal_water_yield_workspace`

emlys:
Some observations:
* InVEST processes are not sandboxed, according to activity monitor
* Manually giving InVEST full disk access in System preferences > Security and privacy does not fix the problem
* The issue of validation not working the first time is related. There are also several similar messages: `Sandbox: invest(19141) System Policy: deny(1) file-read-xattr /Users/emily/Downloads/DelineateIt/DEM_gura.tif.xml` indicating that the invest app isn't being allowed to read the input files.
[...]

PR #1882, opened 2025-04-08
HRA update to go along with pygeoprocessing #322

### LINKED ISSUE #1881: Update HRA for pygeoprocessing zonal_stats geometry change
The changes made in https://github.com/natcap/pygeoprocessing/pull/428 caused an HRA test to fail, because HRA's geometry simplification function was always setting the vector geometry type to `Unknown`. `pygeoprocessing.zonal_statistics` will now require a type of `Polygon` or `MultiPolygon`, so this raised an error. 

I think we can copy over the geometry type from the input vector, which should always be `Polygon` or `MultiPolygon` for the AOI. It may still be `Unknown` in the case of habitat or stressor vectors that contain multiple geometry types.

emlys:
## Description
Fixes #1881

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -61,9 +61,14 @@
 
 
 
-..
-  Unreleased Changes
-  ------------------
+Unreleased Changes
+------------------
+
+HRA
+===
+* The intermediate simplified vectors will now inherit their geometry type from
+  the input vectors, rather than using ``ogr.wkbUnknown``
+  (`#1881 <https://github.com/natcap/invest/issues/1881>`_).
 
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Looks great, thanks @emlys !

phargogh:
Hmm something seems to be a little funky with GHA.  I'll let it cook a little longer and see if the actions complete.  Otherwise, this change is so simple I'm happy to just merge.

emlys:
I fixed a whitespace issue that was causing the HISTORY.rst check to fail. The ReadTheDocs failure seems unrelated, so I'm going to merge

PR #1886, opened 2025-04-11
Plugin usage logging: python side

emlys:
## Description
In #1873, I forgot that the workbench doesn't log directly to the usage logging server - it passes info to the invest server, which then sends it on to the usage logging server. This PR updates the python side of things to pass on the new `type` and `source` attributes. 

I tested and confirmed that a plugin run is indeed logged and stored in the database with the `type` and `source` attributes correctly set. And, regular invest runs are continuing to show up in the database without errors.

## Checklist


dcdenu4 (commented):
Thanks @emlys. Looks like there's something funky going on with installing the `natcap.invest` wheel, which is causing tests to break.

emlys:
@dcdenu4 I fixed the failing test, the remaining failures on ubuntu are expected and a fix for that is included in #1795

emlys:
Checks passing now!

dcdenu4 (approved):
Thanks!

PR #1889, opened 2025-04-15
Fix Workbench dialog for selecting location to extract datastack

### LINKED ISSUE #1879: "Choose location to extract archive" window is confusing
A while back I asked about this in Slack, apparently forgot to make an issue for it, and just encountered it again, so here's that conversation, lightly edited. 

> Stacie Wolny
  Jan 7th at 8:35 AM
I'm trying to work with a forum user's .tgz datastack, and it seems to me that something has changed about how they're handled in Workbench, which is confusing. I swear I used to be able to simply drag and drop the .tgz into the Workbench model screen, and it automagically unpacked it and filled in the data entries. But now when I drag and drop the .tgz into Workbench, it gives a window saying "Choose location to extract archive".  Through trial and error I've found that it ends up creating a folder, but it won't let me select an existing folder (even if I made a new one in the "Choose location" window itself) as my location, I can only type a new one in the "File name" line that it then goes on to create. All of this is very unintuitive.

> Doug Denu
  Jan 7th at 10:54 AM
It looks like we touched on datastack extraction in [this issue](https://github.com/natcap/invest/issues/1308), which would have been updated in the 3.14.1 release. I think generally asking the user where they would like the compressed datastack archive to be unpacked is good practice.
I've found that it ends up creating a folder, but won't let me select an existing folder (even if I made a new one in the "Choose location" window itself) as my location, I can only type a new one in the "File name" line that it then goes on to create.
This seems a little funky and would be worth checking out! It should exactly like you'd expect your OS to handle such things. 
[@Stacie](https://stanford-natcap.slack.com/team/WCAMU7QT1)
, would you mind creating an issue for this?

> Stacie Wolny
  Jan 7th at 11:51 AM
Agreed that it is good practice to ask the user (although it worked fine before when I think it just extracted in the same folder where the tgz was located, a good default). If we keep asking the user, I'd suggest making it more clear what to do in the window that pops up. And of course, try to make folder creation work as expected, like it does when creating the Workspace.

Meanwhile, I'm updating the User Guide with an explanation of what it's asking for and how to enter it.

davemfish:
We needed to change the way we use the `electron.shell.dialog` API to allow users to select a folder instead of a file. I also added a check to see if the folder is writeable. This now matches how we handle selecting the download location from the sample data `DownloadModal`.

Fixes #1879 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/SetupTab/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/index.jsx
@@ -299,18 +299,29 @@ class SetupTab extends React.Component {
     const { pyModuleName, switchTabs, t } = this.props;
     let datastack;
     try {
-      if (filepath.endsWith('gz')) {  // .tar.gz, .tgz
+      if (filepath.endsWith('gz')) { // .tar.gz, .tgz
         const extractLocation = await ipcRenderer.invoke(
-          ipcMainChannels.SHOW_SAVE_DIALOG,
-          { title: t('Choose location to extract archive') }
+          ipcMainChannels.SHOW_OPEN_DIALOG,
+          {
+            title: t('Choose location to extract archive'),
+            properties: ['openDirectory'],
+          }
         );
-        if (extractLocation.filePath) {
-          datastack = await fetchDatastackFromFile({
-            filepath: filepath,
-            extractPath: extractLocation.filePath});
-        } else {
-          return;
-        }
+        if (extractLocation.filePaths.length) {
+          const directoryPath = extractLocation.filePaths[0];
+          const writable = await ipcRenderer.invoke(
+            ipcMainChannels.CHECK_FILE_PERMISSIONS, directoryPath);
+          if (writable) {
+            datastack = await fetchDatastackFromFile({
+              filepath: filepath,
+              extractPath: directoryPath,
+            });
+          } else {
+            alert( // eslint-disable-line no-alert
+              `${t('Permission denied extracting files to:')}\n${directoryPath}`
+            );
+          }
+        } else { return; } // dialog closed without selection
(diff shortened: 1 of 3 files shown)


workbench/tests/renderer/investtab.test.js
```diff
@@ -125,6 +125,9 @@ describe('Run status Alert renders with status from a recent run', () => {
 
 describe('Open Workspace button', () => {
   const spec = {
+    pyname: 'natcap.invest.foo',
+    model_name: 'Foo Model',
+    userguide: 'foo.html',
```
davemfish, line 130: props validation was issuing warnings about missing values.

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -299,18 +299,29 @@ class SetupTab extends React.Component {
...
+          const directoryPath = extractLocation.filePaths[0];
+          const writable = await ipcRenderer.invoke(
+            ipcMainChannels.CHECK_FILE_PERMISSIONS, directoryPath);
+          if (writable) {
+            datastack = await fetchDatastackFromFile({
+              filepath: filepath,
+              extractPath: directoryPath,
+            });
+          } else {
+            alert( // eslint-disable-line no-alert
```
emilyanndavis, line 320: Ha, I didn't realize eslint complained about JS alerts! I think this is fine for what's likely to be a rare scenario, especially since we seem to be handling other datastack-related errors in the same way. If we find ourselves doing this elsewhere, that might be a good time to come up with a solution that's more integrated into the UI.

emilyanndavis (approved):
Thanks, @davemfish!

Issue #1698, opened 2024-11-25
Wind Energy: Raster result values returned outside of wind data


newtpatrol:
Testing the Wind Energy model with sample data (populated in Workbench with the .json file we provide), both density_W_per_m2.tif and harvested_energy_MWhr_per_yr.tif (and possibly others related to valuation) return values that are outside of the input wind data extent, which they shouldn't do.  

Here are the relevant bits from the Slack thread between Stacie and Rob Griffin:

> SW: Also, the student noticed that the raster outputs have data outside of the wind_energy_points area, which seems strange - most of the darkest blue area in this screenshot.

![wind_energy_raster_result_extent](https://github.com/user-attachments/assets/6721fe1d-bd47-4a2a-b613-7d53da15f1a8)

> RG: Raster values outside the wind data are definitely an issue for both density_W_per_m2.tif and harvested_energy_MWhr_per_yr.tif and any of the valuation outputs that are not included here. Looks like the bilinear interpolation is filling in areas in the results, which is why you get results in places where there is no underlying wind data. Seems like a missed masking step in the model to constrain outputs to only exist in areas with wind data points.

megannissel:
After some initial exploration of the model, the shape of the issue seems to be: 

The wind data are input in the form of a CSV, with each row representing a discrete geographic coordinate point. This CSV gets converted to a point vector and then the data are interpolated onto a raster. If we were to simply mask the output raster with the point vector created from the input wind data CSV, the output raster would only contain data at the discrete points covered by the wind data input. 

If some interpolation is desired, we considered the possibility of introducing a maximum interpolation distance, which we could then use along with the wind data point vector to create a mask. This value could either be static, derived from user input data, or otherwise user-supplied.

Based on this assessment of the problem, I reached out to Rob Griffin and asked for his opinion as a model expert. His response was:

> After some thought, I think the cleanest way to address this is to get rid of the raster interpolation and only return the point vector. That would mean getting rid of:
carbon_emissions_tons.tif
density_W_per_m2.tif
harvested_energy_MWhr_per_yr.tif
levelized_cost_price_per_kWh.tif
npv.tif
> 
> where all these would retain their values in wind_energy_points.shp but just not be interpolated to raster outputs. Interpolation introduces too many potential violations of the constraints of the model and requires additional assumptions that may not be helpful for the user.

He also mentioned that he's recently been able to find high-resolution wind data that minimizes the need for interpolation. I've asked him to send along an example.

[...]

PR #1892, opened 2025-04-16
ADR for Wind Energy update: remove raster outputs

megannissel:
## Description
Related to #1698 

ADR documenting the proposal to remove raster outputs (which rely on interpolation) from the Wind Energy model.

## Checklist

### DIFF
--- a/doc/decision-records/ADR-0004-Remove-Wind-Energy-Raster-Outputs.md
+++ b/doc/decision-records/ADR-0004-Remove-Wind-Energy-Raster-Outputs.md
@@ -19,7 +19,7 @@ Users noticed the raster outputs included data in areas outside of those covered
 
 ## Decision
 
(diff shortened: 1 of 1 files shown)


rmgriffin:
Having interpolated rasterized results is increasingly unnecessary as high resolution wind data becomes more available. Raster interpolation also has several downsides with the most salient for our purposes being interpolating over invalid areas that violate model constraints, such as depth, distance from shore, and areas outside of the gridded wind speed data. If users are interested in higher spatial resolution results for areas that have coarse spatial wind speed data, they could either: 1) interpolate wind speed inputs (as points) before running the model and use the corresponding CSV as an input; or 2) rasterize the outputs as a post-processing step, while making sure to then filter out areas that violate model constraints.

dcdenu4 (approved):
Thanks Megan!

PR #1895, opened 2025-04-18
Handle validation runtime error in `check_spatial_overlap`

### LINKED ISSUE #1849: `RuntimeError` in validation `check_spatial_overlap`
A user reported this, I don't have the data to reproduce it yet. 
https://community.naturalcapitalproject.org/t/seasonal-water-yield-not-working-no-output-files-generated-previously-used-crop-pollination-successfully/5888

```python
[11:30:35.473] [undefined] 03/25/2025 11:30:35  natcap.invest.ui_server ERROR    Exception on /api/validate [POST]
  File "flask/app.py", line 1511, in wsgi_app
  File "flask/app.py", line 919, in full_dispatch_request
  File "flask_cors/extension.py", line 176, in wrapped_function
  File "flask/app.py", line 917, in full_dispatch_request
  File "flask/app.py", line 902, in dispatch_request
  File "natcap/invest/ui_server.py", line 105, in get_invest_validate
  File "natcap/invest/validation.py", line 1146, in _wrapped_validate_func
  File "natcap/invest/seasonal_water_yield/seasonal_water_yield.py", line 1421, in validate
  File "natcap/invest/validation.py", line 1070, in validate
  File "natcap/invest/validation.py", line 808, in check_spatial_overlap
  File "src/pygeoprocessing/geoprocessing_core.pyx", line 77, in pygeoprocessing.geoprocessing_core.gdal_use_exceptions.wrapper
  File "src/pygeoprocessing/geoprocessing_core.pyx", line 78, in pygeoprocessing.geoprocessing_core.gdal_use_exceptions.wrapper
  File "pygeoprocessing/geoprocessing.py", line 3594, in transform_bounding_box
  File "pygeoprocessing/geoprocessing.py", line 3574, in _transform_point
  File "osgeo/osr.py", line 2112, in TransformPoint
RuntimeError: Point outside of projection domain
```

Perhaps we should catch this one within `validation` and then return a validation message. That's the normal pattern for validation. In this specific case of `check_spatial_overlap`, one validation messages is associated with multiple args, which is not ideal for this error.

Also, we should consider catching exceptions like this at the flask app layer as well, for the benefit of Workbench users. See also #1610

davemfish:
When a model allows inputs to be in different projections, `check_spatial_overlap` tries to transform all input layer bounding boxes to EPSG:4326 in order to test that they all overlap. `transform_bounding_box` can raise `ValueError` or `RuntimeError` if a bounding box cannot be transformed. We need to handle those exceptions so that the process running validation does not crash.

I think it makes sense to skip the overlap check for the layer when this occurs, rather than issue a validation error. That matches the current behavior of `check_spatial_overlap` where there's already another case where a layer can be skipped. We don't want to prevent people from a running a model unless we're 100% certain the data is invalid.

Fixes #1849 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -64,6 +64,7 @@
 ..
   Unreleased Changes
   ------------------
+
 General
 =======
 * Handle exceptions that can arise during ``validation.check_spatial_overlap``
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -64,6 +64,11 @@
 ..
   Unreleased Changes
   ------------------
+General
+=======
+* Handle exceptions that can arise during ``validation.check_spatial_overlap``
+  when a layer's bounding box cannot be transformed to EPSG:4326.
+  (`#1849 <https://github.com/natcap/invest/issues/1849>`_).
 
 Workbench
 =========
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -811,7 +811,7 @@ def check_spatial_overlap(spatial_filepaths_list,
             except (ValueError, RuntimeError) as err:
                 LOGGER.debug(err)
                 LOGGER.warning(
-                    f'Skipping spatial overlap check for {filepath} '
+                    f'Skipping spatial overlap check for {filepath}. '
                     'Bounding box cannot be transformed to EPSG:4326')
                 continue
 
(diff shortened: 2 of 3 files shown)


claire-simpson (approved):
Looks like a good change!

PR #1896, opened 2025-04-18
Fix nodata handling in monthly QF output for SWY model

### LINKED ISSUE #1592: SWY monthly QF outputs have 0 values instead of nodata
@newtpatrol was testing 3.14.2 and noticed that `QF.tif` had some 0 values around the watershed, where it looks like `nodata` values should be. I found that the monthly QF is where this is introduced. I'm not yet sure if it's part of the logic of the function `_calculate_monthly_quickflow_flow` or if it's a nodata masking bug.

image

claire-simpson:
## Description
This PR fixes a bug in `_calculate_monthly_quick_flow` where pixels with nodata values in precipitation or number of rain events were incorrectly assigned a value of 0 in the QF output raster. Now, such pixels are correctly masked as nodata in the output. I also changed two variable names so that masking could be more easily interpreted and added a test to confirm that nodata values are treated correctly. 

Fixes #1592

## Checklist

### DIFF
--- a/src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
+++ b/src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
@@ -1097,8 +1097,8 @@ def _calculate_monthly_quick_flow(precip_path, n_events_path, stream_path,
         valid_p_mask = ~pygeoprocessing.array_equals_nodata(p_im, p_nodata)
         valid_n_mask = ~pygeoprocessing.array_equals_nodata(n_m, n_nodata)
         # precip mask: both p_im and n_m are defined and greater than 0
-        precip_mask = valid_p_mask & valid_n_mask & (p_im > 0) & (n_m > 0)
-        stream_mask = stream == 1
+        precip_ok = valid_p_mask & valid_n_mask & (p_im > 0) & (n_m > 0)
+        on_stream = stream == 1
         # stream_nodata is the only input that carries over nodata values from
         # the aligned DEM.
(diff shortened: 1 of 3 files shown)


src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -1163,14 +1163,15 @@ def qf_op(p_im, s_i, n_m, stream):
         # qf_im is the quickflow at pixel i on month m
         qf_im = numpy.full(p_im.shape, TARGET_NODATA, dtype=numpy.float32)
 
-        # case 1: where there is no precipitation
-        qf_im[~precip_mask] = 0
+        # case 1: where precip or n are <=0 but not nodata
+        case_1_mask = ~precip_ok & valid_p_mask & valid_n_mask
```
emlys, line 1167: There could be an argument made that, because QF = 0 where precipitation = 0 _regardless of_ `n`, you could leave out the `valid_n_mask` filter. But I think it makes more sense to be consistent and filter out the nodata areas.

emlys (approved):
Thanks @claire-simpson!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
claire-simpson: Thats a good point! I think it could go either way. If this change ends up resulting in an excessive amount of nodata values because of rain events I could see changing this.

PR #1897, opened 2025-04-21
Object-oriented model specs

### LINKED ISSUE #1891: Proposal: Object-oriented model specs
## Background/Motivation
InVEST models store key information about the model, its inputs, and outputs, in the `MODEL_SPEC` dictionary that is an attribute of the python module. The `MODEL_SPEC` has evolved over the last few years to contain much more information, but its structure hasn't been fundamentally changed or re-evaluated.

In the context of developing the plugins feature and associated documentation, I think it's time to consider overhauling the `MODEL_SPEC` from a dictionary to an object-oriented structure. When something is difficult to explain in writing, that can be a sign that it's unnecessarily complicated, and I'm certainly experiencing that trying to document the `MODEL_SPEC` as it currently exists. 

## Description
An object-oriented `MODEL_SPEC`, where the `MODEL_SPEC` is an object of type `ModelSpec`, which would have all the same attributes that the dictionary currently has. Arg specs would become `InputSpec` objects and output specs would become `OutputSpec` objects. (names subject to change)

## Benefits
- Syntactically, object attributes read more nicely than dictionary keys (`foo.bar` vs. `foo["bar"]`)
- Documentation is easier to write when we have objects to talk about, rather than explaining a nested dictionary structure. Particularly in the case of nested arg specs, I found this very difficult to explain without a concept of an `InputSpec` object to refer to. 
- Users will be more familiar with reading documentation and using APIs that are object-oriented 
- By instantiating an object, rather than writing out a dictionary, we can set constraints in the `__init__` method (making sure the necessary attributes are provided, types are correct, etc). With the dictionary approach, a separate script would be needed to check the specs for validity.

## Mitigation of side effects
Additional workload ahead of the plugins feature release - while more work upfront, it may save time in the long run vs. releasing something we know we'll want to change soon.

## Implementation details

Rather than manually editing each model's `MODEL_SPEC`, we should be able to make this conversion programmatically. I would start by writing the `ModelSpec` class with an `__init__` function that takes in an existing `MODEL_SPEC` dictionary.

It's expected that this change would affect these parts of our codebase:

- `natcap.invest` python package: many non-model files will be affected (`cli.py`, `datastack.py`, `models.py`, `spec_utils.py`, `ui_server.py`, `usage.py`, `validation.py`, `userguide_filevalidator.py`). A few test modules will need changes. Nearly all changes are trivially swapping out dictionary access for object attribute access syntax. The `invest getspec` command will need to be updated to correctly print out the object as JSON or plain text.
- The trivial demo invest plugin will need to be updated
- The `investspec` sphinx extension that we use to auto-generate parts of the user's guide will need updating to parse the new spec format.
- Workbench: no changes expected. Since `MODEL_SPEC`-related information is passed through the server, the same request/response JSON format can be preserved.
- No changes are anticipated to github actions workflows; sample data; test data; auto-generated `natcap.invest` API docs; rec model server; or build and release process.

emlys:
## Description
Fixes #1891 

First attempt at moving from dictionary-based to class-based model specs.
- Wrapped all model specs in a new function, `build_model_spec`, that converts the old dictionary style to the new classes. Eventually this function won't be necessary. In a future PR, we'll rewrite the model specs by instantiating the classes directly.
- Created the `ModelSpec` class, and classes to represent each input and output type
- Moved the validation check functions into their corresponding input classes
- Updated syntax throughout
- Renamed `spec_utils` to `spec` because it now has a lot more going on than just utility functions
- Renamed `args` to `inputs` in the model spec
- Updated JSON serialization to work with the new data structure
- Converted the validation `timeout` function to work as a decorator that can be applied to each input class's `validate` function
- ~~The API interface with the workbench is unchanged, so no changes to the workbench were needed.~~ I made _one_ small change on the workbench side, moving the `ui_spec.order` property into a standalone property, `input_field_order`.
- Re-enabled the compiled models so that changes to those modules can be tested. We disable them before merging.

## Checklist


tests/test_model_specs.py
```diff
@@ -57,59 +121,57 @@ def test_model_specs_are_valid(self):
...
-                    required_keys.issubset(model.MODEL_SPEC),
+                    required_keys.issubset(set(dir(model.MODEL_SPEC))),
                     ("Required key(s) missing from MODEL_SPEC: "
-                     f"{set(required_keys).difference(model.MODEL_SPEC)}"))
-                extra_keys = set(model.MODEL_SPEC).difference(required_keys)
-                if (extra_keys):
-                    self.assertEqual(extra_keys, set([optional_spatial_key]))
+                     f"{set(required_keys).difference(set(dir(model.MODEL_SPEC)))}"))
+
+                if model.MODEL_SPEC.args_with_spatial_overlap:
```
emlys, line 135: This whole module could be obsoleted by adding good type checking to the model spec classes, with or without pydantic

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+        projected: Defaults to None, indicating a projected (as opposed to
+            geographic) coordinate system is not required. Set to True if a
+            projected coordinate system is required.
+        projection_units: Defaults to None. If `projected` is `True`, and a
+            specific unit of projection (such as meters) is required, indicate
+            it here.
+    """
+    band: typing.Union[Input, None] = None
+    geometries: set = dataclasses.field(default_factory=dict)
+    fields: typing.Union[typing.Iterable[Input], None] = None
```
dcdenu4, line 435: Do we need to define these here since they're already defined in the parent classes of `SingleBandRasterInput` or `VectorInput`?

src/natcap/invest/validation.py
```diff
@@ -81,7 +77,7 @@ def _evaluate_expression(expression, variable_map):
     Args:
         expression (string): A string expression that returns a value.
         variable_map (dict): A dict mapping string variable names to their
-            python object values.  This is the variable map that will be used
+            python object values.  This is the variable map that will be usedf
```
dcdenu4, line 80:
```suggestion
            python object values.  This is the variable map that will be used
```

src/natcap/invest/validation.py
```diff
@@ -998,55 +303,37 @@ def validate(args, spec, spatial_overlap_opts=None):
...
+                    if (isinstance(nested_spec.required, str)):
+                        nested_spec.required = (
                             bool(_evaluate_expression(
-                                nested_spec['required'], expression_values)))
-
-        type_validation_func = _VALIDATION_FUNCS[param_type]
-
-        if type_validation_func is None:
-            # Validation for 'other' type must be performed by the user.
-            continue
```
dcdenu4, line 1033: Huh, was this something from a leftover idea we once had?

tests/test_delineateit.py
```diff
@@ -26,8 +26,8 @@
     'delineateit')
 
 # Skipping all compiled model tests temporarily for feature/plugins
-pytestmark = pytest.mark.skip(
-        reason="Temporarily ignoring compiled models for feature/plugins")
+# pytestmark = pytest.mark.skip(
```
dcdenu4, line 29: Is this just for testing for the PR, do we plan to uncomment these again?

tests/test_model_specs.py
```diff
@@ -6,47 +6,111 @@
...
+        spec.PercentInput,
+        spec.StringInput,
+        spec.OptionStringInput,
+        spec.BooleanInput,
+        spec.SingleBandRasterInput,
+        spec.VectorInput,
+        spec.RasterOrVectorInput,
+        spec.CSVInput,
+        spec.DirectoryInput,
+        spec.FileInput
```
dcdenu4, line 31: Annoying request, can we alphabetize these?

dcdenu4: And within the grouped lists below?

tests/test_validation.py
```diff
@@ -18,8 +18,45 @@
 from osgeo import ogr
 from osgeo import osr
 
+from natcap.invest import spec
```
dcdenu4, line 21: Do we need this import if you're specifying components below?

dcdenu4: Other cases of `spec` might be locally imported to the tests themselves?

dcdenu4 (changes requested):
Holy toledo, thanks @emlys ! Here's my first pass review. One thing that jumped out is around using `__class__` in a few places as opposed to `isinstance` to check class types. I was also curious about a few class inheritance related things and needing to respecify attributes on the child class. One last thing is that I noticed the removal of [NUTRIENT] syntax in NDR. I think this is great and was wondering if we do it anywhere else still and how this effects the UG builds?

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+        Args:
+            col: Column of a pandas dataframe to format
+
+        Returns:
+            Transformed dataframe column
+        """
+        return col.astype(float)
+
+@dataclasses.dataclass
+class IntegerInput(Input):
```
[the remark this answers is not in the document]
emlys: In the old model spec design, integers don't support `unit` and `expression`. I think it could make sense to change that, but for the sake of consistency I tried to keep everything functionally the same as before.

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+            regexp=arg.get('regexp', None))
+
+    elif t == 'number':
+        return NumberInput(
+            **base_attrs,
+            units=arg['units'],
+            expression=arg.get('expression', None))
+
+    elif t == 'integer':
+        return IntegerInput(**base_attrs)
```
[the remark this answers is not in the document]
emlys: Correct, integers have no units. (Which makes sense most of the time)

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+    def validate(self, filepath: str):
+        """Validate a raster file against the requirements for this input.
+
+        Args:
+            filepath (string): The filepath to validate.
+
+        Returns:
+            A string error message if an error was found.  ``None`` otherwise.
+        """
+        file_warning = FileInput.validate(self, filepath)
```
davemfish, line 301: In the `VectorInput` case you called `super().validate(filepath)` for this purpose. That seems like the more general solution, though they may be equivalent in all cases we have so far.

src/natcap/invest/coastal_vulnerability.py
```diff
@@ -2871,10 +2868,8 @@ def assemble_results_and_calculate_exposure(
             with open(pickle_path, 'rb') as file:
                 final_values_dict[var_name] = pickle.load(file)
 
-    habitat_df = validation.get_validated_dataframe(
-        habitat_protection_path, **MODEL_SPEC['outputs']['intermediate'][
-            'contents']['habitats']['contents']['habitat_protection.csv']
-    ).rename(columns={'r_hab': 'R_hab'})
+    habitat_df = utils.read_csv_to_dataframe(
```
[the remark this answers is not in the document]
emlys: Yes, I realized that `get_validated_dataframe` wasn't really adding anything here that `read_csv_to_dataframe` can't do. And `CSVOutput` objects don't currently have the `get_validated_dataframe` method.

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+        projection_units: Defaults to None. If `projected` is `True`, and a
+            specific unit of projection (such as meters) is required, indicate
+            it here.
+    """
+    band: typing.Union[Input, None] = None
+    geometries: set = dataclasses.field(default_factory=dict)
+    fields: typing.Union[typing.Iterable[Input], None] = None
+    projected: typing.Union[bool, None] = None
+    projection_units: typing.Union[pint.Unit, None] = None
+    type: typing.ClassVar[str] = 'raster_or_vector'
```
davemfish, line 438: I'm not sure where to look for this so I'll just ask, where are we using this `type` attribute? Could we use builtin `type(foo)` or `isinstance` instead? Where `foo = RasterOrVectorInput(...)`

src/natcap/invest/datastack.py
```diff
@@ -347,7 +342,7 @@ def build_datastack_archive(args, model_id, datastack_path):
             target_arg_value = target_csv_path
             files_found[source_path] = target_arg_value
 
-        elif input_type == 'file':
+        elif input_spec.__class__ is spec.FileInput:
```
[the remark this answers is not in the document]
emlys: We don't want `isinstance` here because this block is specific to `FileInput`, and not other classes that inherit from `FileInput`. Example:
```
>>> a = 1
>>> isinstance(a, int)
True
>>> isinstance(a, object)
True
>>> a.__class__ is int
True
>>> a.__class__ is object
False
```

tests/test_validation.py
emlys: We can remove the redundant imports from the old test cases. AFAIK there is no reason to re-import the module within each test function. I think we have those extra imports left over from a time when it did matter?

tests/test_delineateit.py
emlys: This was just for this PR. Now that #1915 is merged, I need to resolve the merge conflict.

src/natcap/invest/validation.py
emlys: We used to support the "other" type but never used it. Now that we're using classes, if someone wanted to make another input type, they can subclass from `Input`!

src/natcap/invest/datastack.py
```diff
@@ -258,22 +260,16 @@ def build_datastack_archive(args, model_id, datastack_path):
                 rewritten_args[key] = files_found[source_path]
                 continue
 
-        if input_type == 'csv':
+        if input_spec.__class__ is spec.CSVInput:
```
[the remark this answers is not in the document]
emlys: Either one would work here. Currently we don't have any classes that inherit from `CSVInput`, so it wouldn't matter. But if a class did inherit from `CSVInput`, should this block apply to it?

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+        Args:
+            col: Column of a pandas dataframe to format
+
+        Returns:
+            Transformed dataframe column
+        """
+        return col.astype(float)
+
+@dataclasses.dataclass
+class PercentInput(Input):
```
davemfish, line 852: Perhaps this could inherit from `NumberInput`? Then might not need the `format_column` static method. And `validate` could call `super().validate()`, if desired.

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+    This represents a vector file output (all GDAL-supported vector file types
+    are allowed). It is assumed that only the first layer is used.
+
+    Attributes:
+        geometries: A set of geometry type(s) that are produced in this vector
+        fields: An iterable of `Output`s representing the fields created in
+            this vector. The `key` of each input must match the corresponding
+            field name.
+    """
+    geometries: set = dataclasses.field(default_factory=set)
```
davemfish, line 1044: Same comment here about possibly renaming to "geometry_types"

davemfish (commented):
I did an initial review of `spec.py`. Thanks, @emlys !

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+            vector is expected to have. The `key` of each input must match the
+            corresponding field name.
+        projected: Defaults to None, indicating a projected (as opposed to
+            geographic) coordinate system is not required. Set to True if a
+            projected coordinate system is required.
+        projection_units: Defaults to None. If `projected` is `True`, and a
+            specific unit of projection (such as meters) is required, indicate
+            it here.
+    """
+    geometries: set = dataclasses.field(default_factory=dict)
```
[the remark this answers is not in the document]
emlys: Sure 👍

src/natcap/invest/spec.py
emlys: You're right, this can be `super()`. I think this was left over from a multiple inheritance setup I tried.

src/natcap/invest/spec.py
emlys: Sure, I did not do this originally because in the old model spec, `number`s have attributes that don't apply to `percent` and `ratio`s. But I think it's fine to add those attributes for the sake of a sensible inheritance structure.

tests/test_model_specs.py
emlys: Done in 57f91dfd8

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+        # Check that an overview .ovr file wasn't opened.
+        if os.path.splitext(filepath)[1] == '.ovr':
+            return get_message('OVR_FILE')
+
+        srs = gdal_dataset.GetSpatialRef()
+        projection_warning = _check_projection(srs, self.projected, self.projection_units)
+        if projection_warning:
+            return projection_warning
+
+
```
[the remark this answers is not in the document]
emlys: I think we want double blank lines between classes (which I did not do consistently elsewhere, but fixed in a1dde022f)

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+            base_path: Base path of the source CSV. Relative file path values
+                will be expanded relative to this base path.
+
+        Returns:
+            Transformed dataframe column
+        """
+        return col.apply(
+            lambda p: p if pandas.isna(p) else utils.expand_path(str(p).strip(), base_path)
+        ).astype(pandas.StringDtype())
+
```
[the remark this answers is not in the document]
emlys: Fixed in a1dde022f

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+    Attributes:
+        units: The units of measurement for this numeric value
+    """
+    units: typing.Union[pint.Unit, None] = None
+
+@dataclasses.dataclass
+class IntegerOutput(Output):
+    """An integer output, or result, of an invest model."""
+    pass
```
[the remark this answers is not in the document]
emlys: We might want to add more to them in the future, but for now, the only purpose they serve is to indicate the data type of the output (integer, raster, etc.)

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+
+@dataclasses.dataclass
+class ModelSpec:
+    model_id: str
+    model_title: str
+    userguide: str
+    input_field_order: list[str]
+    inputs: typing.Iterable[Input]
+    outputs: typing.Iterable[Output]
+    args_with_spatial_overlap: dict
```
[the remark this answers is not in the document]
emlys: Yeah, I'd be in favor of renaming this, and maybe there's a better design for it entirely... hmm

src/natcap/invest/datastack.py
```diff
@@ -376,22 +371,17 @@ def build_datastack_archive(args, model_id, datastack_path):
             target_arg_value = target_directory
             files_found[source_path] = target_arg_value
 
-        elif input_type in spatial_types:
+        elif input_spec.__class__ in spatial_types:
             # Create a directory with a readable name, something like
             # "aoi_path_vector" or "lulc_cur_path_raster".
-            spatial_dir = os.path.join(data_dir, f'{key}_{input_type}')
+            spatial_dir = os.path.join(data_dir, f'{key}_{input_spec.__class__.__name__}')
```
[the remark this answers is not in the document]
emlys: I ended up changing this to use the `input_spec.type` property instead

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2042 @@
...
+
+@dataclasses.dataclass
+class SingleBandRasterInput(FileInput):
+    """A single-band raster input, or parameter, of an invest model.
+
+    This represents a raster file input (all GDAL-supported raster file types
+    are allowed), where only the first band is needed.
+
+    Attributes:
+        band: An `Input` representing the type of data expected in the
```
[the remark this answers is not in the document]
emlys: I updated this like we discussed in [14d696e](https://github.com/natcap/invest/pull/1897/commits/14d696e90e672a83b4c7a5be48a2e61d729d78b8)

src/natcap/invest/spec.py
emlys: You're right, they do not need to be redefined here!

src/natcap/invest/spec.py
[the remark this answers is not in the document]
emlys: I looked at this more, and I'm going to suggest changing it to two separate boolean attributes - 
`validate_spatial_overlap`: Whether to validate that all top-level spatial inputs overlap. For all invest models, the list of `spatial_inputs` includes _all_ spatial inputs, except for 2 cases where an input was left out seemingly accidentally. So I don't see a need to continue supporting checking just a sub-group of inputs. (And if someone _did_ want that, they can set this attribute to `False` and validate it on their own).

`different_projections_ok` - This attribute can move up to the model spec itself.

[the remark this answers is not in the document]
davemfish: Will checking _all_ spatial inputs for overlap include ones that are referenced within CSVs? I think it should not because it's quite common for those layers to not necessarily overlap with one another, or even with the AOI in some cases where models are interested in distance to nearest feature, for example (Rec, CV, HRA?). In CV, the continental shelf input and optional SLR input are also not included in the overlap list for a similar reason: they are about proximity, not overlap.

src/natcap/invest/spec.py
davemfish: Hmm interesting. If it's working for `VectorInput.validate` to call `super().validate()` then shouldn't it be okay for `SingleBandRasterInput.validate` to do the same? I guess I see why `RasterOrVectorInput.validate` cannot call `super()`, but not sure I understand the others.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -118,15 +118,16 @@ export function setupAddPlugin(i18n) {
             upath.join(path, 'pyproject.toml')
           ).toString());
         }
-
+        console.log(pyprojectTOML);
```
dcdenu4, line 121: Leaving in or just for debugging?

tests/test_validation.py
```diff
@@ -259,12 +294,11 @@ def validate(args, limit_to=None):
 
     def test_n_workers(self):
         """Validation: validation error returned on invalid n_workers."""
-        from natcap.invest import spec_utils
+        from natcap.invest import spec
```
dcdenu4, line 297: Can remove if we're importing at the top level.

tests/test_validation.py
```diff
@@ -541,12 +569,12 @@ def test_raster_not_projected(self):
...
         raster = None
 
-        error_msg = validation.check_raster(filepath, projected=True)
+        error_msg = SingleBandRasterInput(projected=True).validate(filepath)
         self.assertEqual(error_msg, validation.MESSAGES['NOT_PROJECTED'])
 
     def test_raster_incorrect_units(self):
         """Validation: test when a raster projection has wrong units."""
-        from natcap.invest import spec_utils
+        from natcap.invest import spec
```
dcdenu4, line 577: Remove since importing at top level module

tests/test_validation.py
```diff
@@ -619,16 +650,20 @@ def test_missing_fieldnames(self):
...
+                Input(id='col_c')]
+            ).validate(filepath)
         expected = validation.MESSAGES['MATCHED_NO_HEADERS'].format(
             header='field', header_name='col_c')
         self.assertEqual(error_msg, expected)
 
     def test_vector_projected_in_m(self):
         """Validation: test that a vector's projection has expected units."""
-        from natcap.invest import spec_utils
+        from natcap.invest import spec
```
dcdenu4, line 666: Dare I comment on line 666? Remove import if importing at top level

tests/test_validation.py
dcdenu4: Yea, from my understanding if we're using the import everywhere then importing at top level makes sense. If only a few test cases used an import I could see isolating that import to those functions specifically.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -151,11 +176,13 @@ export function setupAddPlugin(i18n) {
...
             type: 'plugin',
             source: installString,
-            env: pluginEnvPrefix,
+            env: newPluginEnvPrefix,
+            version: version,
           }
         );
         logger.info('successfully added plugin');
       } catch (error) {
+        console.log(error);
```
dcdenu4, line 185: Leave in or for debugging?

dcdenu4 (changes requested):
Thanks @emlys , back to you!

src/natcap/invest/spec.py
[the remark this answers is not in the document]
emlys: > Will checking all spatial inputs for overlap include ones that are referenced within CSVs?

No, just the top-level spatial inputs, like it does currently

> top-level args that are also not included in the overlap list 

Good point. I updated it so that `validate_spatial_overlap` can be a list representing a subset of inputs to check, if that applies. So `validate_spatial_overlap=True` just serves as a shorthand rather than listing out all the inputs.

src/natcap/invest/datastack.py
```diff
@@ -34,10 +34,11 @@
 
 from osgeo import gdal
 
-from . import spec_utils
+from . import spec
```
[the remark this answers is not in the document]
emlys: Fixed!

workbench/src/main/setupAddRemovePlugin.js
emlys: I think this will be useful in the future for debugging

emlys:
@dcdenu4 Sorry to pile on to this PR, but I added a few important updates - 
- reorganized the `args_with_spatial_overlap` attribute into `validate_spatial_overlap` and `different_projections_ok` attributes
- Uniquely identify plugins on the workbench side by a combination of model ID and package version so that multiple versions of the same plugin won't overwrite each other
- Deduplicated `model_id` and `model_title` from the pyproject.toml - they are now pulled from the `MODEL_SPEC` only. This required updates to the plugin install process.

src/natcap/invest/spec.py
emlys: I made `RasterOrVectorInput` a separate class that does not inherit from `SingleBandRasterInput` or `VectorInput` so that we don't have these issues.

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2227 @@
...
+                return required_field_warning
+
+        srs = layer.GetSpatialRef()
+        projection_warning = _check_projection(srs, self.projected, self.projection_units)
+        return projection_warning
+
+
+@dataclasses.dataclass
+class RasterOrVectorInput(FileInput):
+    """An invest model input that can be either a single-band raster or a vector."""
```
dcdenu4, line 497: I think this gets us under 80 chars.

```suggestion
    """An invest model input of either a single-band raster or a vector."""
```

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2227 @@
...
+        # rename 'inputs' to 'args' to stay consistent with the old api
+        spec_dict.pop('inputs')
+        spec_dict.pop('inputs_dict')
+        spec_dict.pop('outputs_dict')
+        spec_dict['args'] = self.inputs_dict
+        spec_dict['outputs'] = self.outputs_dict
+        return json.dumps(spec_dict, default=fallback_serializer, ensure_ascii=False)
+
+
+def build_model_spec(model_spec):
```
dcdenu4, line 1334: Docstring?

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2227 @@
...
+        userguide=model_spec['userguide'],
+        aliases=model_spec['aliases'],
+        inputs=inputs,
+        outputs=outputs,
+        input_field_order=model_spec['ui_spec']['order'],
+        validate_spatial_overlap=validate_spatial_overlap,
+        different_projections_ok=different_projections_ok)
+
+
+def build_input_spec(argkey, arg):
```
dcdenu4, line 1367: Docstring?

src/natcap/invest/spec.py
```diff
@@ -0,0 +1,2227 @@
...
+            data_type=int if arg['bands'][1]['type'] == 'integer' else float,
+            units=arg['bands'][1].get('units', None),
+            projected=arg.get('projected', None),
+            projection_units=arg.get('projection_units', None))
+
+    else:
+        raise ValueError
+
+
+def build_output_spec(key, spec):
```
dcdenu4, line 1467: Docstring?

src/natcap/invest/validation.py
```diff
@@ -1007,55 +312,46 @@ def validate(args, spec, spatial_overlap_opts=None):
...
+    if model_spec.validate_spatial_overlap:
+
+        # validate_spatial_overlap can be True, meaning validate all spatial keys,
+        # or a list, representing a subset of keys to validate
+        if isinstance(model_spec.validate_spatial_overlap, list):
+            spatial_keys = set(model_spec.validate_spatial_overlap)
+        else:
+            spatial_keys = set()
+            for i in model_spec.inputs:
+                if i.type in['raster', 'vector']:
```
dcdenu4, line 353: @davemfish, here's an instance where `type` is used!

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -143,19 +147,38 @@ export function setupAddPlugin(i18n) {
            'python', '-m', 'pip', 'install', installString]
         );
         logger.info('installed plugin into its env');
+
+        event.sender.send('plugin-install-status', i18n.t('Importing plugin...'));
+        // Access plugin metadata from the MODEL_SPEC
+        const modelID = execSync(
```
dcdenu4, line 153: Some serious inception.

dcdenu4 (changes requested):
Hey @emlys, I just had some minor suggestions and pointed out a few missing docstrings.

src/natcap/invest/spec.py
emlys: I'll add a short docstring! This function (and `build_input_spec` and `build_output_spec`) are just temporary as we transition over from the model spec dictionaries, so I didn't put as much effort into documenting them.

src/natcap/invest/validation.py
emlys: Its main purpose is to indicate the input type when we JSON serialize it and send it to the workbench. (Maintaining the old API)

dcdenu4 (approved):
Thanks @emlys !

Issue #1699, opened 2024-11-25
Wind energy returns harvested energy values for masked out points


newtpatrol:
From a Slack thread between Stacie and Rob Griffin:

> SW: wind_energy_points.shp is based on the Wind Data Points input, and each point seems to have output data for fields Dens_W/m2 and Harv_MWhr, even though density_W_per_m2.tif and harvested_energy_MWhr_per_yr.tif have values of NoData at these locations. Why is this?

> RG: It is ok that Dens_W/m2 returns values even for locations that are masked out based on depth/distance constraints, as this is a metric that is based on wind resource independent of any interaction with turbines. _**Returning values for harvested energy (or any valuation outputs) for those same locations does not make sense though. I suspect this would be straightforward to fix for the software team using distance/depth indicators.**_

Also, this point needs clarification in the User Guide, so it's easier to understand the relationship between the rasters and wind input, and where the wind point output gets its fields from.  In case it's useful later, when I asked for clarification on the first point, Rob said this:

> RG: The depth and distance constraints in the input fields control which areas are viable for placing turbines. All modeled results (points and rasters) should only be given for these areas, with the possible exception of Dens_W/m2 as this is a metric that is based on wind resource independent of any interaction with turbines. It should probably be an intermediate output (as a raster), as it is really just restating in different terms the wind resource input file.

Rob says that he has a lot of UG updated queued up, so let's make sure this is one of them.

PR #1898, opened 2025-04-25
Wind Energy: Remove Raster Outputs

### LINKED ISSUE #1698: Wind Energy: Raster result values returned outside of wind data

megannissel:
## Description
Fixes #1698 and #1699 

ADR: #1892 
User's Guide PR: https://github.com/natcap/invest.users-guide/pull/178
Test Data PR: [invest-test-data#36](https://bitbucket.org/natcap/invest-test-data/pull-requests/36)

The Wind Energy model has historically interpolated Harvested and Density values to rasters, which are then used to calculate NPV, Levelized Cost, and Carbon Emissions. The method of interpolation resulted in data existing outside of the bounds of the input wind data. 

Based on Rob's use of the model, and review and evaluation of the problem by the team, the consensus is that the model's current use of interpolation introduces too many potential violations of the constraints of the model (e.g. interpolating over areas that are invalid due to ocean depth or distance from shore, or are outside of the areas included in the input wind speed data) and requires assumptions that may not be helpful for users. 

We are therefore removing the raster outputs and retaining the associated values in the output `wind_energy_points.shp` vector. (In cases where Valuation is run, there will still be intermediate rasters for Harvested, NPV, Levelized Cost, and Carbon Emissions.)

The process of doing so has required careful consideration on a few points:
- We want to do away with interpolation but still be able to make use of the efficiency that comes along with raster calculations. As such, Harvested values do still need to be rasterized so that the raster can be used in the NPV, Levelized Cost, and Carbon Emissions calculations. This PR moves rasterization of Harvested values into the Valuation half of the model, since this raster is only needed for Valuation calculations. 
- Higher-resolution wind datasets are now available, some with horizontal resolution as fine as 2km. As such, when rasterizing the Wind Data, we need to make sure our pixel size is fine enough. This PR manually sets a (90, -90) pixel size for the rasters.
- To write raster values back to the wind data vector, this PR uses pixel-picking. The intermediate `unmasked_wind_energy_points.shp` contains Density and Harvested values for all points, as well as an additional field, `Masked`, that indicates whether or not a point has been masked out. The final output only contains valid points, with values for Density, Harvested, Depth, Distance, NPV, Levelized Cost, and Carbon Emissions. 

These changes also address #1699; data points in the final wind vector will now be appropriately masked out by depth and distance. 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -89,6 +89,7 @@ Wind Energy
 * The model no longer returns results as rasters; instead, values are
   written to the output ``wind_energy_points`` shapefile for each point
   (`#1698 <https://github.com/natcap/invest/issues/1698>`_).
+  Any Decision Record (ADR): `ADR-0004: Remove Wind Energy Raster Outputs <https://github.com/natcap/invest/blob/main/doc/decision-records/ADR-0004-Remove-Wind-Energy-Raster-Outputs.md>`_
 * The output ``wind_energy_points.shp`` no longer returns Harvested or
   Valuation-related values for points that are invalid wind farm locations
   due to depth or distance constraints
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -77,6 +77,16 @@ Urban Flood Risk
   ``Runoff_retention_m3.tif``
   (`#1837 <https://github.com/natcap/invest/issues/1837>`_).
 
+Wind Energy
+===========
+* The model no longer returns results as rasters; instead, values are
+  written to the output ``wind_energy_points`` shapefile for each point
+  (`#1698 <https://github.com/natcap/invest/issues/1698>`_).
+* The output ``wind_energy_points.shp`` no longer returns Harvested or
+  Valuation-related values for points that are invalid wind farm locations
+  due to depth or distance constraints
+  (`#1699 <https://github.com/natcap/invest/issues/1699>`_).
+
 
 3.15.1 (2025-05-06)
 -------------------
--- a/src/natcap/invest/wind_energy.py
+++ b/src/natcap/invest/wind_energy.py
@@ -18,6 +18,7 @@ from osgeo import ogr
 from osgeo import osr
 from scipy import integrate
 from shapely import speedups
+from rtree import index
 
 from . import gettext
 from . import spec_utils
@@ -429,6 +430,24 @@ MODEL_SPEC = {
         "output": {
             "type": "directory",
             "contents": {
+                "wind_energy_points.shp": {
+                    "about": gettext("Map of summarized data at each point."),
+                    "geometries": spec_utils.POINT,
+                    "fields": OUTPUT_WIND_DATA_FIELDS
+                }
+            }
+        },
+        "intermediate": {
+            "type": "directory",
+            "contents": {
+                "aoi_raster.tif": {
+                    "about": "Empty raster covering the extent of the AOI",
+                    "bands": {1: {"type": "integer"}}
+                },
+                "bathymetry_projected.tif": {
+                    "about": "Clipped and reprojected bathymetry map",
+                    "bands": {1: {"type": "number", "units": u.meter}}
+                },
                 "carbon_emissions_tons.tif": {
                     "about": gettext(
                         "Map of offset carbon emissions for a farm centered "
@@ -444,6 +463,22 @@ MODEL_SPEC = {
                         "type": "number",
                         "units": u.watt/u.meter**2
                     }}
+                },
+                "depth_mask.tif": {
+                    "about": (
+                        "Bathymetry map masked to show only the pixels that "
+                        "fall within the allowed depth range"),
+                    "bands": {1: {"type": "number", "units": u.meter}}
+                },
+                "distance_mask.tif": {
+                    "about": (
+                        "Distance to shore, masked to show only the pixels "
+                        "that fall within the allowed distance range"),
+                    "bands": {1: {"type": "number", "units": u.meter}}
+                },
+                "distance_trans.tif": {
+                    "about": "Distance to shore from each pixel",
+                    "bands": {1: {"type": "number", "units": u.meter}}
                 },
                 "harvested_energy_MWhr_per_yr.tif": {
                     "about": gettext(
@@ -467,40 +502,6 @@ MODEL_SPEC = {
                     "about": gettext(
                         "Map of the net present value of a farm centered on each pixel."),
                     "bands": {1: {"type": "number", "units": u.currency}}
-                },
-                "wind_energy_points.shp": {
-                    "about": gettext("Map of summarized data at each point."),
-                    "geometries": spec_utils.POINT,
-                    "fields": OUTPUT_WIND_DATA_FIELDS
-                }
-            }
-        },
-        "intermediate": {
-            "type": "directory",
-            "contents": {
-                "aoi_raster.tif": {
-                    "about": "Empty raster covering the extent of the AOI",
-                    "bands": {1: {"type": "integer"}}
-                },
-                "bathymetry_projected.tif": {
-                    "about": "Clipped and reprojected bathymetry map",
-                    "bands": {1: {"type": "number", "units": u.meter}}
-                },
-                "depth_mask.tif": {
-                    "about": (
-                        "Bathymetry map masked to show only the pixels that "
-                        "fall within the allowed depth range"),
-                    "bands": {1: {"type": "number", "units": u.meter}}
-                },
-                "distance_mask.tif": {
-                    "about": (
-                        "Distance to shore, masked to show only the pixels "
-                        "that fall within the allowed distance range"),
-                    "bands": {1: {"type": "number", "units": u.meter}}
-                },
-                "distance_trans.tif": {
-                    "about": "Distance to shore from each pixel",
-                    "bands": {1: {"type": "number", "units": u.meter}}
                 },
                 "projected_clipped_land_poly.shp": {
                     "about": "Clipped and reprojected land polygon vector",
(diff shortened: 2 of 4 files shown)


src/natcap/invest/wind_energy.py
```diff
@@ -1259,45 +1297,274 @@ def execute(args):
...
+
+    # Add the mask field to the base vector
+    if mask_keys:
+        base_vector = gdal.OpenEx(
+            base_point_vector_path, gdal.OF_VECTOR | gdal.GA_Update)
+        base_layer = base_vector.GetLayer()
+        field_defn = ogr.FieldDefn(mask_field, ogr.OFTReal)
+        field_defn.SetWidth(24)
+        field_defn.SetPrecision(11)
+        base_layer.CreateField(field_defn)
```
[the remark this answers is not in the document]
dcdenu4: Oh, but maybe `mask_keys` is entirely optional, so if you don't want this functionality you can leave it blank.

src/natcap/invest/wind_energy.py
```diff
@@ -1259,45 +1297,274 @@ def execute(args):
...
+                try:
+                    block_value = block_matrix[j][i]
+                except IndexError:
+                    # It is possible for an index to be *just* on the edge of a
+                    # block and exceed the block dimensions.  If this happens,
+                    # pass on this point, as another block's bounding box should
+                    # catch this point instead.
+                    continue
+                else:
+                    if block_value == _TARGET_NODATA:
```
dcdenu4, line 1536: Should we be using the input raster's nodata values from raster_info instead?

src/natcap/invest/wind_energy.py
```diff
@@ -949,144 +972,158 @@ def execute(args):
...
-    temp_harvested_raster_path = os.path.join(
-        inter_dir, 'temp_harvested%s.tif' % suffix)
+    align_dependent_task_list = [create_depth_mask_task]
+    rasters_to_align_list = [depth_mask_path]
+    aligned_depth_mask_path = depth_mask_path.replace(
+        '%s.tif' % suffix, '_aligned%s.tif' % suffix)
+    aligned_rasters_list = [aligned_depth_mask_path]
+
+    # If we will be masking by distance, align the distance raster as well
+    if mask_by_distance:
```
dcdenu4, line 982: I'll have to think about this some more. But instead of having to worry about aligning everything on the fly, I'm wondering if we can just get a common bounding box to use upfront and create all our rasters using that bounding box. I'm not yet sure if it makes sense or is possible, but my instinct is to want to handle as much "table setting" up front as we can.

src/natcap/invest/wind_energy.py
```diff
@@ -949,144 +972,158 @@ def execute(args):
...
+        aligned_dist_mask_path = dist_mask_path.replace(
+            '%s.tif' % suffix, '_aligned%s.tif' % suffix)
+        aligned_rasters_list.append(aligned_dist_mask_path)
+    else:
+        # No mask_by_distance_task is added to taskgraph
+        LOGGER.info('NO Distance Mask to add to list')
+        # No need to align bathy
+
+    if not run_valuation:
+        if mask_by_distance:
```
dcdenu4, line 1006: We don't want to do this regardless of `run_valuation` because we'll have to do it again if running valuation. I'll check that assumption below.

dcdenu4 (changes requested):
Thanks @megannissel, I took a first pass at this and left some comments. I think I still want to review some of the alignment and set up for that. Maybe see if there's a way to reduce some of the mask by distance conditionals and run valuation conditionals. There might not be! Thanks for taking this on!

src/natcap/invest/wind_energy.py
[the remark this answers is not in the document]
megannissel: Yeah, this whole function was cleaner before we decided we ought to remove invalid points from the final output vector and include a `Masked` field in the intermediate vector. 

Adding the mask value to the base vector is intentional, since I figured we probably didn't want two intermediate vectors that only differ by the single `Masked` field (it seemed to me that having two would be more confusing to users). `mask_keys` is entirely optional; I've updated the docstring to make this clearer. 

There is a part of me that wonders if we ought to be including the actual depth and distance values in the intermediate file as well, so that it's easy to determine WHY a point was masked out. But if we do want to do that, I think I need to do a larger rework of this function. And perhaps that would be a reason to not mutate the base vector at all and to instead create two vector copies: one that retains all points as an intermediate output and contains Depth, Distance, and Masked fields, and one that retains only valid points as the final output? But I may be over-complicating things.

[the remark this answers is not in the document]
megannissel: I gave this some additional thought; if we want to keep the "mutate the base vector" functionality, it would make more sense if that behavior was dependent upon passing a value for `mask_field` (perhaps renamed for clarity). That way, `mask_keys` would be responsible only for the clipping behavior, but wouldn't mutate the base vector. If base vector mutation is desired, values would need to be passed for `mask_keys` AND `mask_field`.

src/natcap/invest/wind_energy.py
megannissel: It would be awesome if we could do that. I'll give it some thought as well.

dcdenu4 (changes requested):
We chatted about removing some legacy operations around reprojecting and clipping for the bathymetry raster that hopefully will clean some of this up for us.

src/natcap/invest/wind_energy.py
```diff
@@ -1259,45 +1214,301 @@ def execute(args):
...
 
+def _index_raster_values_to_point_vector(
+        base_point_vector_path, raster_fieldname_list,
+        target_point_vector_path, overwrite=False,
+        mask_keys=[], mask_field=None):
+    """Add raster values to vector point feature fields.
+
+    This function does two things:
+        - Creates a copy of the base point vector and updates the copy to include
+          the fields in raster_fieldname_list, the values of which are pixel-picked
```
dcdenu4, line 1291:
```suggestion
          the fields in ``raster_fieldname_list``, the values of which are pixel-picked
```

src/natcap/invest/wind_energy.py
```diff
@@ -1259,45 +1214,301 @@ def execute(args):
...
+          from their corresponding raster. If `mask_keys` are provided, points that
```
dcdenu4, line 1292:
```suggestion
          from their corresponding raster. If ``mask_keys`` are provided, points that
```

dcdenu4 (changes requested):
Thanks @megannissel , just a few comments and questions. Back to you!

tests/test_wind_energy.py
```diff
@@ -400,6 +400,107 @@ def test_calculate_npv_levelized_rasters(self):
...
+            output_layer.CreateFeature(output_feature)
+            output_feature = None
+
+        output_layer.SyncToDisk()
+        output_layer = None
+        target_vector = None
+
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(32619)
+        projection_wkt = srs.ExportToWkt()
```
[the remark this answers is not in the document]
megannissel: I think I was initially trying to, as closely as possible, mimic the way our vector starts out with coordinates in lat/long, but now that I'm looking at this again that seems silly. Since I'm not even constructing the vector from the coordinates in the same way that the model does, and the point of this test is to check the pixel-picking, there's really no reason to do it this way; I could just define points in projected space to begin with.

tests/test_wind_energy.py
```diff
@@ -400,6 +400,107 @@ def test_calculate_npv_levelized_rasters(self):
         numpy.testing.assert_allclose(
             actual_levelized_array, desired_levelized_array)
 
+    def test_raster_values_to_point_vector(self):
```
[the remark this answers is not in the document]
megannissel: Definitely worth testing!

HISTORY.rst
```diff
@@ -84,6 +84,16 @@ Visitation: Recreation and Tourism
   option is not selected.
   (`#1893 <https://github.com/natcap/invest/issues/1893>`_).
 
+Wind Energy
+===========
+* The model no longer returns results as rasters; instead, values are
```
dcdenu4, line 89: Let's get that ADR PR in and then link to the ADR here.

tests/test_wind_energy.py
```diff
@@ -400,6 +400,76 @@ def test_calculate_npv_levelized_rasters(self):
...
+            {'id': 3}, {'id': 4}, {'id': 5}]
+        valid_ids = [1, 5]
+
+        geometries = [
+            Point(pos_x + 250, pos_y - 150),    # out of extent
+            Point(pos_x + 50, pos_y),           # valid
+            Point(pos_x + 150, pos_y),          # invalid: nodata
+            Point(pos_x + 250, pos_y),          # out of extent
+            Point(pos_x + 50, pos_y - 150),     # invalid: nodata
+            Point(pos_x + 150, pos_y - 150)]    # valid
```
dcdenu4, line 427: Excellent!

dcdenu4 (commented):
A couple more quick thoughts!

dcdenu4 (changes requested):
Thanks Megan! One trivial suggestion.

dcdenu4 (approved):

dcdenu4:
Hooray tests and binary builds pass!

PR #1899, opened 2025-04-28
Generate/copy metadata when creating a datastack

### LINKED ISSUE #1774: Metadata: "Save datastack" should include metadata documents in the archive
Saving a datastack refers to saving the input data for an model. There will not necessarily be metadata documents present, but if there are, we might consider including them in the datastack archive. We want to promote sharing metadata along with datasets.

We might also want to generate a metadata document for the archive itself.

claire-simpson:
## Description
- Copies any existing metadata for model inputs into the .tgz archive
- Creates metadata for model inputs (without overwriting any existing user-added attributes) using descriptive info from the `MODEL_SPEC`
- To do: generate a metadata document for the archive itself.

Fixes #1774 

## Checklist


src/natcap/invest/spec_utils.py
```diff
@@ -644,23 +670,101 @@ def write_metadata_file(datasource_path, spec, lineage_statement, keywords_list)
...
 
     resource.write()
 
 
+def _walk_spec_and_write_metadata(
+        invest_spec, workspace, keywords, lineage, file_suffix=None,
+        args_dict=None, use_relative_path=False):
+    """
+    Recursively walk an InVEST MODEL_SPEC and write metadata files for
+    matching input or output files found in the given workspace.
```
claire-simpson, line 708: Because I'm generating metadata by iterating through the model spec/args (rather than simply calling `describe_dir` on the temp workspace before datastack compression), when an argument is a directory containing a set of files (e.g., the precip and et0 folders in SWY, which contain monthly files), metadata isn't generated for those files. I think this could be a use case for the "generate metadata for a collection of files without compression" that we've discussed. What do you think? Should each file within a directory input to a model have its own metadata?

claire-simpson: @davemfish Another instance where this model_spec iteration approach is suboptimal is when the input arg is a csv that references other spatial files, which are included in the datastack but don't have metadata generated for them (e.g., HRA, rec model) - now rethinking iterating through the temp directory and would love to chat about this

src/natcap/invest/datastack.py
```diff
@@ -456,6 +487,9 @@ def build_datastack_archive(args, model_name, datastack_path):
     build_parameter_set(
         rewritten_args, model_name, param_file_uri, relative=True)
 
+    spec_utils.generate_metadata_for_datastack(module, rewritten_args,
+                                               data_dir)
+
```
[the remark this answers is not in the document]
claire-simpson: As discussed, this will be a future PR after the next geometamaker release

src/natcap/invest/spec_utils.py
davemfish: We talked about this and decided it's okay for those "sub-arguments" that are not directly included in the `MODEL_SPEC` to be omitted from metadata creation since we don't really have any useful metadata to add for them anyway. And there are some other benefits to creating the metadata in place, before transferring everything to the temp folder.

src/natcap/invest/datastack.py
```diff
@@ -417,7 +448,7 @@ def build_datastack_archive(args, model_name, datastack_path):
                 if os.path.isdir(src_path):
                     shutil.copytree(src_path, dest_path)
                 else:
-                    shutil.copyfile(src_path, dest_path)
+                    _copy_flat_file(src_path, dest_path)
```
davemfish, line 451: Now that our approach at the end of this function is to create metadata by reading from the dataset (and any pre-existing metadata) in its original location, and writing the metadata doc to the target temp folder, do we still have a need to copy `yml` documents to the temp folder here? Or will they always just get overwritten in the end anyway?

src/natcap/invest/spec_utils.py
```diff
@@ -695,9 +735,37 @@ def _walk_spec(output_spec, workspace):
...
+                            full_path, spec_data, keywords, lineage_statement)
                     except ValueError as error:
                         # Some unsupported file formats, e.g. html
                         LOGGER.debug(error)
 
     _walk_spec(model_module.MODEL_SPEC['outputs'], args_dict['workspace_dir'])
+
+
+def generate_metadata_for_datastack(model_module, args_dict, param_set,
+                                    temp_dir):
```
davemfish, line 747: There's definitely more complexity to this task than I originally realized! I know we've gone back and forth on whether to create the metadata first, in-place, and then copy it during datastack creation, or to create the metadata after all the files get transferred to the temp dir for compression. 

I think we settled on creating the metadata first, then moving it. But I didn't foresee the problem of, "What if metadata docs already exist...we don't want to move those. And what if they don't already exist, we don't want to create them without the user's consent".

So I really like your idea to generate the metadata with the data in it's original location, but then use the `out_workspace` argument to write out the document to the new location. That leaves the user's filesystem in the same state as it was before and allows us to re-use any existing metadata they might have.

I'm getting a little hung up on the parameters of this function though:
```
def generate_metadata_for_datastack(model_module, args_dict, param_set, temp_dir):
```
It's a little complicated because there's essentially two different `args` `dict` parameters. And the `temp_dir` parameter is used in conjunction with directory names from `param_set`. It's challenging to understand the purposes of those parameters without the context of the rest of the `datastack.build_datastack_archive` function. So I wonder if makes sense to just move the body of this function to be in-line code at the end of `datastack.build_datastack_archive`. 

Then for testing, we could add some assertions to existing tests in `test_datastack.DatastackArchiveTests` to assert there is metadata. Or make a new test there, if that seems more appropriate.

src/natcap/invest/datastack.py
claire-simpson: I believe this function is needed in the instance where the user has existing metadata that is invalid (and therefore `geometamaker.describe` gets skipped). This probably begs the question, "if a user has invalid metadata, do we want to copy it into the datastack? Or do we want to create new valid metadata?" I elected to copy a user's invalid metadata into the datastack, the reason being if the metadata is invalid, it means the user has edited it in some way (or created metadata through a different method?), so it would potentially contain some useful info. I'm curious if you have a different opinion on this. 

Perhaps this could be implemented better to avoid needlessly copying metadata for most cases, and instead I could alter the `try/except` block in `write_metadata_file` to copy the existing (invalid) metadata into the temporary directory, something like:
```
try:
        resource = geometamaker.describe(datasource_path)
except ValidationError:
        LOGGER.debug(
            f"Skipping metadata creation for {datasource_path}, as invalid "
            "metadata exists.")
        shutil.copyfile(datasource_path, out_workspace)
        return None  
```

src/natcap/invest/spec_utils.py
```diff
@@ -613,26 +616,56 @@ def describe_arg_from_name(module_name, *arg_keys):
...
+        names = {field.name.lower(): field.name
+                 for field in resource.data_model.fields}
+        return names[key]
+
+    try:
+        resource = geometamaker.describe(datasource_path)
+    except ValidationError:
+        LOGGER.debug(
+            f"Skipping metadata creation for {datasource_path}, as invalid "
+            "metadata exists.")
```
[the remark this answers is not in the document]
claire-simpson: I really like this idea and think it makes sense to be able to `describe` a dataset even if it already has metadata the geometamaker can't parse, given that `describe` doesn't change anything on disc

src/natcap/invest/spec_utils.py
```diff
@@ -644,24 +677,31 @@ def write_metadata_file(datasource_path, spec, lineage_statement, keywords_list)
...
                 # from model results because they are conditional.
                 LOGGER.debug(error)
     if 'bands' in spec:
         for idx, value in spec['bands'].items():
-            try:
-                units = format_unit(spec['bands'][idx]['units'])
-            except KeyError:
-                units = ''
-            resource.set_band_description(idx, units=units)
+            if len(resource.get_band_description(idx).description) < 1:
```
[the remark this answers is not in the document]
claire-simpson: I think my intention was to only overwrite the description if it was empty, but to always set the units based on the model spec, so the code should be like: `resource.set_band_description(idx, description = value['about'], units=units)`

But this actually brings up a good question: if there is pre-existing unit metadata for a band or vector field, should that be overwritten by the unit info in the model_spec? As I have it set up now, it will be overwritten, but now I'm not sure if thats the right call..

src/natcap/invest/spec_utils.py
claire-simpson: I definitely agree that the parameters are overly complicated.. 

> So I wonder if makes sense to just move the body of this function to be in-line code at the end of datastack.build_datastack_archive.

My thinking in keeping this as a standalone function was so it could be reused if/when we add functionality to write in-place metadata for a user's input data. That being said, this function is obviously _not_ currently adapted for that task given that `param_set` and `temp_dir` are required args. I have a slight preference for keeping this as a standalone function but making these args optional (or rethinking the args/design if you have a better suggestion), but ultimately would also be fine to move this function into `build_datastack_archive` if you think thats best.

src/natcap/invest/datastack.py
davemfish: >I believe this function is needed in the instance where the user has existing metadata that is invalid (and therefore geometamaker.describe gets skipped).

Ah I see, now, thanks.

> "if a user has invalid metadata, do we want to copy it into the datastack? Or do we want to create new valid metadata?"

I think we want to create valid metadata. I wouldn't want invest to be responsible for passing around invalid metadata. However, I think it's the responsibility of `geometamaker` to get this right. Currently it can't create a valid document while the invalid document exists. I think we should fix that in `geometamaker` if we want to, rather than work around that limitation here.

> I elected to copy a user's invalid metadata into the datastack, the reason being if the metadata is invalid, it means the user has edited it in some way (or created metadata through a different method?), so it would potentially contain some useful info. I'm curious if you have a different opinion on this.

I agree with this in theory, I just don't think it's worth the extra complexity of checking for and copying the files over. Without that, `build_datastack_archive` is nearly unchanged. It continues to be all about building the datastack archive, and then at the end, we also create metadata. That's a nice separation of concerns, and an indication of good design, in my opinion.

src/natcap/invest/spec_utils.py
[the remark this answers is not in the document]
davemfish: I agree, I'll make a `geometamaker` issue for this.

src/natcap/invest/spec_utils.py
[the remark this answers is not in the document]
davemfish: I think it's okay to treat `units` the same way you're treating the `description`, where any pre-existing data takes precedence over the value in the spec. For the `args` spec, the spec says what units the dataset _should_ be using, but that's not a guarantee that the user is abiding by that.

So in this case, for a `band` I think it would be,
```
if len(resource.get_band_description(idx).units) < 1:
...
resource.set_band_description(idx, units=units)
```
Because `units` are the only attribute being set on the band. The description was set on the `raster` instead of the `band`, which is fine. That matches how the model specs are organized.

src/natcap/invest/datastack.py
claire-simpson: Ok I'm on board with all of this. I will remove the `.yml` copying, which will mean the current functionality is: only valid geometamaker metadata will be included in the datastack, so at this point, some input data may _not_ have metadata . Once `describe` is changed (as described below) however, all data in the `args` will have metadata.

src/natcap/invest/spec_utils.py
davemfish: >My thinking in keeping this as a standalone function was so it could be reused if/when we add functionality to write in-place metadata for a user's input data.

This was my original hope also, but I think I was ignoring some of the complexities around the pre-existing metadata. 

I'm concerned that trying to re-use this function for that other use-case will result in a function that has a lot of `if/else` logic going on that requires the reader of `spec_utils.generate_metadata_for_datastack` to know about how `datastack.build_datastack_archive` works. I think it's reasonable to start out without a function, and then when we add that other use-case, if we find ourselves really wanting to re-use some existing code, then that's a good time to create an abstraction.

src/natcap/invest/datastack.py
```diff
@@ -33,7 +33,9 @@
 import warnings
 
 from osgeo import gdal
+import geometamaker
```
davemfish, line 36: It looks like we might not be using `geometamaker` in this module anymore

src/natcap/invest/spec_utils.py
```diff
@@ -613,26 +616,56 @@ def describe_arg_from_name(module_name, *arg_keys):
...
+    """Write a metadata sidecar file for an invest dataset.
+
+    Create metadata for invest model inputs or outputs, taking care to
+    preserve existing human-modified attributes.
+
+    Note: We do not want to overwrite any existing metadata so if there is
+    invalid metadata for the datasource (i.e., doesn't pass geometamaker
+    validation in ``describe``), this function will NOT create new metadata.
+    Instead, if ``out_workspace`` specified, the existing metadata will be
+    copied here.
```
davemfish, line 630: I think the last sentence of this docstring may no longer apply.

src/natcap/invest/spec_utils.py
```diff
@@ -613,26 +616,56 @@ def describe_arg_from_name(module_name, *arg_keys):
...
+        spec (dict) - the invest specification for ``datasource_path``
         keywords_list (list) - sequence of strings
-
+        lineage_statement (str, optional) - string to describe origin of
+            the dataset
+        out_workspace (str, optional) - where to write metadata if different
+            from data location
     Returns:
-        None
+        None: if metadata could not be created due to validation or file errors.
```
davemfish, line 641: I think technically it returns `None` in all cases. A function with no `return` statement will return `None`.

src/natcap/invest/spec_utils.py
```diff
@@ -2,6 +2,7 @@
 import json
 import logging
 import os
+import pathlib
```
davemfish, line 5: Is this an unused import?

davemfish (changes requested):
Thank you @claire-simpson . This looks great. Just a few minor comments about unused imports and docstrings, etc.

src/natcap/invest/spec_utils.py
claire-simpson: Good catch, and will eventually need to update this whole block once the geometamaker issue proposed above is dealt with

davemfish (approved):

PR #1903, opened 2025-04-30
Hide model status indicator when a parameter is changed

### LINKED ISSUE #1655: Workbench: changing parameters from a previous run does not change "model complete" state
If I open a previously run model, then change one or more parameters, the workbench still seems to indicate that the model has been completed. For instance, if I change the workspace directory, the log tab still displays the original log, and the "Model Complete" indicator is still there. I think the "model complete" state should be removed if I change any parameters from a previously run model.

emilyanndavis:
## Description
Fixes #1655 by clearing job status on arg change. This hides the model status indicator (if one is present), to limit confusion when updating params after a model run.

Also updates an existing focus handler from `handleChange` to `handleFocus`.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -76,6 +76,10 @@ Workbench
 * Fixed a bug that did not allow users to select a folder as the location
   to extract a datastack archive.
   (`#1879 <https://github.com/natcap/invest/issues/1879>`_).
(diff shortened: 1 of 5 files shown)


emlys (approved):
Thanks @emilyanndavis, I tried this out locally and it looks great!

PR #1906, opened 2025-05-02
make the setup.py GDAL library path stuff a little more bulletproof on Windows

davemfish:
Getting the hang of the new gdal build requirement on Windows. I'm doing this, which was failing to create the proper path due to trailing whitespace on the env var.

`set NATCAP_INVEST_GDAL_LIB_PATH=%CONDA_PREFIX%\Library && pip install .`


dcdenu4 (approved):
Thanks @davemfish!

PR #1910, opened 2025-05-06
HISTORY file needs an expected layout

davemfish:
## Description
Our autorelease workflow didn't work because it didn't find the expected header in HISTORY.rst


emilyanndavis (approved):
Thanks!

PR #1918, opened 2025-05-12
REC: always write out all userday metrics

### LINKED ISSUE #1893: Rec: calculate all userday metrics even if not computing a regression model.
The Rec model produces several different variables that can be used as proxies for visitation rates. These are spread across outputs: `PUD_results.gpkg`, `TUD_results.gpkg`, and `regression_data.gpkg`.

In particular, `regression_data.gpkg` has `pr_PUD`, `pr_TUD`, and `avg_pr_UD`. This last variable is used as the response variable in the regression model.

`regression_data.gpkg` is only created if users opt for the "Compute Regression" option. But it's likely possible that these variables are useful to have even without computing the regression model. We should consider always calculating and writing out these 3 variables.

davemfish:
Always write out final userday variables to `regression_data.gpkg` even if not proceeding to the `compute_regression` portion of the model.

Fixes #1893 

UG update: https://github.com/natcap/invest.users-guide/pull/179

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -23,9 +23,9 @@
   - Urban Flood Risk
   - Urban Nature Access
   - Urban Stormwater Retention
+  - Visitation: Recreation and Tourism
   - Wave Energy
   - Wind Energy
-  - Visitation: Recreation and Tourism
 
   Workbench fixes/enhancements:
   - Workbench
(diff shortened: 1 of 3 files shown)


davemfish:
The test failures here are unrelated. They are the HRA failure that is awaiting #1881

claire-simpson (approved):
This all looks great! Are we happy to merge even with the failing tests, since we know that will be fixed in PR 1882?

PR #1923, opened 2025-05-14
Update CBC docs to describe implications of accum/decay --> no carbon change

### LINKED ISSUE #671: Should CBC guard against NCC transitions in blue carbon habitat?
In a recent [forums post](https://community.naturalcapitalproject.org/t/coastal-blue-carbon-understanding-outputs/2182/3?u=jdouglass), a user had some curiously malformed input data where:

1. Between 2 transition years, the blue carbon habitat `mangue` would "transition" to `mangue`
2. The carbon transition type identified in the transitions table was "NCC", no carbon change.

So, as a result, the output rasters showed carbon accumulation from the baseline year up until the first transition and then after that _nothing_.  The question is: should this be allowed?  Or maybe we should log a warning against this because it's probably malformed data?  Or maybe there's a case where someone would want to have the transition be "NCC"?

phargogh:
This PR updates the CBC documentation to clarify what happens when a transition shifts from a state of accumulation or decay to one of no carbon change.

I couldn't bring myself to make assumptions about what the user intends, so I'd rather clarify the model's logic through docs than assume intent.

Fixes #671 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -70,6 +70,12 @@ Workbench
   existing user-added metadata preserved)
   (`#1774 <https://github.com/natcap/invest/issues/1774>`_).
 
+Coastal Blue Carbon
+===================
+* Updated the Coastal Blue Carbon documentation to clarify what happens when a
+  class transitions from a state of accumulation or decay to a No Carbon Change
+  ("NCC") state. (`#671 <https://github.com/natcap/invest/issues/671>`_).
+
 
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Thanks @phargogh, I like the documentation clarification approach myself.

dcdenu4:
Since no code was changed I think it's safe to merge this in.

PR #1925, opened 2025-05-14
Add load type parameter to NDR

### LINKED ISSUE #1044: NDR: which is correct for the `load_n` parameter - fertilizer application or modified export load?
A user brought up a question on [the forum](https://community.naturalcapitalproject.org/t/interpreting-the-n-load-parameter/2913) about interpreting the `n_load` parameter. A discussion in a private [Slack thread](https://stanford-natcap.slack.com/archives/CECFUU1DW/p1658952941920249) took place between @newtpatrol , @adrianvogl , and @schmittrjp. I'll attempt to capture the questions.

- Should we be providing an export coefficient, such that fertilizer application would need to be adjusted by the retention coefficient to get the load_n parameter?
- Some assume that the n_load parameter would be the amount of N directly applied to a pixel (sum of fertilizer, atmospheric deposition, etc.). 

Action items:
- [ ] Look over the NDR source code and provide a report of how the `n_load` parameter is used and what's happening on the load-originating pixel

dcdenu4:
## Description
Add columns to the biophysical table, `load_type_p` and `load_type_n`, that flags whether the nutrient loads should be treated as an application rate or measured runoff. The columns values specifically expected are: `measured-runoff | application-rate`. Currently the model acts as if the load is measure runoff.

General notes:
- These are required columns, no default value. Error if columns are not present.
- Allow for mix behaviors within the model run.
- If application rate given, then model will do the adjustment: `load_n * (1 - retention_efficiency)`

Test data updated: https://bitbucket.org/natcap/invest-test-data/pull-requests/35
Sample data updated: https://bitbucket.org/natcap/invest-sample-data/pull-requests/41
~UG update: https://github.com/natcap/invest.users-guide/pull/181 New PR coming~
UG update: https://github.com/natcap/invest.users-guide/pull/186
Fixes #1044 

## Checklist


tests/test_ndr.py
```diff
@@ -124,18 +124,12 @@ def test_crit_len_0(self):
         # use predefined directory so test can clean up files during teardown
         args = NDRTests.generate_base_args(self.workspace_dir)
         new_table_path = os.path.join(self.workspace_dir, 'table_c_len_0.csv')
-        with open(new_table_path, 'w') as target_file:
```
dcdenu4, line 127: Rewrote as a pandas operation to more reliably update a specific column, by name, instead of trying to update by index be mutating each line.

src/natcap/invest/ndr/ndr.py
```diff
@@ -1362,8 +1408,7 @@ def _map_surface_load(
     """Calculate surface load from landcover raster.
 
     Args:
-        modified_load_path (string): path to modified load raster with units
-            of kg/pixel.
```
dcdenu4, line 1366: @emilyanndavis I think this unit reference is now outdated since your per hectare PR update.

src/natcap/invest/ndr/ndr.py
```diff
@@ -362,14 +379,14 @@
                     "about": "Above ground nitrogen loads",
                     "bands": {1: {
                         "type": "number",
-                        "units": u.kilogram/u.year
```
dcdenu4, line 365: I _think_ these are now hectare after the per hectare PR

Makefile
```diff
@@ -2,11 +2,11 @@
...
-GIT_SAMPLE_DATA_REPO_REV    := ecdab62bd6e2d3d9105e511cfd6884bf07f3d27b
+GIT_SAMPLE_DATA_REPO_REV    := b84517619d94d4e277cd1271c28e5f1f6e7f1226
 
 GIT_TEST_DATA_REPO          := https://bitbucket.org/natcap/invest-test-data.git
 GIT_TEST_DATA_REPO_PATH     := $(DATA_DIR)/invest-test-data
-GIT_TEST_DATA_REPO_REV      := f0ebe739207ae57ae53a285d0fd954d6e8cfee54
+GIT_TEST_DATA_REPO_REV      := 371d8d0f3f20a965716978cc49cf942e87652b3e
 
 GIT_UG_REPO                 := https://github.com/natcap/invest.users-guide
 GIT_UG_REPO_PATH            := doc/users-guide
```
dcdenu4, line 12: Still need to update the UG revision, but that PR is still pending from my fork.

src/natcap/invest/ndr/ndr.py
```diff
@@ -77,6 +77,23 @@
...
             "columns": {
                 "lucode": spec_utils.LULC_TABLE_COLUMN,
+                "nut_load_type": {
+                    "type": "option_string",
+                    "required": True,
+                    "options": {
+                        "application-rate": {
+                            "description": gettext(
+                                "Treat the load values as an application rate."
+                                "applied_nutrient * (1-retention_efficiency)")},
```
emilyanndavis, line 87: Can we work this into a sentence that provides some context around this expression? Maybe something like `The model will adjust the load based on retention efficiency: load_[NUTRIENT] * (1 - eff_[NUTRIENT]).`? (I think I've understood the behavior correctly?)

tests/test_ndr.py
```diff
@@ -537,3 +531,56 @@ def test_synthetic_runoff_proxy_av(self):
...
+        """Test ``_calculate_load`` raises ValueError on bad nut_load_types."""
+        from natcap.invest.ndr import ndr
+
+        lulc_path = os.path.join(self.workspace_dir, "lulc-load-type.tif")
+        target_load_path = os.path.join(self.workspace_dir, "load_raster.tif")
+
+        # Calculate load
+        lucode_to_params = {
+            1: {'load_n': 10.0, 'eff_n': 0.5, 'nut_load_type': 'measured-runoff'},
+            2: {'load_n': 20.0, 'eff_n': 0.5, 'nut_load_type': 'cheese'},
```
emilyanndavis, line 579: Definitely one of my favorite nutrient load types! 🤣

emilyanndavis (changes requested):
Thanks, @dcdenu4! I left a few questions/suggestions, and I also have a more general question:

Do you see any potential need for _two_ load type columns, i.e., one for N and another for P? Or is it safe to assume that, if a user is modeling N and P simultaneously, the load type will be the same for each nutrient?

src/natcap/invest/ndr/ndr.py
```diff
@@ -1325,23 +1344,50 @@ def _normalize_raster(base_raster_path_band, target_normalized_raster_path,
...
+        else:
+            lucode_to_values[key] = value_list
+
     def _map_load_op(lucode_array):
         """Convert unit load to total load & handle nodata."""
         result = numpy.empty(lucode_array.shape)
         for lucode in numpy.unique(lucode_array):
             try:
-                result[lucode_array == lucode] = (lucode_to_load[lucode])
+                if lucode_to_values[lucode][type_index] == load_types[0]:
```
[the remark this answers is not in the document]
dcdenu4: Yep, that makes sense.

src/natcap/invest/ndr/ndr.py
```diff
@@ -1325,23 +1344,50 @@ def _normalize_raster(base_raster_path_band, target_normalized_raster_path,
...
+    # restructure the lookup dict to easier access load, eff, and type
+    # {lucode: [load, eff, type], ... }
+    load_types = ['measured-runoff', 'application-rate']
+    load_index = 0
+    eff_index = 1
+    type_index = 2
+    lucode_to_values = {}
+
+    for key, value in lucode_to_load.items():
+        value_list = list(value.values())
```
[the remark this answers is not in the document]
dcdenu4: Took a stab at cleaning this up a bit: https://github.com/natcap/invest/pull/1925/commits/9906d9060d9a596a1a4cd6a42e780b466d09a1dc

Instead of passing in 3 dictionaries or data structures I keep the one, but pass in the nutrient key so we know what the columns / dict keys are.

src/natcap/invest/ndr/ndr.py
dcdenu4: Reworked here: https://github.com/natcap/invest/pull/1925/commits/4ed0ac882c8b6ed2ab32647dc653f64074afa6c0

dcdenu4:
>Do you see any potential need for two load type columns, i.e., one for N and another for P? Or is it safe to assume that, if a user is modeling N and P simultaneously, the load type will be the same for each nutrient?

Thanks for thinking about this. @newtpatrol thinks it would be great to have separate columns for this.
>I think it's typical for users to collect their loading data from a variety of sources, including separate sources for both N and P. So I would say yes, we should provide separate load types for both N and P.

src/natcap/invest/ndr/ndr.py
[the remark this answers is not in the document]
emilyanndavis: Ah, I see. Much easier to follow now, thanks! Hopefully it won't be too tricky to extend this to handle nutrient-specific load type columns.

src/natcap/invest/ndr/ndr.py
```diff
@@ -77,6 +101,8 @@
             "index_col": "lucode",
             "columns": {
                 "lucode": spec_utils.LULC_TABLE_COLUMN,
+                "load_type_p": LOAD_TYPE_OPTIONS,
+                "load_type_n": LOAD_TYPE_OPTIONS,
```
emilyanndavis, line 105: I don't think this matters a lot, but since we seem to be using the `prefix_[NUTRIENT]` pattern for other columns, it might be simpler to replace these two entries with a single `load_type_[NUTRIENT]` entry, and then you wouldn't need to define `LOAD_TYPE_OPTIONS` separately.

emilyanndavis (changes requested):
Thanks for the updates! Just one more suggestion, and I'll also keep an eye out for the UG PR.

src/natcap/invest/ndr/ndr.py
dcdenu4: The `[NUTRIENT]` syntax was removed for NDR in some of Emily S. work on the OO Model Spec. We thought that with only 2 repeated items, n and p, it read cleaner for the user to explicitly name them. I think the intent of the `[VARIABLE]` syntax is to really cover an unknown or many type situation. So, I might leave it this way for now.

dcdenu4: Okay, given the model spec refactor coming in and that we did away with the `[NUTRIENT]` notation in that, I think for now it's actually best to just spell all this out for p and n in the model spec directly. It carries over some redundancy in text for sure, which nags at me, but I think it's the cleanest for now.

https://github.com/natcap/invest/pull/1925/commits/243eff99285825a6a0261af20a610f7151f522bd

emilyanndavis (approved):
Looks good, thanks! The UG PR has been merged; once you've updated the revision hash here and verified the UG build succeeds, feel free to merge this.

PR #1931, opened 2025-05-19
[For review only—do not merge] Diff between SDR and SDR USLE C Raster Plugin

emilyanndavis:
## Description
Replaces contents of sdr.py and test_sdr.py with contents from [invest-sdr-usle-c-raster](https://github.com/natcap/invest-sdr-usle-c-raster), for the sole purpose of illustrating differences between the plugin and the core SDR model.

For review only; not to be merged. Although the tests pass in the plugin repo, we should expect test failures here.

## Checklist
- n/a

### DIFF
--- a/src/natcap/invest/sdr/sdr.py
+++ b/src/natcap/invest/sdr/sdr.py
@@ -33,7 +33,7 @@ LOGGER = logging.getLogger(__name__)
 MODEL_SPEC = {
     "model_id": "invest_sdr_usle_c_raster",
     "model_name": "SDR with USLE C raster",
-    "userguide": "",
+    "userguide": "https://github.com/natcap/invest-sdr-usle-c-raster/blob/main/README.md",
     "aliases": set(),
     "args_with_spatial_overlap": {
         "spatial_keys": ["dem_path", "erosivity_path", "erodibility_path",
--- a/src/natcap/invest/sdr/sdr.py
+++ b/src/natcap/invest/sdr/sdr.py
@@ -1,6 +1,9 @@
-"""InVEST Sediment Delivery Ratio (SDR) module.
+"""Sediment Delivery Ratio (SDR) with USLE C raster.
 
-The SDR method in this model is based on:
+This is an InVEST plugin based on the core InVEST SDR model. This alternative
+version expects USLE C values in a raster instead of in the biophysical table.
+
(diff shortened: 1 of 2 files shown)


src/natcap/invest/sdr/sdr.py
```diff
@@ -557,15 +577,31 @@ def execute(args):
...
-        for lulc_code, row in biophysical_df.iterrows():
-            if row[key] < 0 or row[key] > 1:
-                raise ValueError(
-                    f'A value in the biophysical table is not a number '
-                    f'within range 0..1. The offending value is in '
-                    f'column "{key}", lucode row "{lulc_code}", '
-                    f'and has value "{row[key]}"')
+    # Test to see if C values are outside of 0..1
+    def update_out_of_range(any_out_of_range, block):
+        if numpy.any(block[numpy.any([block < 0, block > 1])]):
```
emlys, line 582: I think this does work, but the logic could be simplified:
```suggestion
        if numpy.any((block < 0) | (block > 1)):
```

emlys: Or even further:
```suggestion
        return any_out_of_range or numpy.any((block < 0) | (block > 1))
```
which has the neat advantage of using short-circuit evaluation: if `any_out_of_range` is already `True`, everything after the `or` doesn't need to be evaluated at all!

src/natcap/invest/sdr/sdr.py
```diff
@@ -557,15 +577,31 @@ def execute(args):
...
+            any_out_of_range = True
+        return any_out_of_range
+
+    any_out_of_range = pygeoprocessing.raster_reduce(
+        function=update_out_of_range,
+        raster_path_band=(args['usle_c_path'], 1),
+        initializer=False)
+
+    if any_out_of_range:
```
emlys, line 591: It's worth noting that historically we haven't done validation on rasters like this. Checking that every pixel value meets some condition has the potential to add significant compute time on large datasets. I think I'm in support of it, but I'd be curious to get more opinions.

emlys: It definitely doesn't belong in `validate`, because that needs to return near-instant results to the workbench. Maybe we should have a "`slow_validate`" function for validation steps that may take a long time, and the user could opt out of.

PR #1935, opened 2025-05-20
Migrate recent jobs data to new format

### LINKED ISSUE #1922: Plugins branch does not populate the list of recent model runs
The Workbench should display the list of recent model runs. This data should persist across versions of the Workbench. The plugins branch does not populate the list of recent runs that I already have completed from other other versions of the Workbench.

If we changed the way this data is stored (I don't remember if we have or not), perhaps we should consider migrating the database over to the new system.

emlys:
## Description
Fixes #1922 

Updates the job property names in any old job data, mapping `modelRunName` to `modelID` and `modelHumanName` to `modelTitle`, and re-saves them in the new format.

## Checklist

### DIFF
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -72,6 +72,8 @@ MESSAGES = {
     'WRONG_GEOM_TYPE': gettext('Geometry type must be one of {allowed}')
 }
 
+def get_message(key):
+    return gettext(MESSAGES[key])
 
 def _evaluate_expression(expression, variable_map):
     """Evaluate a python expression.
--- a/workbench/src/renderer/InvestJob.js
+++ b/workbench/src/renderer/InvestJob.js
@@ -32,6 +32,19 @@ export default class InvestJob {
         (key) => investJobStore.getItem(key)
       ));
     }
+    // Migrate old-style jobs
+    // We can eventually remove this code once it's likely that most users
+    // will have updated and ran a newer version of the workbench
+    jobArray = await Promise.all(jobArray.map(async (job) => {
+      if (job.modelID === undefined) {
+        job.modelID = job.modelRunName;
+        job.modelTitle = job.modelHumanName;
+        delete job.modelRunName;
+        delete job.modelHumanName;
+        await investJobStore.setItem(job.hash, job);
+      }
+      return job;
+    }));
     return jobArray;
   }
 
--- a/workbench/src/renderer/app.jsx
+++ b/workbench/src/renderer/app.jsx
@@ -63,20 +63,6 @@ export default class App extends React.Component {
   async componentDidMount() {
     const investList = await this.updateInvestList();
     let recentJobs = await InvestJob.getJobStore();
-    // Migrate old-style jobs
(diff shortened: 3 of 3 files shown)


src/natcap/invest/validation.py
```diff
@@ -72,6 +72,8 @@
     'WRONG_GEOM_TYPE': gettext('Geometry type must be one of {allowed}')
 }
 
+def get_message(key):
```
emlys, line 75: This function was missing, it must have been lost in a merge.

davemfish (changes requested):
@emlys I had an idea for doing the migration in `InvestJob.getJobStore` instead of in `componentDidMount`. Let me know if you think that makes sense too.

workbench/src/renderer/app.jsx
```diff
@@ -62,7 +62,21 @@ export default class App extends React.Component {
...
+    recentJobs = recentJobs.map((job) => {
+      if (job.modelID === undefined) {
+        job.modelID = job.modelRunName;
+        job.modelTitle = job.modelHumanName;
+        delete job.modelRunName;
+        delete job.modelHumanName;
+        InvestJob.saveJob(job);
+      }
+      return job;
+    });
```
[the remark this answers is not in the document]
emlys: Oh good catch. Yes, the `InvestJob` API isn't ideal for what I'm trying to do here. I'll try using `investJobStore.setItem`.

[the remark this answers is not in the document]
emlys: That worked! I tested and confirmed it displays both old and new job data.

workbench/src/renderer/InvestJob.js
```diff
@@ -32,6 +32,19 @@ export default class InvestJob {
         (key) => investJobStore.getItem(key)
       ));
     }
+    // Migrate old-style jobs
+    // We can eventually remove this code once it's likely that most users
+    // will have updated and ran a newer version of the workbench
+    jobArray = await Promise.all(jobArray.map(async (job) => {
```
davemfish, line 38: It seems like we could condense this mapping into the `Promise.all(sortedJobHashes.map(` from above and then only map over the jobs one time?

workbench/src/renderer/app.jsx
```diff
@@ -62,7 +62,7 @@ export default class App extends React.Component {
   /** Initialize the list of invest models, recent invest jobs, etc. */
   async componentDidMount() {
     const investList = await this.updateInvestList();
-    const recentJobs = await InvestJob.getJobStore();
+    let recentJobs = await InvestJob.getJobStore();
```
davemfish, line 65: I think this can probably revert back to `const`, right?

davemfish (commented):
Thanks for making this update @emlys , one more small suggestion.

workbench/src/renderer/InvestJob.js
emlys: We could, I had kept it separate to make it easier to remove in the future, but it would be worth condensing if you think it's a performance concern?

davemfish (approved):

Issue #1928, opened 2025-05-15
Workbench does not support "User's guide entry" links for plugins


emilyanndavis:
## Description
Closely related to #1927; tracking this separately to document additional considerations for the "User's guide entry" links, and in case we want to apply a different solution here.

Currently, the workbench does not support user's guide links for plugins. Although it is possible to specify an external URL in the model spec, the workbench assumes all user's guide links are local, i.e., pointing to a location in the local copy of the user's guide that is bundled with the workbench.

In addition, the "User's guide entry" link that is rendered in the "info" modal for each model input is configured as a deep link that points directly to the relevant list item in the "Data Needs" section of the user's guide. (For example, the link for the Biophysical Table input to the SDR model will take you directly to the Biophysical Table entry in the SDR user's guide.) Such links will navigate gracefully to the main URL if the deep link ID does not exist (e.g., pointing to [https://example.com#banana](https://example.com#banana) will still load example.com even though the page has no element with ID banana). However, it's easy to argue that providing a link called "User's guide entry"—in the context of a specific input—that does not navigate directly to an entry about that input, could be misleading, or at least confusing.

## To replicate
1. Install a plugin.
2. Open the plugin in a workbench tab.
3. Select an "info" button next to any input in the Setup tab, then follow the "User's guide entry" link from the resulting modal.

## Observed behavior
The link opens a new browser window (or tab, depending on your browser settings), but the window/tab is blank, because the URL—having been prepended with the prefix needed for the local user's guide—is invalid.

## Expected behavior
See #1927 for some possible solutions.

emlys:
If we support linking to an arbitrary external URL for plugin documentation, like suggested in #1927, we can't know what if any section of that documentation applies to a specific input. I'd support removing the "User's guide entry" link for plugins, I think it only really applies to core models.

emilyanndavis:
We discussed this in the software team meeting today and agreed that the "User's guide entry" links don't make a lot of sense for plugins. We'll remove them for plugins.

dcdenu4:
I believe this was handled in #1939

PR #1939, opened 2025-05-21
Handle in-workbench UG/docs links for plugins

### LINKED ISSUE #1927: Workbench does not support user's guide links for plugins
## Description
Currently, the workbench does not support user's guide links for plugins. Although it is possible to specify an external URL in the model spec, the workbench assumes all user's guide links are local, i.e., pointing to a location in the local copy of the user's guide that is bundled with the workbench.

## To replicate
1. Install a plugin.
2. Open the plugin in a workbench tab.
3. Follow the "User's Guide" link in the sidebar menu.

## Observed behavior
The link opens a new window, but the window is blank, because the URL—having been prepended with the prefix needed for the local user's guide—is invalid.

## Expected behavior
There are a few possible ways to handle this, each with tradeoffs.
1. The User's Guide link could simply be absent from a tab when the model is a plugin. This would be simple to implement, but it would leave the user on their own when it comes to finding more info about how to use the plugin.
2. The workbench could assume the user's guide for a plugin lives at an external URL, and generate the href for the User's Guide accordingly. For a plugin, this would mean the User's Guide link would open a new window (perhaps even a browser window instead of an Electron window) and navigate to the full URL specified in the plugin's model spec. This would not be very challenging to implement, but there are risks involved in allowing the workbench to navigate to an arbitrary URL. What if the URL has a typo, or is outdated, or otherwise results in a broken link because it is an invalid URL? Or worse: what if the URL points to a malicious site? How can we protect our users from unwanted consequences?
3. The workbench could make some initial assumption about where the plugin's user's guide is located—for example, at the public GitHub site for the plugin repo, determined from the git URL that was used to install it—then configure the User's Guide link to point to that URL. This would afford a way to protect against malicious links or broken links, but it would require retaining the git URL used for installation (which I doubt we do already), it would exclude plugins installed from a local path (even if it is a local clone of a public GitHub repo), and it could be seen as too opinionated.
4. The workbench could fetch the plugin's README and render it in a new window (first converting from Markdown or RST to HTML). This would prevent the workbench from directly navigating to a broken or malicious link, but a README could itself contain any arbitrary external URLs, and more importantly, we would essentially be injecting arbitrary HTML into the workbench, which is generally Not Advisable.

emilyanndavis:
## Description
Fixes #1927 and #1928.

I didn't update HISTORY, since I see these changes as parts of the forthcoming plugins feature, rather than isolated changes to existing behavior.

## Checklist

### Example of "Plugin Documentation" link in lieu of "User's Guide" link in sidebar
plugin-documentation-link_2025-05-20

### Example of arg info modal without "user's guide entry" link
plugin-input-info-modal_2025-05-20


workbench/tests/invest/flaskapp.test.js
```diff
@@ -214,6 +214,8 @@ describe('Test building model UIs and forum links', () => {
         modelID={modelID}
         argsSpec={argsSpec.args}
         userguide={argsSpec.userguide}
+        isCoreModel={true}
+        uiSpec={argsSpec.ui_spec}
```
davemfish, line 218: I think as of the latest `feature/plugins`, `uiSpec` is no longer needed as a prop and `inputFieldOrder` is used instead.

workbench/tests/renderer/investtab.test.js
```diff
@@ -88,6 +91,7 @@ describe('Run status Alert renders with status from a recent run', () => {
       status: status,
       argsValues: {},
       logfile: 'foo.txt',
+      type: 'core',
```
davemfish, line 94: Thanks for catching this!

workbench/tests/renderer/setuptab.test.js
```diff
@@ -356,13 +357,25 @@ describe('Arguments form interactions', () => {
...
     await userEvent.click(link);
     await waitFor(() => {
       const calledChannels = spy.mock.calls.map(call => call[0]);
       expect(calledChannels).toContain(ipcMainChannels.OPEN_LOCAL_HTML);
     });
   });
+
+  test('Open info dialog, expect text but no link if model is a plugin', async () => {
+    const spy = jest.spyOn(ipcRenderer, 'send')
+      .mockImplementation(() => Promise.resolve());
```
davemfish, line 371: Is this `mockImplementation` needed for this test?

davemfish (commented):
Thanks, @emilyanndavis , this looks great to me, just a couple small comments.

workbench/tests/renderer/setuptab.test.js
emilyanndavis: Nope, it isn't, thanks!

davemfish (approved):
Thanks!

PR #1940, opened 2025-05-22
Bugfix for NDR: OverflowError when filling nodata

### LINKED ISSUE #1904: Nodata casting error.
## Description
There is a potential mismatch between the nodata assigned to the numpy array and the numpy type at line 1187 (from task at line 672).

## To replicate

1. Use a LULC raster of type "Byte - Eight bit unsigned integer" with "n/a" nodata (as reported by QGIS).
2. Run NDR model in Python - I could not replicate this behavior with the Workbench

## Expected behavior
I expect the model to run! More specifically, I expect the LULC raster to be successfully masked by the watersheds.

## Observed behavior
OverflowError: Python integer -2147483648 out of bounds for uint8

## Logfile(s)
No log file output from python run.

## Screenshots
If applicable, add screenshots to help illustrate the problem.

## Environment
- Operating System: macOS Sequoia 15.4.1
- InVEST version number: 3.14.3 (although the code seems to be unchanged in 3.15.0)
- InVEST interface (Workbench, CLI, or Python API): Python API

## Additional context
I think this is a fairly straightforward fix. When passing a `source raster` of Int8 with undefined nodata and defining `default_nodata` as `numpy.iinfo(numpy.int32).min` in the `_mask_raster()` function (line 1160), the function tries to create a Byte array full of -2147483648 values, which breaks numpy.
This happens in particular with `mask_lulc_task`, line ~ 672.

Image

davemfish:
`_mask_raster` was trying to fill a numpy array with a module-defined (as opposed to user-defined) default nodata value, but using the source raster's native dtype for the array.

This only occurred when the source raster didn't have it's own nodata value defined. Depending on the source raster's datatype, the default nodata value might overflow it.

Since `_mask_raster` is always specifying the target dtype for the raster, it didn't make sense to use the source raster's dtype to initialize the numpy array.

Also, since we're always specifying the target dtype, I didn't see a reason to try to re-use the source raster's nodata value. So I changed the parameter from `default_nodata` to `target_nodata`.

Fixes #1904 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -86,6 +86,8 @@ NDR
 ====
 * ``stream.tif`` is now saved in the main output folder rather than the
   intermediate folder (`#1864 <https://github.com/natcap/invest/issues/1864>`_).
+* Fixed a bug where input rasters (e.g. LULC) without a defined nodata value could
+  cause an OverflowError. (`#1904 <https://github.com/natcap/invest/issues/1904>`_).
 
 Seasonal Water Yield
 ====================
--- a/src/natcap/invest/ndr/ndr.py
+++ b/src/natcap/invest/ndr/ndr.py
@@ -9,6 +9,7 @@ import pygeoprocessing
 import pygeoprocessing.routing
 import taskgraph
 from osgeo import gdal
+from osgeo import gdal_array
 from osgeo import ogr
 
 from .. import gettext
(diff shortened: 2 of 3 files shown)


claire-simpson (approved):
Looks good @davemfish!

PR #1945, opened 2025-05-22
Remove accidental duplication (and overwriting) of definition of args passed to cythonize

emilyanndavis:
## Description
Hoping this will resolve some (if not all) of the build failures we're seeing in some open PRs!


emlys (approved):
Good catch! I didn't notice that block of code had gotten duplicated. 

The failing test is the puppeteer plugin test, which I'm not sure why it's running here at all, but it's skipped upstream.

PR #1949, opened 2025-05-27
Mitigate setuptools invalid metadata for twine upload

### LINKED ISSUE #1913: Autorelease Part 2 fails on twine upload
The failure: https://github.com/natcap/invest/actions/runs/14869550476/job/41754573373

The explanation: https://github.com/pypa/twine/issues/1216

There are bugs in `twine` and/or `setuptools` that generate invalid package metadata. A workaround is making sure the `packaging` package is upgraded after installing `twine`.

But a better fix might be to figure out which version of `setuptools` and which format of the license metadata in our `pyproject.toml` will not produce invalid metadata.

dcdenu4:
## Description
In the autorelease part 2 step, twine uploads were failing on a invalid metadata from setuptools. Specifically the `license-files` attribute. Twine seems to have a mitigation for this in 6.1.0, but requires `packaging` 24.2 to work. So, it seems like the most straightforward patch is to update the `packaging` library.

Another possible workaround might be to include
```python
[tools.setuptools]
license-files = [License.txt]
```
In `pyproject.toml`. But, since we have `license-files` listed under `[project]`, I wasn't sure how this would resolve.
Fixes #1913 

## Checklist

### DIFF
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -42,7 +42,11 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install dependencies
-        run: pip install twine
+        run: |
+          # Workaround for setuptools generating invalid metadata
+          # https://github.com/natcap/invest/issues/1913
+          pip install "twine>=6.1.0"
+          pip install -U packaging
 
       - name: Extract version from autorelease branch name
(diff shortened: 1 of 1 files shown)


davemfish (approved):
That looks like it should work, thanks, @dcdenu4 !

PR #1966, opened 2025-05-28
Call the right micromamba executable in plugin install commands

### LINKED ISSUE #1965: Micromamba invoked incorrectly in plugin install
## Description
In the 3.16.0a1 binary build, I discovered that a micromamba command mistakely calls `micromamba` directly instead of using the path to the micromamba executable that is distributed with the workbench. This was not found earlier because it's not a problem in dev mode, and the puppeteer test for plugins is currently skipped.

emlys:
## Description
Fixes #1965 

## Checklist

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -147,15 +147,14 @@ export function setupAddPlugin(i18n) {
            'python', '-m', 'pip', 'install', installString]
         );
         logger.info('installed plugin into its env');
-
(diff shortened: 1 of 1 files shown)


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -178,7 +177,7 @@ export function setupAddPlugin(i18n) {
         );
         logger.info('successfully added plugin');
       } catch (error) {
-        console.log(error);
+        logger.info(error);
```
emlys, line 180: For some reason, the `logger.info` gave more useful information when I was debugging this.

PR #1973, opened 2025-05-28
"Add a plugin" form updates: user acknowledgment checkbox, form validation

### LINKED ISSUE #1941: A user should be required to acknowledge risks when installing a (third-party) plugin
## Description
It's in a user's best interest to know there are risks inherent to installing third-party software, such as an InVEST plugin that is not maintained by the NatCap team. We should make a good-faith effort to ensure users are aware of and accept these risks before installing a plugin.

## To replicate
1. Install a plugin via the Workbench.

## Expected behavior
Before installing any plugin, I should see a message about the nature of installing a third-party plugin and be required to select a checkbox indicating I acknowledge and accept the risks involved.

## Observed behavior
The installation process is so seamless, a user could easily forget they are installing software that is not maintained by the NatCap team. In the vast majority (hopefully 100%!) of cases, such software will not cause harm, but by acknowledging the possibility, we can encourage users to pause and think carefully about what they are installing.

emilyanndavis:
## Description
Resolves #1941 by adding a plugin installation risk statement and checkbox, along with form validation and error messages.

I did not update HISTORY, since this is part of the plugins feature rather than a standalone update.

## Checklist

### Full plugin modal with new risk statement & acknowledgment checkbox
add-plugin-risk-statement_2025-05-27

### Example form validation scenario: git URL empty, checkbox unchecked
add-plugin-url-and-acknowledgment-errors_2025-05-27

### Example form validation scenario: path empty, checkbox checked
add-plugin-path-error_2025-05-27

### DIFF
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -34,13 +34,60 @@ export default function PluginModal(props) {
   const [needsMSVC, setNeedsMSVC] = useState(false);
   const [plugins, setPlugins] = useState({});
   const [installFrom, setInstallFrom] = useState('url');
+  const [userAcknowledgment, setUserAcknowledgment] = useState(false);
+  const [userAcknowledgmentError, setUserAcknowledgmentError] = useState(false);
+  const [pluginSourceMissingError, setPluginSourceMissingError] = useState(false);
 
   const handleModalClose = () => {
     setURL('');
     setRevision('');
     setInstallErr('');
     setUninstallErr('');
+    clearFormErrors();
     closeModal();
+  };
+
+  const clearFormErrors = () => {
+    setUserAcknowledgmentError(false);
+    setPluginSourceMissingError(false);
+  };
+
+  useEffect(() => {
+    clearFormErrors();
+  }, [installFrom]);
+
+  useEffect(() => {
+    if (pluginSourceMissingError) {
+      setPluginSourceMissingError(false);
+    }
+  }, [url, path]);
+
+  useEffect(() => {
+    if (userAcknowledgment) {
+      setUserAcknowledgmentError(false);
+    }
+  }, [userAcknowledgment]);
+
+  const handleAddPluginClick = () => {
+    clearFormErrors();
+    if (validateAddPluginForm()) {
+      addPlugin();
+    }
+  };
+
+  const validateAddPluginForm = () => {
+    let formValid = true;
+    if ((installFrom === 'url' && !url)
+        || (installFrom === 'path' && !path)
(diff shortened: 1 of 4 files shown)


workbench/tests/renderer/plugin.test.js
```diff
@@ -45,7 +47,95 @@ describe('Add plugin modal', () => {
...
+    });
+
+    test('Should render an error on submit if git URL is empty', async () => {
+      const {
+        findByText, findByLabelText, findByRole,
+      } = render(<App />);
+
+      await userEvent.click(await findByRole('button', { name: 'menu' }));
+      const managePluginsButton = await findByText(/Manage plugins/i);
+      userEvent.click(managePluginsButton);
```
davemfish, line 78: I think `userEvent` actions always returns a promise, so best to `await` it in all these cases.

davemfish (changes requested):
Hi @emilyanndavis , this looks great, thanks. One small suggestion to `await userEvent.*`. And I'll let @emlys also decide which is the best target branch at this point.

workbench/tests/renderer/plugin.test.js
davemfish: I have also been around and around on this in the past and don't fully understand it. The version 14 release notes state that the APIs always return a promise, so `awaiting` seems appropriate. https://github.com/testing-library/user-event/releases/tag/v14.0.0

But it's possible that some tests that did not `await` would proceed to `await` something else, like a `findBy` query, that allowed the promise to resolve in time. Not awaiting the userEvent actions could sometimes lead to the dreaded `not wrapped in act()` warning from React. Adding `await` has resolved that for me in the past.

I also wonder if the resolution of the promise also depends on the event handler, which maybe makes the 'Add' click a unique case.

davemfish (approved):
Approving the changes, but I think @emlys suggested targeting `main` at this point, so I won't merge yet.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -34,15 +34,62 @@ export default function PluginModal(props) {
...
+    clearFormErrors();
+    if (validateAddPluginForm()) {
+      addPlugin();
+    }
+  };
+
+  const validateAddPluginForm = () => {
+    let formValid = true;
+    if ((installFrom === 'url' && !url)
+        || (installFrom === 'path' && !path)
```
emlys, line 81: I'm not sure about needing to show this warning when installing from a local path. I don't know much about security, but it seems that if the malicious code is already downloaded onto someone's computer, it may be too late? And the more likely case is that someone is installing a plugin that they've developed themselves

PR #1989, opened 2025-06-02
Workaround for gsutil python 3.13

### LINKED ISSUE #1980: `natcap.invest` sdist is not posted to the releases bucket
## To replicate
To replicate, view a recent release.  Example: [3.16.0a3](http://releases.naturalcapitalproject.org/?prefix=invest/3.16.0a3)

## Expected behavior

The `natcap-invest-3.16.0a3.tar.gz` archive should be there.

## Observed behavior

The `.tar.gz` is not there, just the wheels and everything else related to the release.

dcdenu4:
## Description
I think the issue here is that we're running into the same gsutil error with mac and python 3.13. Here's an [example](https://github.com/natcap/invest/actions/runs/15355465721/job/43213430679#step:10:24.) of the failure. So, this PR adds that same gsutil workaround to the Check sdist workflow, since in that workflow we're specifically waiting for the mac and latest python version to deploy.

I'm a bit out of practice in these github workflow environments, so let me know if I"m way off base!

Fixes #1980 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -255,8 +255,17 @@ jobs:
         if: github.event_name != 'pull_request' && matrix.os == 'macos-13' && matrix.python-version == env.LATEST_SUPPORTED_PYTHON_VERSION
         uses: google-github-actions/setup-gcloud@v2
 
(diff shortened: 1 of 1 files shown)


dcdenu4:
Looks good on my fork: https://github.com/dcdenu4/invest/actions/runs/15398973178/job/43326713436#step:11:47.

PR #2000, opened 2025-06-04
Identify plugins by `model_id@version`; display plugin version in models list

### LINKED ISSUE #1986: Bug: Loading parameters from file fails for plugins
## Description
When attempting to load parameters from a file for a plugin, the operation fails with the following error in the console: `TypeError: Cannot read properties of undefined (reading 'modelTitle') at SetupTab.loadParametersFromFile`

## To replicate

1. Install a plugin into workbench
2. In the Datastack Options, save parameters to json
3. (Try to) load parameters from file using the file you just created

## Expected behavior
I expect the fields to populate

## Observed behavior
No fields are populated and I get error: `TypeError: Cannot read properties of undefined (reading 'modelTitle') at SetupTab.loadParametersFromFile`

## Environment
- Operating System: MacOS Sequoia (15.5)
- InVEST version number: 3.16.0a2

## Additional context
 @davemfish noted that the issue seems to stem from this block in `SetupTab.loadParametersFromFile`:
```
if (datastack.model_id === modelID) {
  this.batchUpdateArgs(datastack.args);
  switchTabs('setup');
  this.triggerScrollEvent();
} else {
  alert(
    t(
      'Datastack/Logfile for model {{modelTitle}} does not match this model.',
      { modelTitle: investList[datastack.model_id].modelTitle }
    )
  );
}

```
According to the error, `investList[datastack.model_id]` is undefined. This results in the `modelTitle` access throwing the `TypeError`.

Note that in `renderer/app.jsx`, `updateInvestList()` sets the `modelTitle` and `type` for core models, but for plugins, only the `type` is set:
```

const coreModels = {};
const investList = await getInvestModelIDs();
Object.keys(investList).forEach((modelID) => {
  coreModels[modelID] = { modelTitle: investList[modelID].model_title, type: 'core' };
});
const plugins = await ipcRenderer.invoke(ipcMainChannels.GET_SETTING, 'plugins') || {};
Object.keys(plugins).forEach((plugin) => {
  plugins[plugin].type = 'plugin';
});
this.setState({
  investList: { ...coreModels, ...plugins },
});
```

emlys:
## Description
Fixes #1986

Instead of hashing the `model_id@version` identifier string, just escape the periods when using `settingsStore`. For the purpose of saving and opening datastacks, each combination of `model_id` and `version` is a unique model. 

This means that datastacks are not reusable between different versions of the same plugin, which I think is fine for now, but we might want to support differently in the future.

I tested and confirmed that datastacks created from a plugin (both JSON datastacks and TGZ archives) can be opened using the "Load parameters from file" button in the sidebar.

Also adding the plugin version to the displayed title in the list of models (if it is not unique). I'm not sure if this is the right design, vs. always displaying the version.

## Checklist

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -162,11 +162,11 @@ export function setupAddPlugin(i18n) {
         // Write plugin metadata to the workbench's config.json
         logger.info('writing plugin info to settings store');
         // Uniquely identify plugin by a hash of its ID and version
-        // Hashing because the version may contain dots,
-        // which doesn't work well as a key for electron-store's set and get methods
+        // Replace dots with underscores in the version, because dots are a
+        // special character in keys for electron-store's set and get methods
         const pluginID = `${modelID}@${version}`;
         settingsStore.set(
-          `plugins.${pluginID.replaceAll('.', '\\.')}`,
+          `plugins.${pluginID.replaceAll('.', '_')}`,
           {
             modelID: modelID,
             modelTitle: modelTitle,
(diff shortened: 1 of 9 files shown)


workbench/src/renderer/server_requests.js
```diff
@@ -208,7 +208,6 @@ export async function saveToPython(payload) {
  */
 export async function archiveDatastack(payload) {
   const { port, id } = await getPortAndID(payload.model_id);
-  payload.model_id = id;
```
emlys, line 211: For these two endpoints, use the original id in the format `model_id@version` rather than stripping it down to the `model_id`.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -164,9 +164,9 @@ export function setupAddPlugin(i18n) {
         // Uniquely identify plugin by a hash of its ID and version
         // Hashing because the version may contain dots,
```
davemfish: It looks like this comment could use updating

davemfish (commented):
This looks good @emlys . My only minor concern is that repeating the string-escaping in so many places is a bit cumbersome and prone to us missing a spot in the future.

I wonder if it would work to replace the dots with another character (e.g. underscore) when the `pluginID` is first created and entered into the `settingsStore`, near the end of `setupAddPlugin`. Would that avoid the need to deal with dots at all the subsequent `set` and `get` calls?

workbench/src/main/setupAddRemovePlugin.js
emlys: That's a good idea. I updated it to replace dots with underscores where the plugin ID is first set in the settings store, and then we can avoid all the escaping elsewhere.

PR #2008, opened 2025-06-04
rename variables in NDR extension

emlys:
## Description
After talking with @chrisnootenboom about NDR, and going over the code again, I realized that I was getting confused by some of these variable names. This PR contains no functional changes, I'm just renaming variables in hopes that it'll help me understand the model better! Most significantly, I'm removing everything with `eff` because it has been used to mean both "efficency" and "effective", which has caused a lot of confusion for me. While "effective" retention might have some meaning in a scientific context, I don't think it's useful to include in the code - just "retention" is fine.

## Checklist


phargogh (approved):
Gotta say, this is indeed a lot easier to read and understand.  Thanks @emlys !

PR #2013, opened 2025-06-05
catch ImportError in plugins discovery

### LINKED ISSUE #2012: handle ImportError when discovering plugins
Plugin discovery occurs in `natcap.invest.models.py` and involves import any package found by `pkgutil.iter_modules()` whose name `startswith(invest)`. 

Since we don't really know what will be discovered, we should wrap those imports in a `try/except`. This is especially necessary because of a circular import error witnessed in Windows conda environments during our feedstock builds, because `invest-script.py` is created as the invest entrypoint, and it tries to import itself anytime the cli is used.

davemfish:
Fixes #2012 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -60,9 +60,16 @@
   6. InVEST model Z (model names should be sorted A-Z)
 
 
-..
-  Unreleased Changes
-  ------------------
+Unreleased Changes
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks!

PR #2017, opened 2025-06-05
Move egg_info from to src to build

dkav:
## Description
Keep src directory clean by having natcap.invest.egg_info saved in build folder.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ env:
 # compatible with pip>=7.0.0
 # REQUIRED: Need to remove natcap.invest.egg-info directory so recent versions
 # of pip don't think CWD is a valid package.
-install: $(DIST_DIR)/natcap.invest%.whl
+install: $(BUILD_DIR) $(DIST_DIR)/natcap.invest%.whl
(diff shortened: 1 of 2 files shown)


dkav:
From my limited knowledge it seems like you could also remove the two instances of ` -$(RMDIR) natcap.invest.egg-info` from the [Makefile](https://github.com/dkav/invest/blob/a3224e787456d784e307987427604fb344aae650/Makefile).

emlys:
@dkav, I'm not sure I agree that this is worth it to deviate from the default setting. Is there a reason that this is a best practice and not just a personal preference?

dkav:
My perspective is similar to the initial question in this [pypa discussion](https://github.com/pypa/setuptools/discussions/3348). It was also the source for how I implemented the change. 

In the absence of a project `.gitignore`, I have been building my own `.gitignore` and I was surprised that the `egg-info` folder was being placed in `src` and not a build or similar type directory. As I was already ignoring the `build` directory it made sense to me to have `egg-info` in there too. It also made it easier to clean up after a build run.

PR #2024, opened 2025-06-06
Remove deprecated setuptools_scm version scheme

### LINKED ISSUE #2023: Our setuptools_scm versioning scheme is deprecated
## Description
We have had `setuptools_scm` configured to use `post-release` versioning, which is apparently deprecated (https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmversion_scheme). I see no reason not to switch to the suggested alternative and default scheme, `guess-next-dev`.

emlys:
## Description
Fixes #2023

Switching from the deprecated `post-release` version scheme to the suggested replacement, `guess-next-dev` (default):

> Automatically guesses the next development version (default). Guesses the upcoming release by incrementing the pre-release segment if present, otherwise by incrementing the micro segment. Then appends :code:.devN. In case the tag ends with .dev0 the version is not bumped and custom .devN versions will trigger a error.

The `local_scheme` remains the same, but it's the default, so I removed the whole config block.

Also bumping to `setuptools>=64`, which is suggested in the setuptools_scm docs.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,8 +46,7 @@ invest = "natcap.invest.cli:main"
 # available at runtime.
 requires = [
     'setuptools>=64', 'wheel', 'setuptools_scm>=8.0', 'cython>=3.0.0', 'babel',
-    'oldest-supported-numpy; python_version<="3.8"',
-    'numpy>=2; python_version>="3.9"',  # numpy 2 only available for 3.9+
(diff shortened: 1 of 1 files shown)


davemfish (changes requested):
Thanks, @emlys ! Just looks like there's a conflict in pyproject.toml

Issue #1543, opened 2024-03-25
Transition to functional React components


emlys:
When we began developing the workbench, class components were the standard and functional components were new. React has moved away from class components, and functional components are the standard. I haven't seen any indication that class components will be deprecated any time soon, but there are some reasons it would make sense to convert our code to use functional components:

- Documentation about class components has been moved to a [legacy site](https://legacy.reactjs.org/docs/components-and-props.html) and is no longer updated
- It's increasingly hard to find tutorials and code examples using class components
- `i18next` is easier to use with functional components; exported class components require a workaround
- We have written our newer components using the functional style, and found it to work well and require less boilerplate code
- When working on code that uses both styles of components, I find it confusing having to switch back and forth

R-Stev:
I should be able to adjust the class components.

PR #2030, opened 2025-06-09
Fix recent jobs filtering and refactor app.jsx as functional component

### LINKED ISSUE #1977: After removing a plugin, its jobs still appear in the recent jobs list
## Description
After removing a plugin, the model list refreshes so that the plugin is no longer displayed, but the recent jobs data does not get updated to filter out jobs from that plugin.

## To replicate

1. Install and run the demo plugin
2. Remove the demo plugin
3. Observe that the recent job remains displayed

### LINKED ISSUE #2028: Recent jobs are not filtered after saving or removing a job
## Description
On first load, the "Recent jobs" list is filtered to include only models that are currently available in the workbench. But when a job is added or deleted, the list is refreshed without filtering, causing other jobs to appear.

emlys:
## Description
Fixes #1977 by updating the recent jobs list when the `investList` changes. This way the recent jobs data is filtered according to the current set of invest models.

Fixes #2028 by applying the same filtering the recent jobs list when initializing it OR updating it.

I tried to make these changes using a callback in the old class-style component, but I ran into an issue where the callback didn't seem to wait for state to update as it should. The logic seemed more intuitive to express with `useEffect` so I refactored the whole `App` component into the modern functional style. We've been meaning to make this switch anyway (#1543), so I think it was worth the effort. 

The invest test failures are unrelated, see #2032 

## Checklist

### DIFF
--- a/workbench/src/renderer/app.jsx
+++ b/workbench/src/renderer/app.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import i18n from 'i18next';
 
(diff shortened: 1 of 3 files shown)


workbench/src/renderer/app.jsx
```diff
@@ -36,446 +36,369 @@ const { ipcRenderer } = window.Workbench.electron;
...
       new Uint32Array(1)
     ).toString();
-    const { openJobs, openTabIDs } = this.state;
-    openTabIDs.push(tabID);
-    openJobs[tabID] = job;
-    this.setState({
-      openTabIDs: openTabIDs,
-      openJobs: openJobs,
-    }, () => this.switchTabs(tabID));
+    const newOpenJobs = new Map(openJobs);
```
emlys, line 99: I also took the opportunity to combine the `openTabIDs` and `openJobs` state variables into a single `Map` which serves the purpose of both (tracking both the order of open tabs and their corresponding jobs).

workbench/src/renderer/app.jsx
```diff
@@ -36,446 +36,369 @@ const { ipcRenderer } = window.Workbench.electron;
...
-    const recentJobs = await InvestJob.getJobStore();
-    this.setState({
-      // filter out models that do not exist in current version of invest
-      recentJobs: recentJobs.filter((job) => (
-        Object.keys(investList)
-          .includes(job.modelID)
-      )),
-      showDownloadModal: this.props.isFirstRun,
+  useEffect(() => {
+    async function componentDidMount() {
```
emilyanndavis: Could we rename this function so that it no longer has the same name as a component lifecycle hook? Maybe something generic like `setup` or similar? While we still have a mix of functional and class components, it's a little confusing to see a `componentDidMount` function defined inside a `useEffect` hook.

workbench/src/renderer/app.jsx
emilyanndavis: Great improvement, thanks!

workbench/src/renderer/app.jsx
```diff
@@ -36,446 +36,369 @@ const { ipcRenderer } = window.Workbench.electron;
...
-      showPluginModal,
-      showChangelog,
-      showSettingsModal,
-      showMetadataModal,
-      downloadedNofN,
-    } = this.state;
+  async function updateRecentJobs() {
+    if (investList) {
+      const recentJobs = await InvestJob.getJobStore();
+      // filter out models that do not exist in current version of invest
```
emilyanndavis, line 187: I don't think this is crucial right now, but it seems like it would be worth exploring solutions that would enable skipping this filtering step when it's not needed. I'm assuming we're not directly removing outdated models from the job store so that we can continue supporting multiple versions of InVEST on the same machine? but in scenarios where that's not relevant (e.g., if we check for outdated models in the job store and find none), we could probably set a flag to skip that check on every subsequent call to updateRecentJobs. Anyway, I'm just thinking out loud here; this is not likely to cause a noticeable performance lag unless a user accumulates a large number of recent (or maybe not-so-recent!) jobs and rarely or never clears them.

workbench/src/renderer/app.jsx
emlys: Sure! Renamed to `setup`

workbench/src/renderer/app.jsx
emlys: We could also consider automatically deleting jobs after some limit is reached. With the current design, having hundreds of saved jobs would not be very useful, since you scroll through them.

emilyanndavis (approved):

PR #2057, opened 2025-06-24
Use pydantic for model specs

### LINKED ISSUE #2051: Use Pydantic to enforce model spec class structure
As @davemfish suggested, we can use `pydantic` to enforce that the `ModelSpec` and `Input` and `Output` classes have the right attributes of the right types. This will make the `test_model_specs` tests obsolete, and be helpful for future plugin development.

emlys:
## Description
Fixes #2051

- Use pydantic with all classes defined in `spec`. This means inheriting them from `pydantic.BaseModel` (and removing the dataclass decorator). This enables a lot of type checking automatically.
- For more complex constraints, add pydantic validator functions.
- Remove `test_model_specs.py`. Those tests are superseded by pydantic validation.
- Create a new `Option` class to represent options in an `OptionStringInput` or `OptionStringOutput`. This lets us use pydantic to enforce option properties, and simplifies some logic.
- Refactor RST formatting functions as class methods
- Remove some default values for `Input` and `Output` attributes, enforcing that the user define them explicitly
- Update tests where needed to meet the new pydantic requirements  
- Removed the `IterableWithDotAccess` class, instead using getter methods to get nested inputs. I had thought that it would be useful having the flexibility to access nested inputs with dot notation, but I don't think it's worth the trouble for now.

## Checklist


tests/test_ui_server.py
```diff
@@ -291,3 +291,8 @@ def test_set_geometamaker_profile(self, mock_user_config_dir):
         self.assertEqual(
             response.json,
             {'message': 'Metadata profile saved', 'error': False})
+
+    def test_model_specs_serialize(self):
```
emlys, line 295: Moved this here from `test_model_specs.py`.

src/natcap/invest/scenic_quality/scenic_quality.py
```diff
@@ -282,10 +282,6 @@ def execute(args):
             'a': float(args['a_coef']),
             'b': float(args['b_coef']),
         }
-        if (args['valuation_function'] not in
-                MODEL_SPEC.get_input('valuation_function').options):
```
emlys, line 286: This block was redundant to the automatic validation done in `validate`.

src/natcap/invest/spec.py
```diff
@@ -147,49 +147,47 @@ def _check_projection(srs, projected, projection_units):
...
-            key: the item id
-
-        Returns:
-            the corresponding item
-        """
-        return self.inputs_dict[key]
-
-    def to_json(self):
-        """Return a JSON serializable representation of self.
+def check_permissions(permissions):
```
davemfish: Could we consider changing this name to `validate_permissions` or `validate_permissions_str`? I got tripped up for a minute when I saw this function being applied and thought, "how can we check permissions of a file when `FileInput` is initialized?"

workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -58,12 +58,10 @@ function initializeArgValues(argsSpec, inputFieldOrder, argsDict) {
...
-      } else { // default to first
-        if (Array.isArray(argsSpec[argkey].options)) {
-          value = argsSpec[argkey].options[0];
-        } else {
-          value = Object.keys(argsSpec[argkey].options)[0];
-        }
+      } else if (argsSpec[argkey].options.length > 0) { // default to first
+        value = argsSpec[argkey].options[0].key;
+      } else {
+        value = undefined; // for dynamic dropdowns before the values are initialized
```
davemfish: It may be slightly better to initialize to an empty string. Sometimes if an `input` value changes from `undefined` to something else, React issues a warning about an uncontrolled component being changed to a controlled component.

src/natcap/invest/spec.py
emlys: Sure, renamed to `validate_permissions_string`

workbench/src/renderer/components/SetupTab/index.jsx
emlys: Good point!

davemfish (approved):

davemfish:
Looks like the binary builds are failing because of a scipy-pyinstaller packaging issue. I'll go ahead and merge since the fix for that will almost certainly be in scipy.

Issue #1547, opened 2024-04-02
Integrate DOIs into the InVEST release process


phargogh:
Per the [DataCite Best Practices documentation](https://support.datacite.org/docs/best-practices-for-datacite-members#versioning):

> The [DataCite Metadata Schema](https://schema.datacite.org/) documentation recommends that a new DOI should be assigned when there is a new major version of the resource you are sharing

Zenodo comes to mind as a research product repository that supports versioning like this, and this has been used for some other research-focused software projects.  For example:

* Xarray: https://zenodo.org/records/10895413
* cffconvert: https://zenodo.org/records/5171937

[Zenodo even offers a github integration](https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content), where a new DOI is minted and the source code archived on each github release.  While this may be fine for most releases, there's a bit of a chicken-and-egg problem because it'd be great to have the DOI be available _as a part of the research artifacts_ and not just tacked on after the fact.  I believe a DOI can be reserved in advance through zenodo, but this would need to be confirmed and then integrated into the release automation if so.  Otherwise, we'll just have to be OK with noting the DOI after the fact.

phargogh:
I am currently working on uploading our artifact history to zenodo for the complete release history of InVEST, not including development builds.  I bumped this issue to the next release milestone to make sure that this won't interfere with the 3.15.0 release.

PR #2066, opened 2025-07-01
Update CITATION.cff as a part of the release process

phargogh:
With this PR, we now update CITATION.cff with the expected DOI and the date of the release.  At this time, the DOI is not actually minted during the release process, as this is still handled as a manual process over in https://github.com/natcap/invest.release-metadata, but since our DOIs have a predictable pattern, we can just make sure it's formatted correctly and then it should line up with what we create over in the other repo.

RE:#1547

## Checklist

### DIFF
--- a/.github/workflows/release-part-1.yml
+++ b/.github/workflows/release-part-1.yml
@@ -110,7 +110,7 @@ jobs:
           package-json-path: workbench/package.json
 
       - name: Update CITATION.cff
-        if: env.RUN == 'true
+        if: env.RUN == 'true'
         run: python scripts/update-citation-cff.py CITATION.cff $VERSION $(date '+%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


.github/workflows/release-part-1.yml
```diff
@@ -104,12 +104,17 @@ jobs:
           ignore-semver-check: "true"  # allow PEP 440 alpha/beta/rc releases
           package-json-path: workbench/package.json
 
-      - name: Commit updated HISTORY.rst, changelog.html, and package.json
+      - name: Update CITATION.cff
+        if: env.RUN == 'true
```
emlys:
```suggestion
        if: env.RUN == 'true'
```

emlys (approved):
👍

Issue #1413, opened 2023-09-18
LS Factor: restore the weighted mean of proportional flow


phargogh:
In our recent LS Factor update (ADR here: https://github.com/natcap/invest/blob/main/doc/decision-records/ADR-0001-Update-SDR-LS-Factor.md), we removed the weighted mean of proportional flow, replacing it with the dot-product `sin(alpha) + cos(alpha)` to represent the flow length across a pixel.

In revisiting [Desmet & Govers (1996)](https://searchworks.stanford.edu/articles/edsgac__edsgac.A18832564), it turns out that they actually use _both_ of these methods in the calculation of the LS Factor ... but I probably took out the weighted mean part because SAGA doesn't do it because they're assuming D8 flow.

In any case, look back into D&G(1996) and make sure we're doing the right LS Factor calculations, see how much this changes things, run the new build by Rafa and be sure to note the new change in an ADR, HISTORY and the User's Guide.

phargogh:
FYI @schmittrjp

schmittrjp:
Thanks - keep me posted on this @phargogh

dcdenu4:
Hey @phargogh, just checking in on this issue and status and where we might want to place it for a Milestone.

phargogh:
No progress has been made yet except for this issue.  I've removed the in progress tag for now until I (or someone else) can get to it.

dcdenu4:
Hey @phargogh , this is a great rundown. Thanks for taking the time to deep dive on it. Do you think it'd be worth amending the [ADR here](https://github.com/natcap/invest/blob/main/doc/decision-records/ADR-0001-Update-SDR-LS-Factor.md) with your findings?

phargogh:
# I was wrong

In my comment above (https://github.com/natcap/invest/issues/1413#issuecomment-3010629821), I say that we are already closely following Desmet & Govers (1996).  I was wrong.

In working up an ADR summarizing my notes above (which is now in PR at #2068), I realized several key things:

1. D+G do, indeed use a weighted mean of proportional flow calculation (eq. 2) as a part of the calculation of the unit contributing area (eq 4), which is then used in the LS Factor calculation (eq 9).
2. Because the complaint in #915 was comparing our LS Factor to SAGA, we went with one of SAGA's estimates for the unit contributing area instead of using D+G's formulation.  This estimate is used in place of D+G's unit contributing area calculations.
[...]

PR #2068, opened 2025-07-10
ADR for documenting SDR's lack of use of Weighted Mean of Proportional Flow

phargogh:
This PR adds an ADR with findings related to the question of whether SDR should have the Weighted Mean of Proportional Flow (WMPF, yes it's fun to say, like the sound of a clump of dense, wet snow falling out of a tree to the ground).

In short, the current version of SDR uses an alternative, inexpensive estimate for unit contributing area in place of the WMPF, and this appears to be fine and is close to what SAGA uses.  If the user is using D8 routing in SDR, then the LS factor should be functionally the same as what SAGA produces.

Additionally, although we were using _a_ WMPF calculation between InVEST 3.8.0 - 3.13.0, it was not _the_ WMPF calculation that D+G use, so we were still wrong there.

RE:#1413

## Checklist

### DIFF
--- a/doc/decision-records/ADR-0001-Update-SDR-LS-Factor.md
+++ b/doc/decision-records/ADR-0001-Update-SDR-LS-Factor.md
@@ -1,8 +1,14 @@
 # ADR-0001: Update the InVEST SDR LS Factor
 
 Author: James
-
 Science Lead: Rafa
+
+## Changelog
+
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Thanks @phargogh for taking the time to detail this out!

PR #2080, opened 2025-07-28
Specify `NoData` when calculating CC if CC method = `intensity`

### LINKED ISSUE #2079: UCM: `NoData` not set correctly if CC Method is `intensity`
## Description
A forum user reported an issue (https://community.naturalcapitalproject.org/t/problemi-con-metodo-intensity-nel-modello-urban-cooling-raster-t-air-incompleto/6117/3) where, if `intensity` was selected as the Cooling Capacity Calculation Method, `T_air.tif` is cropped and does not extend to the full AOI. 

When I looked into the layers used in the calculation of `t_air`, I found that `HM.tif` has incorrectly set `NoData` value, and instead the max float value (`3.4028e38`) registers as a valid value. This is likely because, when `intensity` is set as the CC method, the CC index is calculated using `pgp.raster_map` *with no* `target_nodata` *value explicitly specified*

This problem does not appear if `factors` is the CC method used, as `target_nodata` is explicitly specified in the calculation of `cc`.

## Additional context
The user tried using a larger LULC raster, which they reported solved this issue. This may have to do with an issue we previously documented, where `HM.tif` was missing areas of data when the aligned/resized ET raster (`ref_eto`) did not fully overlap with `lulc`. A workaround was to reproject `ET` to match `lulc` before running the model.

From the slack thread:
> It looks like the issue [of missing data] is present in 3.15 and less so in 3.14.3 because in 3.15, the `cc_park` calculation (correctly) incorporates `cc`, which is calculated using `eti`, `shade` and `albedo` (the latter 2 which are derived from `lulc`). `eti` is calculated based on `kc` (derived using `lulc`) and the aligned/resized ET raster, `ref_eto`, which, in this example, does not fully overlap with `lulc`. Any layers calculated using both `eto` and `lulc` (or derivatives) will have no data where these layers don't intersect, which we see in `eti`, `cc`, and `cc_masked_green_areas`.
> In 3.14.3, the `cc_park` calculation incorrectly only used `green_area` which is just based off `lulc` so the ET layer wasn't causing issues.
> Then, not sure about this, but my guess in that the convolution step is causing the `nodata` values in `cc_masked_green_areas` to propagate outward, creating a larger `nodata` area in `cc_park`?
> I think this is indeed an issue that originates in `align_and_resize_raster_stack` . Quick fix: If you reproject your ET raster to match the projection of your LULC raster before running the model, you should see better results.
> TLDR; the issue is that after realignment/resizing, the ET raster doesn't fully overlap with the lulc raster. This issue can be seen in invest 3.14.3, but the effect is greater in 3.15, as `cc_park` is now calculated correctly using a layer that was previously ignored

claire-simpson:
## Description
`NoData` was not explicitly specified when calculating cooling capacity (in `raster_map` operation), leading to default float nodata value in the output cc raster. However `3.4e38` did not register as `NoData` in the calculation of `HM.tif`, which ultimately caused issues with `T_air.tif` (as `HM` is used in the calculation of `T_air`).

This is similar to a Stormwater bug that was also reported on the forum (#1850).

Fixes #2079

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -65,6 +65,12 @@
   Unreleased Changes
   ------------------
 
+Urban Cooling
+=============
+* Fixed a bug where ``NoData`` in ``cc`` was not correctly set if the Cooling
+  Capacity Calculation Method was set to ``intensity``
+  (`#2079 <https://github.com/natcap/invest/issues/2079>`_).
+
 3.16.1 (2025-07-01)
 -------------------
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -74,7 +74,7 @@ Urban Cooling
 =============
 * Model validation now requires that the "UHI effect" is >= 0 degrees Celsius,
   meaning that the urban air temperature is greater than the rural reference
-  temperature. (`#2076 <https://github.com/natcap/invest/issues/2076>`_
+  temperature. (`#2076 <https://github.com/natcap/invest/issues/2076>`_)
 * Fixed a bug where ``NoData`` in ``cc`` was not correctly set if the Cooling
   Capacity Calculation Method was set to ``intensity``
   (`#2079 <https://github.com/natcap/invest/issues/2079>`_).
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Interestingly, the `calc_eti_op` uses `raw` value inputs to pass along NODATA into the `raster_calculator` operation. This is an alternative better practice approach to wrapping the function.

PR #2087, opened 2025-08-01
Support GDAL VSI paths for remote raster and vector inputs

### LINKED ISSUE #2077: Support remote input paths in all models
Now that `pygeoprocessing` supports remote input paths using `/vsicurl/`, what if any additional measures do we need to take for `natcap.invest` to support them? Changes are likely needed to validation and utils, and maybe some model-specific code.

## Description
All models should accept remote input paths using `/vsicurl/`.

Do we need to do anything else to support the other network-based file system handlers such as `/vsis3/` and `/vsigs/`?

Eventually we may want to support authentication with those handlers, or have a UI wrapper around that, but I'm excluding that from this initial task.

emlys:
## Description
Fixes #2077

## Checklist

### DIFF
--- a/src/natcap/invest/hra.py
+++ b/src/natcap/invest/hra.py
@@ -2037,9 +2037,10 @@ def _parse_criteria_table(criteria_table_path, target_composite_csv_path):
                     float(attribute_value)
                 except ValueError:
                     # If we can't cast it to a float, assume it's a string path
-                    # to a raster or vector.
-                    attribute_value = utils.expand_path(
-                        attribute_value, criteria_table_path)
+                    # to a raster or vector. Don't expand remote paths.
+                    if '://' not in attribute_value:
+                        attribute_value = utils.expand_path(
+                            attribute_value, criteria_table_path)
 
                     try:
                         _ = pygeoprocessing.get_gis_type(attribute_value)
--- a/src/natcap/invest/validation.py
+++ b/src/natcap/invest/validation.py
@@ -165,9 +165,6 @@ def load_fields_from_vector(filepath, layer_id=0):
         A list of string fieldnames within the target layer.
 
     """
-    if not os.path.exists(filepath):
-        raise ValueError('File not found: %s' % filepath)
-
     vector = gdal.OpenEx(filepath, gdal.OF_VECTOR)
     layer = vector.GetLayer(layer_id)
     fieldnames = [defn.GetName() for defn in layer.schema]
--- a/src/natcap/invest/datastack.py
+++ b/src/natcap/invest/datastack.py
@@ -578,6 +578,10 @@ def extract_parameter_set(paramset_path):
                 # Probably not a boolean, so continue checking paths.
                 pass
 
+            # Don't expand or modify remote paths
+            if '://' in args_param:
+                return args_param
+
             # Convert paths to whatever makes sense for the current OS.
             expanded_param = os.path.expandvars(
                 os.path.expanduser(
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -384,9 +384,10 @@ class RasterInput(FileInput):
         Returns:
             A string error message if an error was found.  ``None`` otherwise.
         """
-        file_warning = super().validate(filepath)
-        if file_warning:
-            return file_warning
+        if not filepath.startswith('/vsi'):
+            file_warning = super().validate(filepath)
+            if file_warning:
+                return file_warning
 
         try:
             gdal_dataset = gdal.OpenEx(filepath, gdal.OF_RASTER)
@@ -445,9 +446,10 @@ class SingleBandRasterInput(FileInput):
         Returns:
             A string error message if an error was found.  ``None`` otherwise.
         """
-        file_warning = super().validate(filepath)
-        if file_warning:
-            return file_warning
+        if not filepath.startswith('/vsi'):
+            file_warning = super().validate(filepath)
+            if file_warning:
+                return file_warning
 
         try:
             gdal_dataset = gdal.OpenEx(filepath, gdal.OF_RASTER)
@@ -522,9 +524,10 @@ class VectorInput(FileInput):
         Returns:
             A string error message if an error was found.  ``None`` otherwise.
         """
-        file_warning = super().validate(filepath)
-        if file_warning:
-            return file_warning
+        if not filepath.startswith('/vsi'):
+            file_warning = super().validate(filepath)
+            if file_warning:
+                return file_warning
 
         try:
             gdal_dataset = gdal.OpenEx(filepath, gdal.OF_VECTOR)
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -321,9 +321,13 @@ class FileInput(Input):
         Returns:
             Transformed dataframe column
         """
-        return col.apply(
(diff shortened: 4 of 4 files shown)


src/natcap/invest/validation.py
```diff
@@ -165,9 +165,6 @@ def load_fields_from_vector(filepath, layer_id=0):
         A list of string fieldnames within the target layer.
 
     """
-    if not os.path.exists(filepath):
```
emlys, line 168: Removing this so we don't have to handle remote vs. local paths. `gdal.OpenEx` below will raise an error if the file isn't found.

src/natcap/invest/cli.py
```diff
@@ -465,6 +465,11 @@ def main(user_args=None):
             LOGGER.info('Imported target %s from %s',
                         model_module.__name__, model_module)
 
+            for arg_key, val in parsed_datastack.args.items():
+                if type(model_module.MODEL_SPEC.get_input(arg_key)) in {
+                        spec.RasterInput, spec.SingleBandRasterInput, spec.VectorInput}:
+                    parsed_datastack.args[arg_key] = utils._GDALPath.from_uri(val).to_string()
```
emlys: Convert spatial paths to gdal-ready format. Eventually this should happen in a standardized preprocessing step at the beginning of model execution.

src/natcap/invest/hra.py
```diff
@@ -1878,8 +1878,6 @@ def _parse_info_table(info_table_path):
                   ``{stressor_name: {'path': path to spatial layer, 'buffer':
                       buffer distance}}``
     """
-    info_table_path = os.path.abspath(info_table_path)
```
emlys, line 1881: This was redundant and causes problems with remote paths

src/natcap/invest/utils.py
```diff
@@ -841,3 +843,127 @@ def copy_spatial_files(spatial_filepath, target_dir):
...
+    Attributes
+    ----------
+    path : str
+        Parsed path. Includes the hostname and query string in the case
+        of a URI.
+    archive : str
+        Parsed archive path.
+    scheme : str
+        URI scheme such as "https" or "zip+s3".
+    """
```
dcdenu4, line 865: Since we started using classes in `spec.py` I'm curious your thoughts on if we should keep a standardized approach for our class docstrings across InVEST? I know this is largely pulled from rasterio._path, but maybe we want to document the attributes in the class body like we do in spec?

src/natcap/invest/utils.py
```diff
@@ -841,3 +843,127 @@ def copy_spatial_files(spatial_filepath, target_dir):
...
+        self.schemes = {
+            'ftp': 'curl',
+            'gzip': 'gzip',
+            'http': 'curl',
+            'https': 'curl',
+            's3': 's3',
+            'tar': 'tar',
+            'zip': 'zip',
+            'file': 'file',
+            'oss': 'oss',
```
dcdenu4, line 882: What's `oss`?

src/natcap/invest/utils.py
```diff
@@ -841,3 +843,127 @@ def copy_spatial_files(spatial_filepath, target_dir):
         return_filepath = target_filepath
 
     return return_filepath
+
+
+class _GDALPath:
```
dcdenu4, line 848: I wonder if we could include a comment somewhere that highlights the structure for a URL. From urlparse docs: 
>This corresponds to the general structure of a URL:` scheme://netloc/path;parameters?query#fragment`

Looking this up helped me walk through this class, so it might be helpful to future maintainers too.

src/natcap/invest/utils.py
```diff
@@ -841,3 +843,127 @@ def copy_spatial_files(spatial_filepath, target_dir):
...
+    def from_uri(cls, uri):
+        """Instantiate a _GDALPath for a given URI.
+
+        Args:
+            uri: URI string to parse
+
+        Returns:
+            _GDALPath object
+        """
+        parts = urlparse(uri)
```
dcdenu4, line 908: Do we handle the case where someone might be passing in an already formatted GDAL VSI path?

src/natcap/invest/utils.py
emlys: It's a bit different because we're not using pydantic here, there's no attributes definition section. The attributes are initialized in `__init__` as regular lines of code. I don't think we have any precedent of how we document attributes like this.

src/natcap/invest/utils.py
emlys: Alibaba cloud storage, apparently! https://gdal.org/en/stable/user/virtual_file_systems.html#vsioss Did not know they offered that

src/natcap/invest/utils.py
emlys: Added a note about this to the docstring

src/natcap/invest/utils.py
emlys: No, at first I thought we should. But I noticed that in the rasterio source code these are described as "legacy GDAL filenames" and they are handling them as a special case. I'm guessing it's easier to support just the one format, and it prevents any ambiguity where a path starting with `/vsizip/` could actually be a regular directory.

src/natcap/invest/utils.py
```diff
@@ -841,3 +843,130 @@ def copy_spatial_files(spatial_filepath, target_dir):
...
+    Attributes
+    ----------
+    path : str
+        Parsed path. Includes the hostname and query string in the case
+        of a URI.
+    archive : str
+        Parsed archive path.
+    scheme : str
+        URI scheme such as "https" or "zip+s3".
+    """
```
dcdenu4, line 865:
```suggestion
    """
    
```

dcdenu4: Looks like PEP257 would like a blank line after all docstrings that document a class.

src/natcap/invest/utils.py
dcdenu4: Oh right, thanks for that reminder of why those classes with pydantic are a bit of a unique case.

tests/test_cli.py
```diff
@@ -126,6 +126,34 @@ def test_run_coastal_blue_carbon_invalid_datastack(self):
             ])
         self.assertEqual(exit_cm.exception.code, 1)
 
+    def test_run_carbon_remote_files(self):
```
dcdenu4, line 129: Awesome, thanks!

tests/test_hra.py
```diff
@@ -450,6 +451,41 @@ def test_criteria_table_missing_section_headers(self):
         self.assertIn('HABITAT NAME', str(cm.exception))
         self.assertIn('HABITAT RESILIENCE ATTRIBUTES', str(cm.exception))
 
+    def test_criteria_table_remote_filepath(self):
+        """HRA: verify exception when a required section is not found."""
```
dcdenu4: Update docstring

tests/test_validation.py
```diff
@@ -1007,6 +1007,22 @@ def test_get_validated_dataframe(self):
         self.assertEqual(df['header'][0], 'a')
         self.assertEqual(df['header'][1], 'b')
 
+    def test_remote_csv_with_local_path(self):
+        """validation: an error is raised if a remote csv reference local paths"""
```
dcdenu4:
```suggestion
        """validation: an error is raised if a remote csv reference local paths."""
```

Issue #2054, opened 2025-06-19
Provide a key for `Output` objects separate from the file path


emlys:
## Background/Motivation
In the current design of the `spec.Output` class and subclasses, the output's `id` is its path (for instance `aggregate_results.gpkg`). This was nice and simple to start with, but now that we are doing more with the `ModelSpec`, it could be helpful to have separate `id` and `path` attributes. @megannissel and I discussed this last week.

Currently the file format is inferred from the file extension in the `id`, which works but it would be cleaner to explicitly document the format. Perhaps the `id` and `format` could replace the `path` - given an id (`aggregate_results`) and a format (`GPKG`), generate the file name (`aggregate_results.gpkg`).

#2053 is closely related.

PR #2089, opened 2025-08-04
Output Spec Updates: Separate `path` and `id`; remove `DirectoryOutput`s

### LINKED ISSUE #2053: Don't use `DirectoryOutput`s to organize groups of `Output`s
I am rethinking the design of using `DirectoryOutput`s to organize groups of `Output`s. While we do want to organize the files within directories, it's perhaps not the best way to think of documenting those files. Historically this design derives from when we used much simpler dictionaries, but things are getting more complicated!

The directories themselves are usually not meaningful outputs. So it would be nice to remove those `DirectoryOutput`s and instead have all the file-based `Output`s at the top level. Perhaps adding a new attribute to track their location within the workspace (e.g. `location='outputs/intermediate'`).

megannissel:
## Description
Fixes #2053 & #2054

While doing some initial work on updating `build_file_registry`, I started altering the Wind Energy output spec. From there, I figured there wasn't much reason to hold off on making these changes until the larger file registry issue was sorted; adding `path`s and removing `DirectoryOutput`s from our model specs is a separate enough issue to warrant its own PR (though doing these two things together made more sense to my brain than separating them out). 

In this PR:
- A `path` attribute has been added to the `FileOutput` class
- All file-type output classes now inherit from `FileOutput`
- All file outputs in our model specs now have separate `path` and `id` attributes
- Model output specs have been updated to remove grouping outputs within `DirectoryOutput`s. Location information within the workspace is now included in / can be derived from the `path`
- The spec test `_generate_files_from_spec` has been updated to work with the new `path` attribute
- `test_write_metadata_for_outputs` has been updated; I made the example model spec slightly more complicated, to make sure it will still handle files nested within `DirectoryOutput`s for backwards-compatibility. 

Remaining questions: Under what use cases might we still want to use `DirectoryOutput`? If we keep it around as a class, should it also have a `path` attribute? For the moment, I didn't add one (figuring `id` is sufficient), but I could see it being handy in a nested-directory context (especially if we want to automate making a directory that doesn't have any specific files in it, the paths of which could be parsed to construct the full directory tree). 

## Checklist

### DIFF
--- a/.github/workflows/release-part-1.yml
+++ b/.github/workflows/release-part-1.yml
@@ -6,7 +6,7 @@ on:
     # cron doesn't support "first tuesday of the month", so we will use github
     # actions syntax below to skip tuesdays that aren't in the first week.
     # it is recommended not to start on the hour to avoid peak traffic
-    - cron: "20 18 * * 2"
+    - cron: "20 20 * * 2"
   workflow_dispatch:
     inputs:
       version:
@@ -15,7 +15,6 @@ on:
 
 env:
   VERSION: ${{ inputs.version }}
-  AUTORELEASE_BRANCH: autorelease/${{ inputs.version }}
   # use a personal access token here which has permissions to trigger further actions
   # this is necessary for the pr checks
   GITHUB_TOKEN: ${{ secrets.AUTORELEASE_BOT_PAT }}
@@ -28,11 +27,12 @@ jobs:
       # skip scheduled runs that aren't the first tuesday of the month
       - name: Check day of month
         run: |
-          if [[ $GITHUB_EVENT_NAME == "workflow_dispatch" || $(date +%d) -le 7 ]]; then
-            echo "RUN=true" >> $GITHUB_ENV
-          else
-            echo "RUN=false" >> $GITHUB_ENV
-          fi
+          #if [[ $GITHUB_EVENT_NAME == "workflow_dispatch" || $(date +%d) -le 9 ]]; then
+          #  echo "RUN=true" >> $GITHUB_ENV
+          #else
+          #  echo "RUN=false" >> $GITHUB_ENV
+          #fi
+          echo "RUN=true" >> $GITHUB_ENV
 
       - uses: actions/checkout@v4
         if: env.RUN == 'true'
@@ -44,24 +44,29 @@ jobs:
         if: env.RUN == 'true'
         run: pip install rst2html5 setuptools_scm
 
+      # Assuming there has been at least one commit since the last tag,
+      # the bugfix number in the setuptools_scm version will be one
+      # greater than that of the last tag.
+      # https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme
       - name: Get current version number (scheduled runs only)
         if: env.RUN == 'true' && github.event_name == 'schedule'
-        run: echo "CURRENT_VERSION=$(python -m setuptools_scm)" > $GITHUB_ENV
+        run: echo "SETUPTOOLS_SCM_VERSION=$(python -m setuptools_scm)" > $GITHUB_ENV
 
       - name: Get next bugfix version number (scheduled runs only)
         if: env.RUN == 'true' && github.event_name == 'schedule'
         shell: python
         run: |
           import os
-          major, minor, bugfix, *_ = os.environ['CURRENT_VERSION'].split('.')
-          with open(os.environ['GITHUB_ENV']) as env_file:
-            env_file.write(f'VERSION={major}.{minor}.{int(bugfix) + 1}')
+          major, minor, bugfix, *_ = os.environ['SETUPTOOLS_SCM_VERSION'].split('.')
+          with open(os.environ['GITHUB_ENV'], "a") as env_file:
+            env_file.write(f'VERSION={major}.{minor}.{bugfix}')
 
-      - name: Configure git
+      - name: Configure git and define branch name
         if: env.RUN == 'true'
         run: |
           git config user.name "GitHub Actions"
           git config user.email "<>"
+          echo "AUTORELEASE_BRANCH=autorelease/$VERSION" > $GITHUB_ENV
 
       # Members of the natcap software team can push to the autorelease branch on
       # natcap/invest; this branch is a special case for our release process.
@@ -100,7 +105,7 @@ jobs:
         if: env.RUN == 'true'
         uses: BellCubeDev/update-package-version-by-release-tag@v2
         with:
-          version: ${{ inputs.version }}
+          version: $VERSION
           ignore-semver-check: "true"  # allow PEP 440 alpha/beta/rc releases
           package-json-path: workbench/package.json
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -70,6 +70,10 @@ General
 * Updated readme to fix broken links, and also add links to repositories
   containing older source code and downloads, for future reference.
   (`#2029 <https://github.com/natcap/invest/issues/2029>`_)
+* Fixed a bug in ``build_datastack_archive`` where the raster filepath
+  included in a datastack JSON for an ArcGIS GRID would be unusable.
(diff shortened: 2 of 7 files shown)


src/natcap/invest/hra.py
```diff
@@ -222,321 +222,327 @@
...
-            contents=[
-                spec.SingleBandRasterOutput(
-                    id="aligned_[HABITAT/STRESSOR/CRITERIA].tif",
-                    about=gettext(
-                        "Copy of the input, aligned to the same projection and extent"
-                    ),
-                    data_type=float,
-                    units=u.none
+        spec.SingleBandRasterOutput(
+            id="RECLASS_RISK_[HABITAT]",
```
emlys: I'm not sure why these output file names were written in all caps, but I think it would make sense to standardize all ids to lowercase.

src/natcap/invest/ndr/ndr.py
```diff
@@ -348,292 +352,334 @@
...
-                spec.FileOutput(
-                    id="surface_load_p.pickle",
-                    about=gettext("Pickled zonal statistics of phosphorus surface load")
-                ),
-                spec.FileOutput(
-                    id="total_export_n.pickle",
-                    about=gettext("Pickled zonal statistics of total nitrogen export")
-                )
-            ]
+        spec.STREAM,
```
emlys, line 364: Does this need to be updated with a path?

tests/test_spec.py
```diff
@@ -275,11 +275,15 @@ def _generate_files_from_spec(output_spec, workspace):
     """A utility function to support the metadata test."""
     for spec_data in output_spec:
         if spec_data.__class__ is spec.DirectoryOutput:
```
emlys: Can we remove this conditional now that DirectoryOutputs are removed?

tests/test_spec.py
```diff
@@ -380,7 +393,7 @@ def test_write_metadata_for_outputs(self):
         spec.generate_metadata_for_outputs(model_module, args_dict)
 
         files, messages = geometamaker.validate_dir(self.workspace_dir)
-        self.assertEqual(len(files), 3)
+        self.assertEqual(len(files), 4)
```
emlys, line 388: Where is the extra file coming from?

src/natcap/invest/spec.py
```diff
@@ -1726,28 +1727,31 @@ def fallback_serializer(obj):
 )
 
 # Specs for common outputs ####################################################
-TASKGRAPH_DIR = DirectoryOutput(
+TASKGRAPH_DIR = FileOutput(
```
emlys: Can we rename this variable to reflect that it's no longer representing a directory (maybe `TASKGRAPH_DB` or `TASKGRAPH_CACHE`)?

src/natcap/invest/hra.py
emlys: Possibly with an exception for the variable portion in square brackets?

emlys (changes requested):
Thanks @megannissel! In addition to a few specific comments above, I am wondering generally what you think about this design of having both `id` and `path` attributes vs. other possible designs. In nearly every case, the `id` is the file name without the extension, which creates some redundancy. You could store the same information, for example:

```
  spec.SingleBandRasterOutput(
      id="eff_p",
      path="intermediate_outputs/eff_p.tif"
```
or 
```
  spec.SingleBandRasterOutput(
      id="eff_p",
      location="intermediate_outputs",
      format="GeoTIFF"
```
or even
```
  spec.SingleBandGeoTIFFOutput(
      id="eff_p",
      location="intermediate_outputs"
```

which would remove the redundancy between `id` and `path`. But it does rely on knowing the file extension that corresponds to the file type. Do you think there's value in doing something like that, or should we stick with this approach that solves our immediate use case?

megannissel:
Thanks, @emlys! I agree that with this approach, we have some redundancy with `id` and `path`. I did consider something like your second suggestion, 
```
  spec.SingleBandRasterOutput(
      id="eff_p",
      location="intermediate_outputs",
      format="GeoTIFF"
```
but landed on being explicit about the `path` for two reasons:
- The `id` doesn't _have_ to be the filename. Thinking ahead to one potential way of building the file registry (where we use the `id` as the key and full path constructed from `path` as the value), this leaves room for us to use shorter, easier-to-reference `id`s than what we want the filename to be. The current state of `id`s being so redundant was largely a byproduct of the migration to separate attributes, but there are certainly `id`s I'd be in favor of standardizing / shortening. As you pointed out, there are some instances of strange capitalization in our current file-naming for some models, for instance; I'm also not in love with `id`s that have brackets in them (but that might just be a personal preference).
- I personally think being explicit about the `path` is a strength. Looking at the model spec, you know _exactly_ where a given output file should be located, and what extension it should have. I could see this level of explicitness being helpful for some of our future Project Arches-related development, as well as for the User's Guide integration. 

All that said, I wouldn't be opposed to going with something more like example 2. I would hesitate to go so far as example 3, though, since trying to infer the file type from the output class itself could cause problems; both `.shp` and `.geojson` are valid vector extensions, for instance, and there wouldn't be a clear way to determine which it is / should be.

tests/test_spec.py
megannissel: Do we want to remove `DirectoryOutput` as a class entirely? I haven't done so -- I wanted to see if we can foresee any potential use-cases, first -- and if the class stays, I wanted to make sure we were continuing to test a model spec that includes files nested in this way. If we do want to remove the class entirely, this can definitely go!

tests/test_spec.py
megannissel: I made the model spec we're testing against a little more complex (added an additional intermediate raster).

src/natcap/invest/ndr/ndr.py
megannissel: I had given these reusable outputs default ids and paths, and in this case the default path matches the desired path (just `stream.tif` within the main workspace directory). However, now I'm rethinking that approach in favor of always being explicit about the path within each individual model spec.

tests/test_spec.py
dcdenu4: IMO this redesign removes the major use case for the DirectoryOutput class and we should feel free to remove it. I'd prefer we keep the code lean and add any functionality back in if the use case does arise again in the future.

dcdenu4 (changes requested):
Hey @megannissel thanks for all this! I don't have much to add but do suggest we go ahead and remove the DirectoryOutput class entirely. Hopefully that simplifies some code, some tests, and keeps the code lean. We can always add back in the functionality in a future use case, but I think we're explicitly removing THE current use case with this refactor.

src/natcap/invest/spec.py
```diff
@@ -2139,21 +2122,16 @@ def generate_metadata_for_outputs(model_module, args_dict):
 
     def _walk_spec(output_spec, workspace):
```
megannissel, line 2123: Hey @davemfish, wanted to get your opinion on something. If we're removing `DirectoryOutput` as a class entirely in this PR, and therefore removing the handling of them in `generate_metadata_for_outputs`, does this seem like a big enough breaking change, backwards-compatibility-wise, that this should be included in a minor release rather than a bugfix release? Or do you think adoption beyond NatCap is likely small enough at this point that it isn't a big concern?

src/natcap/invest/ndr/ndr.py
megannissel: You sold me on avoiding additional redundancy. Since this is a top-level output, and the filename and id match the default, it shouldn't need any update here!

src/natcap/invest/spec.py
davemfish: Hey @megannissel , I think it's pretty safe to make these changes in a bugfix release. I can't imagine there are any other users of `natcap.invest.spec.generate_metadata_for_outputs` besides us. Thanks for asking!

emlys (approved):
Thanks, @megannissel, this will be really useful to have especially as I'm working on the compute API!

PR #2090, opened 2025-08-04
Fix missing validation warnings for boolean types in Workbench

### LINKED ISSUE #2074: NDR UI appears valid but Run button is not enabled.
## Description
NDR UI appears valid but Run button is not enabled.

## To replicate
Load the sample data (or any datastack) and toggle off `Calculate Phosphorous` and `Calculate Nitrogen`. No validation messages appear anywhere in the UI, but there is one:
```
Validation warnings: [(['calc_n', 'calc_p'], 'Either calc_n or calc_p must be True')]
[1] 07/21/2025 09:54:41  natcap.invest.ui_server DEBUG    [(['calc_n', 'calc_p'], 'Either calc_n or calc_p must be True')]
```

So, it's correct that the Run button is disabled, but this validation messages needs to appear in the UI. Do our toggle switches not support the alert text where messages usually appear?

dcdenu4:
## Description
Boolean or Switch type validation was not set like the rest of the Workbench inputs. Specifically boolean types were excluded from receiving validation messages and also did not include `isValid` or `isInvalid`. Also adding `onFocus` to boolean types so that they benefit from `touched` being updated like the other inputs.

I don't quite have the testing working properly and could use some feedback for testing this. But, now, if a switch / boolean is required and unchecked it should show a red circle outline around the switch. When touching and unselecting, if it's invalid it should display the validation message.

NOTE: currently when not valid, there is some CSS markup that pushes the switch left, which doesn't look great. Could use some quick tips on where to make this adjustment.

Fixes #2074

image

image

## Checklist

### DIFF
--- a/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -218,6 +218,8 @@ export default function ArgInput(props) {
       />
     );
   }
+  // show validation message for boolean (switch) types regardless of touched
+  // because touching them necessarily changes their state.
   else if (validationMessage && argSpec.type === 'boolean') {
     feedback = (
       <Feedback
@@ -227,7 +229,6 @@ export default function ArgInput(props) {
       />
     );
   }
-
 
   let fileSelector = <React.Fragment />;
   if (['csv', 'vector', 'raster', 'directory', 'file'].includes(argSpec.type)) {
@@ -260,8 +261,8 @@ export default function ArgInput(props) {
         name={argkey}
         checked={value}
         onChange={() => updateArgValues(argkey, !value)}
-	isValid={enabled && isValid}
-	isInvalid={enabled && validationMessage}
+        isValid={enabled && isValid}
+        isInvalid={enabled && validationMessage}
         disabled={!enabled}
         bsCustomPrefix="form-switch"
       />
--- a/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -209,7 +209,7 @@ export default function ArgInput(props) {
   const className = enabled ? null : 'arg-disable';
 
   let feedback = <React.Fragment />;
-  if (validationMessage && touched && argSpec.type !== 'boolean') {
+  if (validationMessage && touched) {
     feedback = (
       <Feedback
         argkey={argkey}
--- a/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -250,6 +250,8 @@ export default function ArgInput(props) {
         name={argkey}
         checked={value}
         onChange={() => updateArgValues(argkey, !value)}
+	isValid={enabled && touched && isValid}
+	isInvalid={enabled && validationMessage}
         disabled={!enabled}
         bsCustomPrefix="form-switch"
       />
--- a/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -209,7 +209,7 @@ export default function ArgInput(props) {
   const className = enabled ? null : 'arg-disable';
 
   let feedback = <React.Fragment />;
-  if (validationMessage && touched) {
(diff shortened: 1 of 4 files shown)


dcdenu4:
@davemfish, I simplified the test and got it to pass. I also defaulted to showing the validation message for switches without needing "touched".

workbench/tests/renderer/setuptab.test.js
```diff
@@ -1,975 +1,1034 @@
...
+    });
+    const input1Group = input1.closest('.input-group');
+    await waitFor(() => {
+      expect(within(input1Group).queryByText(RegExp(VALIDATION_MESSAGE)))
+        .toBeInTheDocument();
+    });
+
+    // An optional input with no value is valid, but green check
+    // does not display until the input has been touched.
+    expect(input2).not.toHaveClass('is-valid', 'is-invalid');
```
davemfish: I don't think `touched` matters anymore for boolean inputs, right? So I think this comment could use updating, and I think I would `expect(input2).toHaveClass('is-valid')`

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -218,6 +218,16 @@ export default function ArgInput(props) {
       />
     );
   }
+  else if (validationMessage && argSpec.type === 'boolean') {
```
davemfish, line 223: Maybe we could add a comment here about why we're treating `boolean` separately? Like, we want to display the validation feedback for switches even before they are touched, because touching them necessarily changes their state.

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -218,6 +218,16 @@ export default function ArgInput(props) {
...
+    feedback = (
+      <Feedback
+        argkey={argkey}
+        argtype={argSpec.type}
+        message={validationMessage}
+      />
+    );
+  }
+
```
davemfish, line 232: extra newline here?

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -250,6 +260,8 @@ export default function ArgInput(props) {
         name={argkey}
         checked={value}
         onChange={() => updateArgValues(argkey, !value)}
+	isValid={enabled && isValid}
+	isInvalid={enabled && validationMessage}
```
davemfish: Forgive me, but since I was looking at this in my text editor, I noticed these two lines appear to be indented with a single tab instead of 8 spaces. I'm not sure why the indent looks correct here, and I'm guessing it looks correct in your editor too, but do you mind double-checking the indent characters?

workbench/src/renderer/styles/style.css
```diff
@@ -605,7 +605,8 @@ input[type=text]::placeholder {
 /* Default sized toggle switch too small */
 .form-switch {
   transform:scale(1.5);
-  margin-left: 1rem;
+  transform-origin:0%;
+  /*margin-left: 1rem;*/
```
davemfish: Safe to remove the commented line I think

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
dcdenu4: Done: https://github.com/natcap/invest/pull/2090/commits/11199ceff12127207350f1ae19d12df4f95abf6c

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
dcdenu4: Should be good now.

workbench/tests/renderer/setuptab.test.js
dcdenu4: Yep, comment should updated. I've struggled with getting the right `is-valid / is-invalid` flags set. Because mocking validation with warning messages for both args, I THINK, will mean this will show up as `is-invalid`, because there's a warning message tied to the input.

dcdenu4: I moved this particular check to after we remove the mocking of validaiton messages.

dcdenu4:
> Do you know why the entire test file appears to be changed?

I think this has to do with line endings. I have my settings set to convert CRLF (Windows) to LF on commit... I'm not sure if this file therefore was committed as CRLF when I pulled it? That or something else funky happened with line endings... I tried reverting the file, re-committing it, and then updating it again. But, got the same result.

PR #2094, opened 2025-08-05
Define AUTORELEASE_BRANCH for scheduled releases

### LINKED ISSUE #2093: Scheduled autorelease workflow fails on autorelease branch creation
During the workflow run, `AUTORELEASE_BRANCH` is defined at the top in `env` section using `inputs.version`. But, when _scheduled_, we don't have `inputs.version`, so `AUTORELEASE_BRANCH` ends up being `autorelease/`. When creating that branch we get an error. I think we can just update the `AUTORELEASE_BRANCH` variable when we get the next bugfix version automatically.

https://github.com/natcap/invest/actions/runs/16758197214/job/47446294330#step:8:1

emlys:
## Description
Fixes #2093

Define `AUTORELEASE_BRANCH` using `$VERSION` rather than `inputs.version` so that it's defined for both scheduled and manually triggered runs.

Bumped the cron schedule to Thursdays to give us time to review, and since I'll be out tomorrow.

## Checklist

### DIFF
--- a/.github/workflows/release-part-1.yml
+++ b/.github/workflows/release-part-1.yml
@@ -6,7 +6,7 @@ on:
     # cron doesn't support "first tuesday of the month", so we will use github
     # actions syntax below to skip tuesdays that aren't in the first week.
     # it is recommended not to start on the hour to avoid peak traffic
-    - cron: "20 18 * * 2"
+    - cron: "20 18 * * 4"
   workflow_dispatch:
     inputs:
       version:


.github/workflows/release-part-1.yml
```diff
@@ -6,7 +6,7 @@ on:
     # cron doesn't support "first tuesday of the month", so we will use github
     # actions syntax below to skip tuesdays that aren't in the first week.
```
emlys, line 7: I think we should change this back to Tuesdays afterward, so I didn't update the comment.

.github/workflows/release-part-1.yml
```diff
@@ -57,11 +56,12 @@ jobs:
           with open(os.environ['GITHUB_ENV'], "a") as env_file:
             env_file.write(f'VERSION={major}.{minor}.{int(bugfix) + 1}')
 
-      - name: Configure git
+      - name: Configure git and define branch name
         if: env.RUN == 'true'
         run: |
           git config user.name "GitHub Actions"
           git config user.email "<>"
+          echo "AUTORELEASE_BRANCH=autorelease/$VERSION" > $GITHUB_ENV
```
dcdenu4, line 64: Okay, so updating the `env_file` above during a scheduled run will have updated our `$VERSION` env variable here.

PR #2120, opened 2025-08-14
More robust logic to parse release notes from HISTORY.rst

### LINKED ISSUE #2119: Fix release notes on GitHub release pages
It looks like our release process has not been properly setting release notes on the [GitHub release pages](https://github.com/natcap/invest/releases).

We should fix this in our autorelease process and manually edit releases that have already gone out.

emlys:
## Description
Fixes #2119 

Instead of using blank lines to identify the end of the section, print lines from the version string until the next line that looks like a header.

I have tested the python script locally, but not in the workflow environment.

## Checklist

### DIFF
--- a/.github/workflows/release-part-2.yml
+++ b/.github/workflows/release-part-2.yml
@@ -107,7 +107,8 @@ jobs:
                       if line.startswith(os.environ['VERSION']):
                           do_write = True
                       # match the header pattern X.XX.X (XXXX-XX-XX)
-                      elif re.match(r'\d+\.\d+\.\d+ \(\d{4}-\d{2}-\d{2}\)', line):
+                      # using \w to allow for alpha and post releases
+                      elif re.match(r'\d+\.\d+\.\w+ \(\d{4}-\d{2}-\d{2}\)', line):
                           do_write = False
                       if do_write:
                           notes.write(line)


.github/workflows/release-part-2.yml
```diff
@@ -93,12 +93,27 @@ jobs:
...
+          with open('notes.rst', 'w') as notes:
+              notes.write('This release includes the following fixes and features:\n\n')
+              with open('HISTORY.rst') as f:
+                  do_write = False
+                  for line in f.readlines():
+                      # match the header starting with $VERSION
+                      if line.startswith(os.environ['VERSION']):
+                          do_write = True
+                      # match the header pattern X.XX.X (XXXX-XX-XX)
+                      elif re.match(r'\d+\.\d+\.\d+ \(\d{4}-\d{2}-\d{2}\)', line):
```
dcdenu4: I think we'll need to add some additional logic for alpha releases. Maybe something like

`r'\d+\.\d+\.\d+\w?\d? \(\d{4}-\d{2}-\d{2}\)'`

emlys: Good suggestion. I think `\w` includes letters and numbers, so I replaced the `\d+` with `\w+`.

PR #2129, opened 2025-08-25
Refactor habitat quality aligned lulc outputs to have static names

### LINKED ISSUE #2127: Habitat Quality aligned lulc output names could be simplified
Habitat Quality uses up to 3 different LULC inputs: `lulc_cur_path`, `lulc_fut_path`, `lulc_bas_path`. It creates an aligned version of each LULC raster that is provided. The aligned raster paths are named after the filename of the original file. For instance, if `lulc_cur_path` is `foo.tif`, a file named `foo_aligned.tif` will be created in the intermediate output directory. 

This makes the model spec and file naming logic unnecessarily complicated. Because there are only 3 possible LULC rasters, it would be simpler to use static output names (eg. the aligned version of `lulc_cur_path` would be called `aligned_lulc_cur.tif` regardless of the input file name). I suggest that we replace the `[LULC]_aligned` output with three statically-named outputs (`aligned_lulc_cur`, `aligned_lulc_fut`, `aligned_lulc_bas`).

emlys:
## Description
Fixes #2127 

Replaced the dynamically-named aligned LULC outputs with statically-named `lulc_cur_aligned`, `lulc_fut_aligned`, `lulc_bas_aligned`. This led to some light refactoring of how we track the file paths, which I was able to simplify quite a bit. No functional changes here. The user guide does not document these intermediate outputs, so nothing needs updating there.

## Checklist

### DIFF
--- a/.github/workflows/release-part-1.yml
+++ b/.github/workflows/release-part-1.yml
@@ -27,12 +27,11 @@ jobs:
       # skip scheduled runs that aren't the first tuesday of the month
       - name: Check day of month
         run: |
-          #if [[ $GITHUB_EVENT_NAME == "workflow_dispatch" || $(date +%d) -le 9 ]]; then
-          #  echo "RUN=true" >> $GITHUB_ENV
-          #else
-          #  echo "RUN=false" >> $GITHUB_ENV
-          #fi
-          echo "RUN=true" >> $GITHUB_ENV
+          if [[ $GITHUB_EVENT_NAME == "workflow_dispatch" || $(date +%d) -le 7 ]]; then
+            echo "RUN=true" >> $GITHUB_ENV
+          else
+            echo "RUN=false" >> $GITHUB_ENV
+          fi
 
       - uses: actions/checkout@v4
         if: env.RUN == 'true'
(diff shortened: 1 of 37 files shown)


src/natcap/invest/habitat_quality.py
```diff
@@ -433,7 +449,7 @@
 # Scaling parameter from User's Guide eq. 4 for quality of habitat
 _SCALING_PARAM = 2.5
 # To help track and name threat rasters from paths in threat table columns
-_THREAT_SCENARIO_MAP = {'_c': 'cur_path', '_f': 'fut_path', '_b': 'base_path'}
+_THREAT_SCENARIO_MAP = {'cur': 'cur_path', 'fut': 'fut_path', 'bas': 'base_path'}
```
emlys, line 452: I think these should be standardized (change "base_path" to "bas_path" to match "cur" and "fut", or use the full word for all of them), but I'm not sure if it's worth renaming columns at this point.

megannissel: I agree; standardizing would be nice, but it's probably not worth the potential for confusion / user error (especially if we were to make the very subtle shift from `base_path` to `bas_path`).

Issue #1905, opened 2025-05-02
Wind Energy: Workbench conditionals validation logic doesn't match tooltips


megannissel:
## Description
The Wind Energy model has a few inputs that are conditionally required: 
- Land Polygon
- Minimum Distance
- Maximum Distance

The Land Polygon tooltip says "Required if the Minimum Distance and Maximum Distance inputs are provided." However, this field only becomes editable if an AOI has been selected.

The Minimum Distance and Maximum Distance tooltips say "Required if Run Valuation is selected," but they only become editable if a Land Polygon has been selected -- and don't become editable if Run Valuation is toggled to True.

Based on the tooltip wording, a user would assume toggling "Run Valuation" to True would make these fields editable; instead, it's selecting an AOI that makes Land Polygon editable, and selecting a Land Polygon makes Min and Max Distance editable.  

Additionally, the tooltip for the AOI says "Required if Run Valuation is selected _and the Grid Connection Points table is provided_," but the workbench won't let you run the model unless the AOI is provided (but also doesn't show any validation errors). Running the model from the command line without the Grid Connection Points table and without an AOI causes the model to crash (the model looks for information based on the Land Polygon, but the Land Polygon is processed only if an AOI is present). 

It seems that the AOI should always be required if Run Valuation is selected. And the tooltips for Land Polygon, Min Distance, and Max Distance should be updated to follow the actual requirements structure per the model's functionality.

<img width="1441" alt="Image" src="https://github.com/user-attachments/assets/0e4bd15d-7b50-4863-a28f-7468a3488c59" />

megannissel:
The remaining problems here were addressed by #2135

PR #2131, opened 2025-09-02
Implement file registry and use in all models

emlys:
## Description
Fixes #

## Checklist
- [ ] Updated HISTORY.rst and link to any relevant issue (if these changes are user-facing)
- [ ] Updated the user's guide (if needed)
- [ ] Tested the Workbench UI (if relevant)

PR #2135, opened 2025-09-04
Wind Energy: Always Require AOI, Land Polygon, Min and Max Distances

### LINKED ISSUE #1944: Future Wind Energy Enhancements
Circle back with Rob about additional Wind Energy enhancements, including this point from [User's Guide PR #178](https://github.com/natcap/invest.users-guide/pull/178):

> For foundation-based wind farms, the distance/depth constraint should apply to all outputs, as they are all conditional on actually being able to build a farm there. For floating wind farms, a distance constraint is probably still warranted. In which case an AOI and land polygon should generally be required inputs for all model runs. I could see a check box for floating/traditional foundations and then requirements conditional on that for the min/max distance and depth.

Further User's Guide updates; from Rob via Slack:

> Looking forward, it would be great to update the sample data. With the newly released paper, it might be worth using that exact case study, as folks would then have that paper as a reference too.

megannissel:
## Description
Fixes #1944 and #1905

Following the updates to the Wind Energy model in #1898, Rob had recommended some additional updates to the model. These included updating the sample data and also making the AOI, Land Polygon, and Min/Max Distances requirements for all model runs. Since the model should only provide outputs for locations where a wind farm could feasibly exist, it makes sense to always account for distance-from-shore constraints. As an added benefit, this change simplifies the model's logic quite a bit. 

Corresponding updates to the User's Guide reflecting these changes can be found [here](https://github.com/natcap/invest.users-guide/pull/199). 

Additional User's Guide updates incorporating @rmgriffin's new sample data are forthcoming, but aren't strictly required for these changes to be included in the 3.17.0 release. 

## Checklist

### DIFF
--- a/src/natcap/invest/wind_energy.py
+++ b/src/natcap/invest/wind_energy.py
@@ -897,20 +897,20 @@ def execute(args):
             args['bathymetry_path'])['pixel_size']
         mean_pixel_size, _ = utils.mean_pixel_size_and_area(bathy_pixel_size)
         target_pixel_size = (mean_pixel_size, -mean_pixel_size)
-        LOGGER.debug('Target pixel size: %s' % (target_pixel_size,))
+        LOGGER.debug(f'Target pixel size: {target_pixel_size}')
         bathymetry_path = args['bathymetry_path']
         # The task list would be empty for clipping and reprojecting bathymetry
         bathy_dependent_task_list = None
 
     except ValueError:
         LOGGER.debug(
-            '%s has pixels that are not square. Resampling the raster to have '
-            'square pixels.' % args['bathymetry_path'])
+            f"{args['bathymetry_path']} has pixels that are not square. "
+            "Resampling the raster to have square pixels.")
         bathymetry_path = file_registry['bathymetry_path']
         mean_pixel_size = numpy.min(numpy.absolute(bathy_pixel_size))
         # Use it as the target pixel size for resampling and warping rasters
         target_pixel_size = (mean_pixel_size, -mean_pixel_size)
-        LOGGER.debug('Target pixel size: %s' % (target_pixel_size,))
+        LOGGER.debug(f'Target pixel size: {target_pixel_size}')
 
         resample_bathymetry_task = task_graph.add_task(
             func=pygeoprocessing.warp_raster,
@@ -937,10 +937,9 @@ def execute(args):
     parameters_dict = global_params_dict.copy()
     parameters_dict.update(turbine_dict)
 
-    LOGGER.debug('Biophysical Turbine Parameters: %s', parameters_dict)
+    LOGGER.debug(f'Biophysical Turbine Parameters: {parameters_dict}')
 
-    if ('valuation_container' not in args or
-            args['valuation_container'] is False):
+    if ('valuation_container' not in args or args['valuation_container'] is False):
         LOGGER.info('Valuation Not Selected')
         run_valuation = False
     else:
@@ -959,8 +958,8 @@ def execute(args):
             if year_count != time + 1:
                 raise ValueError(
                     "The 'time' argument in the Global Wind Energy Parameters "
-                    "file must equal the number of years provided in the price"
-                    " table.")
+                    "file must equal the number of years provided in the price "
+                    "table.")
 
             # Save the price values into a list where the indices of the list
             # indicate the time steps for the lifespan of the wind farm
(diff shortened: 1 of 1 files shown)


src/natcap/invest/wind_energy.py
```diff
@@ -364,7 +360,7 @@
                 spec.NumberInput(
                     id="turbine_rated_pwr",
                     about="The turbine's rated power output.",
-                    units=u.kilowatt
+                    units=u.megawatt
```
dcdenu4, line 361: Should we add a quick history note for this?
>Units for Turbine Rated Power were incorrectly listed as kilowatt when the model expected megwatt

Or something.

src/natcap/invest/wind_energy.py
```diff
@@ -827,15 +820,12 @@ def execute(args):
             projected in linear units of meters. The polygon specifies the
             area of interest for the wind data points. If limiting the wind
             farm bins by distance, then the aoi should also cover a portion
-            of the land polygon that is of interest (optional for biophysical
-            and no distance masking, required for biophysical and distance
-            masking, required for valuation)
+            of the land polygon that is of interest (required)
```
dcdenu4, line 821: Thanks for remembering to update this docstring!

src/natcap/invest/wind_energy.py
```diff
@@ -1005,155 +991,91 @@ def execute(args):
...
+
+    # The wind energy points shapefile constructed from the dictionary will
+    # need to be clipped and projected to the AOI, so save it as an intermediate
+    # file. The final clipped and projected shapefile will be written to the
+    # output directory.
+    wind_point_vector_path = os.path.join(
+        inter_dir, 'wind_energy_points_from_data%s.shp' % suffix)
+
+    # Use the projection from the projected bathymetry as reference to
+    # create wind point vector from wind data dictionary
```
dcdenu4: Outdated? Should this refer to the AOI instead of bathymetry?

src/natcap/invest/wind_energy.py
```diff
@@ -1005,155 +991,91 @@ def execute(args):
...
+
+    # Create point shapefile from wind data
+    LOGGER.info('Create point shapefile from wind data')
+
+    # The wind energy points shapefile constructed from the dictionary will
+    # need to be clipped and projected to the AOI, so save it as an intermediate
+    # file. The final clipped and projected shapefile will be written to the
+    # output directory.
+    wind_point_vector_path = os.path.join(
+        inter_dir, 'wind_energy_points_from_data%s.shp' % suffix)
```
dcdenu4: I know this is a bit out of scope, but since we're in here, could you update the use of `%` string replacement with `f'{}` replacement? We've been trying to move in this direction generally.

src/natcap/invest/wind_energy.py
```diff
@@ -1005,155 +991,91 @@ def execute(args):
...
+        func=_clip_vector_by_vector,
+        args=(wind_point_vector_path, aoi_vector_path,
+              clipped_wind_point_vector_path, inter_dir),
+        target_path_list=[clipped_wind_point_vector_path],
+        task_name='clip_wind_point_by_aoi',
+        dependent_task_list=[wind_data_to_vector_task])
+
+    # Handle distance constraints:
+    min_distance = float(args['min_distance'])
+    max_distance = float(args['max_distance'])
```
dcdenu4, line 1019: These are required now, so we don't need any error handling.

src/natcap/invest/wind_energy.py
dcdenu4: Actually you might want to wait on this, because I think the updated File Registry PR will be merged first. And so these paths might go away.

dcdenu4 (changes requested):
Thanks @megannissel . I had a few suggestions and then think we should hold off before merging. I want to merge the [File Registry PR](https://github.com/natcap/invest/pull/2131) before this one, and I think unfortunately that means this PR will need a bit of refactoring... Hopefully it's not too bad!

We'll also want to make sure we get the UG changes lined up too.

Happy to chat about this more if needed.

src/natcap/invest/wind_energy.py
megannissel: Good catch! Yes, this should be AOI now.

src/natcap/invest/wind_energy.py
megannissel: Correct. It seemed most in keeping with the handling of other required inputs to not have any specific error handling here. However, since this is a fundamental change to inputs, I could see an argument for keeping some error handling for the benefit of CLI users (who don't get the workbench's built-in validation).

dcdenu4: I think in the CLI case here this would throw a `KeyError` if an argument was missing, which is a helpful error.

PR #2150, opened 2025-09-15
added a `ModelSpec.get_output` method.

### LINKED ISSUE #2138: `ModelSpec` method for accessing `outputs` by id.
This method does not exist yet. It could follow the pattern of `ModelSpec.get_input`, or we could come up with a better design for both, if there is one.

davemfish:
We're not using this method yet, though it will be useful to us in a feature related to generating model reports. See #2138, I think we will probably want this method even if we redesign the API at some point.

Fixes #2138 

## Checklist


emlys (approved):
Thanks @davemfish!

PR #2151, opened 2025-09-16
Plugin installation error-handling & cleanup

### LINKED ISSUE #2071: Plugins: installation error crashes Workbench
## Description
Installing a plugin from local path with a syntax error in the model spec crashes the Workbench. It brings up the "Something Went Wrong" page in the Workbench.

## To replicate

1. Introduce a syntax error in the Model Spec of a local plugin and try installing said plugin in the Workbench.

## Expected behavior
I thought we were catching errors in the plugin modal but still able to recover the Workbench session.

## Observed behavior
The Workbench brings up the Something Went Wrong page and I have to manually restart InVEST.

## Logfile(s)
[main.log](https://github.com/user-attachments/files/21185913/main.log)

## Screenshots
If applicable, add screenshots to help illustrate the problem.

## Environment
- Operating System: Windows 11 Pro
- InVEST version number: 3.16.1

## Additional context
Please provide any additional details that might help us understand and debug the issue, including what you think the cause might be, and any solutions or workarounds you have tried.

### LINKED ISSUE #2104: Plugins: garbage collect mamba env on plugin installation failure
## Description
When testing a plugin during development I had the installation process fail a few times when trying to install it into the Workbench. The envs that were being created were being left behind and piling up.

## To replicate

1. Try to install a plugin that has a syntax error or is non conformant in some way.
2. Look at the env left behind.

## Expected behavior
During a plugin installation failure I expect the created env to be cleaned up.

## Observed behavior
Env is not cleaned up.

davemfish:
Remove micromamba environments if the plugin installation failed. I created several named functions just to make the main installation function a bit shorter and easier to parse. Then I added the error-handling logic to that function.

Also fixed a bug where we were resolving a `Promise` with an `Error` object instead of throwing the error and catching it. And we needed to convert that `Error` to a string also before rendering it in React.

Fixes #2104 
Fixes #2071 

## Checklist

I tested installing a plugin from a git url and from a local path. And for both, I tested with a plugin that failed to install, and one that succeeded.

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -294,7 +294,9 @@ export function setupAddPlugin(i18n) {
           logger.info('successfully added plugin');
         } catch (error) {
           logger.info('Cleaning up failed installation environment.');
-          fs.rmSync(pluginEnvPrefix, { recursive: true, force: true });
+          await spawnWithLogging(
+            micromamba, ['env', 'remove', '--yes', '--prefix', `"${pluginEnvPrefix}"`]
+          );
           throw error;
         }
       } catch (error) {


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -143,72 +283,23 @@ export function setupAddPlugin(i18n) {
...
+            micromamba, pluginEnvPrefix, condaDeps, installString, i18n, event
+          );
+          event.sender.send('plugin-install-status', i18n.t('Importing plugin...'));
+          storePluginMetadataSync(
+            micromamba, pluginEnvPrefix, packageName, installString
+          );
+          logger.info('successfully added plugin');
+        } catch (error) {
+          logger.info('Cleaning up failed installation environment.');
+          fs.rmSync(pluginEnvPrefix, { recursive: true, force: true });
```
emilyanndavis: Just curious: It looks like when we remove a plugin we're invoking the micromamba command to remove its environment (line 321). Would it be possible to do the same thing here, or is there a reason we need to invoke the file system command directly when plugin installation has failed?

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -57,10 +54,170 @@ function spawnWithLogging(cmd, args, options) {
...
+    `plugins.${pluginID.replaceAll('.', '_')}`,
+    {
+      modelID: modelID,
+      modelTitle: modelTitle,
+      type: 'plugin',
+      source: installString,
+      env: pluginEnvPrefix,
+      version: version,
+    }
+  );
```
emilyanndavis, line 214: What happens if the settings store ends up with invalid data? For example, if a plugin's model spec is malformed, and an attempt to print the model_id results in an error message? Would that result in a failed import?

emilyanndavis: Also, is there any failure scenario (such as, but perhaps not limited to, the one I mentioned above) in which we need to remove the entry from the settings store?

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -301,9 +300,7 @@ export default function PluginModal(props) {
             {
               installLoading ? (
                 <div className="adding-button">
-                  <Spinner animation="border" role="status" size="sm" className="plugin-spinner">
-                    <span className="sr-only">{t('Adding...')}</span>
-                  </Spinner>
+                  <Spinner animation="border" role="status" size="sm" className="plugin-spinner" aria-label="loading"/>
```
emilyanndavis: I don't think the Spinner component supports aria-label, which is why the docs recommend using a visually hidden (or "screenreader-only") element instead ([React-Bootstrap 4: Spinner Accessibility](https://react-bootstrap-v4.netlify.app/components/spinners/#accessibility)). I'm assuming you made this change for testing purposes; maybe this would be a good candidate for adding a `data-testid` attribute and using a [ByTestId](https://testing-library.com/docs/queries/bytestid) query instead?

emilyanndavis (changes requested):
Thanks for taking care of these issues, @davemfish! Just a few questions & suggestions!

workbench/src/main/setupAddRemovePlugin.js
davemfish: Good question. `micromamba env remove` might work too. I'm not sure if it does anything different, other than the fact that it would work with named environments as well as prefixed ones. If there's an exception while creating the environment, and it failed to finish creating in the first place, I'm not sure what `micromamba env remove` would do. While `fs.rmSync` with these options should be pretty reliable: https://nodejs.org/api/fs.html#fsrmsyncpath-options

I just tried manually removing a prefixed env (`rm -rf`) and then checking `micromamba env list` and the environment is no longer listed. So it seems like it might be okay to do it that way. But I'm happy to switch to using `micromamba env remove`.

@emlys do you have any advice?

workbench/src/main/setupAddRemovePlugin.js
davemfish: > What happens if the settings store ends up with invalid data? For example, if a plugin's model spec is malformed, and an attempt to print the model_id results in an error message? Would that result in a failed import?

If there's a python exception during one of these `execSync` processes, then `execSync` will `throw` and we will catch it as a failed installation. (This happened to me many times and was the reason I updated those dependencies in `invest_reports`!)

If `execSync` doesn't throw but returns bad data somehow...I'm not sure if that's possible given our pydantic validation of the `ModelSpec`.

davemfish: > Also, is there any failure scenario (such as, but perhaps not limited to, the one I mentioned above) in which we need to remove the entry from the settings store?

Hmm, not that I can think of. If the plugin installs successfully (storing the settings is the last part of installation) then the user will be able to use the "remove plugin" menu to remove it if they wish.

workbench/src/renderer/components/PluginModal/index.jsx
davemfish: Oops, thanks! I reverted this back. I was trying to pin down the behavior in the test and was incorrectly skeptical about the `sr-only` text.

emilyanndavis (approved):

workbench/src/main/setupAddRemovePlugin.js
davemfish: @emilyanndavis I switched over to the `micromamba` command. I tested it again with a failed installation and it worked fine.

emilyanndavis (approved):

PR #2161, opened 2025-09-22
Don't limit percent inputs to [0, 100] by default

### LINKED ISSUE #2160: Carbon model validation does not accept negative values for rate change
A forum user pointed out that the workbench won't allow you to enter a negative value for the `rate_change` parameter, even though the user guide describes using both positive and negative values. This is happening because the percent input type automatically restricts values to the range [0, 100].

https://community.naturalcapitalproject.org/t/carbon-storage-and-sequestration-model-annual-price-change-value/6202

emlys:
## Description
Fixes #2160 

I removed the default range limit for percent inputs. I reviewed all percent inputs across invest and found a few other that also shouldn't be limited to [0, 100]. For those that should be limited to [0, 100], I added an `expression` to enforce that.

We should separately review all ratio inputs and check if there's any that shouldn't be limited to [0, 1].

## Checklist

### DIFF
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -1102,8 +1102,10 @@ class RatioInput(NumberInput):
 class PercentInput(NumberInput):
     """A percent input, or parameter, of an invest model.
 
-    A percent is a proportion expressed as a value from 0 to 100 (in contrast to
-    a ratio, which ranges from 0 to 1). Values are restricted to the range [0, 100].
+    A percent is a proportion expressed as a value from 0% to 100% (in contrast
+    to a ratio, which ranges from 0 to 1). By default there is no restriction on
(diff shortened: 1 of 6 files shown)


src/natcap/invest/spec.py
```diff
@@ -1121,11 +1121,6 @@ def validate(self, value):
         message = super().validate(value)
         if message:
             return message
```
emilyanndavis, line 1125: Probably worth updating the docstring for the PercentInput class, now that values are _not_ necessarily restricted to the range [0, 100].

emilyanndavis (changes requested):
Just one small request!

src/natcap/invest/spec.py
emlys: Good catch!

emilyanndavis (approved):
👍🏻

PR #2163, opened 2025-09-22
Model input preprocessing

### LINKED ISSUE #1451: Abstraction for model input pre-processing
In #1419 we introduced standard pre-processing for CSV inputs with `get_validated_dataframe`. 
`csv` inputs are parsed into a `pandas.DataFrame` with the appropriate structure and dtypes.

We pre-process some other inputs too: `boolean`, `integer`, `number`, `ratio`, and `percent` inputs are cast to the appropriate python type (`bool`, `int`, `float`). Some inputs have default values if none is provided.

The other types usually aren't pre-processed, but there are possible use cases:
- cast an option string to an enum value
- expand file paths; handle python `Path` objects
- open a vector as a `geopandas.GeoDataFrame`

An abstraction for pre-processing all model inputs would deduplicate code and help standardize our models. This could look something like
```
def execute(args):
    args = validation.preprocess_args(args, MODEL_SPEC)
    ...
```

or as a decorator:
```
@validation.preprocess_args(MODEL_SPEC)
def execute(args):
    ...
```

There's a lot of overlap between pre-processing and validation, so it might make sense to combine them. For instance, if the `n_workers` parameter can't be cast to an integer, that's a validation issue.

```
try:
    args['n_workers'] = int(args['n_workers'])
except ValueError:
    raise validation message
return args
```

emlys:
## Description
Fixes #1451

## Checklist


src/natcap/invest/cli.py
```diff
@@ -465,46 +465,23 @@ def main(user_args=None):
             model_module = importlib.import_module(name=target_model)
             LOGGER.info('Imported target %s from %s',
                         model_module.__name__, model_module)
-            for arg_key, val in parsed_datastack.args.items():
```
emlys, line 468: This block is replaced by path normalization that now happens in the Input.preprocess methods

src/natcap/invest/datastack.py
```diff
@@ -96,34 +95,6 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
         safe_extract(tar, dest_dir_path)
 
 
-def format_args_dict(args_dict, model_id):
```
emlys, line 99: Moved to utils to avoid circular import.

src/natcap/invest/validation.py
```diff
@@ -66,51 +65,10 @@
     'WRONG_GEOM_TYPE': gettext('Geometry type must be one of {allowed}')
 }
 
+
 def get_message(key):
     return gettext(MESSAGES[key])
 
-def _evaluate_expression(expression, variable_map):
```
emlys, line 72: Moved to utils to avoid circular import.

tests/test_cli.py
```diff
@@ -126,34 +127,6 @@ def test_run_coastal_blue_carbon_invalid_datastack(self):
             ])
         self.assertEqual(exit_cm.exception.code, 1)
 
-    def test_run_carbon_remote_files(self):
```
emlys, line 129: The remote path functionality no longer lives in cli.py, so it should be tested elsewhere.

src/natcap/invest/hra.py
```diff
@@ -1129,85 +1089,77 @@ def execute(args):
...
     # refactoring it out (should we end up doing more such visualizations).
     LOGGER.info('Generating visualization outputs')
-    visualization_dir = os.path.join(args['workspace_dir'],
-                                     'visualization_outputs')
-    utils.make_directories([visualization_dir])
     shutil.copy(  # copy in the summary table.
         summary_csv_path,
         file_registry['summary_statistics_viz'])
 
-    # For each raster in reclassified risk rasters + Reclass ecosystem risk:
```
emlys, line 1150: Some light refactoring here to use the file registry more directly

src/natcap/invest/ndr/ndr_core.pyx
```diff
@@ -54,6 +54,7 @@ def ndr_eff_calculation(
         suffix='.tif', prefix='flow_to_process',
         dir=os.path.dirname(effective_retention_path))
     os.close(fp)
+    algorithm = algorithm.lower()
```
emlys, line 57: Ensuring that the algorithm is case-insensitive in these functions, since pre-processing doesn't happen here

src/natcap/invest/spec.py
```diff
@@ -457,25 +492,41 @@ def validate(self, filepath: str):
         Returns:
             A string error message if an error was found.  ``None`` otherwise.
         """
-        gdal_path = utils._GDALPath.from_uri(filepath)
-        if gdal_path.is_local:
-            file_warning = super().validate(filepath)
-            if file_warning:
-                return file_warning
+        with GDALUseExceptions():
```
emlys, line 507: Adding the GDAL exceptions context manager to some functions that should probably have it, just in case.

src/natcap/invest/utils.py
```diff
@@ -126,25 +131,6 @@ def capture_gdal_logging():
         gdal.PopErrorHandler()
 
 
-@contextlib.contextmanager
-def _set_gdal_configuration(opt, value):
-    """Temporarily set a GDAL configuration option.
```
emlys, line 131: This function was unused.

src/natcap/invest/utils.py
```diff
@@ -175,23 +161,22 @@ def prepare_workspace(
     with capture_gdal_logging(), log_to_file(logfile,
                                              exclude_threads=exclude_threads,
                                              logging_level=logging_level):
-        with sandbox_tempdir(dir=workspace):
```
emlys, line 178: This context manager wasn't doing anything here. It yields a temporary directory path, which we weren't assigning to a variable or using in any way.

src/natcap/invest/utils.py
```diff
@@ -286,135 +271,6 @@ def log_to_file(logfile, exclude_threads=None, logging_level=logging.NOTSET,
         root_logger.removeHandler(handler)
 
 
-@contextlib.contextmanager
-def sandbox_tempdir(suffix='', prefix='tmp', dir=None):
```
emlys, line 290: This function was only used in one other place, in datastack.py, and it's easily replaced with the standard library `tempfile.TemporaryDirectory`.

src/natcap/invest/utils.py
```diff
@@ -286,135 +271,6 @@ def log_to_file(logfile, exclude_threads=None, logging_level=logging.NOTSET,
...
-        yield sandbox
-    finally:
-        try:
-            shutil.rmtree(sandbox)
-        except OSError:
-            LOGGER.exception('Could not remove sandbox %s', sandbox)
-
-
-def make_suffix_string(args, suffix_key):
-    """Make an InVEST appropriate suffix string.
```
emlys, line 324: This moved into the `ResultsSuffixInput.preprocess` method

src/natcap/invest/utils.py
```diff
@@ -286,135 +271,6 @@ def log_to_file(logfile, exclude_threads=None, logging_level=logging.NOTSET,
...
-        if file_suffix != "" and not file_suffix.startswith('_'):
-            file_suffix = '_' + file_suffix
-    except KeyError:
-        file_suffix = ''
-
-    return file_suffix
-
-
-def build_file_registry(base_file_path_list, file_suffix):
-    """Combine file suffixes with key names, base filenames, and directories.
```
emlys, line 352: This function was unused now that we have the separate `FileRegistry` class.

src/natcap/invest/utils.py
```diff
@@ -979,3 +819,75 @@ def is_remote(self):
     def is_local(self):
         """Test if the path is a local URI"""
         return not self.is_remote
+
+
+def evaluate_expression(expression, variable_map):
+    """Evaluate a python expression.
```
emlys, line 825: Moved this function here unchanged to avoid a circular import.

src/natcap/invest/stormwater.py
```diff
@@ -234,15 +234,15 @@
                 "Map of percolation ratio derived by cross-referencing the LULC and soil"
                 " group rasters with the biophysical table."
             ),
-            created_if="percolation",
+            # created_if="percolation",
```
emlys, line 237: These `created_if` strings are invalid: there is no input called `percolation` so this can't be evaluated. The state of whether percolation results are calculated is determined by whether the percolation columns are provided in the biophysical table. This is a unique situation among invest models where we don't have a corresponding boolean input for that condition. For now I'm commenting these out since they weren't doing anything anyway. We could consider adding a `percolation` boolean input to more clearly define this.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -1148,21 +1086,20 @@ def execute_transition_analysis(args):
         ``None``.
 
     """
-    task_graph, n_workers, intermediate_dir, output_dir, suffix, file_registry = (
-        _set_up_workspace(args))
+    file_registry = args['file_registry']
```
davemfish, line 1096: This is an interesting case with a second execute function. Do you think the docstring, or a comment, should note that `args` have already been pre-processed? Also it looks like we never updated the `Args` docstring to include `file_registry`.

src/natcap/invest/spec.py
```diff
@@ -1238,6 +1415,29 @@ def format_column(col, *args):
...
+    def preprocess(self, value):
+        """Normalize a value to a string.
+
+        Args:
+            value: value to preprocess
+
+        Returns:
+            string or None
+        """
+        return None if value in {None, ''} else str(value)
```
davemfish, line 1427: I'm not sure if it matters in practice, but this one caught my eye because a valid string `''` gets redefined to `None`. Maybe it's good for consistency with the other input types though.

src/natcap/invest/spec.py
```diff
@@ -1631,6 +1848,193 @@ def fallback_serializer(obj):
         spec_dict['outputs'] = {_output.id: _output for _output in self.outputs}
         return json.dumps(spec_dict, default=fallback_serializer, ensure_ascii=False)
 
+    def preprocess_inputs(self, input_values):
+        """Preprocess a dictionary of input values.
+
+        The resulting dict will contain exactly the input keys in the model spec.
+        Inputs which were not provided will have a value of None. Each provided
```
davemfish, line 1855: I think it ends up being "Inputs which were not provided, _or which have a value of '' (empty string)_, will be assigned a value of `None`". Is that right?

src/natcap/invest/spec.py
```diff
@@ -1631,6 +1848,193 @@ def fallback_serializer(obj):
...
+            generate_metadata (bool): Defaults to False. If True, use
+                geometamaker to create metadata files in the workspace
+                after execution completes.
+            save_file_registry (bool): Defaults to False. If True, the
+                file registry dictionary will be saved to the workspace
+                as a JSON file after execution completes.
+            create_logfile (bool): Defaults to False. If True, all logging
+                from the execute function as well as all other pre- and
+                post-processing will be written to a logfile in the workspace.
+            log_level :
```
davemfish: incomplete item here

src/natcap/invest/spec.py
```diff
@@ -1631,6 +1848,193 @@ def fallback_serializer(obj):
...
+                            expression=f'{output.created_if}',
+                            variable_map=preprocessed_args
+                        )
+                    ) is True
+                ])
+                if outputs_to_be_created != set(registry.keys()):
+                    print('Missing outputs:',
+                        outputs_to_be_created - set(registry.keys()))
+                    print('Extra outputs:',
+                        set(registry.keys()) - outputs_to_be_created)
```
davemfish: The docstring says an error will be raised if there are missing/extra outputs. Would you like to change these `print` calls to exceptions?

src/natcap/invest/spec.py
```diff
@@ -1631,6 +1848,193 @@ def fallback_serializer(obj):
...
+                    # I don't think we want to indicate a model failure
+                    self.generate_metadata_for_outputs(preprocessed_args)
+                except Exception as exc:
+                    LOGGER.warning(
+                        'Something went wrong while generating metadata', exc_info=exc)
+
+            # optionally write the file registry dict to a JSON file in the workspace
+            if save_file_registry:
+                with open(os.path.join(preprocessed_args['workspace_dir'],
+                                       'file_registry.json'), "w") as json_file:
```
davemfish: As much as I hate the `results_suffix`, I think we may need to append it to `file_registry.json`. It's possible to have multiple sets of outputs in the same directory.

src/natcap/invest/recreation/recmodel_client.py
davemfish: This logic relies on "hostname" (and "port") being missing from `args`. That's no longer true after preprocessing. I think it could be,
`if args['hostname'] and args['port']:`

emlys: Oh, good catch

src/natcap/invest/spec.py
emlys: Oh yep, that should raise an error

src/natcap/invest/spec.py
emlys: Well yes, but currently the assignment of `''` to `None` happens in each Input class's `preprocess` method, which the following sentence refers to... but this makes me realize that we may as well do that assignment here as well. I'll update the function and the docstring.

src/natcap/invest/spec.py
emlys: Hmm yeah I'm not thinking of any cases where it matters for us. I suppose if there was a specific input where `''` should be preserved or replaced with something else, you could subclass `StringInput` and overwrite the `preprocess` method.

src/natcap/invest/spec.py
emlys: Actually, after thinking about your comment above about empty strings, I think maybe we _don't_ want to do that assignment in this function, to allow for the possibility of someone doing something else with `''` in a custom `preprocess` method.

davemfish: Yes, that makes sense to keep that functionality in the `preprocess` methods.

PR #2169, opened 2025-09-26
Refactor Crop Production models to use CSVInputs in lieu of DirectoryInput

### LINKED ISSUE #2095: Crop Production: add CSV inputs pointing to crop data
## Description
The Crop Production Regression and Percentile models use a collection of standard base data that consists of CSVs and rasters for each supported crop. Currently, the models take in a `model_data` input which is the path to a directory containing the expected data in a specific layout with specific file names. 

This approach has limitations with regard to supporting remote datasets. Pointing to a remote directory is much more complicated than pointing to specific remote files of a known format. 

We discussed on a team call today and agreed that it makes sense to replace the `model_data` directory input with CSV inputs that point to the exact input files for each crop. This will make it simpler to support remote data. It's also a more explicit and flexible design: it spells out exactly which file is used for which crop, and makes it easier to swap out particular files if the user wishes to do so.

For the percentile model, this would look like:
* Add a `CSVInput` with id `percentile_yield_csv_table`. This CSV will map crop names to the corresponding percentile yield table paths.
* Add a `CSVInput` with id `climate_bin_raster_table`. This CSV will map crop names to the corresponding climate bin raster paths.
* Add a `CSVInput` with id `observed_yield_raster_table`. This CSV will map crop names to the corresponding observed yield raster paths.
* Move the `crop_nutrient.csv` `CSVInput` out of the model data `DirectoryInput`
* Remove the model data `DirectoryInput`

For the regression model,
* Add a `CSVInput` with id `regression_yield_csv_table`. This CSV will map crop names to the corresponding regresion yield table paths.
* Add a `CSVInput` with id `climate_bin_raster_table`. This CSV will map crop names to the corresponding climate bin raster paths.
* Add a `CSVInput` with id `observed_yield_raster_table`. This CSV will map crop names to the corresponding observed yield raster paths.
* Move the `crop_nutrient.csv` `CSVInput` out of the model data `DirectoryInput`
* Remove the model data `DirectoryInput`

Each new CSV will have two columns, `crop_name` and `path`.

New sample data will need to be created to support this change.

emilyanndavis:
## Description
Resolves #2095 by refactoring Crop Production models (Percentile and Regression) to use CSVInputs in lieu of a DirectoryInput for model data.

Also renames existing `crop` table headers to `crop_name`, for consistency.

Note the "Validate Sampledata & User Guide" and "Build binaries" checks will fail until the UG PR is merged. 

## Checklist
- [x] Update Crop Production Percentile model
- [x] Update Crop Production Percentile tests (currently working with locally modified test data)
- [x] Update Crop Production Regression model
- [x] Update Crop Production Regression tests
- [x] Update sample data (see [invest-sample-data PR #46](https://bitbucket.org/natcap/invest-sample-data/pull-requests/46))
- [x] Update test data (see [invest-test-data PR #38](https://bitbucket.org/natcap/invest-test-data/pull-requests/38))
- [x] Update HISTORY.rst and link to any relevant issue (if these changes are user-facing)
- [x] Update the user's guide (see [UG PR #201](https://github.com/natcap/invest.users-guide/pull/201))
- [x] Test the Workbench UI (if relevant)

### DIFF
--- a/src/natcap/invest/crop_production_percentile.py
+++ b/src/natcap/invest/crop_production_percentile.py
@@ -660,21 +660,15 @@ def execute(args):
             "The following lucodes are in the landcover raster but aren't "
             f"in the landcover to crop table: {lucodes_missing_from_table}")
 
-    bad_crop_name_list = []
-
-    for crop_name in crop_to_landcover_df.index:
-        crop_climate_bin_raster_path = get_full_path_from_crop_table(
-            MODEL_SPEC,
-            CROP_TO_PATH_TABLES.climate_bin,
-            args[CROP_TO_PATH_TABLES.climate_bin],
-            crop_name)
-        if crop_climate_bin_raster_path is None:
-            bad_crop_name_list.append(crop_name)
-    if bad_crop_name_list:
+    LOGGER.info("Checking that crops are supported by the model.")
+    user_provided_crop_names = set(list(crop_to_landcover_df.index))
+    valid_crop_names = set([crop.key for crop in CROP_OPTIONS])
+    invalid_crop_names = user_provided_crop_names.difference(valid_crop_names)
+    if invalid_crop_names:
         raise ValueError(
-            "'The following crop names were provided in "
-            f"{args['landcover_to_crop_table_path']} but no corresponding "
-            f"climate bin raster path could be found: {bad_crop_name_list}")
+            "The following crop names were provided in "
+            f"{args['landcover_to_crop_table_path']} but are not supported "
+            f"by the model: {invalid_crop_names}")
 
     file_suffix = utils.make_suffix_string(args, 'results_suffix')
     output_dir = os.path.join(args['workspace_dir'])
@@ -722,6 +716,10 @@ def execute(args):
             CROP_TO_PATH_TABLES.climate_bin,
             args[CROP_TO_PATH_TABLES.climate_bin],
             crop_name)
+
+        if not crop_climate_bin_raster_path:
+            raise ValueError(
+                f'No climate bin raster path could be found for {crop_name}')
 
         LOGGER.info(
             "Clipping global climate bin raster to landcover bounding box.")
--- a/src/natcap/invest/crop_production_percentile.py
+++ b/src/natcap/invest/crop_production_percentile.py
@@ -1,11 +1,10 @@
 """InVEST Crop Production Percentile Model."""
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 import logging
 import os
 import re
 
 import numpy
-from pandas import NA
 import pygeoprocessing
 import taskgraph
 from osgeo import gdal
@@ -15,7 +14,7 @@ from . import gettext
 from . import spec
 from . import utils
 from . import validation
-from .crop_production_regression import NUTRIENTS
+from .crop_production_regression import NUTRIENTS, CROP_TO_PATH_TABLES, get_full_path_from_crop_table
 from .file_registry import FileRegistry
 from .unit_registry import u
 
@@ -241,13 +240,6 @@ nutrient_units = {
     "vitk":        u.microgram/u.hectogram,  # vitamin K
 }
 
-CropToPathTables = namedtuple(
-    'CropToPathTables', ['climate_bin', 'observed_yield', 'percentile_yield'])
-CROP_TO_PATH_TABLES = CropToPathTables(
-    climate_bin='climate_bin_raster_table',
-    observed_yield='observed_yield_raster_table',
-    percentile_yield='percentile_yield_csv_table',
-)
 
 MODEL_SPEC = spec.ModelSpec(
     model_id="crop_production_percentile",
@@ -704,7 +696,8 @@ def execute(args):
     bad_crop_name_list = []
 
     for crop_name in crop_to_landcover_df.index:
-        crop_climate_bin_raster_path = _get_full_path_from_table(
+        crop_climate_bin_raster_path = get_full_path_from_crop_table(
+            MODEL_SPEC,
             CROP_TO_PATH_TABLES.climate_bin,
             args[CROP_TO_PATH_TABLES.climate_bin],
             crop_name)
@@ -757,7 +750,8 @@ def execute(args):
     for crop_name, row in crop_to_landcover_df.iterrows():
         crop_lucode = row[_EXPECTED_LUCODE_TABLE_HEADER]
         LOGGER.info(f'Processing crop {crop_name}')
-        crop_climate_bin_raster_path = _get_full_path_from_table(
+        crop_climate_bin_raster_path = get_full_path_from_crop_table(
+            MODEL_SPEC,
             CROP_TO_PATH_TABLES.climate_bin,
             args[CROP_TO_PATH_TABLES.climate_bin],
             crop_name)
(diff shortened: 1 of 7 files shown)


src/natcap/invest/crop_production_percentile.py
```diff
@@ -290,95 +261,121 @@
...
-                                    id="yield_95th",
-                                    about=None,
-                                    units=u.metric_ton / u.hectare
-                                )
-                            ],
-                            index_col="climate_bin"
-                        )
-                    ]
+        spec.CSVInput(
+            id=CROP_TO_PATH_TABLES.climate_bin,
```
emlys, line 253: If we want to import and reuse these definitions from the other crop production module, may I suggest reusing the entire `Input` object? For example, `CLIMATE_BIN_RASTER_TABLE_INPUT = spec.CSVInput(...)`

src/natcap/invest/crop_production_regression.py
```diff
@@ -67,13 +69,69 @@
...
+    'CropToPathTables', ['climate_bin', 'observed_yield',
+                         'percentile_yield', 'regression_yield'])
+CROP_TO_PATH_TABLES = CropToPathTables(
+    climate_bin='climate_bin_raster_table',
+    observed_yield='observed_yield_raster_table',
+    percentile_yield='percentile_yield_csv_table',
+    regression_yield='regression_yield_csv_table',
+)
+
+CROP_TO_PATH_TABLE_PATH_DESC = ("Each ``path`` may be either a *relative*"
```
emlys: Text in the `about` attribute should not use markup, because it needs to be renderable in different contexts (RST, HTML, plain)

src/natcap/invest/crop_production_regression.py
```diff
@@ -141,85 +199,111 @@
...
+                )
+            ],
+            index_col="crop_name"
+        ),
+        spec.CSVInput(
+            id=CROP_TO_PATH_TABLES.observed_yield,
+            name=gettext("Observed Yield Raster Table"),
+            about=gettext(
+                "A table that maps each crop name to the corresponding"
+                f" observed yield raster. {CROP_TO_PATH_TABLE_PATH_DESC}"
```
emlys: Unfortunately `gettext` strings can't use formatting like this because the translation catalog is extracted from the source code without evaluating it. The resulting string to be translated would look just like it does here: `...observed yield raster. {CROP_TO_PATH_TABLE_PATH_DESC}...` and the translator won't know what to do with it. I'd suggest repeating the text in each `about` paragraph. While it looks repetitive, it's best for the translator to have the full context of the sentence they're translating (in some cases may be necessary in order to match gendered nouns etc.)

src/natcap/invest/crop_production_regression.py
```diff
@@ -483,6 +558,11 @@ def execute(args):
 
     LOGGER.info(
         "Checking if the landcover raster is missing lucodes")
+
+    # It might seem backwards to read the landcover_to_crop_table into a
```
emlys, line 567: Thanks for explaining this. I see that in the spec, `crop_name` is marked as the index column. Would it make sense to rename this input "crop_to_landcover_table" or something to reflect how it's used?

src/natcap/invest/crop_production_regression.py
```diff
@@ -518,15 +598,20 @@ def update_unique_lucodes_in_raster(unique_codes, block):
             f"in the landcover to crop table: {lucodes_missing_from_table}")
 
     LOGGER.info("Checking that crops correspond to known types.")
+    bad_crop_name_list = []
     for crop_name in crop_to_landcover_df.index:
```
emlys: Could this block identifying incorrect crop names be expressed more simply using some dataframe operations? Like perhaps taking the difference of the set of crop names in the `crop_to_landcover_df` index and those in the climate bin table?

emlys (changes requested):
Thanks @emilyanndavis! Just some suggestions/questions

src/natcap/invest/crop_production_percentile.py
emilyanndavis: Hmm, I mainly did this to avoid having to remember each specific string literal when referencing these inputs in the model code, but it's not a bad idea to go ahead and make the entire input reusable as well. (There is a _lot_ of other repetition between these two models, but we'll start small!)

src/natcap/invest/crop_production_regression.py
emilyanndavis: I was definitely thinking exclusively of the user guide context. 🙃

src/natcap/invest/crop_production_regression.py
emilyanndavis: Oops, I entirely forgot about the translation problem here (ironic, I know). Thanks!

src/natcap/invest/crop_production_percentile.py
emilyanndavis: Actually, this turns out to be less straightforward than it seems. Each of those tables refers to CROP_OPTIONS as the options for the crop_name column, but the CROP_OPTIONS differ between the percentile and regression models. We want to make sure the percentile model's spec points to its own CROP_OPTIONS, which would mean modifying each of those CSVInputs after import. I tried this, but I couldn't come up with a succinct way to update just that column, and what I had instead seemed too cumbersome and error-prone to make the code-sharing worthwhile. Perhaps refactoring the crop production models to reduce repetition is a rainy-day project in its own right!

src/natcap/invest/crop_production_regression.py
emilyanndavis: I thought about this, but I also thought there might be some reason the user-facing name and description of this input are the way they are, and changing them could mean opening a can of worms. I also figured that, if it hadn't come up before, maybe it was just me who found it confusing, so I decided to add an explanatory comment for future-me and call it a day. I think I'd prefer to just leave it as is for now.

src/natcap/invest/crop_production_regression.py
emilyanndavis: I realized the stated purpose of this validation was not really about the climate bin rasters at all; it was to ensure all the user-specified crop names are supported by the model. Originally, it seems, the most obvious way to do that was to check for the presence of the associated climate bin raster. Now, it's much simpler to validate directly against CROP_OPTIONS, so I decided to do that instead.

emilyanndavis:
@emlys Once your test data, sample data, and user guide updates have been merged into their respective release branches, I'll sync my test data/sample data/user guide branches and then update the commit hashes here.

src/natcap/invest/crop_production_percentile.py
emlys: Oh right, the tables do have a different specification, so we can't reuse the whole CSVInput. In my opinion, it would be clearer to use the id strings directly rather than sharing the definition between the models, but I'll leave it up to you

src/natcap/invest/crop_production_regression.py
emlys: Okay, sounds good

src/natcap/invest/crop_production_percentile.py
emilyanndavis: I know this is different from what we normally do, but my motivation was twofold: (1) if the id strings I used needed to be changed for some reason, it would be simpler to update them in one place than in multiple locations throughout both models, and (2) defining them in a named tuple prevents us from having to remember (and correctly type) each string literal when referencing them in the model code—as soon as I type CROP_TO_PATH_TABLES, my code editor lets me know which named values are available in that tuple, and I can select the one I want.

emlys (approved):
Thanks @emilyanndavis!

Issue #2155, opened 2025-09-18
`_resample_population_raster` function should be more accessible


claire-simpson:
## Background/Motivation
Urban Nature Access contains the function `_resample_population_raster`, which is necessary for resampling population rasters in a way that preserves the total population count. Resampling population data using nearest neighbor or other common methods is inaccurate. 

## Description
`_resample_population_raster` should be made more accessible for use in other models, e.g., urban mental health, plugins. This function could potentially be moved to `utils.py` or even to pygeoprocessing.

PR #2173, opened 2025-10-01
Urban mental health model option 3 (raw NDVI inputs)

### LINKED ISSUE #2143: Urban Mental Health Model: implement scenario 3 (baseline and alternate NDVI)
This model option is similar to option 2 (#2142), however instead of creating maps using an average/median NDVI value for each LULC class, the user directly supplies baseline and alternate NDVI maps, allowing for more nuanced spatial context.

Required Inputs:
1. Baseline NDVI
2. Alternate NDVI

claire-simpson:
## Description
Implementation of Urban Mental Health Model option 3, which uses baseline and alternate NDVI inputs 

Fixes #2143

## Checklist


src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1320 @@
...
+        else:
+            aoi_bbox = aoi_info["bounding_box"]
+
+        # Expand target bounding box to ensure correct edge pixel calculation
+        aoi_bbox += numpy.array(
+            [-args['search_radius'], -args['search_radius'],
+             args['search_radius'], args['search_radius']])
+        aoi_bbox = list(aoi_bbox)
+
+        # Check if buffered AOI bbox is larger than input base and alt NDVI
```
claire-simpson: This model calculates population-weighted NDVI, and takes the mean NDVI within a circular buffer (of radius `search_radius`) to reflect the typical distance people are likely to access residential green space. To ensure that the edge pixels in the area of interested have the full context for correct calculation, the user should enter an AOI that is smaller than other raster inputs by the `search_radius`. A warning is raised if the input rasters are not large enough, but the model will still run. Does this seem like a good way to go and is this documented clearly?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1320 @@
...
+
+    Returns:
+        File registry dictionary mapping MODEL_SPEC output ids to absolute paths
+
+    """
+    LOGGER.info("Starting Urban Mental Health Model")
+    args, file_registry, task_graph = MODEL_SPEC.setup(args)
+
+    LOGGER.info("Start preprocessing")
+    if args['scenario'] == 'ndvi':
```
claire-simpson: The logic is currently all wrapped up in this `if scenario == ndvi` but most of these functions will be pulled out of this block once the other scenarios are implemented and its more clear which functions are common across scenarios. Likely all the logic beyond and including the `calc_delta_ndvi` function as well as some of the aligning and resizing (for population) with the AOI bounding box. Feel free to provide feedback on this but definitely know everything in this if/else is not final!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1320 @@
...
+        if args['lulc_base']:
+            input_align_list.append(args['lulc_base'])
+            output_align_list.append(file_registry['lulc_base_aligned'])
+            resample_method_list.append('near')
+            check_raster_bounds_against_aoi(
+                aoi_bbox, pygeoprocessing.get_raster_info(
+                    args['lulc_base'])['bounding_box'], "LULC_base")
+
+        ndvi_align_task = task_graph.add_task(
+            func=pygeoprocessing.align_and_resize_raster_stack,
```
claire-simpson, line 626: The aligning and resizing of various rasters in this model is pretty convoluted because (1) the population raster has to be resampled in a unique way and (2) I'm trying to account for different projections but am not using the 'intersect' method of determining the output bounding box, as we want the target bounding box to be larger than the AOI bbox (by `search_radius`) but still want to clip down NDVI rasters as much as possible to shorten runtime (especially given convolve operation). Would appreciate feedback on if there is a cleaner way to do this!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        ["ndvi_base", "ndvi_alt"],
+        ["lulc_base", "lulc_alt", "lulc_attr_csv"],
+        ["tc_raster", "tc_target"]
+    ],
+    inputs=[
+        spec.WORKSPACE,
+        spec.SUFFIX,
+        spec.N_WORKERS,
+        spec.AOI.model_copy(update=dict(  #  TODO: potentially want to have this req to be census tract pop. shp? or only req that if opt 1 but optional for opts 2-3?
+            id="aoi_vector_path",
```
dcdenu4: I think in Spec this is defaulted to `aoi_path`. Is there a particular reason not to use this or do you think `aoi_vector_path` would be a better descriptor for what we have in `spec.AOI`, given it's a `VectorInput`?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            projection_units=u.meter
+        )),
+        spec.SingleBandRasterInput(
+            id="population_raster",
+            name=gettext("population"),
+            about=gettext(
+                "Gridded population data representing the number of people "
+                "per pixel, ideally at the pixel level. Where available, "
+                "additional demographic attributes (e.g., age, income) can "
+                "enhance model accuracy and equity assessments. For "
```
dcdenu4: I'm confused by the "additional..." statement. Is this implying some population rasters could reflect certain demographic population count and that it would be a good use in this model?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                "population count rasters, users are recommended to use the "
+                "WorldPop data, which has global coverage at 100 meter "
+                "resolution https://hub.worldpop.org/project/categories?id=3"
```
dcdenu4: Does this recommendation come from Yingjie and team? I'm wondering if we want to have explicit recommendations in the About text or have that be communicated in the User Guide? Mostly for maintenance of this recommendation aging well.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                "rates with those projected under improved nature exposure "
+                "scenarios."),
+            geometry_types={"MULTIPOLYGON", "POLYGON"},
+            fields=[
+                spec.RatioInput(
+                    id="risk_rate",
+                    about=gettext("Health risk rate")
+                )
+            ],
+            projected=None
```
dcdenu4: I think the default is `None`, so you could remove this line if you wanted.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+from osgeo import ogr, osr
+
+from . import gettext
+from . import spec
+from . import validation
+from .unit_registry import u
+
+LOGGER = logging.getLogger(__name__)
+
+SCENARIO_OPTIONS = [
```
dcdenu4: I noticed this is only used in the MODEL_SPEC and in Validation. Will this be useful to have as a constant with later options? Otherwise it might read a little better if the options were in the spec itself. But then of course they'd be duplicated for the validation function...

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            options=[
+                spec.Option(key="",
+                            display_name=gettext("--Select a scenario--"))
+            ] + SCENARIO_OPTIONS
+        ),
+        spec.PercentInput(
+            id="tc_target",
+            name="Tree Cover Target",
+            about=gettext(
+                "Target for tree canopy cover within the city or study area."
```
dcdenu4:
```suggestion
                "Target for tree canopy cover within the city or study area. "
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+    module_name=__name__,
+    input_field_order=[
+        ["workspace_dir", "results_suffix"],
+        ["aoi_vector_path", "population_raster", "search_radius"],
+        ["effect_size", "baseline_prevalence_vector",
+         "health_cost_rate"],
+        ["scenario"],
+        ["ndvi_base", "ndvi_alt"],
+        ["lulc_base", "lulc_alt", "lulc_attr_csv"],
+        ["tc_raster", "tc_target"]
```
dcdenu4: I go back and forth with shorthand for variable names. On one hand I don't like long variable names, on the other it can be confusing at first glance. Thoughts on this being `tree_canopy_raster` or `canopy_raster`?

src/natcap/invest/urban_mental_health.py
dcdenu4: Same for a few inputs below too. But, understandable if you think it's better to be explicit!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            # an attribute table
+            required="scenario!='lulc' or not lulc_attr_csv",
+        ),
+        spec.SingleBandRasterInput(
+            id="ndvi_alt",
+            name=gettext("alternate NDVI"),
+            about=gettext(
+                "Map of NDVI under future or counterfactual conditions. "
+                "This raster should extend beyond the AOI by at "
+                "least the search radius distance"
```
dcdenu4:
```suggestion
                "least the search radius distance."
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            required="scenario=='tcc_ndvi'",
+            allowed="scenario=='tcc_ndvi'"
+        ),
+        spec.SingleBandRasterInput(
+            id="ndvi_base",
+            name=gettext("baseline NDVI"),
+            about=gettext(
+                "Map of NDVI under current or baseline conditions. "
+                "This raster should extend beyond the AOI by at "
+                "least the search radius distance"
```
dcdenu4:
```suggestion
                "least the search radius distance."
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            required="scenario=='tcc_ndvi'",
+            allowed="scenario=='tcc_ndvi'"
+        ),
+        spec.SingleBandRasterInput(
+            id="tc_raster",
+            name=gettext("Tree Canopy Cover (TCC)"),
+            about=gettext(
+                "Map of the percentage of pixel area covered by trees. "
+                "This raster should extend beyond the AOI by at "
+                "least the search radius distance"
```
dcdenu4:
```suggestion
                "least the search radius distance."
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        spec.SingleBandRasterInput(
+            id="lulc_alt",
+            name=gettext("alternate land use/land cover"),
+            about=gettext(
+                "Map of land use/land cover codes under future or "
+                "counterfactual conditions. Each land use/land cover type "
+                "must be assigned a unique integer code. If an LULC attribute "
+                "table is used, all values in this raster must have "
+                "corresponding entries. This raster should extend beyond the "
+                "AOI by at least the search radius distance"
```
dcdenu4:
```suggestion
                "AOI by at least the search radius distance."
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            columns=[
+                spec.IntegerInput(
+                    id="lucode",
+                    about=gettext("LULC code")
+                ),
+                spec.BooleanInput(
+                    id="exclude",
+                    about=gettext(
+                        "Whether to exclude the LULC class (e.g., if water) "
+                        "or keep it")
```
dcdenu4:
```suggestion
                        "or keep it.")
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                "A table mapping unique LULC codes to NDVI values and stating "
+                "whether the LULC class should be excluded (0 for keeping, "
+                "and 1 for excluding). Typically, water is excluded and other "
+                "land cover types are kept. NDVI values are only required if "
+                "scenario is 'lulc' and a base NDVI raster is not provided."
+            ),
+            columns=[
+                spec.IntegerInput(
+                    id="lucode",
+                    about=gettext("LULC code")
```
dcdenu4:
```suggestion
                    about=gettext("LULC code.")
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                ),
+                spec.BooleanInput(
+                    id="exclude",
+                    about=gettext(
+                        "Whether to exclude the LULC class (e.g., if water) "
+                        "or keep it")
+                ),
+                spec.NumberInput(
+                    id="ndvi",
+                    about=gettext("NDVI value"),
```
dcdenu4:
```suggestion
                    about=gettext("NDVI value."),
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            required="(scenario=='lulc' and not ndvi_base) or "
+                     "(scenario!='lulc' and lulc_base)",
+            allowed="scenario=='lulc' or lulc_base"
+        )
+        ],
+    outputs=[
+            spec.SingleBandRasterOutput(
+                id="preventable_cases",
+                path="output/preventable_cases.tif",
+                about=gettext("Preventable cases at pixel level"),
```
dcdenu4:
```suggestion
                about=gettext("Preventable cases at pixel level."),
```

dcdenu4: There's a few other about text entries that might need periods. We might not be consistent about this in all our models, but I think it'd be nice.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                about=gettext(
+                    "Difference between baseline and alternate NDVI raster"),
+                data_type=float,
+                units=None
+            ),
+            spec.SingleBandRasterOutput(
+                id="kernel",
+                path="intermediate/kernel.tif",
+                about=gettext(
+                    "Binary raster representing the dichotomous kernel that"
```
dcdenu4:
```suggestion
                    "Binary raster representing the dichotomous kernel that "
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            effect size representing the relationship between nature
+            exposure and the mental health outcome, given as relative
+            risk associated with a 0.1 increase in NDVI.
+        args['baseline_prevalence_vector'] (str): (required) a path to a
+            vector providing the baseline prevalence (or incidence) rate of
+            a specific mental health outcome (e.g., depression or anxiety)
+            across administrative units within the study area. This data allows
+            the model to estimate preventable cases by comparing current rates
+            with those projected under improved nature exposure scenarios. The
+            vector must contain field `risk_rate`.
```
dcdenu4: I _think_ in docstrings we need double back ticks: ``risk_rate`` to be rendered in RST / sphinx properly.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+    elif args['scenario'] == 'lulc':
+        raise NotImplementedError
+
+    elif args['scenario'] == 'ndvi':
+        # TODO rearrange whats in if/else block when implementing scenarios 1-2
+        LOGGER.info("Using scenario option 3: NDVI")
+        base_ndvi_raster_info = pygeoprocessing.get_raster_info(
+            args['ndvi_base'])
+        pixel_size = base_ndvi_raster_info['pixel_size']
+        target_projection = base_ndvi_raster_info['projection_wkt']
```
dcdenu4: If the AOI is required for all scenarios, I wonder if that would make for the best source of truth for the projection? Depending on how we set InVEST models up to handle remote datasets and different projections, it might be more awkward to have the NDVI raster act as the projection target. I'm thinking if someone wants to point to a remote NDVI global layer or something...

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        # alternative option is to have mode='intersection' and use AOI vector
+        # to clip rasters during align/resize task
+        # however, this would mean that edge pixels would be less accurate/
+        # lack spatial context
+
+        # Users should input the AOI to which they want outputs clipped
+        # InVEST will take care of buffering the processing AOI to ensure
+        # correct edge pixel calculation
+        aoi_info = pygeoprocessing.get_vector_info(args['aoi_vector_path'])
+        if aoi_info["projection_wkt"] != target_projection:
```
dcdenu4: Unfortunately we can't rely on comparing the WKT strings directly for projection equality. But I think gdal has an `osgeo.osr.IsSame` function that hopefully does the trick.

dcdenu4: Here's a post about equality checking that Rich, former NatCap software lead, wrote over a decade ago!
https://gis.stackexchange.com/questions/50312/testing-if-two-gdal-datasets-are-in-same-projection

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            aoi_bbox = pygeoprocessing.transform_bounding_box(
+                aoi_info["bounding_box"], aoi_info["projection_wkt"],
+                target_projection)
+        else:
+            aoi_bbox = aoi_info["bounding_box"]
+
+        # Expand target bounding box to ensure correct edge pixel calculation
+        aoi_bbox += numpy.array(
```
dcdenu4: Cool, so this assumes that the AOI is projected in linear units matching that of the search_radius input. This might make having the AOI as the source of truth for projection an even stronger argument.

Do you know if we validate on expected units for a given spatial input?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        check_raster_bounds_against_aoi(aoi_bbox, ndvi_bbox, "NDVI_base")
+        check_raster_bounds_against_aoi(
+            aoi_bbox,
+            pygeoprocessing.get_raster_info(args['ndvi_alt'])['bounding_box'],
+            "NDVI_alt")
+
+        input_align_list = [args['ndvi_base'], args['ndvi_alt']]
+        output_align_list = [file_registry['ndvi_base_aligned'],
+                             file_registry['ndvi_alt_aligned']]
+        resample_method_list = ['cubicspline', 'cubicspline']
```
dcdenu4: Can you talk about why you ended up going with `cubicspline` for interpolation of the NDVI layers?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                aoi_info["bounding_box"], aoi_info["projection_wkt"],
+                target_projection)
+        else:
+            aoi_bbox = aoi_info["bounding_box"]
+
+        # Expand target bounding box to ensure correct edge pixel calculation
+        aoi_bbox += numpy.array(
+            [-args['search_radius'], -args['search_radius'],
+             args['search_radius'], args['search_radius']])
+        aoi_bbox = list(aoi_bbox)
```
dcdenu4: To avoid confusion I might have a separate `aoi_buffered_bbox` variable.

src/natcap/invest/urban_mental_health.py
dcdenu4: I think you're doing it right by using the buffered AOI bbox. I'm assuming towards the end of the model we can then mask out the results with the AOI.

src/natcap/invest/urban_mental_health.py
dcdenu4: I think it's okay to align and resize before doing this check. Then we can include `lulc_base`, `lulc_alt` (which might be missing for alignment currently) up front.

I'm on the fence of whether we should be doing the `check_raster_bounds_against_aoi`. I think it's nice to try and be as helpful as possible to the user, but what if the bbox is okay, but the edges are all just `nodata`? What if the NDVI layer has some `nodata` holes anyway? I just think it can be challenging to ensure perfect data and wonder if this would be better off as a clear User Guide documentation...

If we wanted to keep this check, which definitely could be useful, maybe we could streamline the logic a bit. Could we take the intersection of all non AOI bounding boxes, and then compare that to the AOI buffered bbox? I think warning that "A" input layer is smaller than the AOI extents is sufficient for someone to investigate their input data.

Also thinking to a future where we might not want to require all inputs be projected (global NDVI raster), the bounding box comparison becomes more complicated. Usually in that case we end up transforming everything to geographic wgs84 (lat/lon) for comparison.

These are all just some quick thoughts, let's talk about this more!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            target_path_list=output_align_list,
+            task_name='align NDVI rasters')
+
+        mask_base_inputs = [file_registry['ndvi_base_aligned'],
+                            file_registry['ndvi_base_aligned_masked']]
+        mask_alt_inputs = [file_registry['ndvi_alt_aligned'],
+                           file_registry['ndvi_alt_aligned_masked']]
+        mask_base_outputs = [file_registry['ndvi_base_aligned_masked']]
+        mask_alt_outputs = [file_registry['ndvi_alt_aligned_masked']]
+        if args['lulc_base']:
```
dcdenu4: I haven't put my finger on it, but I wonder if there's a way we can streamline this masking a bit.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                  resample_method_list,
+                  pixel_size,
+                  aoi_bbox),
+            kwargs={
+                'raster_align_index': 0,  # align to base_ndvi
+                'target_projection_wkt': target_projection},
+            target_path_list=output_align_list,
+            task_name='align NDVI rasters')
+
+        mask_base_inputs = [file_registry['ndvi_base_aligned'],
```
dcdenu4: The `mask_` variable names are a bit confusing. Would `ndvi_water_mask_` be too verbose?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        args['results_suffix'] (str): (optional) appended to any output
+            filename.
+        args['aoi_vector_path'] (str): (required) path to a polygon vector
+            that is projected in a coordinate system with units of meters.
+            The polygon should intersect the baseline prevalence vector and
+            the population raster.
+        args['population_raster] (str): (required) a path to a raster of
+            gridded population data representing the number of people per
+            pixel.
+        args['search_radius'] (float): (required) the distance used to define
```
dcdenu4: Might be good to add that `search_radius` is expected to be greater than 0.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+            args=mask_alt_inputs,
+            target_path_list=mask_alt_outputs,
+            dependent_task_list=[ndvi_align_task],
+            task_name="Mask alternate NDVI"
+        )
+
+        pixel_radius = int(round(args['search_radius']/pixel_size[0]))
+        LOGGER.info(f"Search radius {args['search_radius']} results in "
+                    f"buffer of {pixel_radius} pixels")
+        if pixel_radius == 0:
```
dcdenu4: We use `search_radius` above when buffering putting together the bounding box. So, that might be the time to check that the value is great than 0, to avoid computation that is ultimately going to result in raising an error.

But, the Model Spec has an expression for it being greater than 0 and based on our team discussion the other, we are seemingly leaning away from including these checks that validation should catch.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                  file_registry['delta_ndvi']),
+            target_path_list=[file_registry['delta_ndvi']],
+            dependent_task_list=[mean_buffer_base_ndvi_task,
+                                 mean_buffer_alt_ndvi_task],
+            task_name="calculate delta ndvi"  # change in nature exposure
+        )
+
+        pop_raster_info = pygeoprocessing.get_raster_info(
+            args['population_raster'])
+        if pop_raster_info['projection_wkt'] != target_projection:
```
dcdenu4: Might want to look into `IsSame` as mentioned above.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+                func=calc_preventable_cost,
+                args=(file_registry['preventable_cases'],
+                      args['health_cost_rate'],
+                      file_registry['preventable_cost']),
+                target_path_list=[file_registry['preventable_cost']],
+                dependent_task_list=[preventable_cases_task],
+                task_name="calculate preventable cost"
+            )
+
+        task_graph.join()
```
dcdenu4: Yep, if you have a blocking task you can call join before continuing.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+
+        zonal_stats_inputs = [
+            args['aoi_vector_path'],
+            file_registry['preventable_cases_cost_sum_table'],
+            file_registry['preventable_cases_cost_sum_vector'],
+            file_registry['preventable_cases']]
+
+        zonal_stats_dependent_tasks = [preventable_cases_task]
+
+        if args['health_cost_rate']:
```
dcdenu4: I _think_ you could move the above `preventable_cost_task` code under this same conditional, then add to the input and dependent task list, and then remove that `task_graph.join()` above. If I'm reading that right!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1335 @@
...
+        errors = [k for k, v in errors_dict.items() if v]
+        raise UserWarning(
+            "The extent of bounding box of the AOI buffered by the search "
+            f"radius exceeds that of the {raster_name} raster input. The "
+            f"issue is with the following coordinates: {errors}. For "
+            f"reference, the buffered AOI bounding box is: {aoi_bbox}, and "
+            f"the raster bbox is: {raster_bbox}")
+
+
+def _resample_population_raster(
```
dcdenu4: Was this pulled right from UNA? Is it possible to reuse that code or should we move it to a `utils`? If it is pulled directly from somewhere else, I recommend we not duplicate it since it's a large code chunk and would need duplicated testing too.

src/natcap/invest/urban_mental_health.py
claire-simpson: Nope, I also think `aoi_path` is fine, will update

src/natcap/invest/urban_mental_health.py
claire-simpson: Ah, no I don't think demographic information is useful in this model, maybe for analyses after the model has been run... This was a copy+paste from the design doc that I neglected to update. I think all thats needed for the description here is "Gridded population data representing the number of people per pixel."

src/natcap/invest/urban_mental_health.py
claire-simpson: Yes, I agree this is probably more appropriate for the UG

src/natcap/invest/urban_mental_health.py
claire-simpson: Yes I just have this so I didn't have to duplicate the options in both MODEL_SPEC and Validation, don't think it'll be used elsewhere. I can certainly write this directly into the model spec and validation if you think thats better!

src/natcap/invest/urban_mental_health.py
claire-simpson: Yes, that makes sense! I think to me the best option for a more descriptive name would be `tree_cover_raster`. In a similar vein, perhaps it would be better to change the scenario option titled `TCC & Baseline NDVI` to `Tree Cover & Baseline NDVI`? I feel like TCC is a somewhat known acronym but certainly less ubiquitous than NDVI... Thoughts?

src/natcap/invest/urban_mental_health.py
claire-simpson: I'll fix them all! I wasn't sure whether these short not-full-sentence descriptions required a period.

src/natcap/invest/urban_mental_health.py
claire-simpson: I thought about this too. It would definitely be the easiest way to go in terms of writing the model and communicating to users. I chose NDVI as the source of truth as it contains the most important data for the model, and therefore for the sake of accuracy, I wanted to warp it as little as possible. I hadn't thought about what this might mean for remote data sources though, so this is a good point. I could definitely be convinced to reproject everything based on the AOI

src/natcap/invest/urban_mental_health.py
claire-simpson: Ah bummer! Thanks for sharing that function though, I was not aware of it

src/natcap/invest/urban_mental_health.py
claire-simpson: > This might make having the AOI as the source of truth for projection an even stronger argument.

True, or we could require that input rasters also have `projection_units=u.meter`

> Do you know if we validate on expected units for a given spatial input?

Yes, there is a `_check_projection` function in `spec.py` for the purpose of validating `projection_units`

src/natcap/invest/urban_mental_health.py
claire-simpson: Oops I definitely meant to just do `cubic`, as being both accurate and smooth, but I wouldn't mind just using `bilinear` resampling either as its quicker..

src/natcap/invest/urban_mental_health.py
claire-simpson: I'm open to suggestions! Not sure if this is what you're talking about, but there is definitely a trade off between clarity and minimizing the logic we're including within the body of `execute`. Could move all this logic related to which inputs we care about for masking to the `mask_ndvi` task itself

src/natcap/invest/urban_mental_health.py
claire-simpson: Hmm I see your point, but don't think `ndvi_water_mask_` is quite accurate as these lists could include LULC rasters and attribute tables as well. What about `inputs_to_mask_baseline_rasters_task` or `mask_base_ndvi_task_inputs` or `mask_base_task_inputs`? I recognize these are even more verbose...

src/natcap/invest/urban_mental_health.py
claire-simpson: I like the idea to check that `pixel_radius > 0` upfront. 

>But, the Model Spec has an expression for it being greater than 0 and based on our team discussion the other, we are seemingly leaning away from including these checks that validation should catch.

I still include this check to make sure that whatever `search_radius` the user inputs yields a pixel radius of at least 1 (e.g., if the user sets `search_radius=10`, but doesn't realize the resolution of NDVI is 30 m, the pixel_radius would be 0 since the radius doesn't cover at least half of the pixel). An alternative option could be to round up, but my opinion is I'd rather the model error quickly so the user actually understands what their search radius should represent. Or, I could move this check to `validate`, but I think its easier to include in the body of execute.

src/natcap/invest/urban_mental_health.py
claire-simpson: Yep, exact same code. I opened an issue about this (#2155) but haven't got going on it yet. I think it would be great to abstract this out somewhere so its more accessible and `utils` seems appropriate to me!

src/natcap/invest/urban_mental_health.py
claire-simpson: Ah yes, good call on adding to the lists under the same conditional as the `preventable_cost_task`. 

I do think I still have to include the `task_graph.join()` call though before the `zonal_stats_task`, as the `zonal_stats_task` can run _regardless_ of whether the user has entered a `health_cost_rate`, _however_ if the user _has_ entered `health_cost_rate`, `preventable_cost_task` is a dependent task and must be run prior to `zonal_stats_task`. Does this make sense? Am I interpreting how to use `task_graph.join` correctly?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1349 @@
...
+        spec.SingleBandRasterInput(
+            id="tree_cover_raster",
+            name=gettext("Tree Canopy Cover (TCC)"),
+            about=gettext(
+                "Map of the percentage of pixel area covered by trees. "
+                "This raster should extend beyond the AOI by at "
+                "least the search radius distance."
+            ),
+            data_type=float,
+            units=None,
```
emilyanndavis: For rasters where the pixel value represents percentages, it looks like we could use `units=u.percent`. It's only a minor difference, since percentages are still technically unitless (in the colloquial sense of units), but it could be a helpful detail in the metadata, user guide, summary reports, etc.

src/natcap/invest/urban_mental_health.py
emilyanndavis: If it were me, I'd try moving this logic to a new function that returns the appropriate inputs and outputs based on the selected scenario and the model inputs. I think it would reduce the cognitive load of reading the execute method, and it would also cut down on code repetition when it comes time to implement the other two scenarios.

I'm picturing something along the lines of the following:
```
def _get_mask_task_inputs_and_outputs(args, file_registry, ndvi_type='base'):
    ...
    return (inputs, outputs, file_registry)
    
(mask_base_task_inputs, mask_base_task_outputs, file_registry) = _get_mask_task_inputs_and_outputs(args, file_registry)

mask_base_ndvi_task = task_graph.add_task(
    ...
)

(mask_alt_task_inputs, mask_alt_task_outputs, file_registry) = _get_mask_task_inputs_and_outputs(args, file_registry, 'alt')

mask_alt_ndvi_task = task_graph.add_task(
    ...
)
```

One thing that might be slightly annoying about this is, I think your helper function would have to return the file_registry back to execute, because—If I'm remembering correctly how file_registry works—looking up a key in the file_registry can also mutate the file_registry itself, and we want to make sure execute has the latest snapshot of the file_registry.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1331 @@
...
+    PC: preventable cases
+    PF: preventable fraction
+    BC: baseline cases
+    RR: relative risk or risk ratio
+    NE: nature exposure, as approximated by NDVI
+    RR0.1NE: RR per 0.1 NE increase.
+        By default, the model uses the relative risk associated with a
+        0.1 increase in NDVI-based nature exposure. This value must be
+        provided by users in the effect size table, based on empirical
+        studies or meta-analyses relevant to the selected health outcome.
```
emilyanndavis: I'm a little confused by a couple of things in this description.

1. The use of "by default" suggests that it's possible to instruct the model to operate on some other delta in nature exposure. Is this planned? Or would it be accurate to remove "by default"?

2. If I'm understanding the code correctly, the value of `RR_0.1NE` is the user-specified `effect_size`, which is a single value, not a table. Will other scenarios expect a table, or is this a remnant of some earlier description of the model?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1331 @@
...
+    scenario_options = [
+        spec.Option(key="tcc_ndvi", display_name="Tree Cover & Baseline NDVI"),
+        spec.Option(key="lulc", display_name="Baseline & Alternate LULC"),
+        spec.Option(key="ndvi", display_name="Baseline & Alternate NDVI")
+    ]
+
+    error_msg = spec.OptionStringInput(
+        id='', options=scenario_options).validate(args['scenario'])
+    if error_msg:
+        validation_warnings.append((['scenario'], "Must select a scenario."))
```
emilyanndavis, line 1215: It seems like this must be the first/only place we're implementing a dropdown list of options without defaulting to a valid option, but I doubt it'll be the last. I wonder if we might want to consider updating `OptionStringInput.validate` to reject empty-string keys as invalid, or alternatively, omitting the empty-string option from the spec and injecting that option into the UI on the front end instead. Just thinking out loud here; I don't think it warrants addressing immediately, but I might create an issue for it just in case we want to revisit it later.

tests/test_urban_mental_health.py
```diff
@@ -0,0 +1,701 @@
...
+        Test that anywhere delta_ndvi or baseline_cases are 0, output
+        preventable cases is 0.
+        Test that preventable_cases decreases with negative delta_ndvi and
+        increases with positive delta_ndvi (given effect size < 1).
+        """
+        from natcap.invest import urban_mental_health
+
+        ndvi_array = numpy.array(([FLOAT32_NODATA, .3, .1], [.2, 0, 0],
+                                  [-.1, .3, .1], [-0.2, .5, 1]))
+        delta_ndvi = os.path.join(self.workspace_dir, "ndvi_base.tif")
```
emilyanndavis: This doesn't affect the test at all, but I was surprised to see "ndvi_base.tif" as the name of this file.

tests/test_urban_mental_health.py
```diff
@@ -0,0 +1,701 @@
...
+
+    Returns:
+        None.
+    """
+    # UTM Zone 10N
+    srs = osr.SpatialReference()
+    srs.ImportFromEPSG(26910)
+    projection_wkt = srs.ExportToWkt()
+
+    origin = (461251, 4923445)
```
emilyanndavis: I'm honestly not sure if this would be worth the trouble or not, but since these specific values are used many times in the tests, you might consider defining module-level constants that could be referenced by name (easier to remember and harder to mistype than integer literals).

tests/test_urban_mental_health.py
```diff
@@ -0,0 +1,701 @@
+# coding=UTF-8
```
emilyanndavis: I noticed a few—but far from all—of our modules include this `coding=UTF-8` directive (and a couple have `encoding=UTF-8`). Do you have any idea what it's for (and why we're not using it everywhere)?

src/natcap/invest/urban_mental_health.py
claire-simpson: Great catch, this is all legacy from the design doc that I forgot to update when we made changes to the model inputs.

src/natcap/invest/urban_mental_health.py
claire-simpson: Great idea! I could foresee a situation where we wouldn't want to reject an empty string as invalid, so I'd be hesitant to _always_ validate in that way, but having it as an option could be useful. Having an empty string as an option on the front end also seems like a good way to go!

tests/test_urban_mental_health.py
claire-simpson: Whoops definitely a typo

tests/test_urban_mental_health.py
claire-simpson: It declares the encoding which ensures that the Python interpreter correctly interprets non-ASCII characters that could be in our comments/strings. Looks like it isn't needed in Python 3 since UTF-8 is the default! Will remove

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+            id="scenario",
+            name=gettext("scenario"),
+            about=gettext(
+                "Land use scenario options which incorporate the following "
+                "rasters as inputs: (1) Tree Canopy Cover and baseline NDVI, "
+                "(2) baseline and alternate Land Use/Land Cover, or (3) "
+                "baseline and alternate NDVI."
+            ),
+            options=[
+                spec.Option(key="",
```
dcdenu4, line 131: Does validation for `OptionStringInput` handle an empty `key` as invalid or could this be passed to the model? Mostly a mental note for me to check.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+            id="ndvi_base",
+            name=gettext("baseline NDVI"),
+            about=gettext(
+                "Map of NDVI under current or baseline conditions. "
+                "This raster should extend beyond the AOI by at "
+                "least the search radius distance."
+            ),
+            data_type=float,
+            units=None,
+            # require ndvi_base unless scenario is `lulc`` and user enters
```
dcdenu4:
```suggestion
            # require ndvi_base unless scenario is `lulc` and user enters
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+        pop_raster_info = pygeoprocessing.get_raster_info(
+            args['population_raster'])
+        pop_sr = osr.SpatialReference()
+        pop_sr.ImportFromWkt(pop_raster_info['projection_wkt'])
+        if not pop_sr.IsSame(aoi_sr):
+            transformed_bounding_box = pygeoprocessing.transform_bounding_box(
+                pop_raster_info['bounding_box'],
+                pop_raster_info['projection_wkt'],
+                aoi_projection)
+            pop_raster_info['bounding_box'] = transformed_bounding_box
```
dcdenu4: Is this needed or used? Or is everything being handled in `utils.resample_population_raster`? Sorry if I'm missing something here.

src/natcap/invest/urban_mental_health.py
dcdenu4: Wait, sorry. It looks like you're using `zonal_stats_dependent_tasks` below as a dependent task list. So, wouldn't adding this cost task to that list if we're in here work? So we shouldn't need to have this join statement?

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+                        target_base_prevalence_raster, target_base_cases):
+    """Calculate baseline cases via incidence_rate * population
+
+    Args:
+        population_raster (str): path to aligned population raster
+            representing the number of inhabitants per pixel across
+            the study area. Pixels with no population should be
+            assigned a value of 0.
+        base_prevalence_vector (str): path to vector with field
+            `risk_rate` that provides the baseline prevalence
```
dcdenu4:
```suggestion
            ``risk_rate`` that provides the baseline prevalence
```

dcdenu4: I think we need double ticks for it to show up in API docs.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+                                  "ATTRIBUTE=risk_rate", "ALL_TOUCHED=TRUE",
+                                  "MERGE_ALG=REPLACE"])
+
+    raster_list = [target_base_prevalence_raster, population_raster]
+    pygeoprocessing.raster_map(_multiply_op, raster_list, target_base_cases)
+
+
+def calc_preventable_cases(delta_ndvi, baseline_cases, effect_size,
+                           target_preventable_cases, aoi, work_dir):
+    """Calculate preventable cases and clip to AOI
```
dcdenu4, line 990: If we're going to outline the equations the function is implementing, which I think can be helpful. We should circle back and add to the docstring which equation names / references this relates to in the eventual user guide.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+    nodata = raster_info["nodata"][0]
+    target_dtype = raster_info["datatype"]
+
+    pygeoprocessing.raster_calculator(
+        [(preventable_cases, 1), (health_cost_rate, "raw"), (nodata, "raw")],
+        _preventable_cost_op, target_preventable_cost, target_dtype,
+        nodata_target=nodata)
+
+
+def zonal_stats_preventable_cases_cost(
```
dcdenu4, line 1093: Might be nice to add this example use case to the the pygeoprocessing issue, which proposes an option to write out a vector.

https://github.com/natcap/pygeoprocessing/issues/277

src/natcap/invest/urban_mental_health.py
dcdenu4: I commented similarly above about this. I think it's common practice to have a "default" first entry that is essentially asking the user to make a selection. I think it'd be nice for OptionString to have a flag for "prompt_user_to_select" or something... And then have that be an invalid selection.

dcdenu4: Could we at least make sure to make an issue for this?

tests/test_urban_mental_health.py
```diff
@@ -0,0 +1,717 @@
...
+        'search_radius': 100, # 1 pixel
+        'tc_raster': '',
+        'tc_target': '',
+        'workspace_dir': workspace_dir,
+    }
+
+    return args
+
+
+gdal.UseExceptions()
```
dcdenu4: This feels like an odd spot for this... If this is necessary can we move this just after imports?

tests/test_urban_mental_health.py
```diff
@@ -0,0 +1,717 @@
...
+        # this lets us delete the workspace after its done no matter the
+        # the test result
+        self.workspace_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        """Overriding tearDown function to remove temporary directory."""
+        shutil.rmtree(self.workspace_dir)
+
+    def test_ndvi_preprocessing_and_delta_calc(self):
+        """Test NDVI preprocessing and calculation with correct nodata
```
dcdenu4, line 157: Historically we've had something like: "UMH: Test NDVI ..." to help distinguish what test is failing during a test suite run. I'm not sure this convention is necessary anymore, but it might be nice to check an example of a test suite run to see if this is still useful or not.

dcdenu4: It'd be great if we didn't need to take up docstring space with that convention and that pytests clearly identified which tests from which files were failing (or passing)!

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+        ["ndvi_base", "ndvi_alt"],
+        ["lulc_base", "lulc_alt", "lulc_attr_csv"],
+        ["tree_cover_raster", "tree_cover_target"]
+    ],
+    inputs=[
+        spec.WORKSPACE,
+        spec.SUFFIX,
+        spec.N_WORKERS,
+        spec.AOI.model_copy(update=dict(  #  TODO: potentially want to have this req to be census tract pop. shp? or only req that if opt 1 but optional for opts 2-3?
+            id="aoi_path",
```
dcdenu4: Since we're sticking the same `id` name as defined in `spec.AOI` I think we can avoid redefining it here altogether.

dcdenu4:
```suggestion
```

src/natcap/invest/urban_mental_health.py
dcdenu4: This seems good and I like the option title change.

src/natcap/invest/urban_mental_health.py
dcdenu4: Looks like all these have been captured.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+                        args['lulc_attr_csv'],
+                        file_registry[f'lulc_{base_or_alt}_mask']]
+    else:
+        LOGGER.info("Masking NDVI using threshold NDVI<0")
+
+    return mask_inputs, mask_outputs
+
+
+def mask_ndvi(input_ndvi, target_masked_ndvi, input_lulc=None,
+              lulc_attr_table=None, target_lulc_mask=None):
```
dcdenu4, line 890: I don't think in any of the calls to this function that we're using `input_lulc` or `lulc_attr_table`. Based on `_get_masking_inputs_and_outputs`, should we possibly be changing these from defaults? I could be misunderstanding the workflow between these functions.

src/natcap/invest/urban_mental_health.py
```diff
@@ -0,0 +1,1232 @@
...
+
+    Returns:
+        mask_inputs, mask_outputs lists which represent files to input
+        to mask task and target outputs for mask task, respectively.
+
+    """
+    mask_inputs = [file_registry[f'ndvi_{base_or_alt}_aligned'],
+                   file_registry[f'ndvi_{base_or_alt}_aligned_masked']]
+    mask_outputs = [file_registry[f'ndvi_{base_or_alt}_aligned_masked']]
+    if args['lulc_attr_csv']:  # attr table can only be provided if lulc_base
```
dcdenu4, line 871: Related to my comment below, should this function be doing the lulc reclassify step and adding those generated rasters to the returned list to then use in the `mask_ndvi` function?

dcdenu4 (changes requested):
Thanks @claire-simpson ! Sorry for my disjoint follow-up review, but GitHub is seemingly bugged and not showing me the prior conversations in the code diff, which made it really challenging to re-review without repeating prior thoughts.

src/natcap/invest/urban_mental_health.py
claire-simpson: I think an empty string could be valid as per `OptionStringInput`'s `validate`, though `preprocess` would convert it to `None`... To be safe, I'm raising an error in the model's `validate` function if `''` is passed

src/natcap/invest/urban_mental_health.py
claire-simpson: No thats a great catch, thanks! This shouldn't be needed with the updates to resampling/using the AOI as the target

src/natcap/invest/urban_mental_health.py
claire-simpson: Added a quick comment to note this in the issue!

src/natcap/invest/urban_mental_health.py
emilyanndavis: Done! #2260

tests/test_urban_mental_health.py
claire-simpson: I don't think this is needed, here is an example of what GHA tells us during a test failure: https://github.com/natcap/invest/actions/runs/19248776131/job/55215548073?pr=2252 

e.g., we can see `FAILED tests/test_coastal_blue_carbon.py::TestCBC2::test_model - AssertionError: `

src/natcap/invest/urban_mental_health.py
claire-simpson: The reason I'm redefining it is (1) to have a more specific description to inform users of the requirement that the AOI be smaller than the raster inputs by at least the search radius distance (I realize this will be in the UG too, but I think since its a slightly unusual requirement, it would be good to have this stated up front) , and (2) to make sure the AOI is projected in meters.

claire-simpson: Ah, or you're pointing out that I can remove just the `id`

src/natcap/invest/urban_mental_health.py
claire-simpson: I think we are using `input_lulc` and `lulc_attr_table`? E.g., `_get_masking_inputs_and_outputs` will return a `mask_inputs` list containing (in addition to the `_aligned` and `_aligned_masked` files): `[file_registry[f'lulc_{base_or_alt}_aligned'],
                        args['lulc_attr_csv'],
                        file_registry[f'lulc_{base_or_alt}_mask']]` if the user inputs an LULC attribute table

src/natcap/invest/urban_mental_health.py
claire-simpson: I guess it could? Right now, I have the `mask_ndvi` function taking care of this, whereas `_get_masking_inputs_and_outputs` is simply generating the lists of arguments to pass to `mask_ndvi`. To me I'd rather keep `_get_masking_inputs_and_outputs` as a pretty simple helper function, but would be open to moving the LULC reclassification to its own function outside of `mask_ndvi`. I do think it'll be a bit easier to determine where this code should live during the implementation of Option 2 (LULC inputs)

PR #2174, opened 2025-10-01
Document range of (0, 100] for curve number values

### LINKED ISSUE #2164: Add curve number value range to investspec for SWY and UFRM
The Data Needs entry for the biophysical table curve number values currently says this (via :investspec:`seasonal_water_yield.seasonal_water_yield biophysical_table_path.columns.cn_[SOIL_GROUP]`):

> cn_[SOIL_GROUP] ([number](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/input_types.html#number), units: unitless, required): Curve number values for each combination of soil group and LULC class. Replace [SOIL_GROUP] with each soil group code A, B, C, D so that there is one column for each soil group. Curve number values must be greater than 0.

Instead of just saying "Curve number values must be greater than 0", we should specify the complete range of values, which max at 100. So i would recommend changing this line to something like "Curve number values must be between 1 and 100, values of 0 are not allowed".

emlys:
## Description
Fixes #2164

Update documentation and input specs for curve number inputs to require that values are greater than 0 and less than or equal to 100. Note that this will not actually be enforced through validation unless we implement https://github.com/natcap/invest/issues/2175.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -126,6 +126,15 @@ Seasonal Water Yield
   These CSVs must have the columns ``month`` and ``path``, mapping month indexes
   (1-12) to raster paths.
   (`#2096 <https://github.com/natcap/invest/issues/2096>`_)
+* Documentation has been updated to reflect that curve number values must be
+  greater than 0 and less than or equal to 100.
+  (`#2164 <https://github.com/natcap/invest/issues/2164>`_)
+
+Urban Flood Risk Mitigation
+===========================
+* Documentation has been updated to reflect that curve number values must be
+  greater than 0 and less than or equal to 100.
+  (`#2164 <https://github.com/natcap/invest/issues/2164>`_)
 
 Visitation: Recreation and Tourism
 ==================================


src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -144,8 +144,9 @@
                         "Curve number values for each combination of soil group and LULC"
                         " class. Replace [SOIL_GROUP] with each soil group code A, B, C,"
                         " D so that there is one column for each soil group. Curve number"
-                        " values must be greater than 0."
+                        " values must be greater than 0 and less than or equal to 100."
                     ),
+                    expression="0 < value <= 100",
```
emlys, line 149: I'm adding these `expression`s, but for now, the expression isn't actually enforced on the column values (same with all other numeric values in CSV inputs). If we go ahead with https://github.com/natcap/invest/issues/2175, the values will be validated using this expression and the user will see an error if a value is outside of the range.

newtpatrol (approved):
Looks correct to me @emlys .

dcdenu4 (approved):
Thanks @emlys

PR #2178, opened 2025-10-05
I. #2116 update ndr load type text

### LINKED ISSUE #2116: NDR: Clarify load_type_x description in investspec
While trying to help a forum user understand what to enter for the new "load_type_x" biophysical table fields in NDR, I see that the User Guide entry (which is straight investspec) for these fields is not very clear. 

It [currently reads](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/ndr.html#data-needs):

> load_type_p ([option], conditionally required): Whether the nutrient load in column load_p should be treated as nutrient application rate or measured contaminant runoff. ‘application-rate’ | ‘measured-runoff’

I'll make a few suggestions for improvement, although they still don't feel quite as clear as I'd like, so it would be good to have others think on it. Meanwhile, I'll update the example biophysical table in the UG, since I'm now noticing that it's out of date, and having that visual is helpful. 

A type of "option" is confusing, I would call it something like "text string" or "pre-defined text". When I see "option" it makes me think that there should be a drop-down menu of options, which there is not in this case. 

Then, we need to use more words to describe the two allowed options. Instead of 

> ‘application-rate’ | ‘measured-runoff’

How about something like

> The value for load_type_p may be one of two text strings, either 'application-rate' or 'measured-runoff'.

And a minor point, the following option descriptions say:

> application-rate: Treat the load **_values_** as nutrient application rates

But for both of these shouldn't it say "Treat the load **_value_**" (singular value), since the load type is defined for each individual value?

dcdenu4:
## Description
Update NDR load type text.

The load type columns are currently `spec.OptionStringInput`, which has a type name of `option`. @newtpatrol commented that this can be confusing for a text input to a CSV. This PR does not handle that. If it's important we could change this to a `spec.StringInput` with a type name of `text` and add validation to ensure only the two sequences are allowed.

@newtpatrol or @jagoldstein any more thoughts on this?

Fixes #2116 

## Checklist

### DIFF
--- a/src/natcap/invest/ndr/ndr.py
+++ b/src/natcap/invest/ndr/ndr.py
@@ -98,20 +98,21 @@ MODEL_SPEC = spec.ModelSpec(
                     about=(
                         "Whether the nutrient load in column load_p should be treated as"
                         " nutrient application rate or measured contaminant runoff."
-                        " 'application-rate' | 'measured-runoff'"
+                        " The value for load_type_p may be one of two text strings,"
+                        " either 'application-rate' or 'measured-runoff'."
                     ),
                     required="calc_p",
                     options=[
                         spec.Option(
                             key="application-rate",
                             about=(
-                                "Treat the load values as nutrient application rates"
+                                "Treat the load value as nutrient application rates"
                                 " (e.g. fertilizer, livestock waste, ...).The model will"
                                 " adjust the load using the application rate and"
                                 " retention efficiency: load_p * (1 - eff_p).")),
                         spec.Option(
                             key="measured-runoff",
-                            about="Treat the load values as measured contaminant runoff.")
+                            about="Treat the load value as measured contaminant runoff.")
                     ]
                 ),
                 spec.OptionStringInput(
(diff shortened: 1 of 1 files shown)


davemfish:
Here's the current rendering of the UG:

-------------
load_type_p ([option](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/input_types.html#option), conditionally required): Whether the nutrient load in column load_p should be treated as nutrient application rate or measured contaminant runoff. ‘application-rate’ | ‘measured-runoff’

Options:
* application-rate: Treat the load values as nutrient application rates (e.g. fertilizer, livestock waste, …).The model will adjust the load using the application rate and retention efficiency: load_p * (1 - eff_p).
* measured-runoff: Treat the load values as measured contaminant runoff.
--------------
Instead of adding the custom text to each specific item in the spec (e.g. `load_type_p.about`), maybe we should replace the word "Options:" with "Values must be one of the following text strings:" And then perhaps add quotes around the values to set them apart from their descriptions (e.g. "application-rate").

Would that clear it up? It would apply to all `OptionStringInputs` across invest.

Issue #2170, opened 2025-09-29
Move redundant imports out of test cases


emlys:
In most of our test suites we have followed a pattern of importing the modules under test within each test case method, rather than at the top of the test module. This may have been for a historic reason, I'm not sure. But with the way our tests currently run, I am pretty sure that repeating the import in each test function is redundant and has no effect. Pytest runs all tests in the same interpreter by default, so the module will be imported once and subsequent import statements don't do anything. 

I think it's worth going through and moving all these imports to the module level. Doing that will remove the temptation to think that each test is running in a fresh interpreter when it really isn't. It will also remove a lot of lines of code.

phargogh:
With how we are using tests currently in GHA, I think it'd be just fine to move the imports to the module level.

A long time ago when we were building up our first test suite for InVEST, we were running our tests on a self-hosted Jenkins server that had a graph of tests over time, where the graph showed the total number of tests over time, broken down into tests passed, skipped and failed.  Under this setup, the count of tests mattered a lot.

Unfortunately, `nosetests` could only detect the correct number of tests if it could import the test module.  This means that if a module-level import wouldn't run, then the graph would suddenly show a dip in the number of total tests instead of reporting that N tests failed.  Having each test method importing the module under test, even if it failed, allowed us to get an accurate count of how many tests failed.

It's possible that pytest does some more intelligent AST inspection of test source code to determine which methods should be run, but we also just don't have a comparable graph of test failures now.  Plus, with how we maintain good test coverage now with our PR process, such a graph wouldn't even be as important or motivating as it once was.

emlys:
Thanks for that background, @phargogh!

PR #2191, opened 2025-10-10
Validate all data types in CSV columns

### LINKED ISSUE #2175: Validate all types of data in CSV columns
When we validate a CSV input, we validate the column values themselves only for spatial inputs. I don't see any reason why we shouldn't also validate all other types of values - numbers, strings, etc. This should be a fairly simple update to `CSVInput.get_validated_dataframe`. 

The only tricky part might be handling column values that are optional.

emlys:
## Description
Fixes #2175

The main change in this PR is that we now call the `validate` method on all columns in a CSV, not just spatial filepaths.

As a result of that change, some custom validation in various models is now unnecessary and could be removed. The test cases for that custom validation are now arguably redundant to the regular validation tests, but I left them in for now to be sure that everything still works.

In order to validate conditionally-required columns (which were previously unhandled), `get_validated_dataframe` now takes an optional `args` parameter which is used to evaluate the conditional requirement.

At one point while I was working on this, I thought I needed to move the validation messages into a different module to avoid a circular import. I made a new module `validation_messages` and updated everywhere to point to that. In the final implementation, that change wasn't really necessary, but I still think it's an improvement so I left it in. It also removes the need for the (questionable) `get_message` function.

I also did #2199 in this PR (removing the legend from the CBC preprocessor output) because it simplifies validation for that CSV, which would otherwise be non-standard.

## Checklist

### DIFF
--- a/src/natcap/invest/crop_production_percentile.py
+++ b/src/natcap/invest/crop_production_percentile.py
@@ -1013,9 +1013,9 @@ def tabulate_results(
         for yield_percentile_id in sorted(yield_percentile_headers) + [
             'yield_observed']]
 
-    # Since pixel values in observed and percentile rasters are Mg/(ha•yr),
-    # raster sums are (Mg•px)/(ha•yr). Before recording sums in
-    # production_lookup dictionary, convert to Mg/yr by multiplying by ha/px.
+    # Since pixel values in observed and percentile rasters are Mg/ha,
+    # raster sums are (Mg•px)/ha. Before recording sums in
+    # production_lookup dictionary, convert to Mg by multiplying by ha/px.
 
     with open(target_table_path, 'w') as result_table:
         result_table.write(
@@ -1130,9 +1130,9 @@ def aggregate_to_polygons(
         target_aggregate_vector_path,
         driver_name='ESRI Shapefile')
 
-    # Since pixel values are Mg/(ha•yr), zonal stats sum is (Mg•px)/(ha•yr).
+    # Since pixel values are Mg/ha, zonal stats sum is (Mg•px)/ha.
     # Before writing sum to results tables or when using sum to calculate
-    # nutrient yields, convert to Mg/yr by multiplying by ha/px.
+    # nutrient yields, convert to Mg by multiplying by ha/px.
 
     # loop over every crop and query with pgp function
     total_yield_lookup = {}
--- a/src/natcap/invest/crop_production_regression.py
+++ b/src/natcap/invest/crop_production_regression.py
@@ -993,9 +993,9 @@ def tabulate_regression_results(
         for nutrient_id in _EXPECTED_NUTRIENT_TABLE_HEADERS
         for mode in ['modeled', 'observed']]
 
-    # Since pixel values in observed and percentile rasters are Mg/(ha•yr),
-    # raster sums are (Mg•px)/(ha•yr). Before recording sums in
-    # production_lookup dictionary, convert to Mg/yr by multiplying by ha/px.
+    # Since pixel values in observed and percentile rasters are Mg/ha,
+    # raster sums are (Mg•px)/ha. Before recording sums in
+    # production_lookup dictionary, convert to Mg by multiplying by ha/px.
 
     with open(target_table_path, 'w') as result_table:
         result_table.write(
@@ -1101,9 +1101,9 @@ def aggregate_regression_results_to_polygons(
         target_aggregate_vector_path,
         driver_name='ESRI Shapefile')
 
-    # Since pixel values are Mg/(ha•yr), zonal stats sum is (Mg•px)/(ha•yr).
+    # Since pixel values are Mg/ha, zonal stats sum is (Mg•px)/ha.
     # Before writing sum to results tables or when using sum to calculate
-    # nutrient yields, convert to Mg/yr by multiplying by ha/px.
+    # nutrient yields, convert to Mg by multiplying by ha/px.
 
     # loop over every crop and query with pgp function
     total_yield_lookup = {}
--- a/src/natcap/invest/ndr/ndr.py
+++ b/src/natcap/invest/ndr/ndr.py
@@ -461,23 +461,23 @@ MODEL_SPEC = spec.ModelSpec(
         spec.SingleBandRasterOutput(
             id="load_n",
             path="intermediate_outputs/load_n.tif",
-            about=gettext("Nitrogen load (for surface transport) per pixel"),
+            about=gettext("Nitrogen load (for surface transport)"),
             data_type=float,
-            units=u.kilogram / u.year
+            units=u.kilogram / u.hectare / u.year
         ),
         spec.SingleBandRasterOutput(
             id="load_p",
             path="intermediate_outputs/load_p.tif",
-            about=gettext("Phosphorus load (for surface transport) per pixel"),
+            about=gettext("Phosphorus load (for surface transport)"),
             data_type=float,
-            units=u.kilogram / u.year
+            units=u.kilogram / u.hectare / u.year
         ),
         spec.SingleBandRasterOutput(
             id="modified_load_n",
             path="intermediate_outputs/modified_load_n.tif",
             about=gettext("Raw nitrogen load scaled by the runoff proxy index."),
             data_type=float,
-            units=u.kilogram / u.year
+            units=u.kilogram / u.hectare / u.year
         ),
         spec.SingleBandRasterOutput(
             id="modified_load_p",
@@ -486,7 +486,7 @@ MODEL_SPEC = spec.ModelSpec(
                 "Raw phosphorus load scaled by the runoff proxy index."
             ),
             data_type=float,
-            units=u.kilogram / u.year
+            units=u.kilogram / u.hectare / u.year
         ),
         spec.SingleBandRasterOutput(
             id="ndr_n",
@@ -539,7 +539,7 @@ MODEL_SPEC = spec.ModelSpec(
             path="intermediate_outputs/sub_load_n.tif",
             about=gettext("Nitrogen loads for subsurface transport"),
             data_type=float,
-            units=u.kilogram / u.year
+            units=u.kilogram / u.hectare / u.year
         ),
         spec.SingleBandRasterOutput(
             id="sub_ndr_n",
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -74,6 +74,17 @@ General
   (`#2118 <https://github.com/natcap/invest/issues/2118>`_)
   (`#2217 <https://github.com/natcap/invest/issues/2217>`_)
 
+Forest Carbon Edge Effect
+=========================
+* Corrected the units listed for some intermediate outputs
+  (`#2218 <https://github.com/natcap/invest/issues/2218>`_).
+
+NDR
+===
+* Corrected the units listed for some intermediate outputs
+  (`#2218 <https://github.com/natcap/invest/issues/2218>`_).
+
+
 3.17.0 (2025-10-16)
 -------------------
 
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GIT_SAMPLE_DATA_REPO_REV    := db11c7f31c90e3d1e28e85bcaa10a2aabc56d33b
 
 GIT_TEST_DATA_REPO          := https://bitbucket.org/natcap/invest-test-data.git
 GIT_TEST_DATA_REPO_PATH     := $(DATA_DIR)/invest-test-data
-GIT_TEST_DATA_REPO_REV      := 36a447cc132079ed74952bd66bc0c0856d7da7b8
+GIT_TEST_DATA_REPO_REV      := 7e0a51c18c6637637c183c295b0e2874a18f3dda
 
 GIT_UG_REPO                 := https://github.com/natcap/invest.users-guide
 GIT_UG_REPO_PATH            := doc/users-guide
--- a/README_PYTHON.rst
+++ b/README_PYTHON.rst
@@ -13,8 +13,6 @@ quantify and forecast this return. InVEST enables decision-makers to quantify
 the importance of natural capital, to assess the tradeoffs associated with
 alternative choices, and to integrate conservation and human development.
 
-InVEST is licensed under a permissive, modified BSD license.
-
 For more information, see:
   * `InVEST API documentation <http://invest.readthedocs.io/>`_
   * InVEST on `GitHub <https://github.com/natcap/invest>`__
--- a/src/natcap/invest/forest_carbon_edge_effect.py
+++ b/src/natcap/invest/forest_carbon_edge_effect.py
@@ -262,7 +262,7 @@ MODEL_SPEC = spec.ModelSpec(
                 "Carbon stored in the aboveground biomass carbon pool."
             ),
             data_type=float,
-            units=u.metric_ton
+            units=u.metric_ton / u.hectare
         ),
         spec.SingleBandRasterOutput(
             id="c_below_carbon_stocks",
@@ -271,7 +271,7 @@ MODEL_SPEC = spec.ModelSpec(
                 "Carbon stored in the belowground biomass carbon pool."
             ),
             data_type=float,
-            units=u.metric_ton
+            units=u.metric_ton / u.hectare
         ),
         spec.SingleBandRasterOutput(
             id="c_dead_carbon_stocks",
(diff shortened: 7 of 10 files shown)


src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -232,9 +232,11 @@
                 ),
                 spec.NumberInput(
                     id="biomass-half-life",
-                    about=gettext("The half-life of carbon in the biomass pool."),
+                    about=gettext(
+                        "The half-life of carbon in the biomass pool. A value of 0 "
+                        "indicates no emissions."),
                     units=u.year,
-                    expression="value > 0"
+                    expression="value >= 0"
```
emlys, line 239: A comment in the code indicates that this is how 0s are interpreted, and the code is set up to handle values of 0.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -1861,17 +1865,6 @@ def _read_transition_matrix(transition_csv_path, biophysical_df):
...
     for index, row in table.iterrows():
-        # If the user is using the template, all rows have some sort of values
-        # in them until the blank row before the legend.  If we find that row,
-        # we can break out of the loop.
-        if row.isnull().all():
-            LOGGER.info(f"Halting transition table parsing on row {index}; "
-                        "blank line encountered.")
-            break
-
-        # skip rows starting with a blank cell, these are part of the legend
```
emlys, line 1872: See https://github.com/natcap/invest/issues/2199, we're no longer putting a legend in the template (and we already aren't using this logic because the template would fail validation).

src/natcap/invest/coastal_blue_carbon/preprocessor.py
```diff
@@ -454,15 +454,6 @@ def _create_transition_table(landcover_df, lulc_snapshot_list,
                 row.append(column_value)
             csv_file.write(','.join(row) + '\n')
 
-        # Append legend
-        csv_file.write("\n,legend")
```
emlys, line 458: Removing the legend (#2199)

src/natcap/invest/urban_cooling_model.py
```diff
@@ -1610,25 +1609,4 @@ def validate(args, limit_to=None):
...
-        if args['cc_method'] == 'factors':
-            spec.get_column('shade').required = True
-            spec.get_column('albedo').required = True
-        else:
-            # args['cc_method'] must be 'intensity'.
-            # If args['cc_method'] isn't one of these two allowed values
-            # ('intensity' or 'factors'), it'll be caught by
-            # validation.validate due to the allowed values stated in
-            # MODEL_SPEC.
-            spec.get_column('building_intensity').required = True
```
emlys, line 1628: These conditionally-required columns are validated by the default validation, so this block is redundant.

src/natcap/invest/urban_flood_risk_mitigation.py
```diff
@@ -950,25 +950,4 @@ def validate(args, limit_to=None):
...
-        cn_df = MODEL_SPEC.get_input(
-            'curve_number_table_path').get_validated_dataframe(
-            args['curve_number_table_path'])
-        # Check for NaN values.
-        nan_mask = cn_df.isna()
-        if nan_mask.any(axis=None):
-            nan_lucodes = nan_mask[nan_mask.any(axis=1)].index
-            # Convert numpy dtype values to native python types
-            lucode_list = [i.item() for i in nan_lucodes.values]
-            validation_warnings.append((
```
emlys, line 970: This is now validated automatically because now we don't allow NA values in these columns.

src/natcap/invest/wave_energy.py
```diff
@@ -778,21 +778,6 @@ def execute(args):
...
-        # Create a grid_land_df dataframe for later use in valuation
-        grid_land_df = MODEL_SPEC.get_input(
-            'land_gridPts_path').get_validated_dataframe(args['land_gridPts_path'])
-        missing_grid_land_fields = []
-        for field in ['id', 'type', 'lat', 'long', 'location']:
-            if field not in grid_land_df.columns:
-                missing_grid_land_fields.append(field)
-
-        if missing_grid_land_fields:
-            raise ValueError(
```
emlys, line 792: This was redundant to the default validation.

src/natcap/invest/coastal_blue_carbon/coastal_blue_carbon.py
```diff
@@ -2068,24 +2062,4 @@ def validate(args, limit_to=None):
...
-            '[LULC CODE]').list_options()
-        transitions_df = transitions_spec.get_validated_dataframe(
-            args['landcover_transitions_table'])
-        transitions_mask = ~transitions_df.isin(transition_options) & ~transitions_df.isna()
-        if transitions_mask.any(axis=None):
-            transition_numpy_mask = transitions_mask.values
-            transition_numpy_values = transitions_df.to_numpy()
-            bad_transition_values = list(
-                numpy.unique(transition_numpy_values[transition_numpy_mask]))
-            validation_warnings.append((
```
emlys, line 2085: Now redundant to the default validation.

src/natcap/invest/crop_production_percentile.py
```diff
@@ -661,16 +661,6 @@ def update_unique_lucodes_in_raster(unique_codes, block):
             "The following lucodes are in the landcover raster but aren't "
             f"in the landcover to crop table: {lucodes_missing_from_table}")
 
-    LOGGER.info("Checking that crops are supported by the model.")
-    user_provided_crop_names = set(list(crop_to_landcover_df.index))
-    valid_crop_names = set([crop.key for crop in CROP_OPTIONS])
-    invalid_crop_names = user_provided_crop_names.difference(valid_crop_names)
```
emlys, line 667: Redundant to the default validation.

src/natcap/invest/habitat_quality.py
```diff
@@ -1288,41 +1273,6 @@ def validate(args, limit_to=None):
 
             invalid_keys.add('sensitivity_table_path')
 
-        # check that max_dist and weight values are included in the
-        # threats table and that max_dist >= 0
-        invalid_max_dist = []
-        missing_max_dist = []
-        missing_weight = []
-        for threat, row in threat_df.iterrows():
```
emlys, line 1296: Redundant to the new default validation.

src/natcap/invest/spec.py
```diff
@@ -757,10 +795,16 @@ def check_row_and_column_types(self):
             PercentInput, RasterOrVectorInput, RatioInput, FileInput,
             SingleBandRasterInput, StringInput, VectorInput, CSVInput}
         for row in (self.rows or []):
-            if type(row) not in allowed_types:
+            for input_type in allowed_types:
+                if isinstance(row, input_type):
```
emlys, line 797: Allow rows/column types to be subclasses of the allowed types.

src/natcap/invest/spec.py
```diff
@@ -875,45 +928,36 @@ def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}):
...
-                LOGGER.error(f"The column '{index_col}' could not be found "
-                             f"in the table {csv_path}")
-                raise
+            if index_col not in df.columns:
+                # If 'index_col' is not a column then KeyError is raised for
+                # using it as the index column
+                raise KeyError(f"The column '{index_col}' could not be found "
+                               f"in the table {csv_path}")
+            if any(df[index_col].duplicated()):
+                raise ValueError(
```
emlys, line 964: Previously, this case would raise an unhandled exception that did not clearly describe the problem.

src/natcap/invest/spec.py
```diff
@@ -875,45 +928,36 @@ def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}):
                     raise ValueError(
                         f'Value(s) in the "{col}" column could not be interpreted '
                         f'as {type(col_spec).__name__}s. Original error: {err}')
+                if col_spec.id not in self.na_allowed and any(df[col].isna()):
```
emlys, line 937: Check for null values if they are not allowed

tests/test_ufrm.py
```diff
@@ -218,26 +218,6 @@ def test_ufrm_value_error_on_bad_lucode(self):
             f'{[16, 17, 18, 21]}')
         self.assertEqual(expected_message, actual_message)
 
-    def test_ufrm_explicit_zeros_in_table(self):
```
emlys, line 221: Curve number values are no longer allowed to be 0.

tests/test_validation.py
```diff
@@ -37,6 +37,8 @@
     StringInput,
     VectorInput,
     Option)
+from natcap.invest import validation
```
emlys, line 40: Moved these to top-level imports (see #2170) rather than having to update a lot of them to point to `validation_messages`.

src/natcap/invest/crop_production_regression.py
```diff
@@ -1183,44 +1176,6 @@ def aggregate_regression_results_to_polygons(
             aggregate_table.write('\n')
 
 
-def get_full_path_from_crop_table(
```
emlys, line 1186: Redundant to the default CSV processing.

src/natcap/invest/ndr/ndr.py
```diff
@@ -1508,16 +1496,6 @@ def _calculate_load(
     eff_key = f'eff_{nutrient_type}'
     load_type_key = f'load_type_{nutrient_type}'
 
-    # Raise ValueError if unknown load_type
```
dcdenu4, line 1511: I'm surprised I didn't have this in the validate function instead. I can't think of a reason why I would have preferred it here except to also benefit Python users. Huh.

src/natcap/invest/annual_water_yield.py
```diff
@@ -791,13 +792,6 @@ def execute(args):
 
     for lulc_code, row in bio_df.iterrows():
         Kc_dict[lulc_code] = row['kc']
-
-        # Catch invalid LULC_veg values with an informative error.
-        if row['lulc_veg'] not in set([0, 1]):
-            # If the user provided an invalid LULC_veg value, raise an
-            # informative error.
-            raise ValueError(
-                f'LULC_veg value must be either 1 or 0, not {row["lulc_veg"]}')
```
dcdenu4, line 800: It makes sense to me to remove redundant checks that were being done in validation. Do we also want to remove checks that were being done in the model itself? I'm wondering if some of these throughout the models will have been here for the Python users? I know we've always been a little fuzzy on that.

src/natcap/invest/habitat_quality.py
```diff
@@ -1268,7 +1253,7 @@ def validate(args, limit_to=None):
         # Get CSVs as dictionaries and ensure the key is a string for threats.
         threat_df = MODEL_SPEC.get_input(
             'threats_table_path').get_validated_dataframe(
-            args['threats_table_path']).fillna('')
+            args['threats_table_path'], args=MODEL_SPEC.preprocess_inputs(args)).fillna('')
```
dcdenu4, line 1256: What's the need to call preprocess inputs here, instead of just passing in `args=args`?

src/natcap/invest/spec.py
```diff
@@ -391,6 +393,35 @@ def preprocess(self, value):
             return utils._GDALPath.from_uri(value).to_normalized_path()
         return None  # if None or empty string, return None
 
+    @staticmethod
+    def format_column(col: pandas.Series, base_path: str) -> pandas.Series:
```
dcdenu4, line 396: I completely missed that we were doing some type hinting now!

src/natcap/invest/spec.py
```diff
@@ -728,6 +762,9 @@ class CSVInput(FileInput):
     """The header name of the column to use as the index. When processing a
     CSV file to a dataframe, the dataframe index will be set to this column."""
 
+    na_allowed: list[str] = []
```
dcdenu4, line 764: Nice.

src/natcap/invest/spec.py
```diff
@@ -798,7 +841,7 @@ def validate(self, filepath: str):
             except Exception as e:
                 return str(e)
 
-    def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}):
+    def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}, args=None):
```
dcdenu4, line 843: I think we need to update the docstring "Args" to include "args"

src/natcap/invest/spec.py
```diff
@@ -861,12 +903,29 @@ def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}):
...
             matching_cols = [c for c in available_cols if re.fullmatch(pattern, c)]
-            if col_spec.required is True and '[' not in col_spec.id and not matching_cols:
-                raise ValueError(get_message('MATCHED_NO_HEADERS').format(
-                    header=axis,
-                    header_name=col_spec.id))
+            if not isinstance(col_spec.required, bool):
+                raise ValueError(
+                    'get_validated_dataframe called with spec that contains a '
+                    'conditional "required" string. The "required" attribute must '
+                    'be evaluated to a bool before calling this method.')
```
dcdenu4: Is this only relevant to this function? Meaning, for:
>The "required" attribute must be evaluated to a bool before calling this method.

Is this expectation that this could happen outside this function and not only in the `if args` block above? I guess I'm trying to understand when we'd get here, when this function was called without passing `args`?

src/natcap/invest/urban_cooling_model.py
dcdenu4: Awesome!

src/natcap/invest/urban_flood_risk_mitigation.py
dcdenu4: Sweet.

src/natcap/invest/validation.py
```diff
@@ -15,60 +15,15 @@
 from osgeo import osr
 
 from . import gettext
+from . import spec
```
dcdenu4: I couldn't see where `spec` was being used in any of the additions. Is this import needed?

tests/test_seasonal_water_yield_regression.py
```diff
@@ -230,10 +230,10 @@ def make_biophysical_csv(biophysical_csv_path):
             'lucode,Description,CN_A,CN_B,CN_C,CN_D,Kc_1,Kc_2,Kc_3,Kc_4,')
         open_table.write('Kc_5,Kc_6,Kc_7,Kc_8,Kc_9,Kc_10,Kc_11,Kc_12\n')
 
-        open_table.write('0,"lulc 1",50,50,0,0,0.7,0.7,0.7,0.7,0.7,0.7,0.7,')
+        open_table.write('0,"lulc 1",50,50,1,1,0.7,0.7,0.7,0.7,0.7,0.7,0.7,')
```
dcdenu4, line 233: Write, we don't allow curve numbers to be 0 now. Nice.

src/natcap/invest/validation.py
emlys: Ah you're right, it's not being used in this final iteration

src/natcap/invest/spec.py
emlys: This could happen if the function was called without passing in `args` when it's needed. The phrasing _"The "required" attribute must be evaluated to a bool"_ comes from a pattern we use in a couple of models, and in `validation.py`, where we create a copy of the spec and then overwrite its `required` attributes with the evaluated value. But it would be nice to move away from that pattern - I'll see if we can rework that now. And I'll add a comment clarifying this code block.

emlys: I realized that we can do without the custom error and instead rely on `utils.evaluate_expression` to raise an error if a string can't be evaluated with the given args. Updated code and docstring.

src/natcap/invest/spec.py
emlys: I don't necessarily want to commit to doing it everywhere, I guess I just thought it made sense here. Can remove it if we'd rather be consistent!

src/natcap/invest/habitat_quality.py
emlys: It ensures that all of the arg keys are actually present. In this case, the `fut_path` and `bas_path` columns are conditionally required depending on whether the `lulc_fut_path` and `lulc_bas_path` inputs are provided. If they're not provided, those keys may be absent from `args`, which would cause an error when evaluated. `preprocess_inputs` ensures that those keys will exist (with a value of `None` if absent).

src/natcap/invest/spec.py
dcdenu4: I think it's fine to keep.

dcdenu4 (commented):
Thanks @emlys , I think this is good to go but there's a conflict with History. I was also curious if you tested this against invest autotest locally?

emlys:
I fixed the history conflict, and the autotests passed locally. The failing check appears to be random and unrelated.

dcdenu4 (approved):

PR #2220, opened 2025-10-29
Bugfixes for incomplete metadata

### LINKED ISSUE #2118: Generate metadata for outputs with variable names
When we run a model via the CLI, it calls `generate_metadata_for_outputs`, which iterates through the output spec and generates metadata for all files within the workspace matching paths in the spec. Right now, it's looking for an exact match to the path specified in the output spec, but this means it skips files containing a variable in the filename. 

For example, the pollination model output spec contains:
```
        spec.SingleBandRasterOutput(
            id="pollinator_abundance_[SPECIES]_[SEASON]",
            path="pollinator_abundance_[SPECIES]_[SEASON].tif",
            about=gettext("Abundance of pollinator SPECIES in season SEASON."),
            data_type=float,
            units=None
        ),
```
When `generate_metadata_for_outputs` iterates through the spec, it looks for a file matching the exact path `{workspace_dir}/pollinator_abundance_[SPECIES]_[SEASON].tif` but finds none, since the actual filenames generated during the model run end up being, e.g. `{workspace_dir}/pollinator_abundance_apis_spring.tif`. As such, no metadata is generated for these files. 

It would be handy if we could also generate metadata for outputs matching a pattern, in the cases where the filename contains a variable portion.

### LINKED ISSUE #2217: Metadata is missing info from model spec
The `about` and `units` attributes of these fields are not making it into the metadata documents.
```python
        spec.VectorOutput(
            id="wave_energies",
            path="intermediate/wind_wave/wave_energies.gpkg",
            about=gettext(
                "Shore points with associated wave energy data"
            ),
            geometry_types={"POINT"},
            fields=[
                spec.NumberOutput(
                    id="E_ocean",
                    about=gettext(
                        "Average power of oceanic waves at the point"
                        " (equation (18)"
                    ),
                    units=u.kilowatt / u.meter
                ),
```
```
data_model:
  layers:
  - name: wave_energies
    table:
      fields:
      - name: E_ocean
        type: Real
        description: ''
        title: ''
        units: ''
```

davemfish:
I updated the metadata generating function to use the file registry returned by `execute` as the source of truth for which files exist and need metadata. And addressed a couple bugs in the process.

Fixes #2217 
Fixes #2118 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -61,9 +61,16 @@
   7. InVEST model Z (model names should be sorted A-Z)
 
 
-..
-  Unreleased Changes
-  ------------------
+
+Unreleased Changes
+------------------
+
+General
+=======
+* Fixed bugs in incomplete metadata creation.
+  (`#2118 <https://github.com/natcap/invest/issues/2118>`_)
+  (`#2217 <https://github.com/natcap/invest/issues/2217>`_)
+
 
 3.17.0 (2025-10-16)
 -------------------


claire-simpson (approved):
Thanks @davemfish, this looks good!

PR #2225, opened 2025-10-31
Correct `created_if` conditions for several outputs

### LINKED ISSUE #2070: Unused output spec entry in SDR?
In working on updating the invest-sdr-usle-c-raster plugin to the latest Model Spec standard, I noticed that SDR has `weighted_avg_aspect.fit` in the output spec, but it doesn't appear to be used.

emlys:
I tested by temporarily setting `check_outputs=True` and then running the test suite and `invest_autotest`. This revealed a bunch of places where the `created_if` condition was not correct. 

I have fixed them for all outputs except for Wind Energy's `bathymetry_path`. That one is created conditionally on a complex condition that we can't define in the model spec (it's only created if an input has non-square pixels). I'm not sure what the best solution is for that.

Also fixes #2070 

## Checklist

### DIFF
--- a/src/natcap/invest/cli.py
+++ b/src/natcap/invest/cli.py
@@ -477,7 +477,7 @@ def main(user_args=None):
                 create_logfile=True,
                 generate_metadata=True,
                 save_file_registry=True,
-                check_outputs=True)
+                check_outputs=False)
 
         if args.subcommand == 'serve':
             ui_server.app.run(port=args.port)
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -2142,7 +2142,7 @@ class ModelSpec(BaseModel):
 
     def execute(self, args, create_logfile=False, log_level=logging.NOTSET,
             generate_metadata=False, save_file_registry=False,
-            check_outputs=True):
+            check_outputs=False):
         """Invest model execute function wrapper.
 
         Performs additonal work before and after the execute function runs:


src/natcap/invest/wave_energy/wave_energy.py
```diff
@@ -1675,9 +1667,7 @@ def _mask_below_start_value(array):
     table_df = pandas.DataFrame(table_dict)
 
     # Write dataframe to csv, with columns in designated sequence
-    base_attribute_table_path = os.path.splitext(target_raster_path)[0]
-    attribute_table_path = base_attribute_table_path + '.csv'
-    table_df.to_csv(attribute_table_path, index=False, columns=column_names)
```
emlys, line 1667: Modified this function to pass in the CSV path from the file registry rather than creating it from the raster name.

src/natcap/invest/coastal_vulnerability/coastal_vulnerability.py
```diff
@@ -709,16 +710,6 @@ def get_vector_colnames(vector_path):
             path="intermediate/shore_points/landmass_line_index.pickle",
             about=gettext("Pickled landmass index")
         ),
-        spec.FileOutput(
```
emlys, line 712: These outputs are the index files where the rtree data structure is stored on disk. I changed them to be created in a temporary directory so they're no longer outputs of the model. They weren't being reused between model runs, and not likely to be useful for debugging.

PR #2232, opened 2025-11-04
Feature: invest reports

davemfish:
This PR integrates model report-generation and viewing with invest. At the moment, report generators are in a separate package (`invest_reports`). A subsequent PR will move that package into `natcap.invest`.

- Each model's `MODEL_SPEC` can have a `reporter` attribute.
- `ModelSpec.execute` can call the reporter with `generate_report=True`, if the reporter module exists
- The invest CLI `run` command will generate reports
- The Workbench will display a "View Results" button in addition to the "Open Workspace" button when a model completes, if the report html file exists in the workspace.
- The Workbench tracks the report file as part of an `InvestJob`, so results from past completed runs are available to view.
- For each model, at least one regression test will be updated to call `MODEL_SPEC.execute` instead of the `execute` function. If that model has a reporter, the report will be generated and the test will assert the html file exists.
- Tests for the report content and functionality of the reporters & templates are in `invest_reports`.

Partially addresses #2249 

**EDIT:** ~~The Python 3.9 tests are failing due to an incompatibility, but we don't need to continue supporting 3.9 (#2269 ).~~  
We're no longer testing against Python 3.9. But Windows 3.14 tests are failing to setup the python environment. The issue probably stems from the fact that we're temporarily pip-installing `invest-reports` from source from github. It's working in most cases, but not Windows & Python 3.14. Since that's a temporary dependency anyway, I may not try to fix it right now.

## Checklist

### DIFF
--- a/tests/test_annual_water_yield.py
+++ b/tests/test_annual_water_yield.py
@@ -178,8 +178,7 @@ class AnnualWaterYieldTests(unittest.TestCase):
         }
         annual_water_yield.MODEL_SPEC.execute(args, **execute_kwargs)
         assert_complete_execute(
-            args['workspace_dir'], args, annual_water_yield.MODEL_SPEC,
-            **execute_kwargs)
+            args, annual_water_yield.MODEL_SPEC, **execute_kwargs)
 
         raster_results = ['aet_test.tif', 'fractp_test.tif', 'wyield_test.tif']
         for raster_path in raster_results:


tests/test_annual_water_yield.py
```diff
@@ -170,7 +172,14 @@ def test_water_yield_subshed(self):
         args['sub_watersheds_path'] = os.path.join(
             SAMPLE_DATA, 'subwatersheds.shp')
         args['results_suffix'] = 'test'
-        annual_water_yield.execute(args)
+        execute_kwargs = {
+            'generate_report': bool(annual_water_yield.MODEL_SPEC.reporter),
+            'save_file_registry': True
+        }
+        annual_water_yield.MODEL_SPEC.execute(args, **execute_kwargs)
+        assert_complete_execute(
```
davemfish, line 180: If this test pattern makes sense, I'll add it to all the other models also. For now I added it to the three models that have reporters, and to this model, which does not, to serve as an example.

src/natcap/invest/ndr/ndr.py
```diff
@@ -28,6 +28,7 @@
     model_id="ndr",
     model_title=gettext("Nutrient Delivery Ratio"),
     userguide="ndr.html",
+    reporter="invest_reports.jinja_report_generators.ndr_report_generator",
```
davemfish, line 31: I think these `reporter` attributes will be the only thing we need to update again from this PR when we move the `invest_reports` package into `natcap.invest`.

src/natcap/invest/spec.py
```diff
@@ -1956,6 +1956,12 @@ class ModelSpec(BaseModel):
     module_name: str
     """The importable module name of the model e.g. ``natcap.invest.foo``."""
 
+    reporter: str = ''
```
emlys, line 1959: I think it could make sense to add a pydantic validator function that imports this module, to make sure that it is correctly spelled and importable.

src/natcap/invest/cli.py
```diff
@@ -472,12 +472,25 @@ def main(user_args=None):
             # Exceptions will already be logged to the logfile but will ALSO be
             # written to stdout if this exception is uncaught.  This is by
             # design.
+
+            # Not all models will have a reporter; that's okay.
+            has_reporter = False
+            if model_module.MODEL_SPEC.reporter:
+                try:
+                    reporter_module = importlib.import_module(
```
emlys: If we add a pydantic validator like I mentioned in another comment, I think this try/except would not be necessary, and `has_reporter` could be `bool(model_module.MODEL_SPEC.reporter)`.

src/natcap/invest/spec.py
```diff
@@ -2141,8 +2147,8 @@ def setup(self, args, taskgraph_key='taskgraph_cache'):
         return args, file_registry, graph
 
     def execute(self, args, create_logfile=False, log_level=logging.NOTSET,
-            generate_metadata=False, save_file_registry=False,
-            check_outputs=False):
+                generate_metadata=False, save_file_registry=False,
+                check_outputs=False, generate_report=False):
```
emlys, line 2165: Is there ever a situation where we don't want to generate a report for a model that does have a reporting module?

davemfish: Probably not? I would be in favor of changing all of these to `True` by default. Except for `check_outputs`.

src/natcap/invest/spec.py
davemfish: That seems like a good idea!

src/natcap/invest/spec.py
emlys: In that case, would it make sense to determine whether to generate the report within this function, ie. `if self.reporter`, rather than passing it in as a flag?

src/natcap/invest/spec.py
davemfish: Updated in [6e1bdf1](https://github.com/natcap/invest/pull/2232/commits/6e1bdf1539a84b78d5506a6202aae8eefbe0fdff)

src/natcap/invest/cli.py
davemfish: Updated in [6e1bdf1](https://github.com/natcap/invest/pull/2232/commits/6e1bdf1539a84b78d5506a6202aae8eefbe0fdff)

src/natcap/invest/spec.py
davemfish: Well, I think we must have decided at one point that it made sense to have all of these as user-facing options. I don't see the report as different from the metadata or file registry in that respect, so I was just following the precedent we already set.

davemfish: For now, I'm inclined to keep these args as-is. I think the question of what the defaults should be, or whether or not there are user-facing options, is not strictly about this reports feature. So I'd like to address it separately.

tests/test_spec.py
```diff
@@ -504,3 +505,27 @@ def test_option_string_input_preprocess(self):
...
+            'model_id': 'foo',
+            'model_title': 'Foo',
+            'userguide': '',
+            'input_field_order': [],
+            'inputs': [],
+            'outputs': [],
+            'module_name': 'foo',
+        }
+
+        _ = spec.ModelSpec(**data, reporter='')
```
dcdenu4, line 526: Is it worth testing whether `reporter` is included at all as an attribute?

tests/utils.py
```diff
@@ -0,0 +1,29 @@
...
+def assert_complete_execute(workspace_dir, raw_args, model_spec, **kwargs):
+    """Assert that post-processing functions completed.
+
+    This assertion can be used after calling ``model_spec.execute`` with
+    various options to assert that expected files exist.
+
+    Args:
+        workspace_dir (str): path to invest model output workspace
+        raw_args (dict): the args dict passed to ``execute``
+        model_spec (natcap.invest.spec.ModelSpec)
```
dcdenu4: does this need a `:` and one liner description?

tests/utils.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+
+def assert_complete_execute(workspace_dir, raw_args, model_spec, **kwargs):
```
dcdenu4: I was wondering if this should live in `utils` or `test_utils`. It seems like it could be helpful as a general invest api function but wanted to check if that was indeed the intent.

davemfish: It's in `tests/utils.py`, so it is not part of `natcap.invest`.

`test_utils.py` are actually tests of `natcap.invest.utils`. Kinda confusing.

tests/test_spec.py
davemfish: Hmm, this is just a test of the field validator, not of the spec itself. And when the field validator runs, there will always be a `reporter` attribute because it's defined to get a default value if it is missing. I think the general Pydantic validation that happens whenever we import a model covers our bases in terms of testing MODEL_SPECs with and without a `reporter` attribute.

tests/utils.py
dcdenu4: Oops, sorry I didn't catch that!

tests/utils.py
davemfish: updated

PR #2234, opened 2025-11-04
Add getter methods to CSVOutput and VectorOutput

### LINKED ISSUE #2216: `VectorOutput` and `CSVOutput` lack a `get_field` method
Tabular outputs such as `VectorOutput` and `CSVOutput` would benefit from a `get_field` method. The corresponding `*Input` types already have this. Here's a use-case:

```python
model_spec.get_output('wave_energies').get_field('E_ocean').units)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dmf\projects\invest\env\Lib\site-packages\pydantic\main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'VectorOutput' object has no attribute 'get_field'
```

I'm not sure if it is best to simply add the method to these types, or to design some inheritance, either for tabular data structures to share and/or for `*Input/*Output` types to share.

~~Also, a completely reasonable workaround for my use-case is to use `geometamaker` to read the same metadata as can be found in the model spec.~~ https://github.com/natcap/invest/issues/2217
That API looks very similar:
```python
geometamaker.describe(file_registry['wave_energies']).get_field_description('E_ocean').units
```

emlys:
## Description
Fixes #2216 
Adds `get_column` method to `CSVOutput` and `get_field` method to `VectorOutput`.
I also got fed up with the duplication of `rows` and `columns` in the csv classes. So I redesigned that a bit so that the attribute is called `columns` in either case, and the orientation is determined by a separate `orientation` attribute which can be "row" or "column". This only affects Wind Energy, and I don't think any plugins are using row-oriented tables.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -70,6 +70,10 @@ General
 =======
 * Updated codesigning to better handle repeated signing requests.
   (`#2245 <https://github.com/natcap/invest/issues/2247>`_)
+* ``spec.CSVOutput`` now has a ``get_column`` method, and ``spec.VectorOutput``
+  now has a ``get_field`` method. ``CSVInput`` and ``CSVOutput`` no longer have
+  a ``rows`` attribute; ``orientation='row'`` may be used to indicate a row-wise
+  table layout. (`#2216 <https://github.com/natcap/invest/issues/2216>`_)
 
 
 3.17.1 (2025-11-06)
@@ -83,10 +87,6 @@ General
 * Fixed bugs in incomplete metadata creation.
   (`#2118 <https://github.com/natcap/invest/issues/2118>`_)
   (`#2217 <https://github.com/natcap/invest/issues/2217>`_)
-* ``spec.CSVOutput`` now has a ``get_column`` method, and ``spec.VectorOutput``
-  now has a ``get_field`` method. ``CSVInput`` and ``CSVOutput`` no longer have
-  a ``rows`` attribute; ``orientation='row'`` may be used to indicate a row-wise
-  table layout. (`#2216 <https://github.com/natcap/invest/issues/2216>`_)
 
 Forest Carbon Edge Effect
 =========================


src/natcap/invest/spec.py
```diff
@@ -2417,7 +2407,7 @@ def describe_arg_from_name(module_name, *arg_keys):
         elif i > 0 and arg_keys[i - 1] == 'columns':
             spec = spec.get_column(key)
         elif i > 0 and arg_keys[i - 1] == 'rows':
-            spec = spec.get_row(key)
+            spec = spec.get_column(key)
         elif key in {'bands', 'fields', 'contents', 'columns', 'rows'}:
```
davemfish, line 2411: I don't think 'rows' is a possible key anymore, right?

tests/test_validation.py
davemfish, line 1: In `test_spec.py` there are a couple places in the assertions in `test_write_metadata_for_outputs` where the new `get_field` would be useful, instead of indexing into a list. Would you mind updating that for us? Thanks!

```python
        self.assertEqual(
            resource.get_field_description('SUP_DEMadm_cap').description,
            vector_spec.fields[0].about)
        self.assertEqual(
            resource.get_field_description('SUP_DEMadm_cap').units,
            spec.format_unit(vector_spec.fields[0].units))
```

src/natcap/invest/spec.py
```diff
@@ -752,10 +752,11 @@ class CSVInput(FileInput):
     expected to have. The `id` of each input must match the corresponding
     column header."""
 
-    rows: typing.Union[list[Input], None] = None
```
davemfish, line 755: Now that `rows` is not an attribute, should `columns` be required instead of optional? i.e. removing `None` as the default value and acceptable type? The docstring for CSVInput could use updating as well.

src/natcap/invest/spec.py
```diff
@@ -866,29 +849,22 @@ def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}, args=None):
             AssertionError from ``utils.evaluate_expression`` if any conditonal
             requirement string can't be evaluated given the values in ``args``.
         """
-        if not (self.columns or self.rows):
-            raise ValueError('One of columns or rows must be provided')
+        if not self.columns:
+            raise ValueError('Columns must be provided')
```
davemfish, line 853: If the `columns` attribute definition is updated we may just rely on pydantic validation. This check may be obsolete.

src/natcap/invest/spec.py
```diff
@@ -1585,6 +1562,8 @@ class VectorOutput(FileOutput):
     """An iterable of `Output`s representing the fields created in this vector.
     The `key` of each input must match the corresponding field name."""
 
+    _fields_dict: dict[str, Input] = {}
```
davemfish: This should be `dict[str, Output]` I think. Do we also need to use `typing.Dict` in order to set types for keys and values? https://docs.pydantic.dev/2.0/usage/types/dicts_mapping/#dicts-and-mapping-types

src/natcap/invest/spec.py
```diff
@@ -1610,13 +1595,17 @@ class CSVOutput(FileOutput):
...
+    orientation: typing.Literal['column', 'row'] = 'column'
+    """Orientation of the table. Defaults to 'column', meaning that the column
+    headers go horizontally. This is the recommended orientation for most tables.
+    If the table must be oriented row-wise, meaning that the headers go
+    vertically, set this to 'row'."""
 
     index_col: typing.Union[str, None] = None
     """The header name of the column that is the index of the table."""
 
+    _columns_dict: dict[str, Input] = {}
```
davemfish: Same comment about types as for the `VectorOutput`

src/natcap/invest/spec.py
```diff
@@ -1596,6 +1575,12 @@ def check_field_types(self):
                 raise ValueError(f'Field {field} is not an allowed type')
         return self
 
+    def model_post_init(self, context):
+        self._fields_dict = {field.id: field for field in self.fields}
+
+    def get_field(self, key: str) -> Input:
```
davemfish: the type hint here should be `Output` as well.

src/natcap/invest/spec.py
```diff
@@ -1644,6 +1627,13 @@ def check_row_and_column_types(self):
                 raise ValueError(f'Column {col} is not an allowed type')
         return self
 
+    def model_post_init(self, context):
+        if self.columns:
+            self._columns_dict = {col.id: col for col in self.columns}
+
+    def get_column(self, key: str) -> Input:
```
davemfish: type hint needs updating here as well

src/natcap/invest/spec.py
```diff
@@ -2375,7 +2365,7 @@ def describe_arg_from_spec(name, spec):
             indented_block += spec.format_rst()
 
     elif type(spec) is CSVInput:
-        if not spec.columns and not spec.rows:
+        if not spec.columns:
             first_line += gettext(
                 ' Please see the sample data table for details on the format.')
```
davemfish, line 2370: Now this has me wondering if we have a weird HRA table or something that does not have `rows` or `columns`, and maybe my earlier comments about making `columns` a required attribute is not quite possible 🫤

src/natcap/invest/spec.py
emlys: The keys come from the user's guide, which I didn't update for this PR because I'm expecting we'll do away with listing out all the keys in the UG sometime soon. So 'rows' still exists in the UG, we're just mapping it to a different function.

tests/test_validation.py
emlys: Sure!

src/natcap/invest/spec.py
emlys: There's at least one case (HRA criteria table) where we define `columns=None` because the table doesn't follow a standard layout. That disables all the standard column validation etc.

src/natcap/invest/spec.py
emlys: I think this is still relevant in the case where `columns=None`, we can't use this function.

src/natcap/invest/spec.py
emlys: Corrected these to `Output`. It looks like `typing.Dict` was deprecated in python 3.9: https://docs.python.org/3/library/typing.html#typing.Dict

src/natcap/invest/spec.py
emlys: Unfortunately yeah, I would love to do away with that special HRA table some day!

src/natcap/invest/spec.py
davemfish: Oh okay, gotcha.

PR #2237, opened 2025-11-04
Move population resampling helper from Urban Nature Access to utils

### LINKED ISSUE #2155: `_resample_population_raster` function should be more accessible

claire-simpson:
## Description

Moves `_resample_population_raster` function (and associated test) from Urban Nature Access to `utils.py` so the function can be used more widely (e.g., in plugins and new urban health model).

Fixes #2155

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -73,6 +73,10 @@ General
 * Fixed bugs in incomplete metadata creation.
   (`#2118 <https://github.com/natcap/invest/issues/2118>`_)
   (`#2217 <https://github.com/natcap/invest/issues/2217>`_)
+* Moved the ``_resample_population_raster`` helper function from Urban
+  Nature Access to ``utils.py`` for broader reuse, renaming it to
+  ``resample_population_raster`` to indicate public accessibility.
+  (`#2155 <https://github.com/natcap/invest/issues/2155>`_)
 
 Forest Carbon Edge Effect
 =========================


src/natcap/invest/utils.py
```diff
@@ -891,3 +891,118 @@ def format_args_dict(args_dict, model_id):
...
+            in meters.
+        working_dir (string): The path to a directory on disk.  A new directory
+            is created within this directory for the storage of temporary files
+            and then deleted upon successful completion of the function.
+
+    Returns:
+        ``None``
+    """
+    if not os.path.isdir(working_dir):
+        os.makedirs(working_dir)
```
claire-simpson, line 937: > if not os.path.isdir(working_dir):
          os.makedirs(working_dir)

Aside from renaming the arguments from `lulc_pixel_size`, `lulc_bb`, `lulc_projection_wkt` to `target_pixel_size` etc., adding this conditional is the only change I made to the function.

src/natcap/invest/utils.py
```diff
@@ -891,3 +891,116 @@ def format_args_dict(args_dict, model_id):
...
+            Pixel values represent the number of people occupying the pixel.
+            Must be linearly projected in meters.
+        target_population_raster_path (string): The path to where the target,
+            warped population raster will live on disk.
+        target_pixel_size (tuple): A tuple of the pixel size for the target
+            raster.  Passed directly to ``pygeoprocessing.warp_raster``.
+        target_bb (tuple): A tuple of the bounding box for the target raster.
+            Passed directly to ``pygeoprocessing.warp_raster``.
+        target_projection_wkt (string): The Well-Known Text of the target
+            spatial reference fro the target raster.  Passed directly to
```
dcdenu4:
```suggestion
            spatial reference for the target raster.  Passed directly to
```

src/natcap/invest/utils.py
dcdenu4: Cool. Since this does change functionality slightly, before I think you'd get a `FileNotFoundError` if `working_dir` didn't exist, should we add a note to the `working_dir` docstring of something like:
> The path to a directory on disk, if it the path is not a directory one is created at this path

tests/test_utils.py
```diff
@@ -1053,3 +1054,75 @@ def test_print_args(self):
...
+    def setUp(self):
+        """Setup workspace."""
+        self.workspace_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        """Delete workspace."""
+        shutil.rmtree(self.workspace_dir)
+
+    def test_resample_population_raster(self):
+        """UNA: Test population raster resampling."""
```
dcdenu4: Should this be "Utils" instead of "UNA" now? I'm not sure the docstring conventions in utils testing.

dcdenu4 (changes requested):
Thanks @claire-simpson , just a few quick comments / suggestions.

tests/test_utils.py
claire-simpson: Oops yes, good catch!

dcdenu4 (changes requested):
Thanks @claire-simpson , I think I'll be able to use this in the Schisto plugin as well, so nice to have it in utils!

I was about to approve this and then started wondering if we should add a History note. We've talked about this recently, how historically we would only add user facing changes. But I'm starting to lean towards noting all API changes. So I'd be in favor in mentioning this in General, especially since we've moved from this being a function starting with `_`, which in Python indicates a function is not meant for public use and should be considered private.

In C++ and Java you can have true private functions or methods, usually denoted with private or public as part of the function definition.

PR #2250, opened 2025-11-07
Migrate sphinx extension and move RST generating code into class methods

### LINKED ISSUE #2243: RST generating functions should belong to Input classes
The functions we use to generate RST documentation for the users guide, `describe_arg_from_name` and `describe_arg_from_spec`, should be updated to take advantage of the new object-oriented design. This will be useful when we work on generating output documentation too.

### LINKED ISSUE #2246: Migrate `investspec` from user's guide to natcap.invest
The user's guide contains the source code for the `investspec` sphinx extension. I think it should be moved into `natcap.invest` as a new module called `rst_sphinx_ext` or something. This will have the benefit of all rst-related code living in the same repo (I often forget that the `investspec` module exists). And, as part of a proper package, `investspec` won't have to be added to the path in `conf.py`.

emlys:
## Description
This is an effort to set us up for future enhancements to the generated user's guide documentation. The resulting user's guide documentation is unchanged.

Fixes #2243 :  Moved the RST generating code into `Input` methods called `describe_rst`. This makes more sense with the new object-oriented structure, and this way it'll be easier to expand and further customize the documentation for each type.
 
Fixes #2246 : Moving the user's guide sphinx extension into this repo. Adds a new module `rst_generator.py` which defines the `investspec` extension. Code is largely unchanged from how it was in the user's guide repo.

As a result, `docutils` is now a dependency of `natcap.invest`. This is a well established library with no unusual dependencies, so I don't think it's a concern.

I made a minor improvement to how `gettext` translation happens: `gettext` now dynamically references the `LOCALE_CODE` at execution time, which reduces the number of situations in which we have to `importlib.reload` a module in order to switch languages. This was necessary to get the new RST translation test passing.

## Checklist

### DIFF
--- a/src/natcap/invest/rst_generator.py
+++ b/src/natcap/invest/rst_generator.py
@@ -101,7 +101,7 @@ def invest_spec(name, rawtext, text, lineno, inliner, options={}, content=[]):
     The `investspec` extension provides the `:investspec:` role, which can be
     used inline in RST files to insert generated documentation anywhere you want.
 
-    To use in a sphinx project, add 'natcap.invest.investspec' to the list of
+    To use in a sphinx project, add 'natcap.invest.rst_generator' to the list of
     extensions in the conf.py.
 
     Usage:


src/natcap/invest/spec.py
```diff
@@ -306,10 +331,12 @@ class FileInput(Input):
 
     type: typing.ClassVar[str] = 'file'
 
-    display_name: typing.ClassVar[str] = gettext('file')
-
     rst_section: typing.ClassVar[str] = 'file'
 
+    @property
+    def display_name(self):
```
emlys, line 337: `display_name` is now a property method so that the `gettext` call is evaluated in real time using the current locale setting.

src/natcap/invest/rst_generator.py
```diff
@@ -0,0 +1,184 @@
...
+        if i > 0 and arg_keys[i - 1] == 'bands':
+            key = int(key)
+        elif i > 0 and arg_keys[i - 1] == 'fields':
+            spec = spec.get_field(key)
+        elif i > 0 and arg_keys[i - 1] == 'contents':
+            spec = spec.get_contents(key)
+        elif i > 0 and arg_keys[i - 1] == 'columns':
+            spec = spec.get_column(key)
+        elif i > 0 and arg_keys[i - 1] == 'rows':
+            spec = spec.get_column(key)
```
claire-simpson: Is calling `get_columns` on `rows` intentional? If yes, I think a clarifying comment would be helpful!

src/natcap/invest/rst_generator.py
```diff
@@ -0,0 +1,184 @@
...
+        elif key in {'bands', 'fields', 'contents', 'columns', 'rows'}:
+            continue
+        else:
+            try:
+                spec = spec.get(key)
+            except KeyError:
+                keys_so_far = '.'.join(arg_keys[:i + 1])
+                raise ValueError(
```
claire-simpson: Would it be helpful to add a test that asserts this error is raised if an invalid key is passed?

src/natcap/invest/__init__.py
```diff
@@ -37,6 +37,7 @@
 LOCALE_NAME_MAP = {
     locale: babel.Locale(locale).display_name for locale in LOCALES
 }
+LOCALE_CODE = 'en'
```
claire-simpson, line 43: I'm not sure I exactly understand what this is doing. Is the idea that we change this when building the UG for different languages?

claire-simpson: Maybe a comment could help?

src/natcap/invest/spec.py
```diff
@@ -671,6 +734,42 @@ def format_geometry_types_rst(self):
...
+    def describe_rst(self):
+        """Generate RST documentation for this input.
+
+        This is used for documenting:
+            - a single top-level arg
+            - a row or column in a CSV
+            - a field in a vector
+            - an item in a directory
+
+        Args:
```
claire-simpson: Looks like this may be outdated since the only arg is `self`

tests/test_rst_generator.py
```diff
@@ -0,0 +1,79 @@
+import importlib
+import os
+import shutil
+import subprocess
```
claire-simpson: Looks like os, shutil, and subprocess are not used

src/natcap/invest/rst_generator.py
```diff
@@ -0,0 +1,184 @@
+import importlib
+
+from docutils import frontend
+from docutils import nodes
```
claire-simpson:
```suggestion

```

src/natcap/invest/rst_generator.py
```diff
@@ -0,0 +1,184 @@
...
+    # automatically, but lets explicitly replace them here
+    anchor_name = '-'.join(keys).replace('_', '-')
+    rst = '\n\n'.join(_input.describe_rst())
+    return f'.. _{anchor_name}:\n\n{rst}'
+
+
+def invest_spec(name, rawtext, text, lineno, inliner, options={}, content=[]):
+    """Custom docutils role to generate InVEST model input docs from spec.
+
+    This is a custom Sphinx extension that generates documentation of InVEST
```
claire-simpson, line 98: I'm a bit confused by this docstring for `invest_spec`. Maybe its outdated? E.g., this is referring to `'natcap.invest.investspec'`, but isn't this `natcap.invest.rst_generator`?

claire-simpson (changes requested):
Thanks @emlys - generally looks good! I had some clarifying questions and a few requested changes.

src/natcap/invest/rst_generator.py
emlys: Yes it is, that change comes from https://github.com/natcap/invest/pull/2234. I'll add a comment

src/natcap/invest/rst_generator.py
emlys: Sure, I added a test, and also moved the try/except out one layer so it will catch more possible failures.

src/natcap/invest/__init__.py
emlys: Added a comment. `LOCALE_CODE` changes at runtime as needed to update translations to a different language. That matters for anything that uses translated text (UG, validation, server, etc). Previously, it was the `gettext` function defintion that changed at runtime. Now, the `gettext` function definition remains the same, but it reference `LOCALE_CODE` which is dynamic.

src/natcap/invest/rst_generator.py
emlys: You're right, it should say `natcap.invest.rst_generator` instead of `natcap.invest.investspec`. Other than that, I think it's up to date. The docutils role provided by this extension is still called `investspec`.

PR #2259, opened 2025-11-19
Reset release schedule back to Tuesday

emilyanndavis:
## Description
Release schedule was temporarily changed to Wednesday to get around the Tuesday brownouts of GitHub's macos-13 runners. Now that we've upgraded to macos-15-intel, we can reset the schedule back to Tuesday.


claire-simpson (approved):
Thanks @emilyanndavis!

PR #2273, opened 2025-12-10
Implement Urban Mental Health Model Option 2: LULC Inputs

### LINKED ISSUE #2142: Urban Mental Health Model: implement scenario 2 (baseline and alternate LULC)
This model option calculates nature expose (NE) using baseline and scenario (i.e., future/counterfactual) land use and land cover (LULC) maps. The model translates the baseline and alternate LULC maps to baseline and alternate NDVI maps and calculates ΔNE. The user may provide an LULC attribute table to map LULC classes to NDVI values, or may supply a baseline NDVI map and allow InVEST to convert LULC to NDVI (i.e., assign the average or median NDVI value of an LULC class). 

Required inputs:
1. Baseline LULC
2. Alternate LULC
3. LULC attribute table OR baseline NDVI 
(Note: an LULC attribute table should be allowed regardless of NDVI input to allow for LULC classes (e.g., water) to be excluded from analysis)

claire-simpson:
## Description
Implements LULC-based option (scenario 2) in the Urban Mental Health model and unifies NDVI preprocessing across scenarios.

Adds workflow for LULC reclassification to NDVI:
- build LULC-to-NDVI mapping either from attribute table or calculate mean NDVI (from `ndvi_base`) by `lucode` if needed
- auto-masking: map `exclude`d LULC classes to `NODATA`
- write `lulc_to_ndvi_map.csv` to store the final mapping used
- reclassify LULC to mean NDVI

Other changes:

- Refactor preprocessing so LULC and NDVI scenarios share a common align --> mask (and reclassify if `lulc`) --> buffer-mean --> delta_NDVI pipeline.
- Validation for if `ndvi` column and `base_ndvi` both aren't provided for option 2
- Add unit tests

Fixes #2142

## Checklist

### DIFF
--- a/src/natcap/invest/urban_mental_health.py
+++ b/src/natcap/invest/urban_mental_health.py
@@ -1085,15 +1085,15 @@ def _calculate_mean_ndvi_by_lulc_class(lulc_path, ndvi_path, lulc_dict):
 
         # numpy.unique returns unique elements of an array
         # return_inverse returns the indices of unique array
+        # block_counts counts the number of pixels per unique lucode
         # e.g., for lulc array [1, 2, 2, 5]: unique_lucodes = [1, 2, 5]
-        # and inverse_indices = [0, 1, 1, 2]
+        # inverse_indices = [0, 1, 1, 2], and block_counts = [1, 2, 1]
         unique_lucodes, inverse_indices, block_counts = numpy.unique(
             masked_lulc, return_inverse=True, return_counts=True)
 
         # Use bincount to get the NDVI sum for each unique LULC code (in
-        # block) and count the number of pixels for each lucode
-        # e.g., for ndvi_array = [0.2, 0.5, 0.7, 0.1] and same example lulc as
-        # above, block_sums = [0.2, 1.2, 0.1] and block_counts = [1, 2, 1]
+        # block) e.g., for ndvi_array = [0.2, 0.5, 0.7, 0.1] and same example
+        # lulc as above, block_sums = [0.2, 1.2, 0.1]
         block_sums = numpy.bincount(inverse_indices, weights=masked_ndvi)
 
         # accumulate into global sums/counts


src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +602,289 @@ def execute(args):
...
-                  [file_registry['ndvi_base_buffer_mean'],
-                   file_registry['ndvi_alt_buffer_mean']],
-                  file_registry['delta_ndvi']),
-            target_path_list=[file_registry['delta_ndvi']],
-            dependent_task_list=[mean_buffer_base_ndvi_task,
-                                 mean_buffer_alt_ndvi_task],
-            task_name="calculate delta ndvi"  # change in nature exposure
+        buffer_alt_dependencies.append(mask_alt_ndvi_task)
+
+    kernel_task = task_graph.add_task(
```
claire-simpson, line 762: I'd recommend hiding whitespace when reviewing this, as most (all?) of the changes between `kernel_task` and zonal stats are just unindenting

src/natcap/invest/urban_mental_health.py
```diff
@@ -413,6 +415,28 @@
                 units=None,
                 created_if="lulc_alt and lulc_attr_csv"
             ),
+            spec.CSVOutput(
+                id="lulc_to_ndvi_csv",
+                path="intermediate/lulc_to_ndvi_map.csv",
+                about=gettext(
+                    "Table giving mean NDVI by LULC codes, with excluded LULC "
```
emilyanndavis:
```suggestion
                    "Table giving mean NDVI by LULC code, with excluded LULC "
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1003,134 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+    unique_lucodes, inverse_indices = numpy.unique(
+        masked_lulc.astype(numpy.int64), return_inverse=True)
+
+    sums = numpy.bincount(inverse_indices,
+                          weights=masked_ndvi.astype(numpy.float32))
+    counts = numpy.bincount(inverse_indices)
+    means = sums / counts
+
+    mean_ndvi_by_lulc_dict = {lucode: mean for lucode, mean in zip(
+        unique_lucodes, means)}
```
emilyanndavis: Wow, very clever solution! I've never used `bincount` before, so I had to go to the docs and puzzle over this for a bit before I could grasp exactly what's going on here. Not sure if it would be worth adding a comment to help explain it up front? On the one hand, a gloss might have helped me (though I probably would have looked it up in the docs as well); on the other hand, it already seems pretty clear for people who (unlike me) are fluent in numpy. Up to you. Just thought I'd mention it on the off-chance you were already considering adding a comment here.

emilyanndavis (approved):
Looks good!

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +602,289 @@ def execute(args):
...
+
+    raster_to_method_dict = {'ndvi_base': 'cubic', 'ndvi_alt': 'cubic',
+                             'lulc_base': 'near', 'lulc_alt': 'near'}
+    input_align_list = []
+    output_align_list = []
+    resample_method_list = []
+    for input_raster, resample_method in raster_to_method_dict.items():
+        if args[input_raster]:
+            input_align_list.append(args[input_raster])
+            output_align_list.append(file_registry[input_raster+'_aligned'])
```
dcdenu4: We've generally adopted f-strings for string concatenation. I think it's technically "faster" and can be more readable than using the plus operator.

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +602,289 @@ def execute(args):
...
     elif args['scenario'] == 'lulc':
-        raise NotImplementedError
+        LOGGER.info("Using scenario option 2: LULC")
+
+        create_lulc_ndvi_csv_task = task_graph.add_task(
+            func=build_lulc_ndvi_table,
+            args=(args['lulc_attr_csv'],
+                  file_registry['lulc_to_ndvi_csv'],
+                  file_registry['lulc_base_aligned'],  # Note: won't get used if `ndvi` column in `lulc_attr_csv``
+                  file_registry['ndvi_base_aligned'] #TODO: will this get set to None if used doesn't input ndvi_base?
```
dcdenu4: Is this still a TODO to work through for this PR or later?

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1003,134 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+    codes = list(lulc_df['lucode'])
+    excludes = list(lulc_df['exclude'])
+    if 'ndvi' in lulc_df.columns:
+        LOGGER.info("Using NDVI in LULC attribute table to reclassify LULC.")
+        ndvi_means = list(lulc_df['ndvi'])
+        value_map = {}
+        for lu, ndvi, exclude in zip(codes, ndvi_means, excludes):
+            if bool(exclude):
+                value_map[lu] = FLOAT32_NODATA
+            elif numpy.isfinite(lu):
```
dcdenu4: What happens if this fails and there is a positive infinity or NaN value? Given the ModelSpec validation for this should we be worried about it here?

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1003,134 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+                value_map[lu] = ndvi
+    elif base_ndvi_path:
+        LOGGER.info("Using NDVI raster to calculate mean NDVI by LULC class.")
+        lulc_dict = {lu: ex for lu, ex in zip(codes, excludes)
+                     if numpy.isfinite(lu)}
```
dcdenu4: This is used somewhere else too, but is it necessary to check `numpy.isfinite` here?

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1003,134 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+        lulc_path (str): path to baseline LULC raster
+        ndvi_path (str): path to baseline NDVI raster
+        lulc_dict (dict): dictionary mapping lucodes to boolean
+            ``exclude`` values
+
+    Returns:
+        Dict containing the mean NDVI values for each LULC class
+
+    """
+    # TODO: iterblocks?
```
dcdenu4: Yep, we'll want to handle this a little differently to avoid loading the entire raster into memory with `raster_to_numpy_array`. A common way is using `iterblocks` and keeping a running average as you iterate. Habitat Quality does something like this where it counts the number of unique pixel values. But it could be altered to keep a running average too. https://github.com/natcap/invest/blob/98de13b9a9620a10f74a6e53bf292e474071097a/src/natcap/invest/habitat_quality.py#L1119

src/natcap/invest/urban_mental_health.py
claire-simpson: Yes, great idea! It is definitely not completely intuitive so I'll add several comments

src/natcap/invest/urban_mental_health.py
claire-simpson: Oops no, I just forgot to delete this. I determined that for `args`, if the user doesn't enter an optional input, it gets set to `None`, but the value of the `file_registry[key]` is the file path regardless of whether its actually used/created (so, not set to None).

src/natcap/invest/urban_mental_health.py
claire-simpson: This is probably unnecessary. I think I added the `isfinite` check because I was finding that pandas was reading in an empty line at the end of the lulc csv.. but I couldn't figure out why then and now I can't reproduce this behavior.

src/natcap/invest/urban_mental_health.py
claire-simpson: Same as above - I think I added the `isfinite` check because I was finding that pandas was reading in an empty line at the end of the lulc csv.

src/natcap/invest/urban_mental_health.py
claire-simpson: Ok thanks I've implemented this!

src/natcap/invest/urban_mental_health.py
```diff
@@ -219,6 +220,7 @@
             data_type=int,
             units=None,
             required="scenario=='lulc'",
+            # Allow lulc_alt for masking if using scenario 3
```
dcdenu4: Are the scenarios numbered or would it be helpful to use the name descriptor instead?

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +601,289 @@ def execute(args):
...
 
     elif args['scenario'] == 'lulc':
-        raise NotImplementedError
+        LOGGER.info("Using scenario option 2: LULC")
+
+        create_lulc_ndvi_csv_task = task_graph.add_task(
+            func=build_lulc_ndvi_table,
+            args=(args['lulc_attr_csv'],
+                  file_registry['lulc_to_ndvi_csv'],
+                  file_registry['lulc_base_aligned'],  # Note: won't get used if `ndvi` column in `lulc_attr_csv``
```
dcdenu4:
```suggestion
                  file_registry['lulc_base_aligned'],  # Note: won't get used if `ndvi` column in `lulc_attr_csv`
```

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +601,289 @@ def execute(args):
...
+            dependent_task_list=[align_task],
+            task_name='create csv with lulc-to-ndvi mapping'
+        )
+
+        reclassify_base_lulc_task = task_graph.add_task(
+            func=reclassify_lulc_raster,
+            args=(file_registry['lulc_base_aligned'],
+                  file_registry['lulc_to_ndvi_csv'],
+                  file_registry['ndvi_base_aligned_masked']),
+                  # this outputs raster to use as new NDVI to use going forward
```
dcdenu4: This comments wording is a bit unclear and maybe unnecessary?

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +601,289 @@ def execute(args):
...
+            file_registry['ndvi_base_buffer_mean']),
+        kwargs={
+            'ignore_nodata_and_edges': True,
+            'mask_nodata': True,
+            'normalize_kernel': True,
+            'target_datatype': pygeoprocessing.get_raster_info(
+                file_registry['ndvi_base_aligned_masked'])["datatype"],
+            'target_nodata': pygeoprocessing.get_raster_info(
+                file_registry['ndvi_base_aligned_masked'])["nodata"][0]},
+        dependent_task_list=buffer_base_dependencies+[kernel_task],
```
dcdenu4: Just a nit pick that this list concatenation has no spaces and line 794 has a space :)

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1001,154 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+        lulc_attr_table (str): path to lulc attribute csv
+        target_output_csv (str): path to output csv to save lulc:ndvi mapping
+        base_lulc_path (str): path to baseline LULC raster
+        base_ndvi_path (str): path to baseline NDVI raster
+
+    Returns:
+        None
+
+    """
+    lulc_df = pandas.read_csv(lulc_attr_table)
```
dcdenu4: I think there are some benefits to opening input tables using the `MODEL_SPEC.get_input` function. Here's an example from the carbon model.

https://github.com/natcap/invest/blob/98de13b9a9620a10f74a6e53bf292e474071097a/src/natcap/invest/carbon.py#L419

dcdenu4: This also might guard against the behavior you were seeing with trailing empty lines, etc!

src/natcap/invest/urban_mental_health.py
```diff
@@ -568,229 +601,289 @@ def execute(args):
...
     if args['scenario'] == 'tcc_ndvi':
         raise NotImplementedError
 
     elif args['scenario'] == 'lulc':
-        raise NotImplementedError
+        LOGGER.info("Using scenario option 2: LULC")
+
+        create_lulc_ndvi_csv_task = task_graph.add_task(
+            func=build_lulc_ndvi_table,
+            args=(args['lulc_attr_csv'],
```
dcdenu4: I commented in the function as well but the MODEL_SPEC has some handy `get_input` methods that allow us to open CSV tables in a way consistent manner. I gave an example from the carbon model below.

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1001,154 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+
+        # Extract valid pixels into 1D arrays
+        masked_lulc = lulc[mask].astype(numpy.int64)
+        masked_ndvi = ndvi[mask].astype(numpy.float32)
+
+        # numpy.unique returns unique elements of an array
+        # return_inverse returns the indices of unique array
+        # e.g., for lulc array [1, 2, 2, 5]: unique_lucodes = [1, 2, 5]
+        # and inverse_indices = [0, 1, 1, 2]
+        unique_lucodes, inverse_indices = numpy.unique(
```
dcdenu4: `numpy.unique` also has a `return_counts` parameter that might be a little more straightforward than using `bincount`?

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1001,154 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+
+        # Use bincount to get the NDVI sum for each unique LULC code (in
+        # block) and count the number of pixels for each lucode
+        # e.g., for ndvi_array = [0.2, 0.5, 0.7, 0.1] and same example lulc as
+        # above, block_sums = [0.2, 1.2, 0.1] and block_counts = [1, 2, 1]
+        block_sums = numpy.bincount(inverse_indices, weights=masked_ndvi)
+        block_counts = numpy.bincount(inverse_indices)
+
+        # accumulate into global sums/counts
+        for lucode, sum, c in zip(unique_lucodes, block_sums, block_counts):
```
dcdenu4, line 1100: I think the bincount solution is pretty neat and not one that I would have thought of. For the sake of sharing ideas, since we ultimately need to iterate over the unique codes, another solution would be kind of our typical one:

```python
for lucode in unique_lucodes:
    if lucode == nodata:
        continue    
    ndvi_sum = numpy.sum(ndvi[lulc==lucode])
    sum[lucode] = ndvi_sum
    etc...
```

dcdenu4: One positive of this approach is that we might be able to reduce some of the above code that flattens the array and creates the nodata masks?

dcdenu4: Although checking nodata blocks and continuing if there's no relevant data seems like a good thing to keep!

src/natcap/invest/urban_mental_health.py
```diff
@@ -945,6 +1001,154 @@ def _mask_with_ndvi(ndvi, ndvi_nodata):
...
+    Returns:
+        None
+
+    """
+    source_nodata = pygeoprocessing.get_raster_info(lulc)["nodata"][0]
+    target_datatype = gdal.GDT_Float32
+    target_nodata = FLOAT32_NODATA
+
+    # Create lucode: ndvi dict
+    lulc_df = pandas.read_csv(mean_ndvi_by_lulc_csv, index_col='lucode')
```
dcdenu4: For non model spec input CSVs, there is a utils function for opening CSV's that does some sanitization. Could be helpful even though this CSV is curated by the model.

https://github.com/natcap/invest/blob/98de13b9a9620a10f74a6e53bf292e474071097a/src/natcap/invest/utils.py#L293

src/natcap/invest/urban_mental_health.py
```diff
@@ -1214,4 +1418,14 @@ def validate(args, limit_to=None):
     if error_msg:
         validation_warnings.append((['scenario'], "Must select a scenario."))
 
+    # raise error if user enters lulc_attr_csv without 'ndvi' column and also
+    # doesn't provide base_ndvi raster
+    if args['scenario'] == 'lulc' and args.get('lulc_attr_csv'):
+        lulc_df = pandas.read_csv(args['lulc_attr_csv'])
```
dcdenu4: Another use case where using the MODEL_SPEC helper functions to open this could be nice.

dcdenu4 (changes requested):
Hey @claire-simpson ! I took another pass and had some comments and suggestions.

src/natcap/invest/urban_mental_health.py
claire-simpson: Using `return_counts` in `np.unique` is a good idea! However I do think this approach of iterating over the unique codes would be slower because, for each lucode, you're looking at the whole block to build `lucode == nodata` and sum. So I guess the tradeoff is the current approach being faster but the approach you proposed being more readable. What do you think?

claire-simpson: Oh and I think we'd have to nodata mask lulc and ndvi, as not masking ndvi could result in `nodata` values being incorporated into the sum, and not masking lulc would mean the 2 arrays are different shapes (so `ndvi[lulc==lucode]` wouldn't work).

So I think its just considering whether to keep:
`block_sums = numpy.bincount(inverse_indices, weights=masked_ndvi)`
or instead:
```python
block_sums = {}
for lucode in unique_lucodes:
            ndvi_sum = numpy.sum(masked_ndvi[masked_lulc == lucode])
            block_sums[lucode] = ndvi_sum
```

dcdenu4 (approved):
Thanks @claire-simpson ! Do we know why some of the tests are failing? Looks like a few scenic quality tests...

PR #2274, opened 2025-12-11
Pollination: Remove outdated geometry check

### LINKED ISSUE #2262: MultiPolygonZ geometry type causes error in Pollination
## Description
@newtpatrol reported a user was having an issue running Pollination with the farm layer being of geometry type `MultiPolygonZ`:

> Hey software team,  a student in our Stanford class is working with the Pollination model, and getting this error (full logfile attached):
`ValueError: Farm layer not a polygon type`
I'm looking at her farm shapefile now, and it does appear to be a polygon, but it has geometry "MultiPolygonZ" - is this Z thing the problem? Why?

This appears related to the PGP issue: https://github.com/natcap/pygeoprocessing/issues/448. Stacie confirmed that exporting the layer as `MultiPolygon` fixed the issue.

Here's a snippet of the error:
```python
2025-11-29 09:52:46,057 (natcap.invest.pollination) pollination._parse_scenario_variables(1305) INFO Checking to make sure guild table has all expected headers
2025-11-29 09:52:46,059 (natcap.invest.pollination) pollination._parse_scenario_variables(1345) INFO Checking that farm polygon has expected headers
2025-11-29 09:52:46,060 (natcap.invest.utils) utils.prepare_workspace(186) ERROR Exception while executing pollination
  File "natcap/invest/utils.py", line 184, in prepare_workspace
  File "invest/cli.py", line 484, in main
  File "natcap/invest/pollination.py", line 614, in execute
  File "natcap/invest/pollination.py", line 1352, in _parse_scenario_variables
ValueError: Farm layer not a polygon type
```

megannissel:
## Description
Fixes #2262 

There's a [PGP issue](https://github.com/natcap/pygeoprocessing/issues/448) and [related PR](https://github.com/natcap/pygeoprocessing/pull/450) (now merged) for supporting additional geometries beyond Polygon and MultiPolygon in `zonal_statistics`. 

The Pollination model itself has an additional geometry check that contradicts validation behavior, where the spec input geometry_types "POLYGON" and "MULTIPOLYGON" are mapped to several ogr types, i.e.:
```
'POLYGON': [ogr.wkbPolygon, ogr.wkbPolygonM,
            ogr.wkbPolygonZM, ogr.wkbPolygon25D],
'MULTIPOLYGON': [ogr.wkbMultiPolygon, ogr.wkbMultiPolygonM,
                 ogr.wkbMultiPolygonZM, ogr.wkbMultiPolygon25D]
```

I opted to remove the check (and the associated test based around it), since this should be caught by validation. If we think there's value in leaving the check (e.g. for CLI users who might not run validation beforehand), we could instead alter the check to use the same geometry list as validation. But it seems like we've been moving away from that in favor of encouraging folks to use the validation functionality (and this would be a redundant check). 

One additional point: We may want to wait to merge this in until we do a PGP release that supports the expanded geometry list for `zonal_statistics`. That said, if we were to merge this prior to an associated PGP release, this model would fail in a way that is consistent with every other model calling `zonal_statistics`. (HRA, for instance, will also fail if your AOI is a (Multi)Polygon25D/Z/ZM; the failure there will come from PGP itself since the model doesn't have an additional geom check outside of validation, where those geometry types are all allowed.) 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -77,6 +77,13 @@ Coastal Vulnerability
   correcting the units of some intermediate outputs.
   (`#2254 <https://github.com/natcap/invest/issues/2254>`_)
 
+Pollination
+===========
+* Fixed a bug where farm vectors with polygon/multipolygon geometries including
+  Z and/or M coordinates would pass validation but trigger a failure during the
+  model run due to a conflicting geometry check.
+  (`#2262 <https://github.com/natcap/invest/issues/2262>`_)
+
 
 3.17.2 (2025-12-02)
 -------------------
@@ -96,13 +103,6 @@ Coastal Vulnerability
   several tasks that created multiple files, resulting in these tasks always
   evaluating to "not precalculated"
   (`#2231 <https://github.com/natcap/invest/issues/2231>`_)
-
-Pollination
-===========
-* Fixed a bug where farm vectors with polygon/multipolygon geometries including
-  Z and/or M coordinates would pass validation but trigger a failure during the
-  model run due to a conflicting geometry check.
-  (`#2262 <https://github.com/natcap/invest/issues/2262>`_)
 
 
 3.17.1 (2025-11-06)


megannissel:
Workbench test failures are related to #2284 and should be resolved by #2292. We're also planning to update our PGP dependency prior to the next release; see #2295.

HISTORY.rst
```diff
@@ -97,6 +97,13 @@ Coastal Vulnerability
   evaluating to "not precalculated"
   (`#2231 <https://github.com/natcap/invest/issues/2231>`_)
 
+Pollination
```
davemfish: Whoops, this was entered under the "3.17.2" heading instead of the "unreleased" heading.

davemfish (changes requested):
@megannissel this looks good. I support the decision to remove the additional check and just rely on validation and the pygeoprocessing check instead.

Just one fix to history is needed.

HISTORY.rst
megannissel: Whoops, sorry about that; good catch!

PR #2290, opened 2026-01-12
Move models to subpackages and standardize APIs.

### LINKED ISSUE #2277: Reorganize model source code into subpackages
We have discussed giving each model its own subpackage in `natcap.invest`. Many models already have a subpackage because they are composed of multiple modules. With the upcoming invest reports feature, we expect all models to eventually have multiple modules. 

This issue is for doing the re-organization of the existing models, prior to the addition of invest reports. Doug and I decided it makes sense to do the re-organization first because there will be changes to a large number of files that are unrelated to reports, and we can keep the PRs slightly smaller by doing them separately.

### Expected tasks:
- [x] give each model a folder with the same name as the current model's module.
- [x] update import paths throughout test suite
- [x] check the python-script-generating function in `cli.py` to make sure it produces the correct `execute` call.
- [ ] make sure the autotest script still works
- [ ] re-generate sample data json files with updated `model_name` property

### Decisions:
- [x] Decide if we want to maintain backwards-compatibility by inserting `execute` and `validate` attributes onto the parent package for models that are getting a new subpackage. For example, make sure `natcap.invest.carbon.execute` still exists as it does now, in addition to the new, `natcap.invest.carbon.carbon.execute`.
- [x] Similar question for the `MODEL_SPEC`. Do we want to move it to its own module (`natcap.invest.carbon.model_spec`)? Or define it in an `__init__.py` as `natcap.invest.carbon.MODEL_SPEC` (this is the backwards-compatible option)? Or leave it as-is, which will effectively move it to `natcap.invest.carbon.carbon.MODEL_SPEC`?
- [x] * If yes to either, would we also want to add those attributes for models that already have a subpackage?

davemfish:
Model source code was re-organized so that all models are placed in  packages. For example, `natcap.invest.carbon` is now a package containing a `carbon` module. `execute, validate, & MODEL_SPEC` attributes can be found on the package, as well as on the module, to maintain backwards-compatibility. And models that were already in packages now have those same attributes on the top-level package in addition to the module. For example,
* `natcap.invest.sdr.execute` and `natcap.invest.sdr.sdr.execute` are both valid.

For the purposes of internal references, the `pyname` is always the module (e.g. `natcap.invest.sdr.sdr`) even though `natcap.invest.sdr` would also work.

Fixes #2277

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -83,6 +83,13 @@ Coastal Vulnerability
   correcting the units of some intermediate outputs.
   (`#2254 <https://github.com/natcap/invest/issues/2254>`_)
 
+Pollination
+===========
+* Fixed a bug where farm vectors with polygon/multipolygon geometries including
+  Z and/or M coordinates would pass validation but trigger a failure during the
+  model run due to a conflicting geometry check.
+  (`#2262 <https://github.com/natcap/invest/issues/2262>`_)
+
 
 3.17.2 (2025-12-02)
 -------------------
--- a/src/natcap/invest/pollination/pollination.py
+++ b/src/natcap/invest/pollination/pollination.py
@@ -1167,11 +1167,6 @@ def _parse_scenario_variables(args):
         LOGGER.info('Checking that farm polygon has expected headers')
         farm_vector = gdal.OpenEx(args['farm_vector_path'])
         farm_layer = farm_vector.GetLayer()
-        if farm_layer.GetGeomType() not in [
-                ogr.wkbPolygon, ogr.wkbMultiPolygon]:
-            farm_layer = None
-            farm_vector = None
-            raise ValueError("Farm layer not a polygon type")
         farm_layer_defn = farm_layer.GetLayerDefn()
         farm_headers = [
             farm_layer_defn.GetFieldDefn(i).GetName()
--- a/tests/test_pollination.py
+++ b/tests/test_pollination.py
@@ -478,45 +478,6 @@ class PollinationTests(unittest.TestCase):
             "Expected a biophysical, guild, and farm entry for 'fall'",
             str(cm.exception))
 
-    def test_pollination_bad_farm_type(self):
-        """Pollination: ensure detection of bad farm geometry type."""
-        from natcap.invest import pollination
-
-        # make some fake farm points
-        point_geom = [shapely.geometry.Point(20, - 20)]
-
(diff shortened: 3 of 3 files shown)


HISTORY.rst
```diff
@@ -70,6 +70,12 @@ General
 * Now testing and building against Python 3.14.
   No longer testing and building with Python 3.9, which reached EOL.
   (`#2269 <https://github.com/natcap/invest/issues/2269>`_)
+* Model source code was re-organized so that all models are placed in
```
claire-simpson:
```suggestion
* Model source code was reorganized so that all models are placed in
```

tests/test_coastal_vulnerability.py
```diff
@@ -81,7 +81,7 @@ def test_wind_and_wave_exposure(self):
         the wind calculation, so I'm testing them both in this scope.
 
         """
-        from natcap.invest import coastal_vulnerability
```
claire-simpson, line 84: Is the intent for each test module to include both ways of importing, ie. `from natcap.invest import [model]` and `from natcap.invest.[model] import [model]`?

tests/test_habitat_quality.py
```diff
@@ -2383,7 +2384,8 @@ def test_raster_values_in_bounds(self):
 
     def test_decay_distance(self):
         """Test `_decay_distance`"""
-        from natcap.invest.habitat_quality import _decay_distance
+        from natcap.invest.habitat_quality.habitat_quality import \
+            _decay_distance
```
claire-simpson, line 2388: Is there a pattern we want to keep consistent w.r.t our imports in tests? Seems like in all above test files, the private functions were not imported directly. Though it looks like pollination uses this same pattern as well.

claire-simpson (changes requested):
Thanks @davemfish, this looks good

tests/test_habitat_quality.py
davemfish: I realized I wasn't being consistent with that. I think I was more driven by readability than anything else. And I think readability kind of varies depending on how long the names are.

I'm okay with this inconsistency, but open to other opinions.

claire-simpson: I'm ok with the inconsistency as well, I wanted to flag it but no strong feelings about a particular way to import

claire-simpson (approved):
Workbench tests should be solved by #2292

PR #2308, opened 2026-01-21
Actions: tell `gsutil` where to find Python

### LINKED ISSUE #2307: make deploy: ``gsutil`` does not support python 3.14
We cannot rely on the latest supported python when using `gsutil` in our Actions workflows.

davemfish:
These changes essentially revert some changes we made a few months ago when I mistakenly thought we didn't need to jump through these extra hoops to use `gsutil`. Then Python 3.14 was released.

Fixes #2307 

The deploy steps ran on my fork and I observed no warnings from `gsutil` about missing python or incompatibility.
https://github.com/davemfish/invest/actions/runs/21252240444/job/61156736218

Test failures are due to #2310 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -184,7 +184,7 @@ jobs:
       - name: Deploy artifacts to GCS
         if: github.event_name != 'pull_request'
         env:
-          CLOUDSDK_GSUTIL_PYTHON: ${{ steps.python-gsutil-task.outputs.python-path }}
+          CLOUDSDK_PYTHON: ${{ steps.python-gsutil-task.outputs.python-path }}
         run: make deploy
 
   test-source-distribution:
@@ -269,7 +269,7 @@ jobs:
       - name: Deploy artifacts to GCS
         if: github.event_name != 'pull_request' && matrix.os == 'macos-15-intel' && matrix.python-version == env.LATEST_SUPPORTED_PYTHON_VERSION
         env:
-          CLOUDSDK_GSUTIL_PYTHON: ${{ steps.python-gsutil-task.outputs.python-path }}
+          CLOUDSDK_PYTHON: ${{ steps.python-gsutil-task.outputs.python-path }}
         run: make deploy
 
   validate-resources:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -172,8 +174,17 @@ jobs:
...
+      - name: Set up python for gsutil
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.GSUTIL_PYTHON_VERSION }}
+        id: python-gsutil-task
+
       - name: Deploy artifacts to GCS
         if: github.event_name != 'pull_request'
+        env:
+          CLOUDSDK_PYTHON: ${{ steps.python-gsutil-task.outputs.python-path }}
```
dcdenu4, line 187: Is `CLOUDSDK_PYTHON` an environment variable that `gsutil` looks for by default?

.github/workflows/build-and-test.yml
```diff
@@ -248,12 +259,17 @@ jobs:
...
         uses: google-github-actions/setup-gcloud@v3
 
-      - uses: actions/setup-python@v6
+      # gsutil (part of make deploy) never seems to support the latest python version.
+      - name: Set up python for gsutil
+        uses: actions/setup-python@v6
         with:
-          python-version: ${{ env.LATEST_SUPPORTED_PYTHON_VERSION }}
+          python-version: ${{ env.GSUTIL_PYTHON_VERSION }}
+        id: python-gsutil-task
```
dcdenu4, line 267: I don't know much about how `id`'s are used in GHA. Is it okay that this is the same id as what is used above?

.github/workflows/build-and-test.yml
```diff
@@ -516,7 +537,8 @@ jobs:
 
       - uses: actions/setup-python@v6
         with:
-          python-version: ${{ env.LATEST_SUPPORTED_PYTHON_VERSION }}
+          python-version: ${{ env.GSUTIL_PYTHON_VERSION }}
```
dcdenu4, line 540: Is there a reason why this chunk doesn't need a `name` tag like the others above?

.github/workflows/build-and-test.yml
davemfish: That's right. It's not well documented by google, but it's well documented by lots of other people using these tools.

.github/workflows/build-and-test.yml
davemfish: It seems like it is unique to the "step", so we're assigning it in each step where we need to refer to it.

.github/workflows/build-and-test.yml
davemfish: `name` is optional, but I did not think about when I edited this line, I just left it as-is. Without `name`, the "name" of this task is "Run actions/setup-python@v6". https://github.com/natcap/invest/actions/runs/21252242253/job/61156738483?pr=2308

Issue #1800, opened 2025-02-20
Workbench: upgrade electron and puppeteer dependencies


emilyanndavis:
Since [Electron releases a new major version every 8 weeks](https://www.electronjs.org/docs/latest/tutorial/electron-timelines) (in sync with every even-numbered major release of Chromium), the Workbench's `electron` dependency is now several major versions out of date. (Only the 3 most recent stable versions are officially supported, which means a release typically reaches EOL after 24 weeks, or 5–6 months.)

The `puppeteer-core` dependency is also a few major versions out of date.

Since both `electron` and `puppeteer-core` have Chromium dependencies that should be kept in sync, it makes sense to upgrade both packages in tandem to make sure they are compatible.

When I tried upgrading `puppeteer-core` (and `electron`, to a compatible version), the Workbench puppeteer test hangs (and eventually times out). There must be a breaking change introduced in puppeteer v22 (and possibly additional breaking changes in later versions), but I haven't been able to track it down yet.

Some references that may be useful:
- [Electron Breaking Changes](https://www.electronjs.org/docs/latest/breaking-changes), with a section for each major release
- [Puppeteer Changelog](https://pptr.dev/CHANGELOG), with breaking changes listed under each major release

Issue #1798, opened 2025-02-19
Workbench: upgrade to react-bootstrap 2 and Bootstrap 5


emilyanndavis:
To keep the Workbench up to date with the latest bugfixes and security patches, we should upgrade react-bootstrap, and its Bootstrap dependency, to their current major versions.

Since Bootstrap 5 was a major rewrite compared to Bootstrap 4, there will be breaking changes. We should expect all or nearly all of our React components to be affected, and we will need to make several updates to the code.

We should follow the migration guides:

- [React-bootstrap: Migrating to v2](https://react-bootstrap.github.io/docs/migrating)
- [Bootstrap: Migrating to v5](https://getbootstrap.com/docs/5.3/migration/)

... and be sure to perform thorough manual testing of the UI to minimize side effects.

Ideally, this work should be planned while we're not making UI-related updates, to avoid conflicts and re-work.

emilyanndavis:
Here's what I think the React Bootstrap upgrade will entail, based on my reading of the migration guides:

### React Bootstrap component updates
- [x] Change `<Accordion.Toggle>` to `<Accordion.Button>` (and remove now-redundant `as="button"`).
- [x] Update `<Badge>`s to use `bg` instead of `variant`.
- [x] Update `<Dropdown.Menu>` to use `align="end"` instead of `align="right"`.
- [x] Remove `bsCustomPrefix="form-switch"` from `<Form.Check>` (`bsCustomPrefix` has been replaced with `bsSwitchPrefix`, and the default is "form-switch").
- [x] Replace `<Form.Control as="select">` with `<Form.Select>`, and remove `custom` from these elements (currently just one uses `custom`).
- [x] Replace `<Form.Row>` with `<Row>`.
- [x] Remove `onSelect` from `<Nav.Link>`. Use `onSelect` in the parent `<Nav>` instead.
- [x] Add `id`s to `<ToggleButton>`s (now required due to upstream Bootstrap changes).

### Bootstrap utility class updates
- [x] Replace `text-muted` with `text-body-secondary` (deprecated in 5.3.0, will be removed in 6.0.0).
- [x] Replace utility classes based on "left"/"right" with new names based on "start"/"end". (See [Bootstrap Migration Guide: 5.0.0 Utilities](https://getbootstrap.com/docs/5.3/migration/#utilities-3).)
- [x] Replace `sr-only` with `visually-hidden`.

### Testing & cleanup
- [x] Manually test UI and adjust as needed.
[...]

PR #2309, opened 2026-01-21
update to latest electron builder

### LINKED ISSUE #2300: Workbench: Update to latest electron-builder
We've been seeing intermittent, but somewhat frequent, build failures on macos related to `hdiutil`. If there's an upstream fix anywhere, upgrading `electron-builder` would be the likely place to start.

See also #1800

davemfish:
Hoping this yields more reliable DMG builds. I've run the Mac build step several times in prepping this PR, I believe it has worked every time.

I tried using the node LTS version, but that would have required doing a cascade of other package upgrades, so I settled on version 22 for now. We have #1800 and #1798 to do at some point.

Fixes #2300 

Test failures due to #2310 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -312,7 +312,7 @@ jobs:
       - name: Set up node
         uses: actions/setup-node@v6
         with:
-          node-version: "lts/*"
+          node-version: 22
 
       - name: Restore node_modules cache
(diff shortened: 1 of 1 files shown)


emlys (approved):
Thanks @davemfish! Going to go ahead and merge since the test failures are explained by that other issue.

PR #2314, opened 2026-01-24
Implement Urban Mental Health Option 1: Tree Canopy Cover + NDVI Inputs

### LINKED ISSUE #2141: Urban Mental Health Model: implement scenario 1 (NDVI + TCC)
This option involves non-linear mapping of tree cover to population-weighted NDVI (using a Generalized Additive Model [GAM]) and determining the change in greenness from baseline to an alternate scenario, determined by a user-defined target tree cover percent. This scenario effectively tells us the health benefit outcomes of reaching a tree cover goal, approximated by NDVI (in other words, we're translating TC to NDVI to health impact, and specifically looking at the change in nature exposure from baseline NDVI to an alternate scenario calculated by predicting NDVI from the GAM). Required inputs for this scenario (beyond the general inputs required for all scenario options) include:

1. Tree canopy cover (raster)
2. Baseline NDVI (raster)
3. Tree cover target (number)

In summary: Change in Nature Exposure (`delta_NE`) = `NDVI_scenario` (predicted from GAM using the tree-cover target) - `NDVI_baseline` (and negative `delta_NE` values are set to `NoData`).

claire-simpson:
## Description
Implements Option 1 (tree canopy cover–based scenarios) for the Urban Mental Health model by translating a user-defined tree cover target into an NDVI-based nature exposure scenario (to create 'alternate NDVI').

This PR adds a population-weighted, non-linear translation between tree canopy cover (%) and NDVI exposure, following the framework described in the UMH design document.

The steps are as follows:
1. Align input rasters
2. Mask baseline NDVI (mask out water or other excluded LULC classes or mask by 0 threshold)
3. Compute neighborhood NDVI exposure by convolving masked NDVI using the `search_radius`
4. Compute neighborhood tree canopy cover (TCC) exposure by convolving TCC using the same `search_radius`
5. Extract exposure values block-wise: iterate over aligned blocks of buffer-mean NDVI exposure, buffer-mean TCC exposure, and population (masking no data values and pixels with no population)
6. Bin TCC values: assign each valid pixel to a TCC bin (range: [0, 100])
7. Compute population-weighted mean NDVI per bin
8. Fit a linear GAM using population per bin as weights (so bins w/ more people influence the fit more), to get a function mapping TCC exposure to NDVI exposure
9. Evaluate the fitted function at the user-specified tree cover target value to get the NDVI target
10. Generate the alternate NDVI exposure raster via:` NDVI_alt = NDVI_base + (NDVI_target - f(TCC_pixel))`
11. Compute change in nature exposure via `NE_delta = NDVI_alt - NDVI_base`
12. Mask out negative values in NE_delta

**Notes:**
- Population is used only as a weighting factor, not as a spatial transform of TCC or NDVI.
- Both NDVI and TCC are evaluated on the same neighborhood (buffer) scale, so:
    - The GAM learns a relationship between experienced canopy cover and experienced greenness, not pixel-level vegetation
    - Alternate NDVI is generated directly on the exposure scale, so no additional convolution is required after translation.
- Tests are not complete (see #2316), and I haven't added `pygam` as an InVEST dependency (see below)

**Open Questions**
1. Is the above workflow/math correct, specifically w.r.t (1) calculating mean within a buffer distance (i.e., 2d convolution operation) for both NDVI and TCC _before_ fitting the GAM and translating TCC to alternate NDVI and (2) using the population to both compute a population-weighted conditional mean of NDVI for each TCC "bin" and as a _weight_ when fitting the GAM (population is not explicitly used to create a population-weighted TCC layer)
2. Are we ok adding `pygam` as a dependency? There are definitely alternative options like using `scipy.interpolate.UnivariateSpline`. If so, I'd need to add pygam to `requirements.txt`.
3. Are we ok with the binning approach to reduce memory use? In a comment in the design doc, Yingjie clarified that their inputs to the GAM were aggregated at the tract level before fitting this model (to avoid loading the entire NDVI, TCC, and population rasters into memory). However in our implementation, we are not requiring users to input tracts (though we could!). There are certainly other alternative approaches we could take to fitting the TCC-NDVI relationship, including:
- Random spatial sampling of NDVI and TCC rasters (within `iterblocks`)
- Spatial window aggregation: iterate over fixed-sized windows and compute mean NDVI and TCC and fit GAM on window-level summaries (or just downsample both rasters to have fewer pixels)

Fixes #2141

## Checklist

### DIFF
--- a/src/natcap/invest/urban_mental_health/urban_mental_health.py
+++ b/src/natcap/invest/urban_mental_health/urban_mental_health.py
@@ -1296,6 +1296,7 @@ def _translate_tc_target_to_ndvi_target(
 
     Returns:
         None
+
     """
     ndvi_info = pygeoprocessing.get_raster_info(base_ndvi_path)
     tc_info = pygeoprocessing.get_raster_info(tree_cover_path)
@@ -1375,114 +1376,6 @@ def _translate_tc_target_to_ndvi_target(
     LOGGER.info(f"Target NDVI Value: {ndvi_target}")
     df = pandas.DataFrame([ndvi_target[0]])
     df.to_csv(target_output_csv_path, header=False, index=False)
-
-
-    # curve_smooth = gam.predict(centers.reshape(-1, 1))
-
-    # fig, ax = plt.subplots()
-    # ax.plot(centers, curve_smooth, c='r', label='Fitted TC->NDVI curve')
-    # ax.scatter(x, y, c='b', label="Binned means", alpha=0.6)
-    # ax.set_xlabel("Tree cover (%)")
-    # ax.set_ylabel("Predicted NDVI")
-    # fig.tight_layout()
-    # fig.savefig(result_figure_path, dpi=150)
-    # plt.close(fig)
-
-    # return centers, curve_smooth.astype(numpy.float64)
-
-
-# def _apply_tc_target_to_alt_ndvi(base_ndvi_path, population_path,
-#                                  tree_cover_path, tc_target,
-#                                  target_delta_ndvi,
-#                                  nbins=256, nsplines=10):
-#     """Calculate delta_NDVI as difference between target NDVI (calculated from GAM) and baseline_NDVI
-
-#     I.e., computes:
-#     delta_ndvi = target_ndvi - base_ndvi
-
-#     * Note that negative differences are _not_ masked here
-
-#     Args:
-#         base_ndvi_path (str): path to baseline NDVI raster
-#         population_path (str): path to population raster
-#         tree_cover_path (str): path to tree cover raster with pixels expected
-#             as a percentage in range [0, 100]
-#         tc_target (float): target tree canopy cover value as a percentage (in range [0,100])
-#         target_delta_ndvi (str): path to output alternate NDVI raster
-#         nbins (int): number of tree cover bins to use when fitting curve
-#         nsplines (int): number of splines to use in GAM smoothing
-
-#     Returns:
-#         None
-#     """
-
-#     ndvi_target = _fit_tc_to_ndvi_curve(
-#         base_ndvi_path, tree_cover_path, population_path,
-#         tc_target, nbins, nsplines
-#     )
-
-#     ndvi_info = pygeoprocessing.get_raster_info(base_ndvi_path)
-#     tc_info = pygeoprocessing.get_raster_info(tree_cover_path)
-
-#     ndvi_nodata = ndvi_info["nodata"][0]
-#     tc_nodata = tc_info["nodata"][0]
-
-#     # ndvi_target = float(numpy.interp(tc_target, centers, curve))
-
-#     def _calc_delta_using_target_NDVI(ndvi, tcc, ndvi_nodata, tc_nodata,
-#                                       ndvi_target_val):
-#         """
-#         Calculate Delta_NDVI = NDVI_target - baseline NDVI
-        
-#         Args:
-#             ndvi (numpy.ndarray): baseline NDVI array
-#             tcc (numpy.ndarray): tree canopy cover array
-#             ndvi_nodata (float): nodata value for ndvi
-#             tc_nodata (float): nodata value for tree canopy cover
-
-#             ndvi_target_val (float): target NDVI value based on tc_target
-
-#         Returns:
-#             numpy.ndarray: alternate NDVI array
-
-#         """
-
-#         valid_mask = (~pygeoprocessing.array_equals_nodata(ndvi, ndvi_nodata) &
-#                       ~pygeoprocessing.array_equals_nodata(tcc, tc_nodata))
-#         result = numpy.empty(ndvi.shape, dtype=numpy.float32)
-#         result[:] = ndvi_nodata
-
-#         if not numpy.any(valid_mask):
-#             return result
-
-#         # tc_vals = numpy.clip(tc_arr[valid].astype(numpy.float64), 0.0, 1.0)
-#         tcc_vals = tcc[valid_mask].astype(numpy.float64)
-#         tcc_vals = numpy.clip(tcc_vals, 0, 100)
-#         # get predicted NDVI given current tcc, as per the fitted curve
-#         f_tc = numpy.interp(tcc_vals, centers_arr, curve_arr)
-#         # compute difference between target NDVI and predicted NDVI
-#         # i.e., get amount of exposure change need to bring pixel to target NDVI
-#         ndvi_diff = ndvi_target_val - f_tc
-
-#         # apply difference to base NDVI to get alt NDVI
-#         result[valid_mask] = (
-#             ndvi[valid_mask] + ndvi_diff).astype(numpy.float32)
-
-#         return result
-
-#     pygeoprocessing.raster_calculator(
-#         [
-#             (base_ndvi_path, 1),
-#             (tree_cover_path, 1),
-#             (ndvi_nodata, "raw"),
-#             (tc_nodata, "raw"),
-#             (ndvi_target, "raw"),
-#         ],
-#         _calc_delta_using_target_NDVI,
-#         target_delta_ndvi,
-#         ndvi_info["datatype"],
-#         ndvi_nodata
-#     )
 
 
 def calc_baseline_cases(population_raster, base_prevalence_vector,


tests/test_urban_mental_health.py
```diff
@@ -868,3 +881,154 @@ def test_calculate_mean_ndvi_by_lulc_class(self):
         for key in expected_mean_ndvi:
             numpy.testing.assert_allclose(
                 actual_mean_ndvi[key], expected_mean_ndvi[key], atol=1e-6)
+
+    # def test_option1_tcc_input(self):
```
claire-simpson, line 881: Template for test for whole model, but I commented it out because there is still uncertainty around population weighting and whether to convolve TCC before fitting GAM - see #2316

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+    x = centers[has_pop].reshape(-1, 1)
+    y = curve[has_pop]
+    w = pop_sum[has_pop]
+
+    gam = LinearGAM(s(0, n_splines=nsplines)) #TODO - cubic regression spline
+    gam.fit(x, y, weights=w)
+
+    curve_smooth = gam.predict(centers.reshape(-1, 1))
+
+    fig, ax = plt.subplots()
```
claire-simpson: I believe `matplotlib` is a new dependency as of the reports PR so I assume this would be the first time a graph is created within a model. This was an intermediate output of the demo model and seems useful for interpreting the relationship between TCC and NDVI, so I think it'd be great to ultimately include in the report. However maybe saving this as a standalone isn't needed?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -4,18 +4,20 @@
 import shutil
 import tempfile
 
+import matplotlib.pyplot as plt
 import numpy
 import pandas
+from pygam import LinearGAM, s # Are we ok to add pygam as new invest dependency? Alternatively, could us scipy.UnivariateSpline
```
dcdenu4, line 10: I'd be interesting in talking about how `scipy.UnivariateSpline` could be an alternative.

src/natcap/invest/urban_mental_health/urban_mental_health.py
dcdenu4, line 566: Has this docstring been keeping pace with the MODEL_SPEC updates? In terms of descriptive text and required / optional flags.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -782,27 +831,62 @@ def execute(args):
...
-                file_registry['ndvi_alt_aligned_masked'])["datatype"],
-            'target_nodata': pygeoprocessing.get_raster_info(
-                file_registry['ndvi_alt_aligned_masked'])["nodata"][0]},
-        dependent_task_list=buffer_alt_dependencies + [kernel_task],
-        target_path_list=[file_registry['ndvi_alt_buffer_mean']],
-        task_name="calculate mean alternate NDVI within buffer")
+    if args['scenario'] == 'tcc_ndvi':
+        LOGGER.info("Using Tree Canopy Cover and NDVI inputs")
+        mean_buffered_tcc_task = task_graph.add_task(
+            func=pygeoprocessing.convolve_2d,
```
dcdenu4, line 840: Yep, I think using a dichotomous kernel and convolving with `normalize_kernel=True` gets you a mean value within the given kernel radius.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -782,27 +831,62 @@ def execute(args):
...
+            task_name="calculate mean tree cover within buffer")
+
+        mean_buffered_alt_ndvi_task = task_graph.add_task(
+            func=_apply_tc_target_to_alt_ndvi,
+            args=(file_registry['ndvi_base_buffer_mean'],
+                  file_registry['population_aligned'],
+                  file_registry['tree_cover_buffer_mean'],
+                  args['tree_cover_target'],
+                  file_registry['ndvi_alt_buffer_mean'],
+                  file_registry['result_fig_tc_ndvi_plot']),
```
dcdenu4: Should the plot output also be in the `target_path_list` below?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+
+def _apply_tc_target_to_alt_ndvi(base_ndvi_path, population_path,
+                                 tree_cover_path, tc_target,
+                                 target_alt_ndvi, result_figure_path,
+                                 nbins=256, nsplines=10):
+    """Apply a fitted TC --> NDVI curve to create an alternate NDVI raster.
+
+    Writes alt NDVI raster where each pixel's NDVI is increased based on
+    the difference between the target NDVI (based on tc_target) and the
+    NDVI predicted by the TC-->NDVI curve at that pixel's tree cover value.
```
dcdenu4: Should we mention how population is used at all for weighting?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+    I.e., computes:
+    ndvi_target = f(tc_target)
+    alt_ndvi = base_ndvi + (ndvi_target - f(tc_pixel))
+
+    * Note that negative differences are _not_ masked here
+
+    Args:
+        base_ndvi_path (str): path to baseline NDVI raster
+        population_path (str): path to population raster
+        tree_cover_path (str): path to tree cover raster with pixels in
```
dcdenu4: Maybe adding tree cover is expected as a _percentage_ between 0-100.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+
+    I.e., computes:
+    ndvi_target = f(tc_target)
+    alt_ndvi = base_ndvi + (ndvi_target - f(tc_pixel))
+
+    * Note that negative differences are _not_ masked here
+
+    Args:
+        base_ndvi_path (str): path to baseline NDVI raster
+        population_path (str): path to population raster
```
dcdenu4: Be nice to specify if population is expected to be count or density.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+        tree_cover_path (str): path to tree cover raster with pixels in
+            range [0, 100]
+        tc_target (float): target tree canopy cover value (in range [0,100])
```
dcdenu4:
```suggestion
        tc_target (float): target tree canopy cover value as a percentage (in range [0,100])
```

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+        target_alt_ndvi (str): path to output alternate NDVI raster
+        nbins (int): number of tree cover bins to use when fitting curve
+        nsplines (int): number of splines to use in GAM smoothing
+
+    Returns:
+        None
+    """
+
+    centers, curve = _fit_tc_to_ndvi_curve(
```
dcdenu4: I'm wondering if this should be its own taskgraph step in the workflow instead of called in here? Benefits could be avoided re-computation, more modular step by step breakout in `execute`, and maybe more targeted testing?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1156,6 +1245,217 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
+    """Fit a population-weighted TC->NDVI curve using binning.
+
+    Assumptions:
+      - Tree cover values are in range [0, 100]
+      - Population is nonnegative; only pop>0 contributes
+
+    Args:
+        base_ndvi_path (str): path to baseline NDVI raster
+        tree_cover_path (str): path to tree cover raster
+        population_path (str): path to population raster
```
dcdenu4, line 1290: Density or count? For future us mostly.

src/natcap/invest/urban_mental_health/urban_mental_health.py
dcdenu4: If it feels like a useful standalone output, I think it's fine to save on its own along with adding to the report if it'd be useful there too.

It might be worth thinking about whether we should be saving the numpy arrays for `centers`, `curve`, and others as intermediate outputs.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -782,27 +831,62 @@ def execute(args):
...
+            target_path_list=[file_registry['tree_cover_buffer_mean']],
+            task_name="calculate mean tree cover within buffer")
+
+        mean_buffered_alt_ndvi_task = task_graph.add_task(
+            func=_apply_tc_target_to_alt_ndvi,
+            args=(file_registry['ndvi_base_buffer_mean'],
+                  file_registry['population_aligned'],
+                  file_registry['tree_cover_buffer_mean'],
+                  args['tree_cover_target'],
+                  file_registry['ndvi_alt_buffer_mean'],
```
dcdenu4: Maybe its worth updating `ndvi_alt_buffer_mean`, since this function isn't really returning a buffered mean? Right?

PR #2318, opened 2026-01-30
update electron and related packages

### LINKED ISSUE #1800: Workbench: upgrade electron and puppeteer dependencies

davemfish:
This PR updates the Workbench to the latest electron release and updates a few other packages that are used in the electron main process. I did not attempt to update any of the frontend libraries.

According to the electron changelog, users will need at least MacOS 12 and Windows 10. The current Workbench already requires at least Windows 10.

Fixes #1800

## Checklist

### DIFF
--- a/workbench/tests/binary_tests/puppet.test.js
+++ b/workbench/tests/binary_tests/puppet.test.js
@@ -8,7 +8,7 @@ import os from 'os';
 import glob from 'glob';
 import { spawn, spawnSync } from 'child_process';
 
-import { rimrafSync } from 'rimraf';
+import { rimraf } from 'rimraf';
 import puppeteer from 'puppeteer-core';
 
 import pkg from '../../package.json';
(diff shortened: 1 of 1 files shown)


emlys (approved):
Thanks @davemfish! Curious why the `waitForTimeout`s are no longer needed?

Issue #1001, opened 2022-05-23
WB: eliminate some items from the menubar


davemfish:
The menubar has some default menus associated with chrome. We've removed some of the unhelpful ones in the past, but we could further remove `View > Reload` as it's not particularly useful to reload the browser window. Except during dev when I do it all the time, but I always use `ctrl-r` for that.

Also check other menus for other items to cut.

Issue #217, opened 2020-07-10
qtpy error: no Qt bindings could be found


phargogh:
We have 2 forums issues documenting the same exception, that qtpy can't find any Qt bindings.

* https://community.naturalcapitalproject.org/t/i-can-not-start-invest/1040/5
    * Windows 10
    * InVEST 3.8.2 works
    * InVEST 3.8.4, 3.8.5 don't work (qtpy error)
* https://community.naturalcapitalproject.org/t/installation-problems-with-versions-above-3-8-2x86/1140
    * Windows 7
    * InVEST 3.8.2 works
    * InVEST 3.8.{4,5,6} don't work (qtpy error)

The strange thing is that this exception is being after executing these lines in `src/natcap/invest/ui/inputs.py`:

```python
import PySide2

import qtpy
from qtpy import QtWidgets
```

Note how PySide2 is imported _first_ (and successfully!), but then the qtpy import _crashes_ because it can't find a qt API!

- [x] Did PySide2 have a version upgrade between 3.8.2 and 3.8.4?
- [ ] Does setting the qtpy environment variable `FORCE_QT_API=1` do anything?  It's supposed to use an existing Qt API if one has already been imported.  If this does work, we'll need to set an environment variable at runtime.

phargogh:
Looks like the culprit _might_ have been the PySide2 binaries:

* InVEST 3.8.2 used `PySide2==5.14.2.1`, installed via PyPI
* InVEST 3.8.4 used `PySide2==5.15.0`, installed via PyPI

Furthermore, PySide2==5.15.0 has a DLL issue where two required DLLs are not in the correct directory (see [PYSIDE-1323](https://bugreports.qt.io/browse/PYSIDE-1323) and [PYSIDE-1330](https://bugreports.qt.io/browse/PYSIDE-1330)).

The issue has apparently been [fixed as of June 11, 2020](https://wiki.qt.io/Qt_for_Python_Development_Notes), but the fix has not yet been released.

I'll start a development build on my fork on branch `bugfix/217-qtpy-importerror-no-qt-bindings`.  Binary build happening at https://github.com/phargogh/invest/actions/runs/164987485

- [ ] When the build finishes, ask the users to test this build version.

phargogh:
[...]

PR #2320, opened 2026-01-31
Feature/carbon report

### LINKED ISSUE #2331: Carbon report
Since an HTML report already exists for the Carbon model, it seems natural to release an updated Carbon report with the initial launch of InVEST reports.

### LINKED ISSUE #2332: Reports: Raster plot subtitles may appear misaligned
## Description
Raster plots include subtitles that indicate the raster units. Sometimes, the subtitle appears to be positioned ideally: just below the title, and sufficiently above the map itself. At other times, the subtitle appears too far below the title and nearly overlaps the map. Positioning text with matplotlib can be tricky, and given the potential variation in figure layout and the extreme variation possible in raster aspect ratios, there may not be a one-size-fits-all solution here. But it would be good to experiment with a variety of layouts and raster shapes to see if we can dial in a one-size-does-pretty-well-for-most approach.

### NDR (Gura): no problem here!
Image

### NDR (Luzon): looking a little odd
Image

emilyanndavis:
## Description
- Adds new carbon report. (Closes #2331.)
- Removes old carbon report.
- Adds/updates report tests.
- Various updates to raster plotting settings (fig size, text placement, color palettes, etc.) for reports. (Note all snapshot test refs had to be updated due to these changes.) (Closes #2332.)
- Miscellaneous refactoring/reorg/cleanup of raster utils & related code.
- Updates to some text strings for localization support. (Note that it seems babel should be able to extract `gettext`-wrapped messages from f-strings as of v. 2.16.0, according to the conversation on this issue: [python-babel/babel #1001](https://github.com/python-babel/babel/issues/1001)).

For an example, see this [Carbon Report](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/carbon_report_willamette.html) generated with sample data.

Note I am planning to do some additional code cleanup (docstrings, type hints, potentially some refactoring and/or file reorganization). In the interest of preparing a dev build for internal testing, I am saving that work for a separate, near-future PR. 

## Checklist
~~- [ ] Updated the user's guide (if needed)~~ Not yet, but we will: [UG #217](https://github.com/natcap/invest.users-guide/issues/217)

### DIFF
--- a/tests/test_carbon.py
+++ b/tests/test_carbon.py
@@ -5,6 +5,7 @@ import tempfile
 import shutil
 import os
 
+from bs4 import BeautifulSoup
 from osgeo import gdal
 from osgeo import osr
 import numpy
@@ -106,6 +107,7 @@ class CarbonTests(unittest.TestCase):
 
         args = {
             'workspace_dir': self.workspace_dir,
+            'calc_sequestration': True,
             'do_valuation': True,
             'price_per_metric_ton_of_c': 43.0,
             'rate_change': 2.8,
@@ -151,6 +153,23 @@ class CarbonTests(unittest.TestCase):
         # Alternate: 59.00136 * -58 Mg/ha = -3422.079 Mg/ha
         assert_raster_equal_value(
             os.path.join(args['workspace_dir'], 'npv_alt.tif'), -3422.079)
+
+        # Ensure aggregate results are correct.
+        report_path = os.path.join(args['workspace_dir'], 'carbon_report.html')
+        # Raster size is 100 m^2; therefore, raster total is as follows:
+        # (x Mg / 1 ha) * (1 ha / 10000 m^2) * (100 m^2) = (x / 100) Mg
+        with open(report_path, encoding='utf-8') as html_file:
+            soup = BeautifulSoup(html_file, 'html.parser')
+            for (stat_name, expected_value) in [
+                    ('Baseline Carbon Storage', 0.86),
+                    ('Alternate Carbon Storage', 0.28),
+                    ('Change in Carbon Storage', -0.58),
+                    ('Net Present Value of Carbon Change', -34.2208),
+                    ]:
+                stat_name_text = soup.find(string=stat_name)
+                row_heading = stat_name_text.find_parent('th')
+                value_str = row_heading.find_next_sibling('td').string
+                assert float(value_str) == expected_value
 
     def test_carbon_zero_rates(self):
         """Carbon: test with 0 discount and rate change."""


src/natcap/invest/carbon/carbon.py
```diff
@@ -20,6 +18,7 @@
     model_id="carbon",
     model_title=gettext("Carbon Storage and Sequestration"),
     userguide="carbonstorage.html",
+    reporter="natcap.invest.carbon.reporter",
```
emilyanndavis, line 21: In addition to specifying the reporter, I've also removed all code related to the "old" carbon report.

src/natcap/invest/carbon/carbon.py
```diff
@@ -164,7 +163,6 @@
             ),
             required="do_valuation",
             allowed="do_valuation",
-            units=None
         ),
```
emilyanndavis, line 166: I was seeing linting errors on the PercentInputs, which aren't supposed to specify units (because units are defined as a ClassVar and are always None).

src/natcap/invest/coastal_vulnerability/reporter.py
```diff
@@ -186,7 +186,9 @@ def report(file_registry, args_dict, model_spec, target_html_filepath):
         population_spec = model_spec.get_output(
             'coastal_exposure').get_field('population')
         population_checkbox = altair.binding_checkbox(
-            name=f'scale by population ({natcap.invest.spec.format_unit(population_spec.units)})')
+            name=(f"{gettext('scale by population')} "
+                  f"({natcap.invest.spec.format_unit(population_spec.units)})")
+        )
```
emilyanndavis, line 191: According to [this conversation in the `python-babel/babel` repo](https://github.com/python-babel/babel/issues/1001), babel should be able to extract messages from f-strings as of babel 2.16.0 (the latest release is 2.18.0).

src/natcap/invest/reports/raster_utils.py
```diff
@@ -55,6 +87,24 @@
...
+    ('STATISTICS_MEAN', gettext('Mean')),
+    ('STATISTICS_VALID_PERCENT', gettext('Valid percent')),
+]
+
+COUNT_COL_NAME = gettext('Count')
+NODATA_COL_NAME = gettext('Nodata value')
+UNITS_COL_NAME = gettext('Units')
+UNKNOWN_VAL_TEXT = gettext('unknown')
+UNITS_TEXT = gettext('Units')
+
```
emilyanndavis, line 107: I thought about moving all these constants out to a separate module (or maybe even split into raster_constants, table_constants, etc.)—mainly because they take up so much space here—but figured that could wait until later.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -227,16 +285,84 @@ def _choose_n_rows_n_cols(xy_ratio, n_plots):
...
+        downscale_factor = max_figure_height / figure_height
+        figure_height = max_figure_height
+        figure_width *= downscale_factor
+
     fig, axs = plt.subplots(
-        n_rows, n_cols, figsize=(FIGURE_WIDTH, n_rows*sub_height),
-        layout='constrained')
+        n_rows, n_cols, figsize=(figure_width, figure_height),
+        layout='compressed')
+    plt.close()
```
emilyanndavis, line 304: I added this call to `plt.close()` because, when running the test suite, I was seeing warnings from matplotlib indicating more than 20 figures had been opened and may be consuming too much memory. Calling `plt.close()` is the recommendation that accompanies the warning. In practice, this probably won't come up when generating a single report, but there didn't seem to be a downside to preventing it, just in case.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -444,32 +588,21 @@ def geometamaker_load(filepath):
...
 
-# GDAL Metadata keys and corresponding column headers for a table
-STATS_LIST = [
-    ('STATISTICS_MINIMUM', 'Minimum'),
-    ('STATISTICS_MAXIMUM', 'Maximum'),
-    ('STATISTICS_MEAN', 'Mean'),
-    ('STATISTICS_VALID_PERCENT', 'Valid percent'),
-]
-
-
```
emilyanndavis, line 455: Moved these up to join other constants near the top of the file.

src/natcap/invest/reports/sdr_ndr_utils.py
```diff
@@ -31,11 +31,16 @@ def generate_results_table_from_vector(filepath, cols_to_sum):
 
 
 def update_caption_with_stream_map_info(caption: list[str], flow_dir_alg: str):
+    stream_map_info_part_1 = gettext((
+        'Results were generated using the following flow direction '
+        'algorithm:'))
+    stream_map_info_part_2 = gettext(
+        ('The stream network may look incomplete at  this resolution, and '
+         'therefore it may be necessary to view the  full-resolution raster '
+         'in GIS to assess its accuracy.'))
```
emilyanndavis, line 40: I split this into two parts and reworded it slightly to better support localization and hopefully avoid awkward/wrong word ordering (e.g., to avoid forcing a language to say "the MFD flow algorithm" where it wants to say "the flow algorithm MFD").

HISTORY.rst
```diff
@@ -109,6 +109,14 @@ Workbench
   to "touch" the field again in order to see the validation status.
   (`#2149 <https://github.com/natcap/invest/issues/2149>`_)
 
+Carbon Storage and Sequestration
+================================
+* Previously, the Carbon model generated an HTML report consisting of a table
+  of model inputs and a table of aggregate results. That report has been
+  replaced with a new report containing the same features plus many more.
+  (`#2249 <https://github.com/natcap/invest/issues/2249>`_,
+  `#2331 <https://github.com/natcap/invest/issues/2331>`_)
```
emilyanndavis, line 118: I know there's already a mention of the "new" carbon report in the "General" section, but I thought it was worth also mentioning the "old" report is gone (or rather, has been upgraded).

src/natcap/invest/carbon/reporter.py
```diff
@@ -0,0 +1,227 @@
...
+
+    if args_dict['calc_sequestration']:
+        intermediate_output_raster_plot_tuples = [[
+            (f'c_{pool_type}_bas', 'continuous', 'linear'),
+            (f'c_{pool_type}_alt', 'continuous', 'linear')
+         ] for pool_type in ['above', 'below', 'dead', 'soil']]
+    else:
+        intermediate_output_raster_plot_tuples = [[
+            (f'c_{pool_type}_bas', 'continuous', 'linear')
+            for pool_type in ['above', 'below', 'dead', 'soil']]]
```
davemfish, line 53: I think this function could be a little easier to parse if it was structured with only one `if args_dict["calc_sequestration"]:` block. So basically, setup all three base tuples first, then extend them as needed inside the block? Actually, I guess it's not quite that simple for the intermediate rasters, which would get replaced instead of extended. Anyway, perhaps there's some simplification to be found.

src/natcap/invest/carbon/reporter.py
```diff
@@ -0,0 +1,227 @@
...
+
+    return (input_raster_plot_tuples,
+            output_raster_plot_tuples,
+            intermediate_output_raster_plot_tuples)
+
+
+def _get_intermediate_output_headings(args_dict: dict) -> list[str]:
```
davemfish, line 60: I think it could be helpful to note that each pool gets its own section when "calc sequestration" was run, but maps of all pools are plotted under a single section if not. And why that is.

src/natcap/invest/carbon/reporter.py
```diff
@@ -0,0 +1,227 @@
...
+    Returns:
+        ``None``
+    """
+
+    (input_raster_tuples,
+     output_raster_tuples,
+     intermediate_raster_tuples) = _get_raster_plot_tuples(args_dict)
+
+    input_raster_plot_configs = raster_utils.build_raster_plot_configs(
+        args_dict, input_raster_tuples)
```
davemfish, line 175: I keep asking myself if it's possible/desirable to instantiate the `RasterPlotConfig`s here inline, rather than behind the `get_raster_plot_tuples` and `build_raster_plot_configs` abstractions. And I guess the main reason to use that function is that it does the work of constructing the filepaths, allowing us to only reference the raster `id`s here in this script. That's convenient, but the tradeoff is needing to setup the tuples instead of using the `RasterPlotConfig` API directly. No need to change anything, I'm just musing.

src/natcap/invest/carbon/reporter.py
```diff
@@ -0,0 +1,227 @@
...
+    inputs_img_src = raster_utils.plot_and_base64_encode_rasters(
+        input_raster_plot_configs)
+    input_raster_caption = raster_utils.generate_caption_from_raster_list(
+        [(id, 'input') for (id, _) in input_raster_tuples],
+        args_dict, file_registry, model_spec)
```
davemfish, line 180: I'm similarly on the fence about whether this `generate_caption_from_raster_list` function is a useful abstraction. The alternative would be,

```python
input_raster_caption = []
for id, _ in input_raster_tuples:
    filename = os.path.basename(args_dict[id])
    about_text = model_spec.get_input(id).about
    input_raster_caption.append(f'{filename}:{about_text}')
```

And then of course doing it all again for the output rasters. It's just less abstract and easier to read this way IMO. Again, no need to change anything here. I know these functions existed prior to this PR.

src/natcap/invest/coastal_vulnerability/reporter.py
```diff
@@ -220,7 +222,8 @@ def report(file_registry, args_dict, model_spec, target_html_filepath):
     na_count = exposure_geo.exposure.isna().sum()
     if na_count:
         null_checkbox = altair.binding_checkbox(
-            name=f'{na_count} point(s) missing the exposure index. Show:')
+            name=(f"{na_count} "
+                  f"{gettext('point(s) missing the exposure index. Show:')}"))
```
davemfish, line 226: I was a little unsure how well these phrases would translate without more context, but it's probably worth a try. I'm sure we have that same challenge all over invest.

src/natcap/invest/reports/templates/models/carbon.html
```diff
@@ -0,0 +1,137 @@
...
+      (wide_table(
+        output_raster_stats_table | safe,
+        font_size_px=16
+      ), 100)
+    ])
+  )}}
+
+  <h2 class="section-header">Inputs</h2>
+
+  {% if args_dict is defined and args_dict != None %}
```
davemfish: I don't think it hurts anything, but our `StrictUndefined` rule in the jinja2 env should enforce that `args_dict` is at least defined at this point.

src/natcap/invest/reports/templates/styles.html
```diff
@@ -8,9 +8,10 @@
        Helpful random color generator: https://rgbcolorpicker.com/random/pastel */
     /* Make sure each accent color has sufficient contrast with black (#000).
        Minimum 4.5:1; higher is better. Check with https://www.whocanuse.com/. */
+    --accent-color-carbon: #bed4f0;
+    --accent-color-coastal_vulnerability: #fff6d8;
```
davemfish, line 12: I wonder if we should preemptively choose colors for all models. That seems like something we don't need to leave to each individual report author. They could change the color when developing the report if they felt strongly.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -227,16 +285,84 @@ def _choose_n_rows_n_cols(xy_ratio, n_plots):
...
+        figure_width = MAX_FIGURE_WIDTH_2_COL_WIDE_AOI
+    sub_width = figure_width / n_cols
+    sub_height = (sub_width / xy_ratio)
+    figure_height = sub_height * n_rows
+    max_figure_height = MAX_SUBPLOT_HEIGHT_INCHES * n_rows
+    if figure_height > max_figure_height:
+        # Constrain height, then adjust width accordingly.
+        downscale_factor = max_figure_height / figure_height
+        figure_height = max_figure_height
+        figure_width *= downscale_factor
```
davemfish, line 299: neat!

src/natcap/invest/reports/raster_utils.py
```diff
@@ -518,7 +651,26 @@ def raster_inputs_summary(args_dict):
...
                 # Remove 'Units' column if all units are blank
-                if not any(raster_summary[filename]['Units']):
-                    del raster_summary[filename]['Units']
+                if not any(raster_summary[filename][UNITS_COL_NAME]):
+                    del raster_summary[filename][UNITS_COL_NAME]
 
     return pandas.DataFrame(raster_summary).T
+
+
+def rat_to_html(raster_path: str) -> str | None:
```
davemfish: If we think this function might not be working, and if we're not actually using it yet, then it might be best to remove it from this PR and add it back later when we're sure it's working. Just in case we don't get to fixing it right away.

tests/test_carbon.py
```diff
@@ -175,18 +152,6 @@ def test_carbon_full(self):
...
-        report_path = os.path.join(args['workspace_dir'], 'report.html')
-        # Raster size is 100 m^2; therefore, raster total is as follows:
-        # (x Mg / 1 ha) * (1 ha / 10000 m^2) * (100 m^2) = (x / 100) Mg
-        for (stat, expected_value) in [
-                ('Baseline Carbon Storage', 0.86),
-                ('Alternate Carbon Storage', 0.28),
-                ('Change in Carbon Storage', -0.58),
-                ('Net Present Value of Carbon Change', -34.22),
-                ]:
-            assert_aggregate_result_equal(report_path, stat, expected_value)
```
davemfish, line 188: Do you think it's worth reproducing this test on the sums? It's a bit of computation that moved out of the model's `execute` and into the `report`. But we could do a similar test using `bs4`.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -227,16 +285,84 @@ def _choose_n_rows_n_cols(xy_ratio, n_plots):
...
+        'horizontalalignment': 'left',
+        's': f'{UNITS_TEXT}: {units}',
+        'verticalalignment': 'bottom',
+        'x': -0.5,
+        'y': subtitle_offset,
+    }
+    return (ylim_args, text_args)
+
+
+def _get_legend_kwargs(num_patches: int, n_plots: int, xy_ratio: float):
```
davemfish, line 345: All these helper functions for setting up the `kwargs` are a great idea.

src/natcap/invest/carbon/reporter.py
emilyanndavis: I also don't love how this function turned out and went through the same back-and-forth as I was writing it. I'll see if I can streamline it in a "miscellaneous reports cleanup" PR I'm planning for the near future.

src/natcap/invest/carbon/reporter.py
emilyanndavis: Same. I'll keep mulling this over as well.

src/natcap/invest/carbon/reporter.py
emilyanndavis: I feel like this function is worthwhile because of the specific string format that's required for the template to render the caption properly. If we were to build that string in many different places, it would be easy to mess it up in one or more of those places, and equally easy to fail to detect the problem. The way we're handling this now doesn't have to be _the_ way, but I would argue for a helper function for at least the string-formatting part of the process.

src/natcap/invest/reports/templates/styles.html
emilyanndavis: Yes! I actually thought about this early on in the reports development phase, but I quickly decided that was the wrong yak to be shaving at the time. Now that we're closer to release, I agree it's worth pursuing, for the sake of removing the onus from the report developer, and also because I think it might be worthwhile to establish a sort of subtle "branding" for each report that we aim to keep consistent over time.

Generating a set of suitable colors is relatively easy; the tricky part will be assigning a color to each model. Do we want colors to evoke something relevant to the model (e.g., blue = ocean)? Do we group models into categories and assign a color family (or single color?) to each category? Do we group models into categories and assign a thematic icon to each category but more-or-less randomly assign colors within a category? Or do we decide all colors will be (at least technically) distinct but otherwise arbitrary (i.e., unrelated to the context of any particular model)? Either way, how do we ensure we leave space for future models? We won't solve this here, of course, but I think it's worth discussing!

tests/test_carbon.py
emilyanndavis: Yes, it's probably worth keeping this covered with a test. I'll add one!

src/natcap/invest/carbon/reporter.py
emilyanndavis: Good point. I added a docstring here.

src/natcap/invest/reports/templates/models/carbon.html
emilyanndavis: You're right; I removed the `is defined` check.

src/natcap/invest/reports/raster_utils.py
emilyanndavis: I removed it, along with all the related code from the carbon reporter, templates, and tests.

tests/test_carbon.py
emilyanndavis: I added this (back) into the carbon module tests; I hope that's what you were imagining. Feels a little strange since the carbon module neither calculates those sums nor (directly) generates the report, but it also seemed simpler than mocking up all the data I'd need to test the same details from the reports side. I also considered unit-testing the functions in the carbon reporter that generate the table, but it seems more valuable to check what's ultimately rendered in the report. Plus, it's unlikely any other reporters will rely on those same functions, so—at least for now—I don't see a need to test them outside the context of an actual carbon report.

src/natcap/invest/carbon/reporter.py
davemfish: That's a good point, thanks.

tests/test_carbon.py
davemfish: This is what I was imagining. Thanks!

PR #2328, opened 2026-02-03
Workbench: report rendering bugfixes

### LINKED ISSUE #2326: Workbench: "View Results" button not rendered when model is run with a file suffix
## Description
On the `feature/invest-reports` branch, the workbench is set up to render a "View Results" button after a successful run of a model that has a reporter (i.e., CV, NDR, SDR, and soon Carbon). When running one of these models with a non-empty file suffix (e.g., using a sample datastack), the workbench cannot find the generated report and therefore does not render the "View Results" button.

### LINKED ISSUE #2327: Workbench: "View Results" opens new Electron window but does not render report
## Description
After successfully running a model that has a reporter, the "View Results" button opens a new Electron window but does not render the report. The window is completely blank.

The report itself is successfully generated and can be found in the workspace and opened with a web browser.

## To replicate

1. Check out `feature/invest-reports`.
2. Uninstall/re-install `natcap-invest`, then start up the Workbench (e.g., `cd workbench && yarn start`).
3. Run one of the following models using sample data: CV, NDR, or SDR.
4. Press the "View Results" button.

## Expected behavior
The HTML report should appear in the new Electron window.

## Observed behavior
A new Electron window opens, but it has no content.

## Environment
- Operating System: macOS Sequoia 15.7.3
- InVEST branch: `feature/invest-reports`

## Additional context
- I initially observed this issue while running the Workbench on a dev server (i.e., via `yarn start`). I then ran a full dev build on my machine and opened the resulting Workbench app, where I observed the same behavior. 
- All models with reporters (CV, NDR, SDR) are affected.
- In all cases, the HTML report is successfully generated and can be found in the workspace and opened with a web browser.

emilyanndavis:
## Description
- Fixes #2326 (in `setupInvestHandlers.js`).
- Fixes #2327 (in `setupOpenLocalHtml.js`, `InvestTab.index.jsx`, and `investtab.test.js`).
- Also pulls in latest from `main` (all changes other than those in the files listed above have already been reviewed).

## Checklist


workbench/src/main/setupInvestHandlers.js
```diff
@@ -66,7 +66,7 @@ export function setupInvestRunHandlers() {
     const taskgraphLoggingLevel = settingsStore.get('taskgraphLoggingLevel');
     const language = settingsStore.get('language');
     const nWorkers = settingsStore.get('nWorkers');
-    let resultsSuffix = args?.resultsSuffix || '';
+    let resultsSuffix = args?.results_suffix || '';
```
emilyanndavis, line 69: This fixes #2326.

workbench/src/main/setupOpenLocalHtml.js
```diff
@@ -17,7 +20,7 @@ export default function setupOpenLocalHtml(parentWindow, isDevMode) {
         frame: true,
       });
       setupContextMenu(child);
-      child.loadURL(url);
+      child.loadURL(fileUrl.toString());
```
emilyanndavis: These changes fix #2327.

workbench/src/main/ipcMainChannels.js
```diff
@@ -20,8 +20,9 @@ export const ipcMainChannels = {
   IS_NEW_VERSION: 'is-new-version',
   LOGGER: 'logger',
   OPEN_EXTERNAL_URL: 'open-external-url',
-  OPEN_PATH: 'open-path',
+  OPEN_FILE_URL: 'open-file-url',
```
emilyanndavis: Part of the fix for #2327

workbench/src/main/main.js
```diff
@@ -166,6 +167,7 @@ export const createWindow = async () => {
   setupRemovePlugin();
   setupWindowsMSVCHandlers();
   setupOpenLocalHtml(mainWindow, ELECTRON_DEV_MODE);
+  setupOpenFileUrl(mainWindow, ELECTRON_DEV_MODE);
```
emilyanndavis: Part of the fix for #2327

workbench/src/main/setupOpenFileUrl.js
```diff
@@ -0,0 +1,29 @@
...
+        frame: true,
+      });
+      setupContextMenu(child);
+      child.loadURL(fileUrl.toString());
+      if (isDevMode) {
+        child.webContents.openDevTools();
+      }
+    }
+  );
+}
```
emilyanndavis: This file is part of the fix for #2327

workbench/src/renderer/components/InvestTab/index.jsx
```diff
@@ -197,8 +197,8 @@ class InvestTab extends React.Component {
     }
   }
 
-  async handleViewResults(htmlFile) {
-    ipcRenderer.send(ipcMainChannels.OPEN_LOCAL_HTML, htmlFile);
+  async handleViewResults(filepath) {
+    ipcRenderer.send(ipcMainChannels.OPEN_FILE_URL, filepath);
```
emilyanndavis: Part of the fix for #2327

workbench/tests/renderer/investtab.test.js
```diff
@@ -202,7 +202,7 @@ describe('Open Workspace & View Results buttons', () => {
 
     expect(ipcRenderer.send).toHaveBeenCalledTimes(1);
     expect(ipcRenderer.send).toHaveBeenCalledWith(
-      ipcMainChannels.OPEN_LOCAL_HTML, job.htmlfile
+      ipcMainChannels.OPEN_FILE_URL, job.htmlfile
```
emilyanndavis: Part of the fix for #2327

workbench/src/main/ipcMainChannels.js
davemfish: I don't think we need both `OPEN_LOCAL_HTML` and `OPEN_FILE_URL`, do we?

workbench/src/main/setupInvestHandlers.js
davemfish: Thanks!

workbench/src/main/setupOpenFileUrl.js
davemfish: Could we modify the existing `setupOpenLocalHtml.js` instead of adding a new module?

emilyanndavis: In my mind, they were two separate scenarios: for the user guide, you're navigating to a page that is hosted on a local web server; for a report, you're opening a file that lives in your filesystem. But you're right, there's no reason we can't make use of one channel for both purposes. Especially since, as I remind myself now, the primary distinction we're concerned with thus far is whether to load a URL in an Electron window or in a web browser, and both the user guide and reports fall into the "Electron window" category.

davemfish: I forgot that the UG is served on a local webserver, but that's only in dev mode, I believe. In production we're always pointing to files on the local filesystem. I think the same is true for the Workbench's main `index.html` and the splashscreen `splash.html`. In dev mode they are served on an http server, and in production they are on the local filesystem. It's the construction of the URL that varies between dev & production, but then the call to `loadUrl` can be the same. 

>the primary distinction we're concerned with thus far is whether to load a URL in an Electron window or in a web browser, and both the user guide and reports fall into the "Electron window" category.

Right, that's why I figured they could use the same event handler, because they both want to call `win.loadUrl` with the same parameters. If there's a need to customize the calls, then I agree it could make sense to have a separate channel for each use-case.

workbench/src/main/setupOpenLocalHtml.js
```diff
@@ -8,11 +8,14 @@ import setupContextMenu from './setupContextMenu';
 
 export default function setupOpenLocalHtml(parentWindow, isDevMode) {
   ipcMain.on(
-    ipcMainChannels.OPEN_LOCAL_HTML, (event, url) => {
+    ipcMainChannels.OPEN_LOCAL_HTML, (_event, url, isFilePath = false) => {
+      if (isFilePath) {
+        url = require('node:url').pathToFileURL(url).toString();
+      }
```
davemfish, line 14: Oh, I think I finally understand. `url` here is sometimes,
* `http://` - dev mode userguide
* filesystem path - production userguide, production report, dev mode report

PR #2335, opened 2026-02-04
Fix creation of patterned output directories

### LINKED ISSUE #2213: `ModelSpec.setup` cannot create directories based on patterns
## Description
Patterns in a `ModelSpec.Output` path are supported, but only in filenames, not in folder names. When `ModelSpec.setup` creates the necessary output folders, it uses the pattern directly as the folder name, rather than doing the string substitution that the user might expect.

## To replicate
```python
        spec.SingleBandRasterOutput(
            id="diff_sed_deposition_[SCENARIO]",
            path="[SCENARIO]/diff_sed_deposition_[SCENARIO].tif",
            about="Difference in sediment deposition (scenario - baseline).",
            data_type=sdr.MODEL_SPEC.get_output('sed_deposition').data_type,
            units=sdr.MODEL_SPEC.get_output('sed_deposition').units
        ),
```
This creates a folder named `[SCENARIO]` in the output workspace. During model execution, I can get the desired filepath by indexing to `file_registry[('diff_sed_deposition_[SCENARIO]', 'my_scenario')]`. That would give `"my_scenario/diff_sed_deposition_baseline.tif"`. But the "my_scenario" folder was never created.

## Additional context
I don't think there are any examples of a pattern like this in core invest models. My example comes from a plugin where [SCENARIO] is a user-defined name of a scenario. And the plugin can analyze any number of scenarios. It would be possible to put all the files in a single directory, but my preference was to organize them by scenario.

claire-simpson:
## Description

Fix handling of patterned output directories in `ModelSpec` / `FileRegistry` so that directories with unresolved patterns (e.g., `[SCENARIO]`) are not created during model setup, and are instead created lazily once a final output path is resolved. This is because the `create_output_directories` function creates directories _before_ `FileRegistry` exists and so it has no way of know what to substitute for the pattern.

- `create_output_directories` now skips creating folders with square brackets.
- `FileRegistry.__getitem__` ensures that the resolved output path gets created when a patterned output is accessed
- Note that if a `path` to a directory contains a pattern, the `id` **must also contain the pattern**, or a `ValueError` will be thrown. 

e.g., the following is valid:
```python
spec.FileOutput(
    id="foo_[SCENARIO]",
    path="[SCENARIO]_path/foo.tif"
)
```
However, if `id="foo"`, this is invalid.

Fixes #2213

## Checklist

### DIFF
--- a/src/natcap/invest/file_registry.py
+++ b/src/natcap/invest/file_registry.py
@@ -57,6 +57,14 @@ class FileRegistry:
             path_fields = set(re.findall(r'\[(\w+)\]', path))
             # Don't convert id_fields to set, need to preserve order later
             id_fields = re.findall(r'\[(\w+)\]', output.id)
+
+            # If path contains pattern field(s), the ``id`` must also contain
+            # those fields so that users are required to provide substitution
+            # values when indexing the FileRegistry. Otherwise, users could
+            # index the FileRegistry by an ID that doesn't include the pattern
+            # fields, and the FileRegistry wouldn't know what values to
+            # substitute into the path. (*NOTE* The ID can have additional
+            # fields that aren't in the path, but not the other way around.)
             if path_fields and not path_fields.issubset(set(id_fields)):
                 raise ValueError(
                     f"Output id '{output.id}' must include pattern field(s) "
@@ -105,7 +113,7 @@ class FileRegistry:
             raise KeyError(f'Key not found: {key}')
 
         path = self._keys_to_paths[key]
-        if key in self._pattern_fields and self._pattern_fields[key]:
+        if key in self._pattern_fields:
             fields = self._pattern_fields[key]
             if len(field_values) != len(fields):
                 raise KeyError(
@@ -130,6 +138,8 @@ class FileRegistry:
                 raise KeyError('Received field values for a key that has no fields')
             self.registry[key] = path
 
+        # Ensures patterned directories get created when the path is accessed
+        # since they aren't created during model setup like non-patterned paths
         parent_dir = os.path.dirname(path)
         if parent_dir:
             os.makedirs(parent_dir, exist_ok=True)


src/natcap/invest/file_registry.py
```diff
@@ -96,7 +105,7 @@ def __getitem__(self, keys):
             raise KeyError(f'Key not found: {key}')
 
         path = self._keys_to_paths[key]
-        if key in self._pattern_fields:
+        if key in self._pattern_fields and self._pattern_fields[key]:
```
emlys: In what case is this second condition necessary?

src/natcap/invest/file_registry.py
```diff
@@ -120,4 +129,9 @@ def __getitem__(self, keys):
             if field_values:
                 raise KeyError('Received field values for a key that has no fields')
             self.registry[key] = path
+
+        parent_dir = os.path.dirname(path)
```
emlys, line 143: I would add a comment here explaining that this is necessary only for patterened directory paths

src/natcap/invest/file_registry.py
```diff
@@ -54,9 +54,18 @@ def __init__(self, outputs, workspace_dir, file_suffix=None):
         for output in outputs:
             path, extension = os.path.splitext(output.path)
             # Distinguish between paths that are/aren't patterns
+            path_fields = set(re.findall(r'\[(\w+)\]', path))
+            # Don't convert id_fields to set, need to preserve order later
+            id_fields = re.findall(r'\[(\w+)\]', output.id)
+            if path_fields and not path_fields.issubset(set(id_fields)):
+                raise ValueError(
+                    f"Output id '{output.id}' must include pattern field(s) "
```
emlys, line 70: I think this could benefit from a comment explaining why this is an error condition. I'm not sure I fully understand it.

src/natcap/invest/file_registry.py
claire-simpson: Ah yeah good question. This checks to make sure that there are actually fields to substitute for the pattern, i.e., guard against a case where the `path` has brackets but the `id` doesn't. I think that since I'm raising an error above in `__init__`, this is not strictly necessary though so I'll remove it.

Issue #1259, opened 2023-03-17
Browser-like "back" navigation for Workbench UG window


davemfish:
This is not strictly necessary since the TOC is always present, but would be nice to have. I also fell into a trap where I clicked an image in the UG and it opened in the same window, but then I had no way to navigate back to the text and had to close the window.

If this starts to get complicated, we can always consider opening the local HTML in the user's default browser instead of in Electron. See also #1258 which is not an issue if we use another browser.

PR #2354, opened 2026-02-06
Workbench context-menu and browser window enhancements

### LINKED ISSUE #2266: Workbench: Add more useful options to the right-click context menu.
Add more useful options to the right-click context menu. In particular, opening the context menu on an image should at least provide "Save image as" and "Copy image". We may not have any images in the Workbench yet, but we will with reports.

The best way to do this might be to use https://github.com/sindresorhus/electron-context-menu

It probably requires upgrading electron first, which seems like a good idea anyway.

### LINKED ISSUE #2347: Reports: hyperlinks in reports should open in external browser
Right now there's a userguide link at the top of each report. By default, clicking it navigates the electron browser window away from the report and to that page. And there's actually no way to get back because of #1259 .

I think opening the link in an external browser is probably better.

### LINKED ISSUE #2353: Workbench: improve handling of multiple browser windows
There are some misc issues to resolve when the main window opens a new browser window for the UG or a report:

* The child window is always on top of the main window, unless it is minimized
* The child window has the same top menubar (Applciation Menu)as the main window. This is okay, but not sure it's necessary to open the About or Report a problem windows from anything other than the main workbench window. ~~Plus, these windows are child to the window from which they were opened, so it might be more consistent to make sure that parent is always the main window.~~ Actually, the windows opened from the top menubar always have the main window as the parent. And I don't think there is a way to have different windows with different application menus. In particular, that doesn't make sense on MacOS. So we will leave this as-is.

davemfish:
This PR includes various fixes and enhancements related to electron browser windows. It would be great to test all this out on Mac, especially opening and closing multiple UG or report windows. Closing the main window while leaving other windows open. Re-opening the Workbench. And any other iterations we can think of.

Fixes #2266 - uses `electron-context-menu` to provide good defaults and a "Save Image as" option on images.

Fixes #2347 - configures the browser window event handler and adds `target="_blank"` to the `<a>` tag in reports. This does not solve the same problem for the UG because we have not modified the `target` attribute of links in the UG source. See #1259 .

Fixes #2353 - makes UG and report windows independent of the main electron window. This gives more natural window-switching and minimize behavior, at least on Windows.

All of these things are pretty hard/impossible to test in our test suite since they overlap with OS-level functionality. I'm not confident we could get a reliable puppeteer test for this stuff, so I think it's okay to just test it manually.

## Checklist


workbench/src/main/menubar.js
```diff
@@ -126,7 +126,10 @@ function createWindow(parentWindow, isDevMode) {
       additionalArguments: [devModeArg],
     },
   });
-  setupContextMenu(win);
+  contextMenu({
+    window: win,
+    showSearchWithGoogle: false,
```
emilyanndavis, line 131: Some of the content accessed via the menubar contains images (InVEST logo, NatCap logo). Do you think we should enable "save image" and "save image as" in this context? Or, alternatively, do we have reason to disable "copy image" (which seems to be enabled by default)?

src/natcap/invest/reports/templates/base.html
```diff
@@ -15,7 +15,7 @@
 
     {% block content scoped %}
       <h1>{{ self.page_title() }}</h1>
-      <p>To learn more about the {{ model_name }} model, visit the <a href="http://releases.naturalcapitalproject.org/invest-userguide/latest/en/{{ userguide_page }}">InVEST User Guide</a>.</p>
+      <p>To learn more about the {{ model_name }} model, visit the <a target="_blank" href="http://releases.naturalcapitalproject.org/invest-userguide/latest/en/{{ userguide_page }}">InVEST User Guide</a>.</p>
```
emilyanndavis: When a link opens in a new window, it's a good idea to communicate that behavior ahead of time so users know what to expect. There are two main ways to do this:

1. Add visible text to the link that describes the behavior, e.g., `<a>InVEST User Guide (opens in new browser window)</a>`.
2. Add an icon to the link (typically a little square with an arrow pointing out of it, like the one we use in the in-workbench changelog by way of the `link-external` CSS class), along with an aria-label that conveys the same information to screenreaders, e.g., `<a class="link-external" aria-label="InVEST User Guide (opens in new browser window)">InVEST User Guide</a>`.

Typically, "opens in new window" is sufficient, but since this link is designed to open in a web browser, not an Electron window—and that behavior might be surprising—it's worth including the word "browser" here. (If/when someone is viewing a report in a web browser, the word "browser" might be redundant, but it won't be inaccurate.) By that same logic, I'd advocate for option 1, since the visible text clearly communicates the link's behavior. In option 2, the meaning of the icon is ambiguous.

workbench/src/main/menubar.js
davemfish: I think the defaults are okay. I don't feel like it's important to let people save those logos. In order to do so we would need to isolate the `webContents.session` of the window, so that our `will-download` event handler on the main session does not get triggered. "Copy image" I expect does not trigger `will-download`, so it's okay to leave that option.

src/natcap/invest/reports/templates/base.html
davemfish: Thanks, @emilyanndavis . In this case the link should open in a new browser window when clicked from electron, but will likely open in a new tab when clicked from any other browser. Though this tab vs window behavior for `target="_blank"` is configurable in the browser, I think. Nevertheless, it's probably still accurate to just say "opens in new browser window", even if that window is a tab.

emilyanndavis: You're right: in practice, many (most?) people will have their browser configured to open such a link in a new tab, but it's still common to say new "window" instead. The important thing is that people understand that link is going to move them into a new context—whether it's a window or a tab is secondary, and not something we can fully predict.

Issue #2342, opened 2026-02-05
auto PRs into `release` branch: auth error


davemfish:
Here's one example: https://github.com/natcap/invest/actions/runs/21682190031/job/62519474368

```
+ git push origin autopr/main-rev367c2fb-into-release-3.18.0
remote: error: GH006: Protected branch update failed for refs/heads/autopr/main-rev367c2fb-into-release-3.18.0.        
remote: 
remote: - You're not authorized to push to this branch. Visit https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches for more information.        
To https://github.com/natcap/invest
 ! [remote rejected] autopr/main-rev367c2fb-into-release-3.18.0 -> autopr/main-rev367c2fb-into-release-3.18.0 (protected branch hook declined)
error: failed to push some refs to 'https://github.com/natcap/invest'
```

phargogh:
It turned out that we also had a restriction on who could create `autopr/**` branches.  The changes implemented here plus removing that specific branch protection rule allowed the run to pass.  I think this is resolved now!

PR #2355, opened 2026-02-06
Various fixes for autopr branch creation

phargogh:
As proposed, the changes work on my fork, producing the correct behavior.  Having said that, there is a branch protection rule in place on the InVEST repo that I am unable to access on my fork, so it is possible that this is what's preventing branch creation.  I'd like to propose that we merge the PR as-is (pending your review, of course!) and then if the action fails I'll loosen the branch restrictions slightly.

The two main changes here are 1) limiting the workflow to only run on pushes to main and 2) the `gh api` calls now reflect the suggested parameters from the github api docs.  It's possible that the omission of some of these parameters were contributing to the error the api was reporting earlier.

RE: #2342

### DIFF
--- a/.github/workflows/autopr-create.yml
+++ b/.github/workflows/autopr-create.yml
@@ -3,11 +3,13 @@ name: "AutoPR: Create PRs into release branches"
 env:
     SOURCE_BRANCH: main
 on:
-    pull_request:
-        types: [closed]
-        branches:
-            # Triggered when PR into this branch is closed.
-            - main
+    #pull_request:
+    #    types: [closed]
+    #    branches:
+    #        # Triggered when PR into this branch is closed.
+    #        - main
+
+    # Pushes to main (e.g. after a PR) are the only way that this workflow should be triggered.
     push:
         branches:
             - main


.github/workflows/autopr-create.yml
```diff
@@ -3,18 +3,20 @@ name: "AutoPR: Create PRs into release branches"
...
-        branches:
-            # Triggered when PR into this branch is closed.
-            - main
+    #pull_request:
+    #    types: [closed]
+    #    branches:
+    #        # Triggered when PR into this branch is closed.
+    #        - main
+
+    # Pushes to main (e.g. after a PR) are the only way that this workflow should be triggered.
```
davemfish, line 12: Thanks, @phargogh , did you want to remove the commented code? Just in case this actually works and we don't have to change it again 😄

.github/workflows/autopr-create.yml
```diff
@@ -3,18 +3,20 @@ name: "AutoPR: Create PRs into release branches"
...
     push:
         branches:
             - main
 
 jobs:
     create-pr-into-release:
-        if: github.repository == 'natcap/invest'
+        #if: github.repository == 'natcap/invest'
```
davemfish, line 19: This one should probably be un-commented again, right?

PR #2365, opened 2026-02-09
Support placeholder option in model input dropdown list

### LINKED ISSUE #2260: Standardize support for an invalid placeholder option in a model input dropdown list
## Background/Motivation
This came up during development of the Urban Mental Health model. The model will require a user to select one scenario from a dropdown list of three options. Rather than defaulting to one of the three valid options, the dropdown list (as defined in the model spec) is prepended with an invalid option (key='') whose display text instructs a user to "Select a scenario." The model's `validate` function makes sure a user has selected a scenario by using custom logic to accept any of the three valid options and reject the invalid option.

## Description
It seems like this must be the first/only place we're implementing a dropdown list of options without defaulting to a valid option, but it's unlikely it'll be the last. We should consider implementing standardized support for this situation to prevent the need for custom validation logic.

## Possible solutions (in no particular order)
1. Update `OptionStringInput.validate` to reject an empty-string key as invalid.
2. Update `OptionStringInput.validate` to reject some special, invest-defined (non-empty) key (e.g., 'select-an-option', 'OptionStringInputDefault', 'dropdown-placeholder', etc.) as invalid.
3. Omit the invalid/placeholder option from the model spec. Instead, inject a placeholder (e.g., "select an option") option into the UI on the front end instead. This could be designed to be inserted at the beginning of every dropdown list (which would change the behavior of the UI of existing models, but not necessarily for the worse), or a model could opt into this feature through, e.g., a flag in the model spec.
4. ...others?

megannissel:
## Description
Fixes #2260

It would be helpful to have standardized support for a placeholder option in model input dropdown lists. In #2260, Emily outlined a few possible approaches. I elected to handle this primarily on the front end by injecting a placeholder into the list of options. Since the placeholder value isn't in the list of valid input options, our existing validation catches it as an invalid input.

To prevent changing the behavior of existing models, I made the placeholder opt-in by adding an `include_placeholder` attribute to the `OptionStringInput` class. But if we think there's value in always having this behavior (e.g. for consistency), we could remove that attribute. 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -159,7 +159,7 @@ function parseArgType(argtype) {
 export default function ArgInput(props) {
   const uniqueId = useId();
   const inputRef = useRef();
-  const { t, i18n } = useTranslation();
+  const { t } = useTranslation();
 
   const {
     argkey,
--- a/workbench/tests/renderer/setuptab.test.js
+++ b/workbench/tests/renderer/setuptab.test.js
@@ -177,8 +177,6 @@ describe('Arguments form input types', () => {
     const input = await findByLabelText(`${spec.args.arg.name}`);
     expect(input).toHaveDisplayValue('Select an option...');
     expect(input).toHaveValue('placeholderOpt');
-    expect(input).not.toHaveValue('a');
-    expect(input).not.toHaveValue('b');
   });
 
   test('initial arg values can contain extra args', async () => {


workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -159,6 +159,7 @@ function parseArgType(argtype) {
 export default function ArgInput(props) {
   const uniqueId = useId();
   const inputRef = useRef();
+  const { t, i18n } = useTranslation();
```
claire-simpson: What does this do? It looks like `i18n` is unused?

workbench/tests/renderer/setuptab.test.js
```diff
@@ -166,6 +166,21 @@ describe('Arguments form input types', () => {
...
+    spec.args.arg.options = [
+      { key: 'a', display_name: 'Option A' },
+      { key: 'b', display_name: 'Option B' },
+    ];
+    const { findByLabelText } = renderSetupFromSpec(spec, INPUT_FIELD_ORDER);
+    const input = await findByLabelText(`${spec.args.arg.name}`);
+    expect(input).toHaveDisplayValue('Select an option...');
+    expect(input).toHaveValue('placeholderOpt');
+    expect(input).not.toHaveValue('a');
+    expect(input).not.toHaveValue('b');
```
claire-simpson: Are these assertions needed? I.e., if we `expect(input).toHaveValue('placeholderOpt')`, would that necessarily mean the input doesn't have value `a` or `b`?

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
megannissel: I'd been following the pattern used elsewhere in this file, where we're assigning `i18n` even when it isn't used, but upon double-checking the docs, you're right -- we can drop the `i18n` here! 

The `t` is used to wrap "select an option" so that it will be included in translations.

workbench/tests/renderer/setuptab.test.js
megannissel: This was absolutely just a case of me being overly cautious 😅

PR #2366, opened 2026-02-10
Update invest-autotest deployments to get a static URL for sample reports

### LINKED ISSUE #2338: Reports: get static URL for latest model reports deployed from invest-autotest
In the `feature/invest-reports` branch, `invest-autotest` now deploys html reports to GCS, along with other build artifacts.

invest-autotest should deploy reports to a static URL. Similar to how we deploy userguide builds on `main` to https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/index.html, it could be useful to deploy reports from `main` to a "latest" URL. That way we can have static links to the latest report.

We can still also deploy reports to a version-specific URL.

davemfish:
Pushes to `main` or `release/*` branches trigger an `invest-autotest` and then deploy the html reports to GCS alongside other artifacts of a Build Binaries workflow. Now, pushes to `main` also copy those deployed reports over to http://releases.naturalcapitalproject.org/?prefix=invest-reports/latest/ so that we can have static URLs to sample reports like,
* https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-reports/latest/carbon_report_willamette.html

I tested this on my fork and branch by temporarily modifying the conditionals in the workflow, which is why the URLs above exist already.

Fixes #2338 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -73,6 +73,7 @@ General
   Reports exist for the Carbon, Coastal Vulnerability, NDR, & SDR models.
   Reports for other models will be developed and released over time.
   (`#2249 <https://github.com/natcap/invest/issues/2249>`_)
+
     * Reports generated from sample data are available at
       http://releases.naturalcapitalproject.org/?prefix=invest-reports/latest/
       (`#2338 <https://github.com/natcap/invest/issues/2338>`_)


codesigning/enqueue-current-installer.py
```diff
@@ -21,9 +21,9 @@ def main():
     repo_root = os.path.dirname(this_dir)
     dist_url_proc = subprocess.run(
         ['make', '-C', repo_root, '--no-print-directory',
-         'print-DIST_URL_BASE'],
+         'jprint-DIST_URL_BASE'],
```
davemfish, line 24: I modified the `make print` stdout string to remove spaces, which broke this. But the `make jprint-` target seemed more appropriate here anyway.

phargogh: Yep, I agree, `jprint-` is definitely more appropriate here.

PR #2371, opened 2026-02-10
Reports: wrap long raster titles to prevent overlap with adjacent raster titles

### LINKED ISSUE #2356: Reports: text collisions in plot titles
Raster plots use filenames as titles, and those can get very long, ending up wider than the figure axis. Especially if the user appended a long `results_suffix` to their files.

emilyanndavis:
## Description
Fixes #2356.

## Checklist

### One-column layout
text-wrap-1-col_2026-02-09

### Two-column layout, tall AOI
text-wrap-2-col-tall_2026-02-09

### Two-column layout, wide AOI
text-wrap-2-col-wide_2026-02-09

### Three-column layout
text-wrap-3-col_2026-02-09

### DIFF
--- a/src/natcap/invest/reports/raster_utils.py
+++ b/src/natcap/invest/reports/raster_utils.py
@@ -322,7 +322,7 @@ def _get_title_kwargs(raster_path: str, resampled: bool, line_width: int, subtit
     label = f"{filename}{' (resampled)' if resampled else ''}"
     label = textwrap.fill(label, width=line_width)
     if subtitle:
-        subtitle = textwrap.fill(label, width=line_width)
+        subtitle = textwrap.fill(subtitle, width=line_width)
         label = f"{label}\n{subtitle}"
     return {
         'fontfamily': 'monospace',


src/natcap/invest/reports/raster_utils.py
```diff
@@ -307,10 +307,23 @@ def _figure_subplots(xy_ratio, n_plots):
...
+        return 30  # 3-column layout
+
+
+def _get_title_kwargs(raster_path: str, resampled: bool, line_width: int, subtitle: str = ''):
     filename = os.path.basename(raster_path)
     label = f"{filename}{' (resampled)' if resampled else ''}"
-    label = f"{label}\n{subtitle}" if subtitle else label
+    label = textwrap.fill(label, width=line_width)
+    if subtitle:
+        subtitle = textwrap.fill(label, width=line_width)
```
davemfish: Should this one be `textwrapping` the `subtitle`?

emilyanndavis: `subtitle` here is only defined in the context of `plot_raster_facets`, and in that scenario, the `subtitle` is actually appended (after a newline) to the title. It seems right to wrap that text as well, so that's what I've done here. 

Meanwhile, that which we've colloquially called a "subtitle"—the units text—is a matplotlib`text` element, handled separately, and I don't expect that text to require wrapping.

davemfish: > It seems right to wrap that text as well, so that's what I've done here.

@emilyanndavis sorry, I now realize my initial comment was ambiguous. I agree it is correct to wrap the `subtitle` string, but as you have it here, I don't think that's happening:

```python
if subtitle:
    subtitle = textwrap.fill(label, width=line_width)
```
This appears to be mistakenly wrapping the `label` and assigning it to `subtitle`, instead of wrapping the `subtitle`.

emilyanndavis: Oh, I see what you mean now. 🤦🏻‍♀️ Thanks!

davemfish (approved):

PR #2373, opened 2026-02-11
Make `results_suffix` an optional input in ModelSpec

### LINKED ISSUE #2229: make `results_suffix` an optional input
## Background/Motivation
The `results_suffix` is required as a model input and enforced by `ModelSpec` validation. The ability to append a suffix to all output files has been useful to some users, but it has also been annoying to support and maintain IMO. Recently it has gotten much easier with the `FileRegistry` and preprocessing the args dict. Nevertheless, perhaps we should not force plugin developers to include that feature in their plugin?

## Description
A valid `ModelSpec` need not include a `results_suffix` input.

## Benefits
No real benefits to users, but benefits to plugin developers because there would be one less strict rule for them to follow. A plugin developer would probably not include a `results_suffix` input in their model unless they really liked the feature, or were forced to include it.

## Mitigation of side effects
We may still want to mandate that all core models include this feature. And we would no longer be able to rely on validation of the spec to make sure it is there. We would rely on code review instead.

## Alternatives
Developers can copy and modify the `spec.RESULTS_SUFFIX` in order to give it the `hidden=True` attribute in their model spec. Then users cannot see the input, though it still exists. 

## Additional context
I authored a plugin and chose not to support the `results_suffix` feature. Though that was before the FileRegistry and preprocessed args were available to use.

megannissel:
## Description
Fixes #2229

`results_suffix` is an optional input for users to provide, but was a required input to include in the `ModelSpec`. It's probably a good idea to continue following this principle for core InVEST models, but plugin developers may not want to support this feature. Rather than forcing all plugins to do so, we can relax our requirements a bit. 

To help ensure core models continue to support `results_suffix`, I've added a test that checks for the inclusion of a `ResultsSuffixInput` within each `MODEL_SPEC`. 

Implementation note: I considered handling this in `preprocess_inputs` (if no `results_suffix` key in `args`, add it with a value of `''`) but this violates the principle that the dict resulting from `preprocess_inputs` contains exactly the same input keys as the model spec. To me, this seems like a good rule to stand by. 

## Checklist

### DIFF
--- a/tests/test_spec.py
+++ b/tests/test_spec.py
@@ -356,10 +356,9 @@ class MissingResultsSuffixTests(unittest.TestCase):
         """
         from natcap.invest import carbon
 
-        mock_spec = carbon.MODEL_SPEC
-        mock_inputs = carbon.MODEL_SPEC.inputs
-        mock_inputs.pop(1)
-        mock_spec.inputs = mock_inputs
+        # The input at index 1 is the results suffix
+        results_suffix_input = carbon.MODEL_SPEC.inputs.pop(1)
+        assert isinstance(results_suffix_input, spec.ResultsSuffixInput)
 
         args = {
             'workspace_dir': self.workspace_dir,


tests/test_spec.py
```diff
@@ -322,6 +325,78 @@ def test_suffix_string_no_entry(self):
...
+        """Test ModelSpec.execute succeeds without ResultsSuffixInput.
+
+        This test uses the carbon model, as it can be run with minimal
+          inputs and has a reporter
+        """
+        from natcap.invest import carbon
+
+        mock_spec = carbon.MODEL_SPEC
+        mock_inputs = carbon.MODEL_SPEC.inputs
+        mock_inputs.pop(1)
```
claire-simpson: May be worth adding a comment that `.pop(1)` is removing the results suffix just in case we ever change the order of MODEL_SPEC.inputs?

claire-simpson (commented):
Looks good!

tests/test_spec.py
davemfish: Would this test fail/error if it accidentally popped a different input? If not, it might be worth adding an assertion that the popped input is the correct one.

Also, since `mock_spec` is pointing to the same object as `carbon.MODEL_SPEC`, would it be more straightforward to skip the extra assignments and just use the original reference to `MODEL_SPEC`?

e.g. `_ = carbon.MODEL_SPEC.inputs.pop(1)`

megannissel: You're absolutely right; with where this test ended up, the mocking is unnecessary. But we should assert that we're actually removing the results_suffix input, because otherwise the test is useless!

PR #2375, opened 2026-02-11
Add `about` attribute to `ModelSpec` and use it in reports

### LINKED ISSUE #2052: Model specs should include a model description
## Background/Motivation
The `ModelSpec` stores structured model metadata, including both a name and description (the `about` attribute) for model inputs and outputs. The model itself has an id and a title, but no description. As we move towards thinking of models as interchangeable, reusable components, this seems like a useful attribute to add.

This occured to me because of experimenting with `pywps` with the goal of developing an invest server. A pywps `Process` (which would correspond to a single model) takes an `abstract` attribute, for which we currently don't have anything to provide.

The style and length of the description can be pretty open-ended but I would suggest just a few sentences describing the model's general purpose and primary results.

## Description
Add an `about` attribute to the `ModelSpec` class which contains a description or abstract of the model's purpose.

### LINKED ISSUE #2341: Reports: add model descriptions for SDR and NDR
CV and Carbon have model descriptions now as short paragraphs at the top of the report. SDR and NDR should probably have them too.

davemfish:
I set out to just add model descriptions to the SDR and NDR reports, and then decided to do that using the `ModelSpec`. The `about` attribute of the spec is optional, but will instantiate to an empty string if it is missing.

image

Fixes #2341 
Fixes #2052 

Workbench tests failing due to #2376 

## Checklist


tests/reports/test_carbon_template.py
emilyanndavis: Probably safe to remove this mocked `model_description` now.

tests/reports/test_coastal_vulnerability.py
emilyanndavis: I think we can remove this mocked `model_description`, too.

PR #2377, opened 2026-02-12
Reports: add report filename and InVEST version; remove old report from carbon model spec

### LINKED ISSUE #2350: Reports: Include InVEST version in the report
As InVEST reports—and InVEST models—evolve over time, it might be handy to be able to look at a report and quickly determine what version of InVEST was used to generate it.

We already print a line at the end of the page that includes the path to the reporter module and a timestamp. We could add the version number to that line, or maybe even print it in lieu of the reporter filepath, which might not be useful to many users (and may even be something some people wouldn't want to share).

### LINKED ISSUE #2367: Carbon: `report.html` in output spec
`report.html` was formerly an output of the Carbon model. Now, in `release/3.18.0` that report was replaced with `carbon_report_{results_suffix}.html`, which is generated by the reporter module. 

Other models with reporters do not list the html file in the output spec because the file is not actually created by the model's `execute` function.

emilyanndavis:
## Description
Fixes #2350 and fixes #2367.

## Checklist

### With results suffix
report-footer-with-suffix_2026-02-12

### Without results suffix
report-footer-without-suffix_2026-02-12


src/natcap/invest/reports/templates/styles.html
```diff
@@ -152,10 +174,6 @@
   td {
       padding: 0.375rem 0.5rem;
   }
-  .number {
-      text-align: right;
-      font-family: monospace;
-  }
```
emilyanndavis, line 158: We weren't using this class anywhere.

src/natcap/invest/carbon/reporter.py
```diff
@@ -219,7 +220,9 @@ def report(file_registry: dict, args_dict: dict, model_spec: ModelSpec,
 
     with open(target_html_filepath, 'w', encoding='utf-8') as target_file:
         target_file.write(TEMPLATE.render(
-            report_script=__file__,
+            report_script=model_spec.reporter,
+            invest_version=__version__,
+            report_filename=f'{model_spec.model_id}_report{args_dict["results_suffix"]}.html',
```
davemfish: We should be able to just re-use the `target_html_filepath` variable here. Perhaps with `os.path.basename` if we don't want the full path.

src/natcap/invest/reports/templates/base.html
```diff
@@ -25,10 +25,21 @@ <h1>{{ self.page_title() }}</h1>
...
+  {% block footer %}
+  <footer class="report-container">
+    <div class="report-info">
+      <p>
+        This report was generated by {{ report_script }} at {{ timestamp }}
+        using InVEST version {{ invest_version }}.
+      </p>
+      <p>
+        If this is your report, you can find it in your workspace folder.
+        Its name is {{ report_filename }}.
```
davemfish: It comes down to personal preference, but how about something a little less...personal?
>This report is saved in your workspace folder at {{ report_filename }}

Or even,
>This report is saved as {{ full_path_to_html_file }}

The workspace dir already appears elsewhere in the report, so I don't feel like we need to hide it if we don't want to. And if someone has multiple reports open at once, this could be handy to see which is which when `report_filename` would otherwise be the same.

src/natcap/invest/carbon/reporter.py
emilyanndavis: D'oh, I knew I was missing something obvious. Thanks.

src/natcap/invest/reports/templates/base.html
emilyanndavis: Counterpoint: what if we lean into personification and slap a big "Hello, my name is..." sticker on the top of the report? 🤪 

Jokes aside, I think what I was going for was the following:
1. to make it clear that the location of the report is only relevant to the person who ran the model. (If someone posted a report on the web and shared a link with you, it's not automatically saved to your computer, and even if you downloaded it, we don't know where it is or even necessarily what it's called.)
2. to provide just the report filename, without the rest of the path. But you're right, including the path could be helpful here. I do think it might be worth providing an option to let a user control whether full paths are printed in the report, but that would be a separate issue.

davemfish: Yes, these are also good points!

davemfish (approved):

PR #2378, opened 2026-02-12
Update Workbench forum URLs to include tag slug and ID

### LINKED ISSUE #2376: Workbench `Test building model UIs and forum links > test each forum link` failing: 301 instead of 200
The workbench tests that check our forum links have begun failing. Rather than returning a `200` status code, they're returning a `301` redirect (see [example](https://github.com/natcap/invest/actions/runs/21920922755/job/63305324384?pr=2365)).

It looks like Discourse made a change -- now links to `/tag/{tag_name}` redirect to `/tag/{tag_name}/{tag_id}`. Possibly implemented [here](https://github.com/discourse/discourse/commit/737577e8b47feda9535bd89963994851b491dc09)? Though that was only 19 hours ago and I wouldn't have expected it to already have an effect.

Regardless, seems like we need to update either the Workbench or those tests. Two possible approaches that come to mind:
1. Track the Discourse tag IDs in addition to tag names, and update the Workbench FAQ links to point to the tag with the ID (so that the link returns status code `200`)
2. Change the `expect(status).toBeStatus200(address)` with a check based on the URL the user will end up on (e.g. includes `/tag/` and a relevant-to-the-model tag?)

megannissel:
## Description
Fixes #2376

Discourse core [updated tag paths](https://github.com/discourse/discourse/pull/37055) so that they include both the tag slug and ID. Links to the old `/tag/{tagName}` paths now redirect to the new canonical URLs. This led to workbench test failures, where we were expecting a `200` status code. This PR updates the Workbench Forum tag URLs to match the new expected pattern and avoid the redirect. 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/ResourcesLinks/index.jsx
+++ b/workbench/src/renderer/components/ResourcesLinks/index.jsx
@@ -11,48 +11,49 @@ const { ipcRenderer } = window.Workbench.electron;
 
 const FORUM_ROOT = 'https://community.naturalcapitalalliance.org';
 
-// map model names to forum tags:
+// map model names to forum tags
+// forum tag paths include both slug and ID
 const FORUM_TAGS = {
-  annual_water_yield: 'annual-water-yield',
-  carbon: 'carbon',
-  coastal_vulnerability: 'coastal-vulnerability',
-  coastal_blue_carbon: 'blue-carbon',
-  coastal_blue_carbon_preprocessor: 'blue-carbon',
-  crop_production_percentile: 'crop-production',
-  crop_production_regression: 'crop-production',
-  delineateit: 'delineateit',
-  forest_carbon_edge_effect: 'carbon-edge-effects',
-  habitat_quality: 'habitat-quality',
-  habitat_risk_assessment: 'hra',
-  ndr: 'ndr',
-  pollination: 'pollination',
-  recreation: 'recreation',
-  routedem: 'routedem',
-  scenario_generator_proximity: 'scenario-generator',
-  scenic_quality: 'scenic-quality',
-  sdr: 'sdr',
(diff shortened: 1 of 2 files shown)


workbench/src/renderer/components/ResourcesLinks/index.jsx
```diff
@@ -11,48 +11,49 @@ const { ipcRenderer } = window.Workbench.electron;
...
-  stormwater: 'urban-stormwater',
-  urban_cooling_model: 'urban-cooling',
-  urban_flood_risk_mitigation: 'urban-flood',
-  urban_nature_access: 'urban-nature-access',
-  wave_energy: 'wave-energy',
-  wind_energy: 'wind-energy',
+  annual_water_yield: 'annual-water-yield/19',
+  carbon: 'carbon/1',
+  coastal_vulnerability: 'coastal-vulnerability/8',
+  coastal_blue_carbon: 'blue-carbon/3',
```
claire-simpson, line 20: Looks like `blue-carbon` is duplicated for CBC and the preprocessor! And on the forum I see both a `blue-carbon` and `coastal-blue-carbon` tag... For the `coastal-blue-carbon` tag I see: `https://community.naturalcapitalalliance.org/tag/coastal-blue-carbon/26`. On the forum side, I wonder if we should combine these two tags (which could be possible with the unique ID)?

megannissel: I noticed that as well, but the Workbench was already pointing to the `blue-carbon` tag for both the model and preprocessor, so it made sense to me that `blue-carbon/3` would be the appropriate updated path. That said, I think it's worth considering if we can / should merge those two tags -- I tossed a question into Slack to see what everyone thinks.

PR #2383, opened 2026-02-13
Reports: enhancement to `RasterPlotConfig`

### LINKED ISSUE #2372: Reports: Stream maps should include flow algorithm in subtitle
It's nice to remind the reader of the flow algorithm (D8 vs MFD) used to generate the stream network. Right now it's included in the caption of the figure, but it might deserve to be more prominent, like as a subtitle of the plot.

Image

> stream_gura.tif
Stream network, created using flow direction and flow accumulation derived from the DEM and Threshold Flow Accumulation. Values of 1 represent streams, values of 0 are non-stream pixels. Results were generated using the following flow direction algorithm: MFD. The stream network may look incomplete at this resolution, and therefore it may be necessary to view the full-resolution raster in GIS to assess its accuracy.

davemfish:
`raster_utils.RasterPlotConfig` is now responsible for constructing the default caption text for a raster subplot. It requires the `natcap.invest.spec.Input | spec.Output` in order to do so.

* Captions and titles can be overridden by setting those attributes of the instance.
* NDR, SDR, and Carbon reporters were refactored to use the new class.
* The flow direction algorithm is listed in the section header of the Stream Network Maps for SDR and NDR.

Fixes #2372 

## Checklist
- [x] Tested the sample data reports

### DIFF
--- a/src/natcap/invest/ndr/reporter.py
+++ b/src/natcap/invest/ndr/reporter.py
@@ -1,6 +1,7 @@
 from natcap.invest.reports import raster_utils
 from natcap.invest.reports import report_constants
 from natcap.invest.reports import sdr_ndr_report_generator
+from natcap.invest.reports.raster_utils import RasterDatatype, RasterPlotConfig
 
 CALC_N = 'calc_n'
 CALC_P = 'calc_p'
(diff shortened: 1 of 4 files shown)


src/natcap/invest/ndr/reporter.py
```diff
@@ -80,46 +57,75 @@ def report(file_registry, args_dict, model_spec, target_html_filepath):
...
+        datatype='continuous',
+        spec=model_spec.get_output('p_surface_export'))
+
+    output_raster_plot_configs = []
+    if CALC_N in args_dict:
+        output_raster_plot_configs.extend(
+            [n_surface_export_config,
+             n_subsurface_export_config,
+             n_total_export_config])
+    if CALC_P in args_dict:
```
emilyanndavis: I think we want these conditions to be `if args_dict[CALC_N]` and `if args_dict[CALC_P]`, respectively, since—I believe—both of those will always be in the `args_dict`, they just won't always be `True`.

davemfish: Thanks; updated.

tests/reports/test_sdr_ndr_utils.py
```diff
@@ -163,7 +163,7 @@ def test_generate_results_table_and_totals(self):
 
         # Check totals table values.
         # calculated_value_1 is ws_id + 100; calculated_value_2 is ws_id + 200.
-        expected_totals = [101 + 102, 201 + 202]
+        expected_totals = [101.0 + 102.0, 201.0 + 202.0]
```
davemfish, line 166: These tests started failing for some reason because the actual data was a float represented as a string, but the expected data was an integer as string. `("203" != "203.0")`

tests/reports/test_raster_utils.py
```diff
@@ -456,41 +484,35 @@ class RasterCaptionTests(unittest.TestCase):
...
+            spec=spec.Input(
+                id='raster_1',
+                about=about_raster_1))
+
+        about_raster_2 = ('The total amount of sediment exported from each '
+                          'pixel that reaches the stream.')
+        caption_appendix = 'extra text'
+        raster2_config = raster_utils.RasterPlotConfig(
+            raster_path=file_registry['raster_2'],
+            datatype='continuous',
```
emilyanndavis: Minor detail, but for the datatypes, I would use the enums, e.g. `RasterDatatype.nominal`, `RasterDatatype.continuous`, rather than string literals.

tests/reports/test_sdr_ndr_utils.py
emilyanndavis: How strange. I wonder what changed?

src/natcap/invest/reports/report_constants.py
```diff
@@ -12,4 +12,10 @@
     'rasters are available in the output workspace.'
 )
 
+STREAM_CAPTION_APPENDIX = gettext(
+    'The stream network may look incomplete at  this resolution, and '
+    'therefore it may be necessary to view the  full-resolution raster '
```
emilyanndavis: Looks like we ended up with some extra spaces in here.
```suggestion
    'The stream network may look incomplete at this resolution, and '
    'therefore it may be necessary to view the full-resolution raster '
```

src/natcap/invest/sdr/reporter.py
```diff
@@ -40,44 +18,78 @@ def report(file_registry, args_dict, model_spec, target_html_filepath):
...
     """
+    input_raster_plot_configs = [
+        raster_utils.RasterPlotConfig(
+            raster_path=args_dict['dem_path'],
+            datatype='continuous',
+            spec=model_spec.get_input('dem_path')),
+        raster_utils.RasterPlotConfig(
+            raster_path=args_dict['erodibility_path'],
+            datatype='continuous',
+            spec=model_spec.get_input('dem_path')),
```
emilyanndavis: Looks like this line missed an update after copy-paste.
```suggestion
            spec=model_spec.get_input('erodibility_path')),
```

src/natcap/invest/reports/report_constants.py
emilyanndavis: I think we'll also want to include a leading space, since this is appended to the about text from the model spec (which we don't expect to end with a trailing space).
```suggestion
    ' The stream network may look incomplete at this resolution, and '
    'therefore it may be necessary to view the full-resolution raster '
```

emilyanndavis (changes requested):
Great improvements, thanks @davemfish! Just a few small suggestions.

tests/reports/test_sdr_ndr_utils.py
davemfish: I assume there was a change in behavior from some dependency somewhere, but I didn't trace it back.

PR #2392, opened 2026-02-19
Finalize and remove references to TCC inputs (option 1) for urban mental health

### LINKED ISSUE #2391: Implement final pieces of Urban Health Model
This issue is a catch-all for the final pieces of urban mental health, including removing all references to original model option 1 (TCC and NDVI inputs). Implementing option 1 is on hold currently (see #2141), however there are some aspects of PR #2314 that we'd like to incorporate in the model.

claire-simpson:
## Description
Remove original model option 1 (user inputs tree canopy cover and NDVI) and other final changes to the model, including keeping up with 3.18.0

In addition:
- Added tasks to clip input rasters (including population), partly so they look better in the report
- Fixed a bug where population raster wasn't snapping correctly to aligned NDVI
- Handle input NDVI and LULC rasters without a NoData value set
- Added a function to calculate target pixel size for realignment if input rasters aren't linearly projected in meters

Fixes #2391 

## Checklist

### DIFF
--- a/src/natcap/invest/urban_mental_health/urban_mental_health.py
+++ b/src/natcap/invest/urban_mental_health/urban_mental_health.py
@@ -960,9 +960,9 @@ def _get_raster_pixel_size_in_meters(raster_path, vector_path):
         tgt_pixel_size = numpy.mean([abs(raster_info["pixel_size"][0]),
                                      abs(raster_info["pixel_size"][1])])
         LOGGER.info("Baseline raster is projected in meters; will use pixel "
-                    f"size {tgt_pixel_size} as target in align_and_resize,"
+                    f"size {tgt_pixel_size} as target in align_and_resize, "
                     "which is the native resolution of the raster (transformed"
-                    "to have square pixels if it doesn't already)")
+                    "to have square pixels if it doesn't already).")
         return (tgt_pixel_size, -tgt_pixel_size)
     else:
         vector_info = pygeoprocessing.get_vector_info(vector_path)


src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -702,9 +675,8 @@ def execute(args):
             dependent_task_list=[create_lulc_ndvi_csv_task],
             task_name="reclassify base lulc to ndvi and mask"
         )
-        buffer_base_dependencies.append(reclassify_base_lulc_task)
 
-        reclassify_alt_lulc_task = task_graph.add_task(
```
claire-simpson, line 707: Changed the variable name so that `mean_buffered_base_ndvi_task`'s `dependent_task_list` can contain the same task regardless of which scenario is selected

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -11,19 +11,31 @@
...
+from natcap.invest import gettext
+from natcap.invest import spec
+from natcap.invest import utils
+from natcap.invest import validation
+from natcap.invest.unit_registry import u
 
 LOGGER = logging.getLogger(__name__)
 FLOAT32_NODATA = float(numpy.finfo(numpy.float32).max)
 
+_model_description = gettext(
```
claire-simpson, line 23: We probably want Yingjie or Stacie to improve this

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1414,16 +1374,6 @@ def validate(args, limit_to=None):
...
-    scenario_options = [
-        spec.Option(key="tcc_ndvi", display_name="Tree Cover & Baseline NDVI"),
-        spec.Option(key="lulc", display_name="Baseline & Alternate LULC"),
-        spec.Option(key="ndvi", display_name="Baseline & Alternate NDVI")
-    ]
-
-    error_msg = spec.OptionStringInput(
-        id='', options=scenario_options).validate(args['scenario'])
-    if error_msg:
-        validation_warnings.append((['scenario'], "Must select a scenario."))
```
claire-simpson, line 1426: Now that `include_placeholder` exists, I've removed this custom validation. If a Python API user unexpectedly sets `scenario` to something other than `ndvi` or `lulc`, they will probably get an unhelpful error message, but we'd expect the user to validate their inputs args right?

dcdenu4: Yea, this is always something that can end up being a subjective choice. I think we've leaned towards API users being responsible for validating args and then handling Python errors. I think what we wouldn't want is unexpected behavior, where a wrong input could still lead to a successful run, but with really misleading results.

tests/test_urban_mental_health.py
```diff
@@ -423,11 +431,81 @@ def test_population_raster_too_small(self):
...
+        bounding box to the pixel grid of the alignment raster
+        (origin + integer multiples of pixel size).
+
+        This tests that population_aligned bbox == the preprocessed ndvi bbox
+        and both equal the AOI-buffered bbox snapped outward to that grid.
+        """
+        from natcap.invest.urban_mental_health import urban_mental_health
+
+        args = make_synthetic_data_and_params(self.workspace_dir, 'ndvi')
+        urban_mental_health.execute(args)
```
dcdenu4: Just noting something that popped into my head. Now that our models return a file registry, we could use that for getting the output / intermediate paths to test against. I don't think it's necessary to update these tests for that but wanted to mention it!

tests/test_urban_mental_health.py
```diff
@@ -423,11 +431,81 @@ def test_population_raster_too_small(self):
...
+        ndvi_raster_path = os.path.join(
+            intermediate, f"ndvi_base_aligned_{suffix}.tif")
+        ndvi_info = pygeoprocessing.get_raster_info(
+            ndvi_raster_path)
+        ndvi_bbox = ndvi_info['bounding_box']
+
+        # Quick check - population aligned bbox matches the ndvi bbox
+        numpy.testing.assert_allclose(pop_bbox, ndvi_bbox)
+
+        # Build AOI buffered bbox (vector-derived)
```
dcdenu4: The model doesn't output the buffered AOI vector?

tests/test_urban_mental_health.py
```diff
@@ -423,11 +431,81 @@ def test_population_raster_too_small(self):
 
         numpy.testing.assert_allclose(actual_prev_cases, expected_prev_cases)
 
-    def test_AOI_larger_than_lulc_base_option3(self):
+    def test_population_aligned_bbox_matches_snapped_processing_bbox(self):
```
dcdenu4, line 423: This is a fun test, I'm curious to hear what motivated you to add it?

dcdenu4 (commented):
Thanks @claire-simpson ! Just a few comments / questions I was curious about.

tests/test_urban_mental_health.py
claire-simpson: I initially set `population_align_task` further down in the order of events (after `delta_ndvi_task`), so I could use the `delta_ndvi` bounding box as the `target_bb`, to ensure that the population raster was exactly aligned with `delta_ndvi`, that is, exactly aligned with the convolved buffer-mean NDVI rasters. However, when I was rearranging the logic to try to avoid having too many `if args['scenario'] == <some scenario>` clauses, it simplified things to move the population resampling task further up. With this test, I wanted to make sure everything would still align if I just set `target_bb = aoi_buffered_bbox` in the population resampling task, same as the target when aligning all other input rasters. (So I guess my main motivation was to add a test catch any extent issues if the logic were to be rearranged later due to re-adding the TCC option )

claire-simpson: Looking at this now though, I think there is an error, and I intended to be testing that the extent of the aligned population raster matched that of `delta_ndvi.tif` _not_ `ndvi_base_aligned.tif`

tests/test_urban_mental_health.py
claire-simpson: It does not, but maybe it should so then we could leverage the `mask_options` parameter in `align_and_resize_raster_stack` to automatically clip all the input rasters to the buffered AOI. 

Currently I'm not actually using the buffered AOI anywhere, instead I'm just using the buffered AOI bounding box (calculated by adding the `search_radius` distance to the extent of the AOI) as the target bounding box in the align and resample tasks. However I know we talked about actually buffering the AOI instead, which could be beneficial for clipping all the inputs up front, thereby potentially reducing file size down the road and computation on the convolution operation. The cons are that buffering a complex [multi]polygon can take a bit/adds a step, and we really only _need_ to use the bounding box, as the final products are all clipped to the AOI regardless.

tests/test_urban_mental_health.py
claire-simpson: Ooh this is a good suggestion, I'm inclined to just do it now? Unless you'd rather keep this PR simple

tests/test_urban_mental_health.py
claire-simpson: I decided to go for it and use the buffered AOI to clip, but easy enough to roll it back!

tests/test_urban_mental_health.py
dcdenu4: Do it!

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -633,13 +620,13 @@ def execute(args):
...
-    aoi_buffered_bbox = aoi_bbox + numpy.array(
-        [-args['search_radius'], -args['search_radius'],
-            args['search_radius'], args['search_radius']])
-    aoi_buffered_bbox = list(aoi_buffered_bbox)
+    # Buffer AOI by search_radius
+    gdf = geopandas.read_file(args['aoi_path'])
+    gdf["geometry"] = gdf.buffer(args['search_radius'])
+    gdf.to_file(file_registry['aoi_buffered'], driver='GPKG')
+    aoi_buffered_bbox = pygeoprocessing.get_vector_info(
+        file_registry['aoi_buffered'])['bounding_box']
```
dcdenu4: Geopandas sure does make this easy. I know that as a new dependency of the Reports feature, we talked about maybe not relying on geopandas in the future. We could implement this using Shapely or leave as is knowing we might need to refactor it if we ever decided to move away from geopandas as a dependency.

tests/test_urban_mental_health.py
dcdenu4: I left a comment on the `geopandas` implementation.

> However I know we talked about actually buffering the AOI instead, which could be beneficial for clipping all the inputs up front, thereby potentially reducing file size down the road and computation on the convolution operation.

Wouldn't we get this result anyway by just passing the aoi buffered bounding box when using align and resize?

>The cons are that buffering a complex [multi]polygon can take a bit/adds a step, and we really only need to use the bounding box, as the final products are all clipped to the AOI regardless.

Yea, from a user perspective I think it could be helpful to have an output to visually see what the model is doing. But, there are certainly tradeoffs with adding another operation and then depending on `geopandas`. Would we expect a user to upload a complicated multi polygon for this? I don't think we'd want to buffer a multi polygon because I'd assume the overlaps would get pretty weird. We really do only care about buffering the extents. In which case we maybe a simpler solution would be to just create a new polygon output from the calculated extents?

claire-simpson: > Wouldn't we get this result anyway by just passing the aoi buffered bounding box when using align and resize?

I guess the scenario I was thinking of is if you had an AOI that wasn't square-like, like a disparate multipart polygon or long diagonal, so clipping to a buffered bounding box would still mean keeping all the in-between pixels and pixels in the bbox vs. clipping to a buffer would allow you to get rid of these unneeded pixels and not have to convolve over them.

> Yea, from a user perspective I think it could be helpful to have an output to visually see what the model is doing.

To clarify, you'd like to save the buffered bounding box as a vector? 

> Would we expect a user to upload a complicated multi polygon for this? 

If we anticipate any sort of large-scale regional runs (e.g., the US), then we could definitely get into this territory

> I don't think we'd want to buffer a multi polygon because I'd assume the overlaps would get pretty weird. We really do only care about buffering the extents. In which case we maybe a simpler solution would be to just create a new polygon output from the calculated extents?

The way PGP handles clipping to a vector in `align_and_resize_raster_stack` is by converting that vector into a raster and masking, so I don't think overlapping buffers would be an issue (unless `geopandas` is weird about this?). However if you think its better to stay the course with just using a buffered bbox I'm happy to defer.

(I'll also throw out that another reason I'm leaning towards clipping to a buffer is it would make the report look better! Though now that I'm typing this I should maybe just move the "clip to AOI" operation to occur immediately after convolving the ndvi rasters)

tests/test_urban_mental_health.py
```diff
@@ -105,8 +109,10 @@ def make_synthetic_data_and_params(workspace_dir, model_option):
                                     (ORIGIN_X, ORIGIN_Y),
                                     (ORIGIN_X, ORIGIN_Y-250)])])
 
+    # make pop are different shape than ndvi to test if resampled correctly
```
emilyanndavis: I think there might be a typo in this comment?

emilyanndavis (approved):

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -4,26 +4,39 @@
 import shutil
 import tempfile
 
+import geopandas
```
dcdenu4: Remove now?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -951,6 +903,44 @@ def check_raster_against_aoi_bounds(aoi_bbox, aoi_sr, raster):
...
+    Args:
+        input_ndvi (str): path to input NDVI raster
+        kernel (str): path to kernel raster
+        aoi (str): path to AOI vector to clip output raster to
+        work_dir (str): path to workspace directory for temporary files
+        target_raster (str): path to output convolved and clipped raster
+
+    Returns:
+        None
+
```
dcdenu4:
```suggestion
```

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -951,6 +903,44 @@ def check_raster_against_aoi_bounds(aoi_bbox, aoi_sr, raster):
             f"and the raster bbox is: {raster_bbox}")
 
 
+def convolve_ndvi_with_kernel(input_ndvi, kernel, aoi, work_dir,
```
dcdenu4: Since this function is also masking I might add to the function name. Maybe `convolve_and_mask_ndvi_with_kernel`.

I see the advantages of wrapping these operations into a single function, to help reduce code duplication. A couple tradeoffs are we're asking a function to do multiple tasks and hiding that a bit from the execute function and from the user. I don't think it's a problem in this case, but not have that `intermediate_raster` layer could potentially make it more difficult to debug in the future.

dcdenu4 (changes requested):
One small change request and a comment! Thanks @claire-simpson .

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: Good point, I'll separate these two operations and store the unmasked version as well

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -800,52 +824,54 @@ def execute(args):
...
-            'target_projection_wkt': aoi_projection,
-            'working_dir': args['workspace_dir'],
-        },
-        target_path_list=[file_registry['population_aligned']],
-        task_name='Resample population to NDVI resolution')
-
     baseline_cases_task = task_graph.add_task(
         func=calc_baseline_cases,
-        args=(file_registry['population_aligned'],
+        args=(file_registry['population_aligned_clipped'],
```
dcdenu4, line 868: Does this calculation and other calculations need to operate on the buffered AOI space or is it fine to operate on the clipped AOI?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+    """Get raster pixel size in meters; if not projected in m, use vector CRS.
+
+    Use gdal to auto calculate the target pixel size in meters if transforming
+    the raster to the CRS of the input vector AOI. This is necessary because
+    we do not want to force users to initially project their input rasters
+    in meters, however align_and_resize requires a target pixel size.
+
+    Args:
+        raster_path (str): Path to baseline LULC or baseline NDVI raster,
+            which is projected by may not be projected in meters.
```
dcdenu4:
```suggestion
            which may or may not be projected in meters.
```

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+        vector_path (str): Path to AOI vector in a projected CRS with
+            meter units.
```
dcdenu4:
```suggestion
            units in meters.
```

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+
+    Returns:
+        tuple[float, float]: (pixel_width_m, pixel_height_m)
+        pixel_size[float]: target pixel size in meters (to use when
+            aligning and resizing raster stack).
+    """
+    def _raster_projected_in_m(raster_info):
```
dcdenu4: Since only the `projection_wkt` is used, does this helper function need to take in all of `raster_info`?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+    if _raster_projected_in_m(raster_info):
+        LOGGER.info("Baseline raster is projected in meters; will use native "
+                    f"pixel size {raster_info['pixel_size']} as target "
+                    "in align_and_resize")
+        return raster_info["pixel_size"]
+    else:
+        vector_info = pygeoprocessing.get_vector_info(vector_path)
+        vector_wkt = vector_info["projection_wkt"]
+
+        src_ds = gdal.Open(raster_path)
```
dcdenu4, line 971: It's usually good practice to proactively close or clean up opened datasets. After we have what we need from the dataset could we add: `src_ds = None`?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+        transformer = gdal.Transformer(src_ds, None, [f'DST_SRS={vector_wkt}'])
+        target_warp = gdal.SuggestedWarpOutput(src_ds, transformer)
```
dcdenu4, line 973: This is really interesting. We've had to get pixel sizes in meters given a lat/long coordinate system before and it's usually quite the process. Some examples can be found in the wave energy model and pygeoprocessing. I think those use cases are slightly different than here.

`gdal.Transformer` and [gdal.SuggestedWarpOutput](https://gdal.org/en/stable/api/python/utilities.html#osgeo.gdal.SuggestedWarpOutput) are new to me and I can find documentation on the latter, but don't see anything formal for the former. But, `SuggestedWarpOutput` essentially does what we'd often do, but maybe more thoroughly, which is:
> It walks the edges of the input file (approximately 20 sample points along each edge) transforming into output coordinates in order to get an extents box.

In places where we've done this type of thing before we've used [osr.CoordinateTransformation](https://gdal.org/en/stable/api/python/spatial_ref_api.html#coordinatetransformation).

The two-liner you have feels like magic, but I do want to double check to make sure we don't already have a pygeoprocessing or invest utilty for this. Though, if we do, perhaps it should be refactored to use this approach if appropriate.

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -900,6 +924,59 @@ def execute(args):
...
+
+    Args:
+        raster_path (str): Path to baseline LULC or baseline NDVI raster,
+            which is projected by may not be projected in meters.
+        vector_path (str): Path to AOI vector in a projected CRS with
+            meter units.
+
+    Returns:
+        tuple[float, float]: (pixel_width_m, pixel_height_m)
+        pixel_size[float]: target pixel size in meters (to use when
```
dcdenu4: Is this returning two things? Or is it just returning a tuple of the pixel size?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1146,12 +1225,30 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
     mean_ndvi_by_lulc_dict = lulc_df['ndvi'].to_dict()
     if source_nodata is not None:
         mean_ndvi_by_lulc_dict[source_nodata] = target_nodata
+    else:
+        LOGGER.warning("The input LULC raster has no NoData value. This can "
```
dcdenu4:
```suggestion
        LOGGER.warning("The input LULC raster has no NoData value set. This can "
```

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1146,12 +1225,30 @@ def reclassify_lulc_raster(lulc, mean_ndvi_by_lulc_csv, target_path):
...
-                                      'column_name': 'lucode',
-                                      'table_name': 'LULC attribute'})
+    try:
+        utils.reclassify_raster(
+            (lulc, 1), mean_ndvi_by_lulc_dict, target_path, target_datatype,
+            target_nodata, error_details={'raster_name': lulc,
+                                          'column_name': 'lucode',
+                                          'table_name': 'LULC attribute'})
+    except ValueError as e:
+        if any(f"table are: [{val}]" in str(e) for val in ['0', '0.', '0.0']):
```
dcdenu4, line 1245: I'm curious about this extra error handling logic. Is there a specific reason you're looking for 0s here? My below comment might or might not apply depending on the use case you have in mind here.

I think the note on NoData being something to look out for is a good one and I wonder if that should be added to the utility function error message? For me, this custom error message is trying to do to much by going out of its way to highlight a specific potential problem related to the AOI buffering thing and then having a footnote at the end that tries to capture the default error message from the function (though I guess it's dependent on 0 values being identified). I think the note about NoData being something to look out for captures the broader issue and if someone did have NoData set and they had NoData values within the buffered AOI, they wouldn't necessarily see this message but have the same problem.

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: The purpose of adding the buffer is just to ensure full data coverage for the 2d convolution operation (i.e., `mean_buffered_base_ndvi_task` and `mean_buffered_alt_ndvi_task`), so any functions after that don't need to use the buffered AOI space!

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: Good point!

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: Yes definitely

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: Ah whoops, I think I just want to return the average of pixel size height and width, to ultimately make square pixels (so the search radius distance in pixels is accurate)

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: > Is there a specific reason you're looking for 0s here?

If a raster (a) doesn't have a nodata value set and (b) has an extent that is smaller than the target bounding box, when we `align_and_resize` it, the `align_and_resize_raster_stack` function pads the output aligned/resized raster with `0`s (where normally those pixels would be nodata). 

This is particularly problematic for this use case as `0` is a reasonable `lucode` value, which could fall through the cracks and get reclassified as whatever NDVI value `lucode=0` corresponds with. Perhaps there is a better way of consistently detecting this worst-case-scenario (small extent, no NoData value, `0` is an lucode listed in the table), but this case seems like it would be rare and a bit hard to detect, so I went for a warning instead. And  I wanted to add this first warning regardless of whether there is an actual `Values in the lulc_base_aligned.tif raster were found that are not represented under the 'lucode' column of the LULC attribute table` error, so if a user gets a weird output this would help track down the issue. I didn’t raise an error here because an LULC raster without a set nodata value will be fine as long as it has a large enough extent.

> I think the note on NoData being something to look out for is a good one and I wonder if that should be added to the utility function error message?

To clarify: add more text to the “Values in the raster were found that are not represented under the column of the attribute table” error in `reclassify_raster` to explain that this error can occur if a NoData value is not set for the input raster?  

> For me, this custom error message is trying to do to much by going out of its way to highlight a specific potential problem related to the AOI buffering thing and then having a footnote at the end that tries to capture the default error message from the function (though I guess it's dependent on 0 values being identified).

That’s fair. The idea behind the second `ValueError` was to help users actually understand why they're seeing an error, because I think the link between "the LULC raster is too small and doesn't have a nodata value" and "there is no 0 in your lulc attribute table" error wouldn't be obvious. It definitely hinges on `0` being the value that isn’t able to be reclassified as this is a pretty specific scenario I’m trying to help prevent/explain. 

> I think the note about NoData being something to look out for captures the broader issue and if someone did have NoData set and they had NoData values within the buffered AOI, they wouldn't necessarily see this message but have the same problem.

If I am understanding your concern: I don’t think `reclassify_raster` would error when trying to reclassify legitimate NoData pixels, these pixels would just remain NoData. So I think this issue is specific to the case where there are pixels that _should_ be NoData (i.e., they are beyond the extent of the user-input LULC), but haven’t been correctly classified as such (because `align_and_resize_raster_stack` sets these pixels as `0` since there isn't a NoData value speicfied to set them to).

Let me know what you think about all this! These 2 warnings and errors are definitely a bit cumbersome, I felt like it was important to give some context, if for no other reason than to help ourselves debug future issues, particularly given that the default `reclassify_raster` error isn't very helpful in understanding how this "extra" value got added to the dataset

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: I couldn't find a function for this when looking through utils or PGP, but this wave energy function looks relevant! This option is definitely simpler and I found it by looking into how gdal warp determines a target pixel size if one isn't given, so I'd probably vote for refactoring wave energy. (I also couldn't find documentation for `gdal.Transformer` beyond what was included as part of the `SuggestedWarpOutput` description.)

src/natcap/invest/urban_mental_health/urban_mental_health.py
dcdenu4: Thanks for your thoughts.
>add more text to the “Values in the raster were found that are not represented under the column of the attribute table” error in reclassify_raster to explain that this error can occur if a NoData value is not set for the input raster?  

This is what I was thinking would be helpful. The case where NoData is not set on the raster and is not in the table, so then I _think_ it'd cause an error because it would be treated as a valid pixel value.

I think this is all fine to leave in and we can revisit in the future if we need to.

PR #2395, opened 2026-02-19
Resolve conflict to merge `main` in `feature/urban-mental-health`

claire-simpson:
## Description
I wanted to merge `main` into natcap's umh feature branch, but there was a merge conflict with `/workbench/src/renderer/components/ResourcesLinks/index.jsx` which I had to resolve on my fork's branch. So I merged `main` into `claire-simpson/feature/urban-mental-health` and resolved the conflict, so this PR is effectively attempting to merge main into `natcap/invest/feature/urban-health-model` by way of `claire-simpson/invest/feature/urban-health-model`. All the commits here should've been reviewed already except the line 
`urban_mental_health: 'urban-mental-health', //TODO - add ID when forum tag is created` in `/workbench/src/renderer/components/ResourcesLinks/index.jsx`.
 
Fixes #

## Checklist


tests/aoi.geojson
```diff
@@ -0,0 +1,8 @@
+{
```
megannissel: Was this file meant to be included? I'm pretty sure this file gets created by one of the rec tests (which I ought to open an issue to address - I think we're not correctly setting the file path it's not getting cleaned up at the end of the test run)

megannissel (changes requested):
I think `tests/aoi.geojson` isn't meant to be here? But otherwise I don't see any issues!

tests/aoi.geojson
claire-simpson: Ugh I wonder if I somehow added that...

megannissel: I tracked down the test that's to blame for creating that file and opened an issue: #2396

claire-simpson: Ok just deleted this file and the 4.0.7 file

claire-simpson: Ok yep looked through my history and I accidentally added those 2 files by running `git add .` from within the test folder, but at least I can be confident I didn't randomly add anything else in this PR!

claire-simpson: Thanks for catching this!!

PR #2409, opened 2026-02-23
add a custom save as click handler to the context menu.

### LINKED ISSUE #2404: Reports: Workbench crashes upon dismissing download popup after selecting "Save Image As..." from context menu
## Description
Workbench crashes upon dismissing download popup after selecting "Save Image As..." from context menu.

## To replicate

1. Open the Workbench.
2. Run Carbon, NDR, or SDR.
3. Use the "View Results" button to open the report.
4. Right-click on an image to open the context menu, and select "Save Image As...".
5. You should see a popup with some options along with "Cancel" and "Save" buttons. Dismiss this popup using the "Cancel" button or the `esc` key on your keyboard.

## Expected behavior
The popup should disappear. Nothing else should change.

## Observed behavior
The Workbench closes immediately.

## Log
```
[12:46:22.507] [{label}] unhandled rejection at: [object Promise]
[12:46:22.510] [{label}] Error
    at DownloadItem.<anonymous> (file:///Applications/InVEST%203.17.3.dev332+g60927e24c%20Workbench.app/Contents/Resources/app.asar/node_modules/electron-context-menu/node_modules/electron-dl/index.js:131:14)
    at DownloadItem.emit (node:events:508:28)
```

## Environment
- Operating System: macOS Sequoia 15.7.4
- InVEST version number: 3.17.3.dev332+g60927e24c (built from `release/3.18.0` branch)

davemfish:
Fixes #2404

This basically implements the menu item the same way as the built-in item would, but with extra error handling.
https://github.com/sindresorhus/electron-context-menu/blob/b5d4773fe95e9fb270e9a32713dbb17b1b1bcd58/index.js#L136

## Checklist

### DIFF
--- a/workbench/__mocks__/electron-dl.js
+++ b/workbench/__mocks__/electron-dl.js
@@ -1,0 +1,1 @@
+export default jest.fn(() => {});


workbench/src/main/setupOpenLocalHtml.js
```diff
@@ -26,7 +30,29 @@ export default function setupOpenLocalHtml(parentWindow, isDevMode) {
 
       contextMenu({
         window: win,
-        showSaveImageAs: true,
+        append: (defaultActions, parameters, browserWindow) => [
```
emilyanndavis, line 33: I tried this out on my machine, and it definitely solves the problem! I did notice there are two slight differences in the way the context menu appears, though:
1. "Save Image As..." now appears after "Copy Image". Not a big deal, but the default ordering—"Save Image As...", then "Copy Image"—seems to mimic the order in browser context menus, so I think it's worth using`prepend` instead of `append`, just in case people have that browser-default ordering in their muscle memory.
2. Whether we use `prepend` or `append`, a horizontal divider appears, setting that item off from other items in the menu. I don't think the API provides a way to remove those dividers, and I don't feel a need to, but I wanted to mention it just for the sake of being thorough. (If we wanted to get rid of the divider, I suspect we could work around it by hiding the default "Copy Image" and adding a custom "Copy Image" item just after the custom "Save Image As..." item, but I'm not sure it's worth the trouble.)

```suggestion
        prepend: (defaultActions, parameters, browserWindow) => [
```

davemfish: @emilyanndavis I noticed these things as well. With `prepend` the order was "Save Image As", "Select All", "Copy Image", and I didn't like that the two image items were separated so I switched to append. (The "Inspect Element" should only be there in dev mode).

emilyanndavis: Oh, interesting. I agree, it would be strange to separate the image-related items, but it didn't occur to me in this scenario because I don't see the "Select All" item. It looks like the default visibility differs between macOS and Windows/Linux (presumably because that difference exists in browser context menus as well?). I guess append makes more sense in that case (unless we want to redefine all three items as custom menu items, but I don't think we do).

davemfish: I could certainly imagine us wanting to take more control over this menu going forward, for some of these reasons, but for now I think it's okay as-is.

PR #2431, opened 2026-02-26
Restore autopr assignees

### LINKED ISSUE #2407: AutoPRs should have assignees but don't
When an AutoPR is created, it should be assigned to the person who submitted the PR that was just merged.  Right now, the PR is created, but isn't assigned to anyone.

phargogh:
This PR restores assignees by using the github API (via a python script) to fetch the username(s) of whoever created the PR(s) that went into a specific commit.  Github's interface only supports PRs authored by 1 person (as far as I know), but this approach would also support octopus merges from multiple sources, should those ever be supported.

I also took the opportunity to add a little more helpful information to the PR body itself and some additional debug logging.

Fixes #2407

## Checklist

### DIFF
--- a/.github/auto-main-into-release-pr-body.md
+++ b/.github/auto-main-into-release-pr-body.md
@@ -39,4 +39,4 @@ All changes pushed in this way will appear in this PR.
 
 <hr>
 
-<sub> This PR was created because there was a push to `$SOURCE_BRANCH` that triggered this workflow, which is defined at `.github/workflows/autopr-create.yml`. This workflow was triggered by `$PR_USERNAME` with the event `$GH_EVENT`. </sub>
+<sub> This PR was created because there was a push to `$SOURCE_BRANCH` that triggered this workflow, which is defined at `.github/workflows/autopr-create.yml`. This workflow was triggered by `$PR_USERNAMES` with the event `$GH_EVENT`. </sub>


.github/workflows/autopr-create.yml
```diff
@@ -34,16 +34,14 @@ jobs:
...
                   else
                       # event_name is presumed to have been a push.
-                      # A push could include a bunch of commits, so just get
-                      # the login name of the user that pushed
-                      PR_USERNAME="${{ github.event.push.pusher.username }}"
                       export PR_MSG='due to a push to `main`.'
                   fi
+                  export PR_USERNAMES=$(python scripts/who-authored-the-pr.py ${{ github.sha }})
+                  export COMMIT_URL="https://github.com/${{ github.repository }}/commit/${{ github.sha }}"
+                  export GH_EVENT="${{ github.event }}"
```
claire-simpson: `github.event` would be a whole payload object right? Do we want this, or maybe `github.event_name` since it looks like we're just using this in the PR body text?

.github/auto-main-into-release-pr-body.md
```diff
@@ -39,8 +39,4 @@ All changes pushed in this way will appear in this PR.
 
 <hr>
 
-<sub>
-
-This PR was created because there was a push to `$SOURCE_BRANCH` that triggered this workflow, which is defined at `.github/workflows/autopr-create.yml`
-
-</sub>
+<sub> This PR was created because there was a push to `$SOURCE_BRANCH` that triggered this workflow, which is defined at `.github/workflows/autopr-create.yml`. This workflow was triggered by `$PR_USERNAME` with the event `$GH_EVENT`. </sub>
```
claire-simpson: Should this be `$PR_USERNAMES` plural?

.github/workflows/autopr-create.yml
```diff
@@ -91,15 +89,15 @@ jobs:
...
                         # the person who merged the PR that caused this commit to be
                         # created. Others could of course be assigned later.
                         FAILED=0
                         gh pr create \
                             --head ${AUTOPR_BRANCH} \
                             --base ${RELEASE_BRANCH} \
-                            --reviewer "${PR_USERNAME}" \
-                            --assignee "${PR_USERNAME}" \
+                            --reviewer "${PR_USERNAMES}" \
+                            --assignee "${PR_USERNAMES}" \
```
claire-simpson, line 101: I don't _think_ it would ever be the case that `who-authored-the-pr` would return `''`.. but it does seem like an empty string reviewer or assignee would break this, so just wanted to flag this in case you think we should have a fallback

claire-simpson (changes requested):
Thanks @phargogh!

.github/workflows/autopr-create.yml
phargogh: Yep, definitely!  Patched in 95e7a57c2

.github/auto-main-into-release-pr-body.md
phargogh: Yep, sure should.  Patched in 28d615598

.github/workflows/autopr-create.yml
phargogh: You know, I think the action will still run (somewhat counterintuitively) because that's what it's been doing so far and why we don't have any assignees!  I think that behavior is OK, although definitely not ideal, so I _think_ we might be OK to just use this as-is and at least not have a regression of behavior.

claire-simpson: Works for me!

claire-simpson (approved):
Looks good!

PR #43, opened 2020-02-27
Workflow for building python distributions

phargogh:
This PR adds a workflow file for building wheels for `natcap.invest` as well as a source distribution as part of our regular github actions-based workflows.

PR #2439, opened 2026-03-05
SWY Report

### LINKED ISSUE #2321: Report for Seasonal Water Yield
Develop report for Seasonal Water Yield. There is a primary output which is a watershed polygon with zonal statistics results in the table as well as primary and secondary raster outputs. Most likely, this model can follow a similar layout to SDR and NDR. Some consideration should be given to which intermediate outputs are most useful for interpreting results.

Report author guidance: https://github.com/natcap/invest/wiki/InVEST-Reports:-Author's-Guide

Model Supporters to review a draft of the reports:
SWY: Stacie, Adrian, Jesse

megannissel:
## Description
Fixes #2321 

This PR adds a reporter module and associated template for the Seasonal Water Yield model. In addition to including several elements seen in other reports (plotted output and input rasters, tables of raster statistics, and plots of values in the model's aggregate vector output), this report introduces a new element: an interactive map of the AOI linked to a chart of monthly average quickflow, baseflow, and precipitation values for each feature. These charts were requested by the model maintainers during an initial meeting about the contents of the report. 

Since these charts are created from data that is derived from model outputs but not pulled from the directly, a new data source was required. This has taken the form of a CSV, which contains these monthly average quickflow, baseflow, and precipitation values by month, by feature in the AOI. After discussion with Dave, we decided that it would make the most sense for the model's `execute` function to create this CSV output, rather than creating it within the reporter. Since we think it would be beneficial to provide this data source to users to inspect, it makes sense to be able to describe the CSV within the model spec. 

A few additional noteworthy changes introduced by this PR:
- A few of the utilities that previously lived in the CV reporter have been moved to a new shared `vector_utils`.
- Many outputs in the SWY model spec have had a `created_if` condition added, to reflect that they are only created when `user_defined_local_recharge=False`. 

A prototype of the report can be found [here](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/seasonal_water_yield_report_gura.html). 

Additionally, a version of the report when `user_defined_local_recharge` is `True` (and the model therefore skips generating all of the quickflow rasters entirely) can be found [here](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/seasonal_water_yield_report_gura_user_defined_local_recharge.html). Note that this version of the report is sparse; there simply aren't as many model inputs or outputs in this case.

**Related PRs:**
- User's Guide: [PR #224](https://github.com/natcap/invest.users-guide/pull/224)
- invest-test-data snapshots for new `RasterPlotFacetsTests`: [invest-test-data #43](https://bitbucket.org/natcap/invest-test-data/pull-requests/43)

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -61,10 +61,28 @@
   7. InVEST model Z (model names should be sorted A-Z)
 
 
+Unreleased Changes
+------------------
 
-..
-  Unreleased Changes
-  ------------------
+Seasonal Water Yield
+====================
+* The model now generates a report, a visual summary of results, available in
+  the output workspace and also viewable from the Workbench after the model run
+  completes. (`#2321 <https://github.com/natcap/invest/issues/2321>`_)
+* The model now generates an additional output, a CSV containing average monthly
+  quickflow, baseflow, and precipitation values, in cubic meters per month, for
+  each feature in the AOI. This output is used by the report to generate some
+  plots. Note that this CSV is only created when the model is run without
+  inputting a Local Recharge raster.
+  (`#2321 <https://github.com/natcap/invest/issues/2321>`_)
+* Various updates to model output data metadata, including correcting the
+  units of some outputs.
+  (`#2450 <https://github.com/natcap/invest/issues/2450>`_)
+* Updated the naming convention of several monthly intermediate outputs to be
+  1-indexed rather than 0-indexed. This makes filenames consistent throughout
+  the model, where 1=January and 12=December.
+  (`#2451 <https://github.com/natcap/invest/issues/2451>`_)
+
 
 3.18.0 (2026-02-25)
 -------------------


src/natcap/invest/reports/vector_utils.py
```diff
@@ -0,0 +1,57 @@
...
+                         [xmax, ymin],
+                         [xmin, ymin],
+                         [xmin, ymax],
+                         [xmax, ymax]]]},
+        "properties": {}
+    }
+    return extent_feature, xy_ratio
+
+
+def chart_landmass(geodataframe, clip=False, extent_feature=None):
```
davemfish: I'm not sure if you ended up using this function in SWY. I agree it could make for a general-purpose utility function, but maybe we should leave it in CV until we have a need for it elsewhere? At which point we should probably also rename it and write a docstring.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,434 @@
...
+        feature.SetField('geom_fid', feature_id)
+        aoi_layer.SetFeature(feature)
+
+    aoi_layer.SyncToDisk()
+    aoi_layer = None
+    gdal.Dataset.__swig_destroy__(aoi_vector)
+    aoi_vector = None
+
+
+def create_monthly_stats_table(aoi_path, file_registry, output_table_path):
```
davemfish: As we discussed, I think it makes sense to add this computation to `execute` and probably have a new CSV output from the main model. Also, regarding the issue of feature IDs, at this point `zonal_statistics` results will always be indexed by FID (https://github.com/natcap/pygeoprocessing/issues/371), but I think it makes sense to add a dedicated polygon ID to the output vector and also include it in that new CSV.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,434 @@
...
+    b_avg_per_feat_per_month = {k: v['sum'] * 0.001 * pixel_area_m2 / 12
+                                for k, v in zonal_stats_b.items()}
+
+    values_dict = {fid: {month + 1: {'baseflow': b_val / seconds_per_month[month+1]}
+                         for month in range(12)}
+                   for fid, b_val in b_avg_per_feat_per_month.items()}
+
+    for raster_path, month_index, value_name in (
+            monthly_qf_path_list_tuples + monthly_precip_path_list_tuples):
+        zonal_stats = pygeoprocessing.zonal_statistics((raster_path, 1), aoi_path)
```
davemfish: It's not so obvious, but it's now possible to pass a list of aligned rasters to a single zonal_stats call. That saves a ton of work by only needing to rasterize the vector once.

src/natcap/invest/reports/vector_utils.py
megannissel: I didn't end up using this one, so would be totally okay with moving it back into CV for now!

src/natcap/invest/seasonal_water_yield/reporter.py
megannissel: Yeah, the most important thing about adding the FID to the output vector directly is that it guarantees an aligned, matching key between the geopandas geodataframe + CSV that can be used for the map + chart interaction. I'm confident about `zonal_statistics` being indexed by FID, but it's less clear to me if the FID becomes the index of a geodataframe when a vector is read in. And we need to have a matching key for the interaction, regardless, so adding the FID to the vector directly just seems like a more surefire solution.

But yes, 100% agree that we ought to move this computation into `execute`!

src/natcap/invest/seasonal_water_yield/reporter.py
megannissel: OH! I missed that in the PGP docs, but that's SO much better!

src/natcap/invest/seasonal_water_yield/reporter.py
davemfish: >Yeah, the most important thing about adding the FID to the output vector directly is that it guarantees an aligned, matching key between the geopandas geodataframe + CSV that can be used for the map + chart interaction. 

@megannissel Yes, and since this CSV will be a model output, we should support users who want to do their own post-processing that might include a table-join between the CSV and the vector.

Let's also stay away from naming this new index "FID" or even "feature ID", to avoid conflict/confusion with the built-in FID.

megannissel: Agreed -- I ran into an instance during dev where calling it `fid` caused a failure, which is why I changed it to `geom_fid`. But I also don't really love that name and am open to other suggestions.

davemfish: I think `geom_id`, `poly_id`, `area_id` would all be fine.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -559,7 +573,6 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None):
         if numpy.isclose(vmin, 0.0):
             vmin = 1e-6
         normalizer = matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
-        cmap.set_under(cmap.colors[0])  # values below vmin (0s) get this color
```
megannissel, line 562: I'm fairly certain this wasn't doing anything; to the best I could determine, `cmap.get_under` was already returning the same value as `cmap.colors[0]`, so `cmap.set_under` had no effect. 

Notably, `cmap.colors` breaks for many matplotlib colormaps. `viridis` is a `matplotlib.colors.ListedColormap`, but colormaps like `GnBu` (and our custom `truncated_PuOr`) are of type `matplotlib.colors.LinearSegmentedColormap`, which have no attribute `colors`.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -157,6 +158,8 @@ class RasterPlotConfig:
     """For highly skewed data, a transformation can help reveal variation."""
     title: str | None = None
     """An optional plot title. If ``None``, the filename is used."""
+    custom_colormap: str | None = None
+    """The name of a custom matplotlib colormap."""
```
davemfish: Could we just call this attribute `colormap`? And could it be either the string name of a registered matplotlib colormap _or_ a colormap object?

src/natcap/invest/reports/raster_utils.py
```diff
@@ -256,22 +259,25 @@ def _extra_wide_aoi(xy_ratio):
...
+def _choose_n_rows_n_cols(xy_ratio, n_plots, small_plots):
     if _extra_wide_aoi(xy_ratio):
         n_cols = 1
     elif _wide_aoi(xy_ratio):
         n_cols = 2
     else:
         n_cols = 3
 
+    if small_plots:
+        n_cols += 1
```
davemfish, line 275: This is kind of neat if that's all it takes to make smaller plots! We'll probably want some snapshot tests that try different layouts with this option.

src/natcap/invest/reports/raster_utils.py
davemfish: Okay, thanks for catching this. Definitely nice to get rid of it if it was bugged.

I guess I assumed that a value outside the range of `vmin : vmax` would not get a color at all unless we explicitly `set_under`, but I don't remember if I ever tested that or not.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -568,13 +581,23 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None):
...
+                (ylim_kwargs,
+                 text_kwargs) = _get_units_text_kwargs(units, len(arr))
+                ax.set_ylim(**ylim_kwargs)
+                ax.text(**text_kwargs)
+            fig.colorbar(mappable, ax=ax)
+
+    if small_plots:
+        units = _get_raster_units(tif_list[0])
+        legend_label = f"{UNITS_TEXT}: {units}" if units else None
+        fig.colorbar(mappable, ax=axes.ravel().tolist(), label=legend_label)
```
davemfish: I think this approach of a single legend with units label is probably what we always want for this function. It's more critical to do it this way for small plots, but I think it's an improvement for the other case as well. This function requires that all rasters share a colorscale, so it's reasonable to only print one colorbar. It even makes it more obvious that the colorscales are identical across plots.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -568,13 +581,23 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None):
...
+
+    if supertitle:
+        fig.suptitle(f"{supertitle}", fontsize=TITLE_FONT_SIZE)
```
davemfish: This f-string wrapper looks unnecessary.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -636,30 +655,64 @@ def _get_raster_units(filepath):
...
             filename = os.path.basename(resource.path)
             raster_summary[filename] = _build_stats_table_row(
                 resource, band)
 
+    for item in file_registry.values():
+        if isinstance(item, collections.abc.Mapping):
+            for path in item.values():
+                _summarize_output(path)
+        else:
+            _summarize_output(item)
```
davemfish: It looks like this implementation is no longer recursive. It looks like it will work for one level of nesting, but not more. Is that the intent? It may be all we require, though I'm not sure if anything strictly prevents the file registry from having deeper nesting.

I guess the original implementation did not work properly (#2444 ) though it's not obvious to me where the bug is. We do a very similar walk through the file registry in `natcap.invest.spec.ModelSpec.generate_metadata_for_outputs` and I _think_ that one is working properly.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -636,30 +655,64 @@ def _get_raster_units(filepath):
...
 
     return pandas.DataFrame(raster_summary).T
+
+
+def _parse_csv_paths_from_spec(args_dict, spec):
+    table_map_inputs = []
+    for input_ in spec.inputs:
+        if isinstance(input_, CSVInput):
+            table_map_inputs.extend([
+                (input_.id, col.id) for col in spec.get_input(input_.id).columns
```
davemfish: I think `spec.get_input(input_.id)` on this line could be replaced with `input_`.

src/natcap/invest/reports/templates/models/seasonal_water_yield.html
```diff
@@ -0,0 +1,132 @@
...
+
+  <h2 class="section-header">Results</h2>
+  {% if qf_rasters %}
+    {{ accordion_section(
+      'Annual and Monthly Quickflow (QF)',
+      content_grid([
+        (caption(raster_group_caption, pre_caption=True), 100),
+        (raster_plot_img(qf_rasters['annual_qf_img_src'], 'Annual Quickflow values'), 100),
+        (raster_plot_img(qf_rasters['monthly_qf_img_src'], 'Monthly Quickflow values'), 100),
+        (caption(qf_rasters['qf_caption'], definition_list=True), 100)
```
davemfish, line 20: I think it would be nice to include the Annual Quickflow filename somewhere. Either in the plot title, or in the caption.

For the monthly maps, I almost feel like the caption values should be switched, so its like `Maps of monthly quickflow: qf_[MONTH]_gura.tif (1 = January… 12 = December)`

src/natcap/invest/reports/templates/models/seasonal_water_yield.html
```diff
@@ -0,0 +1,132 @@
...
+      ])
+    )}}
+  {% endif %}
+
+  {% if qf_b_charts %}
+    {{ accordion_section(
+      'Average Quickflow and Baseflow by Month',
+      content_grid([
```
davemfish, line 28: Is this extra `content_grid` wrapper necessary here?

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+def _create_aggregate_map(geodataframe, extent_feature, xy_ratio, attribute,
+                          title):
+    attr_map = altair.Chart(geodataframe).mark_geoshape(
+        clip=True,
+        stroke="white",
+        strokeWidth=0.5
+    ).project(
+        type='identity',
+        reflectY=True,
+        fit=extent_feature
```
davemfish: I think we would only need to use the `clip` and `fit` parameters if we wanted this function to clip the `geodataframe` by the extent of some other feature. If we always want to plot the entirety of `geodataframe`, then I don't think this function needs to take an `extent_feature`.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+    extent_feature, xy_ratio = vector_utils.get_geojson_bbox(map_df)
+
+    feat_select = altair.selection_point(fields=["geom_id"], name="feat_select", value=0)
+
+    attr_map = altair.Chart(map_df).mark_geoshape(
+        clip=True, stroke="white", strokeWidth=0.5
+        ).project(
+            type='identity',
+            reflectY=True,
+            fit=extent_feature
```
davemfish: Same comment here about probably not needing to use `clip` or `fit`.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+    ).encode(
+        color=altair.condition(
+            feat_select,
+            altair.value("seagreen"),
+            altair.value("lightgray")
+        ),
+        tooltip=[altair.Tooltip("geom_id", title="FID")]
```
davemfish: I would avoid using "FID" as the label now that this is not strictly the built-in "FID". Maybe just title it "Feature"? And then the bar chart title can be simplified to "Feature 3"

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+    ).properties(
+        width=MAP_WIDTH,
+        height=MAP_WIDTH / xy_ratio,
```
davemfish: It might be nice to give the map a bit more space in this figure. The bar chart really only ever needs so much space, but given a different AOI with many smaller polygons, it could help to allow the map to be larger. And this accordion panel has some empty space to work with.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+        color=altair.value('#0500a3')
+    )
+
+    combined_chart = altair.layer(bar_chart, precip_chart).resolve_scale(
+        y='independent'
+    ).transform_filter(
+        feat_select
+    ).properties(
+        title=altair.Title(altair.expr(
+            f'"Mean Quickflow + Baseflow for Feature, FID " + {feat_select.name}.geom_id')
```
davemfish: I think we can trim ", FID " from the title.

I think it's also worth trying to bump up the font sizes for basically all the text on this chart. You could try using `.configure_axis(**AXIS_CONFIG)` and `.configure_legend(**LEGEND_CONFIG)` if those seem useful.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -0,0 +1,371 @@
...
+                raster_path=file_registry['aet'],
+                datatype=RasterDatatype.continuous,
+                spec=model_spec.get_output('aet')),
+            RasterPlotConfig(
+                raster_path=file_registry['cn'],
+                datatype=RasterDatatype.continuous,
+                spec=model_spec.get_output('cn')),
+            RasterPlotConfig(
+                raster_path=file_registry['l'],
+                datatype=RasterDatatype.continuous,
```
davemfish: It looks like local recharge values are positive and negative where positive indicates more water infiltrating on the pixel than consumed by the veg on that pixel, and negative means more water consumed than infiltrating. Basically, I wonder if makes sense to try a divergent color scale that is centered on 0.

davemfish: I'm wondering the same for `qb` and `vri_sum`, which appear in the vector. Though in the sample data they are all positive values.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -414,24 +447,81 @@
...
+            about=gettext(
+                "Table of average monthly baseflow, quickflow, and precipitation"
+                " values for each watershed (or feature) within the AOI."
+            ),
+            columns=[
+                spec.IntegerOutput(
+                    id="geom_id",
+                    about=gettext(
+                        "The Feature ID for the watershed. This will correspond to"
+                        " the FIDs in the Aggregate Results shapefile."
```
davemfish: Maybe this should say, "will correspond to the 'geom_id' column in the Aggregate Results shapefile".

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
```diff
@@ -1027,6 +1112,21 @@ def execute(args):
...
         task_name='calculate B_sum')
 
+    if not args['user_defined_local_recharge']:
+        monthly_csv_task = task_graph.add_task(
+            func=_generate_monthly_qf_b_p_csv,
+            args=(
+                file_registry['aggregate_vector'],
+                file_registry['b'],
+                [file_registry['qf_[MONTH]', month] for month in range(1, 13)],
+                [file_registry['prcp_a[MONTH]', month] for month in range(12)],
```
davemfish: Are the monthly qf and precip outputs really not indexed at the same number? 0 vs 1? That feels like something we should improve on. We can deal with it separately though if we want to.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
davemfish: Whoops, this comment should go!

tests/test_seasonal_water_yield_regression.py
```diff
@@ -746,6 +780,23 @@ def test_base_regression_nodata_inf(self):
...
+        expected_b = [
+            2.242553e-5, 2.461508e-5, 2.242553e-5, 2.317305e-5, 2.242553e-5, 2.317305e-5,
+            2.242553e-5, 2.242553e-5, 2.317305e-5, 2.242553e-5, 2.317305e-5, 2.242553e-5]
+        expected_p = [
+            4.065860e-5, 4.868549e-5, 4.805108e-5, 5.347222e-5, 5.544355e-5, 6.111111e-5,
+            6.283602e-5, 6.653226e-5, 7.256944e-5, 7.392473e-5, 8.020833e-5, 8.131720e-5]
+        for expected_val, col_name in [(expected_qf, 'quickflow'),
+                (expected_b, 'baseflow'), (expected_p, 'precipitation')]:
+            numpy.testing.assert_allclose(expected_val, actual_result_df[col_name],
+                                          rtol=1e-6)
```
davemfish: These checks on the aggregated monthly values are nice for general regression tests. I wonder if it's enough to do these assertions on the first two `base_regression` tests. I'm not really sure what issue this specific test was covering when it was written, but it feels like the zonal stats results would not fail here unless they also failed in one of the first two tests.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -799,6 +850,23 @@ def test_monthly_alpha_regression(self):
...
+        expected_b = [
+            2.276286e-5, 2.498535e-5, 2.276286e-5, 2.352162e-5, 2.276286e-5, 2.352162e-5,
+            2.276286e-5, 2.276286e-5, 2.352162e-5, 2.276286e-5, 2.352162e-5, 2.276286e-5]
+        expected_p = [
+            4.106930e-5, 4.917726e-5, 4.853644e-5, 5.401235e-5, 5.600358e-5, 6.172840e-5,
+            6.347073e-5, 6.720430e-5, 7.330247e-5, 7.467145e-5, 8.101852e-5, 8.213859e-5]
+        for expected_val, col_name in [(expected_qf, 'quickflow'),
+                (expected_b, 'baseflow'), (expected_p, 'precipitation')]:
+            numpy.testing.assert_allclose(expected_val, actual_result_df[col_name],
+                                          rtol=1e-6)
```
davemfish: Similar comment here, is there any reason to think the alpha table impacts the behavior of the new monthly aggregation function? If not, I would think we could remove this assertion from this test to reduce our test maintenance burden.

src/natcap/invest/reports/raster_utils.py
megannissel: Oh, yep, good catch; this was a holdover from my first iteration of the title that also included units.

src/natcap/invest/reports/raster_utils.py
megannissel: The problem was line 620, in `_get_raster_metadata`: 
```
def _get_raster_metadata(filepath):
    if isinstance(filepath, collections.abc.Mapping):
        for path in filepath.values():
            return _get_raster_metadata(path)
```

The `return` stopped iteration beyond the first value in the dict, which is why we were seeing `qf_1.tif` in the summary stats table, but none of the other months. The implementation in `natcap.invest.spec.ModelSpec.generate_metadata_for_outputs` calls `_generate_metadata` (which in that case is a nested function within `generate_metadata_for_outputs`) recursively, but there's no `return` in there stopping iteration prematurely. I wonder if this function in `raster_utils` originally took a similar form (with `_get_raster_metadata` nested inside of `raster_workspace_summary`), but was later refactored so that it could also be called by `_get_raster_units`. 

You make a good point about how this updated version only works for one layer of nesting, though. I don't think we currently have any instances where we'd have deeper nesting with the file registry, but that doesn't mean we couldn't in the future. I can tweak it to actually work recursively.

src/natcap/invest/reports/templates/models/seasonal_water_yield.html
megannissel: Sure isn't!

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
megannissel: Indeed they are not! I find this incredibly silly as well and agree that consistency would be a nice improvement to the model.

src/natcap/invest/seasonal_water_yield/seasonal_water_yield.py
megannissel: Oops, good catch!

tests/test_seasonal_water_yield_regression.py
megannissel: That seems reasonable to me!

src/natcap/invest/reports/raster_utils.py
```diff
@@ -390,7 +397,7 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
         colorbar_kwargs = {}
         imshow_kwargs['norm'] = transform
         imshow_kwargs['interpolation'] = 'none'
-        cmap = COLORMAPS[dtype]
+        cmap = plt.get_cmap(config.colormap if config.colormap else COLORMAPS[dtype])
```
megannissel: This change is a defensive precaution. I noticed line 408 accesses `cmap.N`, which is fine at the moment because it's inside a conditional where (currently) `cmap` will always be a colormap object. But to help prevent bugs in the future (especially as we expand custom colormap support), it seems wise to always get the object. And `plt.get_cmap()` takes either a string or a Colormap object, so worst-case scenario it just returns the object that was passed to it.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -157,6 +159,8 @@ class RasterPlotConfig:
     """For highly skewed data, a transformation can help reveal variation."""
     title: str | None = None
     """An optional plot title. If ``None``, the filename is used."""
+    colormap: Optional[Union[str, object]] = None
```
megannissel: @claire-simpson, related to #2447 -- `RasterPlotConfig` now accepts either the string name of a matplotlib colormap or a Colormap object. I've also renamed this attribute `colormap`, per Dave's suggestion!

src/natcap/invest/reports/raster_utils.py
```diff
@@ -636,30 +647,64 @@ def _get_raster_units(filepath):
...
+            for path in item.values():
+                _summarize_output(path)
+        else:
+            _summarize_output(item)
+
     return pandas.DataFrame(raster_summary).T
 
 
-def raster_inputs_summary(args_dict):
+def raster_inputs_summary(args_dict, model_spec):
```
megannissel, line 677: These changes address the issue documented in #2440

src/natcap/invest/reports/raster_utils.py
megannissel: I added some snapshot tests in https://github.com/natcap/invest/pull/2439/commits/5e78639f586ba195ad833295dd4fe22398367b66 and opened a [PR in the test data repo](https://bitbucket.org/natcap/invest-test-data/pull-requests/43) with the new images.

src/natcap/invest/seasonal_water_yield/reporter.py
megannissel: @davemfish We could swap to a diverging colorscale for the `qb` and `vri_sum` plots as well; what do you think about this? (sample data edited to include some negatives)
Screenshot 2026-03-20 at 12 05 09 PM

And this is what it would look like when all values are positive:
Screenshot 2026-03-20 at 12 05 23 PM

(I'm not sure I love the `redblue` color scheme, but the built-in options are limited and I did like the idea of choosing one that could reflect bluer=wetter)

src/natcap/invest/reports/raster_utils.py
megannissel: Tweaked this in https://github.com/natcap/invest/pull/2439/commits/68f4641a51bbd657a9967c8b4c13e8a5e8821840 to actually work recursively

src/natcap/invest/seasonal_water_yield/reporter.py
megannissel: Two other color scheme options:

- built-in `brownbluegreen` scheme (top map includes negatives; bottom does not)
Screenshot 2026-03-20 at 12 56 42 PM

- custom scheme which uses the min/max colors from the built-in `blueorange` scheme but swaps them so bluer=wetter
Screenshot 2026-03-20 at 12 57 16 PM

src/natcap/invest/reports/raster_utils.py
davemfish: Instead of `object` can we use `matplotlib.colors.Colormap`? https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap

Also, we don't necessarily need to support Python versions before 3.10, so instead of `Optional` and `Union` I think we can say `str | Colormap | None = None`.

src/natcap/invest/reports/raster_utils.py
davemfish: Cool, looks like it works well. It's ideal when the image ends up close to 1200 px wide, since that's what we specify as a target to fit nicely in our templates. So that first image where there are 2 columns in the small plot layout, it could stand to be larger overall. But I haven't thought about the best approach. We can always make an issue an improve on it later.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -312,16 +313,19 @@ def _get_title_line_width(n_plots: int, xy_ratio: float) -> int:
         return 31  # 3-column layout
 
 
-def _get_title_kwargs(title: str, resampled: bool, line_width: int):
+def _get_title_kwargs(title: str, resampled: bool, line_width: int, facets=False):
     label = f"{title}{' (resampled)' if resampled else ''}"
     label = textwrap.fill(label, width=line_width)
+    padding = 1.5
+    if not facets:
+        padding *= SUBTITLE_FONT_SIZE
```
davemfish, line 326: Can we add a comment in here like, "facetted plots don't need extra padding for title because their units label appears with the legend instead of under the title" if that is accurate.

src/natcap/invest/reports/raster_utils.py
davemfish: Cool, good idea. Another option might be to convert strings to a colormap object in `RasterPlotConfig`, either with a `BeforeValidator` or `__post_init__`, I forget exactly what the approach would be.

src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -124,11 +120,23 @@ def _create_linked_monthly_plots(aoi_vector_path, aggregate_csv_path):
         feat_select
     ).properties(
         title=altair.Title(altair.expr(
-            f'"Mean Quickflow + Baseflow for Feature, FID " + {feat_select.name}.geom_id')
+            f'"Mean Quickflow + Baseflow for Feature " + {feat_select.name}.geom_id')
         )
     )
 
-    chart = attr_map | combined_chart
+    legend_config = vector_utils.LEGEND_CONFIG
```
davemfish: We should consider making a copy of the global variable before modifying it just in case `LEGEND_CONFIG` is used elsewhere.

src/natcap/invest/reports/raster_utils.py
davemfish: Thanks! Recursion is hard.

src/natcap/invest/seasonal_water_yield/reporter.py
davemfish: Thanks for sharing these samples! I like all these fairly well. I'd probably stay away from the Blue-Red because it's so evocative of temperature. Maybe best to get buy-in from the rest of the SWY team, but just choosing one that you like is a great place to start. I also wonder if it makes sense to use the same colors for the local recharge raster as for these maps.

tests/test_seasonal_water_yield_regression.py
```diff
@@ -943,6 +972,49 @@ def test_user_recharge(self):
             os.path.join(args['workspace_dir'], 'aggregated_results_swy.shp'),
             agg_results_csv_path)
 
+    def test_climate_zones(self):
```
megannissel: This is only tangentially related to this PR, but when I updated the SWY model to consistently index monthly rasters, I accidentally introduced a bug that I only caught because one of the sets of sample data I've been using for report demo purposes includes a monthly climate zones raster and table as inputs. It seemed like a good reason to add a test covering this use-case.

src/natcap/invest/seasonal_water_yield/reporter.py
davemfish: I guess I prefer the middle one because despite the names, the `bluegreen` scale seems to use a single hue above 0 whereas the `blue` in `blueorange` looks like it uses multiple hues.

megannissel: Yeah, I couldn't seem to get the custom scale to go between blue and orange without passing through either green or purple (or in one case a truly surprising hot pink) somewhere along the way. I think I'll go with the middle option for now.

src/natcap/invest/reports/raster_utils.py
megannissel: Oh, I like this idea a lot! Looks like we can add this to the existing `RasterPlotConfig.__post_init__`

src/natcap/invest/reports/raster_utils.py
```diff
@@ -158,14 +159,17 @@ class RasterPlotConfig:
...
+    colormap: str | Colormap | None = None
+    """The string name of a registered matplotlib colormap or a colormap object."""
 
     def __post_init__(self):
         if self.title is None:
             self.title = os.path.basename(self.raster_path)
         self.caption = f'{self.title}:{self.spec.about}'
 
+        self.colormap = plt.get_cmap(self.colormap if self.colormap
+                                     else COLORMAPS[self.datatype])
```
davemfish, line 171: Nice

PR #2448, opened 2026-03-20
Refactor `deploy` target to be artifact-specific and fail on missing outputs

### LINKED ISSUE #831: dist/data not uploaded to google cloud
The "Deploy artifacts to GCS" workflow step has been failing to upload the `dist/data` directory. This goes back quite a while, for example [this run on 10/26](https://github.com/natcap/invest/runs/4012688043?check_suite_focus=true#step:11:36).
```
gsutil -m rsync -r dist/data gs://releases.naturalcapitalproject.org/invest/3.9.1.post69+g88536587/data

WARNING: gsutil rsync uses hashes when modification time is not available at
both the source and destination. Your crcmod installation isn't using the
module's C extension, so checksumming will run very slowly. If this is your
first rsync since updating gsutil, this rsync can take significantly longer than
usual. For help installing the extension, please see "gsutil help crcmod".

CommandException: arg (dist/data) does not name a directory, bucket, or bucket subdir.
If there is an object with the same path, please add a trailing
slash to specify the directory.
make: [Makefile:400: deploy] Error 1 (ignored)
```

claire-simpson:
Refactors the `make deploy` target into artifact-specific deploy targets and removes ignored errors from `gsutil` commands (i.e., removed leading `-` so `deploy_[target]` commands now fail on error). `deploy` is retained with the same behavior in case it is used elsewhere.

Previously, `make deploy` attempted to upload all possible artifacts (`dist/`, `dist/data`, `dist/userguide`, `workbench/dist`) regardless of whether they were created in the current job. Missing directories caused `gsutil` rsync to raise errors, which were suppressed, but resulted in noisy logs and could potentially hide real deployment failures.

- Each GHA job now deploys only the artifacts it actually builds
- This eliminates noisy `gsutil` errors for missing directories (which didn't cause issues and could be ignored)
- Prevents real deployment failures from being silently ignored
- Makes deploy behavior explicit

Fixes: #831 

## Checklist


dcdenu4 (approved):
Thanks @claire-simpson, I like that this much more purposeful approach.

PR #2449, opened 2026-03-20
Urban Mental Health report

### LINKED ISSUE #2419: Report for Urban Mental Health
## Description
Create a report for urban mental health model.

1. Primary results:
- Aggregated and per-polygon (e.g., admin unit) preventable cases cost sum table
- Preventable cases raster
- Preventable cost raster
- Preventable cases sum by polygon/admin unit
- Preventable cost sum by polygon/admin unit

2. Useful intermediate results:
- Baseline cases raster
- Baseline prevalence raster
- Raster of change in NDVI between baseline and alternate scenario

[Optional?; only relevant if using LULC inputs]
- Table with average NDVI by lucode (if not input by user)
- Baseline NDVI (i.e., reclassified baseline LULC)
- Alternate NDVI (i.e., reclassified alternate LULC)

3. Output raster stats table
4. Input args table
5. Input raster stats
6. Metadata: output filenames and descriptions

Model supporter: Yingjie

claire-simpson:
## Description
This PR creates the report for Urban Mental Health and makes other relatively minor updates (mostly related to wording as captions and descriptions appear in reports).

Examples of reports:
LULC:
https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_mental_health_report_lulc_040226.html

NDVI:
https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_mental_health_report_ndvi_040226.html 

Fixes #2419 

## Checklist

### DIFF
--- a/src/natcap/invest/reports/templates/models/urban_mental_health.html
+++ b/src/natcap/invest/reports/templates/models/urban_mental_health.html
@@ -52,7 +52,7 @@
   {% endif %}
   
   {{ accordion_section(
-    'Preventable Cases by AOI Vector',
+    'Preventable Cases by AOI Feature',
     content_grid(cases_cost_items)
   ) }}
 


src/natcap/invest/reports/raster_utils.py
```diff
@@ -462,8 +466,9 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
                 len(patches), n_plots, xy_ratio))
             leg.set_in_layout(True)
         else:
-            mappable = ax.imshow(arr, cmap=cmap, **imshow_kwargs)
-            fig.colorbar(mappable, ax=ax, **colorbar_kwargs)
+            config.plot_on_axis(
```
claire-simpson: Wanted an extension hook without changing `plot_raster_list` or duplicating too much code, but curious what you think of this. It also doesn't touch the `nominal` plots, so maybe there is a better design here? Curious what you think.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+LOGGER = logging.getLogger(__name__)
+
+TEMPLATE = jinja_env.get_template('models/urban_mental_health.html')
+
+MAP_WIDTH = 450  # pixels
+
+NEAR_ZERO_RANGE = (-0.01, 0.01)
+
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True))
+class SpecialRangeRasterPlotConfig(RasterPlotConfig):
```
claire-simpson: I stuck this class here to avoid cluttering `raster_utils`, as I think this is a somewhat niche use case that doesn't necessarily need to be available to other reporter modules. Perhaps there is a better location for it though?

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,149 @@
...
+  {{ super() }}
+
+  {% from 'args-table.html' import args_table %}
+  {% from 'caption.html' import caption %}
+  {% from 'content-grid.html' import content_grid %}
+  {% from 'metadata.html' import list_metadata %}
+  {% from 'raster-plot-img.html' import raster_plot_img %}
+  {% from 'wide-table.html' import wide_table %}
+
+
```
dcdenu4:
```suggestion
```

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,149 @@
...
+      'Output Filenames and Descriptions',
+      list_metadata(model_spec_outputs),
+      expanded=False
+    )
+  }}
+
+{% endblock content %}
+
+
+
```
dcdenu4:
```suggestion
```

src/natcap/invest/reports/raster_utils.py
```diff
@@ -170,6 +170,10 @@ def __post_init__(self):
         self.colormap = plt.get_cmap(self.colormap if self.colormap
                                      else COLORMAPS[self.datatype])
 
+    def plot_on_axis(self, fig, ax, arr, cmap, imshow_kwargs, colorbar_kwargs):
```
dcdenu4: Docstring needed?

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -1419,6 +1438,21 @@ def zonal_stats_preventable_cases_cost(
...
+    target_aggregate_vector = gdal.OpenEx(
+        target_aggregate_vector_path, gdal.OF_UPDATE)
+
+    layer = target_aggregate_vector.GetLayer()
+    old_layer_name = layer.GetName()
+    layer = None
+
+    target_aggregate_vector.ExecuteSQL(
+        f'ALTER TABLE "{old_layer_name}" RENAME TO "{custom_layer_name}"'
+    )
```
dcdenu4: I see the convenience of using `CreateCopy` above. Another approach would be to create the vector from scratch and use [CopyLayer](https://gdal.org/en/stable/api/python/raster_api.html#osgeo.gdal.Dataset.CopyLayer), where you could define the new layer name.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+            - lulc_base
+            - lulc_alt
+            - if not lulc_attr_csv: show lulc_attr_table that was built
+            - if ndvi_base: show it as well
+
+        intermediates:
+            - LULC baseline reclassified to NDVI
+            - if lulc_alt: LULC alternate reclassified to NDVI
+
+
```
dcdenu4:
```suggestion
```

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+        - baseline prevalence
+
+    outputs:
+        - preventable_cases
+        - preventable_cost
+        - preventable cases cost sum table (just total cases and total cost)
+        - vector symbolized with field 'sum_cost'
+        - vector symbolized with field 'sum_cases'
+
+
```
dcdenu4, line 121:
```suggestion
```

davemfish:
@claire-simpson could you post a link (in the PR description) to a sample report created from this branch? Thank you!

src/natcap/invest/reports/raster_utils.py
davemfish: @claire-simpson I think this is a neat way of defining your own call to `imshow` with a custom array, while still using the layout and other stuff from `plot_raster_list`.

I do wonder if the particular use-case of `SpecialRangeRasterPlotConfig` is achievable using some existing `imshow` and `colormap` options, rather than doing the special array handling. For example, something like this may be possible:
```python
cmap = plt.get_cmap('viridis')
cmap.set_under('orange')
mappable = ax.imshow(array, cmap=cmap, vmin=0.01)
fig.colorbar(mappable, ax=ax, **colorbar_kwargs)
```

In which case, maybe we could add some optional attributes to `RasterPlotConfig`?
```python
vmin: float
vmax: float
set_under: str
set_over: str
```

I can imagine this functionality could be useful for other datasets too.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+    Returns:
+        str: 'continuous' or 'divergent'
+    """
+    raster_info = pygeoprocessing.get_raster_info(raster_path)
+    nodata = raster_info['nodata'][0]
+    ds = gdal.OpenEx(raster_path, gdal.OF_RASTER)
+    band = ds.GetRasterBand(1)
+
+    stats = band.GetStatistics(False, True)  # (approx_ok, force)
```
davemfish: If the dataset is an invest output then I think it's safe to assume that `geometamaker/invest` will have just calculated the band stats, so we perhaps we could get the stats from the metadata without forcing their re-computation?

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+    min_val = stats[0]
+    LOGGER.info("Stats for %s: min=%s, nodata=%s", raster_path, min_val, nodata)
+
+    if min_val is None:
```
davemfish: I'm curious when/if this would come up because it would be nice to rely on the stats and avoid reading the entire raster into memory if possible.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,500 @@
...
+            valid_values = arr[~numpy.isnan(arr)]
+
+        if valid_values.size == 0:
+            LOGGER.warning(f"No valid pixels found in {raster_path}, "
+                           "defaulting to continuous")
+            return RasterDatatype.continuous
+
+        min_val = round(numpy.nanmin(valid_values), 4)
+
+    return RasterDatatype.divergent if min_val < NEAR_ZERO_RANGE[0] else RasterDatatype.continuous
```
davemfish: It's interesting to think about how to infer this from the data. In this case it looks like a raster of all negative values would classify as divergent, which might not be the intent. Is it better to look for a `mean near zero`, `min < mean` and `max > mean` ?

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Yeah that makes sense! To provide some context and touch on your comment below: when I wrote this function initially I was imagining it being more widely used by other models reports (and possibly on input data without GMM metadata), and was envisioning it living somewhere more accessible, so I wanted to make it as flexible as possible. Now I think that until broader use actually crops up, I'd keep it here and so just relying on metadata seems fine to me.

Also is there an easier way to get a band's minimum value than the following? I didn't notice a getter
```python
resource = geometamaker.describe(raster_path)
min_val = float(resource.data_model.bands[0].gdal_metadata['STATISTICS_MINIMUM'])
```

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: > In this case it looks like a raster of all negative values would classify as divergent, which might not be the intent

Great point. I'm not sure that checking if the mean is near zero is exactly the rule I'd want to follow, thinking about the use case of an NDVI raster with mostly positive values but a few negative values. Even though the mean is above zero, I'd still prefer to see that NDVI raster visualized with a divergent color ramp (and same with the specific case in this report for preventable cases). Maybe checking that `min < 0` and `max > 0` would do the trick?

src/natcap/invest/reports/raster_utils.py
claire-simpson: I have been playing around with this, and I have mixed opinions:
- I definitely like the simplicity of this approach. 
- I personally don't dislike the way matplotlib displays the colorbar in the legend when `set_under` is used (and `extend='min'` is used in the colorbar) - see attached image. That being said, I'm not sure this colorbar is as clear to users about what it's doing whereas the patch approach is pretty explicit (see related concern in next bullet).
- I also have some slight hesitation about using this `set_under` approach for the preventable cases/costs rasters specifically as we'd be setting all values below 0.01 to orange, rather than just values in the range of -0.01 to 0.01. The intention is just to show pixels near 0 with the special color, and I'm concerned this approach doesn't convey that to users. Sticking with the patch legend and a custom label could possibly help mitigate this as we could keep the patch label as "-0.01 to 0.01" or "near zero" or something, but these labels aren't exactly true.

I'm ultimately fine going in either direction! If we used this proposed approach, is your thought to keep the `SpecialRangeRasterPlotConfig` or incorporate this in the existing RasterPlotConfig?

Screenshot 2026-04-07 at 3 01 37 PM

src/natcap/invest/urban_mental_health/reporter.py
davemfish: At the moment that's basically the way to get it using `geometamaker`. Or `resource.get_band_description(0).gdal_metadata['STATISTICS_MINIMUM']`

Feel free to keep using `band.GetStatistics()` from GDAL though. I mainly wanted to caution against using the `force` option there because I don't think we want to force re-calculating the stats when they already exist.

src/natcap/invest/urban_mental_health/reporter.py
davemfish: Yeah you're right, mean near 0 is probably not the ticket. Adding the `max > 0` seems like it will do the job for this use-case.

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+      (caption(outputs_caption, definition_list=True), 100)
+    ])
+  )}}
+
+   {{ accordion_section(
+    'Preventable Cases by AOI Vector',
+    content_grid(
+      [
+        (content_grid([
+          ('<div style="display:flex; justify-content:center;"><div id="cases_map"></div></div>', 100),
```
davemfish: Can you say how this extra `div` wrapper helps? These seem like sensible style attributes, if they are solving a general problem it might be nice to incorporate that solution into our stylesheet (`style.html`) so that other users of `vega-embed` and/or `content_grid` don't have to think about it too much. Or is the problem specific to this report?

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+  
+        (content_grid([
+          ('<div style="display:flex; justify-content:center;"><div id="cost_map"></div></div>', 100),
+          (caption(cost_map_caption, aggregate_map_source_list), 100)
+        ]), 50)
+      ] if cost_map_json else [
+        (content_grid([
+          ('<div style="display:flex; justify-content:center;"><div id="cases_map"></div></div>', 100),
+          (caption(cases_map_caption, aggregate_map_source_list), 100)
+        ]), 100)
```
davemfish: I don't know offhand, but is there a way to structure the conditional to avoid duplicating the `cases_map` definition? Perhaps using an assignment? https://jinja.palletsprojects.com/en/stable/templates/#assignments

Also, In the `else` condition it looks like there will be an extra `content_grid` wrapper around a single element. Perhaps the conditional could could be moved to an outer scope to avoid that?

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+      ]
+    )
+  ) }}
+
+  {% for section in intermediate_raster_sections %}
```
davemfish: In the LULC-based report the "Reclassified Baseline & Alternate LULC (to NDVI)" section comes before "Difference in NDVI between Alternate and Baseline". Since we tend to move backwards from final outputs -> intermediates -> inputs as we move from top down in the report, should those sections be swapped?

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+      args_table(args_dict)
+    )}}
+  {% endif %}
+
+  {% set input_items = [
+    (caption(raster_group_caption, pre_caption=True), 100)
+  ] %}
+  
+  {% if args_dict['lulc_base'] %}
+    {% set input_items = input_items + [
```
davemfish, line 113: Ah this pattern of `set` and appending to a list is kinda what I was thinking about with my comment above about the `cost_map` conditional.

src/natcap/invest/reports/raster_utils.py
davemfish: I think I see what you mean: this legend doesn't clearly indicate that the minimum value is 0 (or near 0). Readers could wonder how low the values go. If it was possible to add two tick marks, one at each end of the orange arrow, that would effectively communicate the same thing as the patch. Or even a single tick at 0 might do the job. `Colorbar.get_ticks()` and `set_ticks()` might make it possible to add those ticks while keeping all the other default ones. https://matplotlib.org/stable/api/colorbar_api.html#matplotlib.colorbar.Colorbar.set_ticks

>If we used this proposed approach, is your thought to keep the SpecialRangeRasterPlotConfig or incorporate this in the existing RasterPlotConfig?

I think I'm in favor of sticking with `RasterPlotConfig` and adding the extra attributes needed to make this possible. Mostly because the `vmin` and `set_under` tools look like they are designed exactly for this type of thing. If we have other cases where we really want to do custom things with the array we're plotting, then I really like the approach you designed to subclass `RasterPlotConfig` and override a plot method.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+        - preventable_cost
+        - preventable cases cost sum table (just total cases and total cost)
+        - vector symbolized with field 'sum_cost'
+        - vector symbolized with field 'sum_cases'
+
+    """
+    ndvi_colorramp = 'viridis_r'
+
+    input_raster_config_list = []
+    intermediate_raster_config_lists = [[
```
davemfish: Would this be easier to read & reason about if this nested list was organized as two separate lists instead?

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+        target_html_filepath (str): path to an HTML file to be generated by
+            this function.
+
+    Returns:
+        ``None``
+    """
+
+    input_raster_config_list, \
+        intermediate_raster_config_lists, \
+            output_raster_config_list = _get_conditional_raster_plot_tuples(
```
davemfish: Is there value in having a function that constructs and returns these lists -- as well as the intermediate output headings -- as opposed to defining them here in the body of `report`? I tend to prefer the latter as it means less flipping back and forth to understand what `report` is doing.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+    input_raster_caption = raster_utils.caption_raster_list(
+        input_raster_config_list)
+
+    outputs_img_src = raster_utils.plot_and_base64_encode_rasters(
+        output_raster_config_list)
+    output_raster_caption = raster_utils.caption_raster_list(
+        output_raster_config_list)
+
+    # There can be multiple sections for intermediate rasters
+    intermediate_img_srcs = [raster_utils.plot_and_base64_encode_rasters(
```
davemfish: It might be easier to reason about this if the intermediate sections were treated as multiple sections with distinct names reflecting their contents. It's up to you if you want to re-organize it or keep it as-is.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+        gettext('Baseline Prevalence & Cases')
+    ]
+    if args_dict['scenario'] == 'lulc':
+        intermediate_captions.insert(0,
+            gettext('Reclassified Baseline & Alternate LULC (to NDVI)'))
+
+    return intermediate_captions
+
+
+def _generate_agg_results_table(file_registry: dict) -> str:
```
davemfish: I suggest passing only the pieces of data used by the function as arguments. In this case, I think that is `table_path: str` and `preventable_cost: bool`. That keeps the function slightly more agnostic to the structure of `file_registry`. And it also makes it more transparent to someone just reading the function call, making it slightly less likely they need to go read the body of the function. It's kind of trivial in this example, but just generally a good pattern to follow. And there will always be exceptions/tradeoffs to evaluate.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+               intermediate_raster_captions)
+    ]
+
+    input_raster_stats_table = raster_utils.raster_inputs_summary(
+        args_dict, model_spec).to_html(na_rep='')
+
+    output_raster_stats_table = raster_utils.raster_workspace_summary(
+        file_registry).to_html(na_rep='')
+
+    agg_results_table = _generate_agg_results_table(file_registry)
```
davemfish: I think this function and `_create_aggregate_map` are good examples of when it's useful to abstract something out to a function. Because if I'm reading `report` I can treat this like a black box and generally know what I'm getting back without going to read the function body. I think that's less true for `_get_conditional_raster_plot_tuples` and that's the main reason behind my earlier comment about that function.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+        extent_feature,
+        xy_ratio,
+        attribute,
+        title):
+    """Create a choropleth map for a given attribute and return Vega JSON."""
+
+    # if the attribute has any negative values, use a divergent color
+    # scale; otherwise, use a continuous color scale
+    if (geodataframe[attribute] < 0).any():
+        scale = altair.Scale(scheme='purpleorange', reverse=True, domainMid=0)
```
davemfish: `domainMid` is very useful to know about!

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+    else:
+        scale = altair.Scale(scheme='purples')
+
+    chart = altair.Chart(geodataframe).mark_geoshape(
+        clip=True,
```
davemfish: I don't think there is any need to clip this layer is there? We always want to plot the whole thing?

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+        stroke="white",
+        strokeWidth=0.5
+    ).project(
+        type='identity',
+        reflectY=True,
+        fit=extent_feature
```
davemfish: This is only useful if the intent is to plot an extent that is different from the native extent of the geodataframe. It's used in conjunction with `clip`.

davemfish:
It looks like the ReadTheDocs build is failing due a few syntax issues in docstrings.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+
+def _get_conditional_raster_plot_tuples(model_spec: ModelSpec,
+                                        args_dict: dict,
+                                        file_registry: FileRegistry) -> tuple[
+        list[tuple[str, ...]],
+        list[tuple[str, ...]],
+        list[list[tuple[str, ...]]]]:
+
+    """
+    inputs
```
davemfish: I think we should consider letting the code speak for itself here regarding these conditionals. Anytime we use a comment to describe what code is doing we run the risk of the comment becoming out of sync with what the code actually does, and then it's worse than no comment at all! A good guiding principle is to use comments to describe "why" rather than "what".

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+    else:
+        return RasterDatatype.continuous
+
+
+def _get_conditional_raster_plot_tuples(model_spec: ModelSpec,
+                                        args_dict: dict,
+                                        file_registry: FileRegistry) -> tuple[
+        list[tuple[str, ...]],
+        list[tuple[str, ...]],
+        list[list[tuple[str, ...]]]]:
```
davemfish: I don't think these types are accurate. Each list contains `RasterPlotConfigs`, right?

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+    'Primary Outputs',
+    content_grid([
+      (caption(raster_group_caption, pre_caption=True), 100),
+      (raster_plot_img(outputs_img_src, 'Primary Outputs'), 100),
+      (caption(outputs_caption, definition_list=True), 100)
+    ])
+  )}}
+
+   {{ accordion_section(
+    'Preventable Cases by AOI Vector',
```
emilyanndavis: Would it make sense to change this heading to 'Preventable Cases by AOI Feature', to match the 'by AOI Feature' terminology that shows up inside this section? ('by AOI Vector' suggests there might be multiple vectors.)

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+  {% from 'caption.html' import caption %}
+  {% from 'content-grid.html' import content_grid %}
+  {% from 'metadata.html' import list_metadata %}
+  {% from 'raster-plot-img.html' import raster_plot_img %}
+  {% from 'wide-table.html' import wide_table %}
+
+  <h2 class="section-header">Results</h2>
+
+  {{ accordion_section(
+    'Aggregate Results',
```
emilyanndavis, line 21: Just dropping a note here to remind me/us to make sure we get this report and the SWY report set up for localization before we send out requests for translation. Can definitely wait until after the release, though!

src/natcap/invest/reports/templates/models/urban_mental_health.html
emilyanndavis: I was also curious, so I played around with dev tools on a sample report. It looks like the vega-embed content here ends up slightly overflowing its container horizontally, and these styles correct the problem. We could achieve the same effect by removing the left/right padding from the containing `.content-grid-item`, but not without adding support for that option to the content grid macro. There is probably lots of room for improvement in the content grid macro (I'm honestly amazed it's worked as well as it has), so I'd like to look into that at some point, but for now, I think Claire's solution is fine.

I might recommend defining a class in the CSS, to (1) clarify the purpose of these styles and (2) make them easily reusable in case they're needed elsewhere. Something like this, maybe?

```
-- styles.html --

.vega-fit-horizontal {
  display: flex;
  justify-content: center;
}

-- urban_mental_health.html --

<div class="vega-fit-horizontal"><div id="cases_map"></div></div>
```

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
...
+{% endblock content %}
+
+
+{% from 'vegalite-plot.html' import embed_vega %}
+
+{% block scripts %}
+  {{ super() }}
+  <script src="https://cdn.jsdelivr.net/npm/vega@5"></script>
+  <script src="https://cdn.jsdelivr.net/npm/vega-lite@5.20.1"></script>
+  <script src="https://cdn.jsdelivr.net/npm/vega-embed@6"></script>
```
emilyanndavis: Might be worth moving these three `<script>` tags out to their own partial (e.g., `vega-js.html` or `vega-dependencies.html`) now that we're using Vega in multiple reports.

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,146 @@
+{% extends 'base.html' %}
+
+{% block styles %}
+  {{ super() }}
+  {% include 'datatable-styles.html' %}
```
emilyanndavis: Did you end up using DataTables in this report? If not, you can remove this.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+            colorbar_kwargs (dict): additional kwargs to pass to fig.colorbar
+
+        Returns:
+            None.
+        """
+
+        if self.special_value_range is None:
+            LOGGER.info(
+                "No special value range for %s, using default plotting",
+                self.raster_path
```
emilyanndavis: Minor detail, but for consistency and readability, I think we've been trying to use f-strings rather than other styles whenever possible, at least when adding new code.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -0,0 +1,509 @@
...
+            special_arr,
+            cmap=matplotlib.colors.ListedColormap([self.special_value_color]),
+            interpolation='none',
+            vmin=1,
+            vmax=1
+        )
+
+        patch = matplotlib.patches.Patch(
+            color=self.special_value_color,
+            label=self.special_value_label or f'{low} - {high}'
```
emilyanndavis: For clarity, might I recommend spelling out "to" rather than using a hyphen (or any kind of dash, really)? Looking at a sample report, I did a double-take to make sure I was reading `-0.01 - 0.01` correctly, even though I already had the context to know what that part of the legend means.

```suggestion
            label=self.special_value_label or f'{low} to {high}'
```

src/natcap/invest/urban_mental_health/reporter.py
emilyanndavis: I'm guessing these types are an artifact of having used an existing reporter as an example before we did away with the tuple intermediates and moved to creating RasterPlotConfigs directly. Along those lines, I'd recommend updating the name of this function, too. :)

emilyanndavis (changes requested):
Looks great, Claire—you clearly put a lot of work into this! I have a handful of small suggestions.

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Sure, I'll also swap the "Difference in NDVI" section with the "Baseline Prevalence & Cases" section so it'll go: "Baseline Prevalence & Cases", then "Difference in NDVI", then "Reclassified Baseline & Alternate LULC (to NDVI)" because that makes more intuitive sense to me, and the `baseline cases` calculation technically comes after the `delta_NDVI` calculation (though they're independent so it doesn't really matter)

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: 1. Relevant for both maps: there is a little kebab menu that allows a user to download the plots and without this extra div, it overlaps the border of the container
2. Relevant for just the `else` condition: If a cost map isn't produced (because the user didn't enter a cost value), there will only be one vector plot and this styling ensures that it is center-justified instead of left-justified

There may certainly be a cleaner way to do this

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Very fair! I'll incorporate directly into `report`

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Ok I think I did this in 3a1be8168

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Yes totally makes sense, I agree. That function really got away from me!

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: yes good catch

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: My maps looks _really_ weird without this!

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Makes sense, thanks!

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Yes, definitely outdated, also I'm deleting this whole function

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Yes, I like it

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Yes, I have that in the back of my mind too!

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Ok I moved these to a new `vega-dependecies.html`, should this file be in a `partials` subdirectory within `templates`? Its currently just in `templates`

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Makes sense, I went back and forth on this, and now have changed to patch legend entirely rendering this obsolete!

src/natcap/invest/urban_mental_health/urban_mental_health.py
```diff
@@ -131,8 +131,8 @@
             required=False
         ),
         spec.OptionStringInput(
-            id="scenario",
-            name=gettext("scenario"),
+            id="model_option",
+            name=gettext("Option"),
```
davemfish, line 135: You may have litigated this elsewhere already, so you can ignore my input if you want to. I think "Option" is less descriptive than "Scenario". "Scenario option" seems even better.

The `id` is less user-facing, so maybe doesn't matter as much as the `name`, but `model_option` seems less descriptive than `scenario_option` in my opinion.

I also noticed that the sample data `json` file will need updating if this `id` changes. (the one that I downloaded a few days ago anyway).

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -136,82 +56,77 @@ def infer_continuous_or_divergent(raster_path: str) -> str:
...
 
 
-def _get_conditional_raster_plot_tuples(model_spec: ModelSpec,
-                                        args_dict: dict,
-                                        file_registry: FileRegistry) -> tuple[
-        list[tuple[str, ...]],
-        list[tuple[str, ...]],
-        list[list[tuple[str, ...]]]]:
+def _generate_agg_results_table(preventable_cases_cost_sum_table_path: str,
+                                cost: str | None) -> str:
```
davemfish: Should `cost` be a boolean? It doesn't look like the value is being used except to check for truthiness.

src/natcap/invest/urban_mental_health/reporter.py
davemfish: @claire-simpson , it looks weird without `fit=extent`? Or without the whole `project` parameters? Sorry my comment wasn't specific, I meant that `fit` probably isn't needed.

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -32,38 +31,59 @@ <h2 class="section-header">Results</h2>
...
-      (raster_plot_img(section['img_src'], section['heading']), 100),
-      (caption(section['caption'], definition_list=True), 100)
+      (raster_plot_img(intermediate_baseline_img_src, 'Baseline Prevalence & Cases'), 100),
+      (caption(intermediate_baseline_img_caption, definition_list=True), 100)
     ])
   )}}
-  {% endfor %}
+
+  {{ accordion_section(
+    'Difference in NDVI between Alternate and Baseline',
```
davemfish: Do you think it's helpful to title this section "Difference in Nature Exposure between ..." as a way to make the connection between NDVI as the proxy for nature exposure? That's assuming you do not change any of these output filenames -- it seems like people are okay to keep them as-is.

src/natcap/invest/urban_mental_health/urban_mental_health.py
claire-simpson: There was concern that "scenario" was confusing alongside our use of "future/counterfactual scenario", i.e. Alternate NDVI or LULC, and yes I have updated the json!

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: Oops yes you're right the `str` type hint is incorrect, but should be `float`. `cost` is just the health cost rate the user enters as a number > 0, but its an optional parameter which is why I'm just treating it like a boolean to check if cost should be added to the table

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Yes, great idea. Saw your comments in the UG google doc and I am also not completely opposed to changing the filenames.. just not sure which is more clear

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson, line 64: If a user enters `cost`, it must be > 0, so this is fine

src/natcap/invest/urban_mental_health/reporter.py
davemfish: Sorry, github didn't show me this thread when I was looking at the changes. If `cost is always >0` then this should be fine to say `if cost`. My original comment was trying to suggest that the function should maybe take a `boolean` parameter instead of the cost value since all the function wants to know is the `true/false`. Anyway, this is fine as-is.

src/natcap/invest/reports/templates/models/urban_mental_health.html
davemfish: I like the filenames as they are now; very precise.

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -0,0 +1,164 @@
...
+      (caption(cost_map_caption, aggregate_map_source_list), 100)
+    ]) %}
+  
+    {% set cases_cost_items = cases_cost_items + [
+      (cost_map_block, 50)
+    ] %}
+  {% endif %}
+  
+  {{ accordion_section(
+    'Preventable Cases by AOI Vector',
```
emilyanndavis: Looks like this update got lost in the course of making other changes 🙃 

```suggestion
    'Preventable Cases by AOI Feature',
```

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Ok great, thanks for your input!

src/natcap/invest/reports/templates/models/urban_mental_health.html
emilyanndavis: No, sorry—"partial" is just a generic term for a template that doesn't represent an entire page. But our file structure doesn't reflect that distinction, and it doesn't need to. In the `templates` directory is great!

src/natcap/invest/reports/templates/models/urban_mental_health.html
claire-simpson: Ahh good catch

PR #2452, opened 2026-03-23
Reports localization

### LINKED ISSUE #2319: Reports: localization of strings defined in report templates
## Background
In our work on InVEST reports, we've been identifying text strings that should be localized and wrapping them in `gettext()`, as we usually do for text that ultimately gets rendered in the Workbench or in the User Guide. While this should work for any text defined in a Python file, we also have some text strings that are defined in the report templates themselves. We should be able to use gettext/babel for this, but there is some additional configuration required.

## Resources/Possible Paths
- In theory, it should be possible to import `natcap.invest.gettext` into a report-dependent Python module (such as a specific `reporter.py`, or `natcap.invest.reports.__init__.py`), and assign it to a [custom filter](https://jinja.palletsprojects.com/en/stable/api/#writing-filters) on the Jinja environment.
- [The `i18n` extension](https://jinja.palletsprojects.com/en/stable/extensions/#i18n-extension) is designed to work with `gettext` and/or `babel`.
- As noted in the [Jinja Integration docs](https://jinja.palletsprojects.com/en/stable/integration/#babel), it may be necessary to create [a configuration file to map files to message extraction methods](https://babel.readthedocs.io/en/latest/messages.html#message-extraction).

emilyanndavis:
## Description
Resolves #2319. Specifically:
- Configures Babel and Jinja to extract translatable messages from report templates.
- Wraps all translatable text in HTML templates in `gettext()` (for string literals passed to macros) or `trans`/`endtrans` tags (everything else), to ensure messages are extracted.
- Uses current locale code in base template to populate `lang` attribute and UG URL.
- Miscellaneous updates to Python modules to ensure localized text is rendered as expected.

## Checklist


src/natcap/invest/internationalization/babel_config.ini
```diff
@@ -0,0 +1,9 @@
+# Extract messages from Python source files
```
emilyanndavis, line 1: I named this file `babel_config.ini` because the [Babel docs on mapping files](https://babel.readthedocs.io/en/latest/messages.html#message-extraction) indicate "[t]he configuration file syntax is based on the format commonly found in .INI files on Windows systems […]." Since I didn't find any requirements specifying how this file should be named, I suspect we could call it whatever we want. Open to suggestions if anyone has a preference for some other name or file type.

src/natcap/invest/reports/templates/models/carbon.html
```diff
@@ -16,18 +16,18 @@
   {% from 'raster-plot-img.html' import raster_plot_img %}
   {% from 'wide-table.html' import wide_table %}
 
-  <h2 class="section-header">Results</h2>
+  <h2 class="section-header">{% trans %}Results{% endtrans %}</h2>
 
   {{ accordion_section(
-    'Aggregate Results',
+    gettext('Aggregate Results'),
```
emilyanndavis, line 22: Note that while we prefer `trans`/`endtrans` tags for readability wherever possible (especially where text contains a variable), `gettext()` is appropriate for string literals, such as those passed as args to macros.

src/natcap/invest/cli.py
```diff
@@ -8,18 +8,14 @@
 import json
```
emilyanndavis, line 8: Removed some unused imports.

src/natcap/invest/spec.py
```diff
@@ -2274,13 +2276,13 @@ def execute(self, args, create_logfile=False, log_level=logging.NOTSET,
...
             if generate_report:
                 LOGGER.info('Generating report for results')
-                reporter_module = importlib.import_module(self.reporter)
                 target_html_filepath = os.path.join(
                     preprocessed_args['workspace_dir'],
                     (f'{self.model_id}_report'
                      f'{preprocessed_args.get("results_suffix", "")}.html'))
 
                 with natcap.invest.reports.configure_libraries():
+                    reporter_module = importlib.import_module(self.reporter)
```
emilyanndavis, line 2282: I had moved this line down because I was experimenting with a second context manager that would handle reloading the modules a particular reporter depends on, before importing the reporter module—until I realized there was a better way (see changes to `report_constants.py`).

I left this line here because it seems like a readability improvement to import the module immediately before invoking its `report` method, but otherwise I don't think it makes any real difference.

src/natcap/invest/__init__.py
```diff
@@ -43,6 +43,11 @@
 LOCALE_CODE = 'en'
 
 
+def get_locale():
+    """Get the current locale code."""
+    return LOCALE_CODE
```
emilyanndavis: This getter is needed so we can pass the current locale code to the base HTML template. If we import and reference `LOCALE_CODE` directly, it will always be 'en' because that was its initial value—it won't change even if it has been modified by `set_locale`.

tests/test_cli.py
```diff
@@ -55,9 +55,11 @@ def test_run_coastal_blue_carbon_workspace_in_json(self):
...
                 json.dumps(datastack_dict, indent=4, sort_keys=True))
 
-        with unittest.mock.patch(
-                'natcap.invest.coastal_blue_carbon.coastal_blue_carbon.execute',
-                return_value=None) as patched_model:
+        target = (
+            'natcap.invest.coastal_blue_carbon.coastal_blue_carbon.execute')
+
+        with (unittest.mock.patch(target, return_value=None) as patched_model,
+              unittest.mock.patch('importlib.reload')):
```
emilyanndavis, line 62: Each CLI test that mocks the CBC `execute` method was failing, because invoking `cli.main` with the `run` command would reload the model module, effectively overwriting the mock and forcing the real `execute` method to, well, execute. (And then the test would fail because there seemed to be some problem with the test datastack and/or the test workspace setup I didn't quite get to the bottom of, but I think it was merely an artifact of this particular test suite's environment and that the model itself is OK since I see no CBC test failures elsewhere.)

I updated `cli.main` so that it reloads the model module only if the model has a reporter, which makes this issue moot as long as CBC doesn't have a reporter. I also updated the affected tests so they mock `importlib.reload`, avoiding a true reload (and overwrite of the mocked `execute` method) so these tests will still pass if/when CBC has a reporter. (I tested and confirmed this locally.)

src/natcap/invest/cli.py
```diff
@@ -463,8 +459,15 @@ def main(user_args=None):
...
-                        model_module.__name__, model_module)
+
+            LOGGER.info(
+                f'Imported target {model_module.__name__} from {model_module}')
+
+            generate_report = bool(model_module.MODEL_SPEC.reporter)
+            if generate_report:
+                # Reload model module to ensure text defined in model spec is
+                # localized before it is passed to the report.
+                importlib.reload(model_module)
```
davemfish, line 474: So in the past the `model_module` would call `execute` without localizing first, and that was okay because localization was not necessary for any reason during/after `execute`? But now it is necessary because the report uses localized strings from `model_module`?

tests/reports/test_base_template.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.assertIsNotNone(soup.main.find(
+            string=render_args['model_description']))
+
+        ug_link = soup.find('a')
+        self.assertIn(render_args['userguide_page'], ug_link['href'])
+
+        for arg in ['report_script', 'invest_version',
+                    'report_filepath', 'timestamp']:
+            # self.assertIn(render_args[arg], soup.footer.string)
```
davemfish: leftover comment here

src/natcap/invest/cli.py
emilyanndavis: Correct. The report uses strings from `model_module`—specifically, from `model_module.MODEL_SPEC`—and although those strings are wrapped in `gettext`, those calls to `gettext` are executed only when the `model_module` loads. When `model_module` first loads, the locale is `en`, so all those strings are in their default English. When the locale changes, we need a mechanism to localize those strings on demand, i.e., by forcing all those calls to `gettext` to run again. I noticed that, when `cli.main` is invoked with the `validate` command, we are reloading `natcap.invest.validation_messages` to ensure the validation messages are localized, so I figured an analogous approach would make sense here. Hypothetically, if we were to move model specs out to their own modules (something we've entertained at least once or twice), we could get away with reloading only the (hypothetical) model spec module. As long as the `MODEL_SPEC` is defined on the `model_module`, however, reloading `model_module` seems to be the way to go.

But also … your question got me wondering how we manage to localize model_spec-defined strings in the Workbench, and it looks like we are actually missing a piece of the puzzle, since I found some text that is appearing in English even though we have a Spanish translation for it. I'll be digging into that separate from this PR.

tests/reports/test_base_template.py
emilyanndavis: Oops, thanks!

tests/reports/test_sdr_ndr_template.py
```diff
@@ -10,6 +10,7 @@
 
 
 def _get_render_args(model_spec):
+    locale = 'en',
```
claire-simpson:
```suggestion
    locale = 'en'
```

emilyanndavis: Whoops, thank you!

src/natcap/invest/carbon/reporter.py
```diff
@@ -230,10 +231,10 @@ def report(file_registry: dict, args_dict: dict, model_spec: ModelSpec,
             outputs_img_src=outputs_img_src,
             outputs_caption=output_raster_caption,
             intermediate_raster_sections=intermediate_raster_sections,
-            raster_group_caption=report_constants.RASTER_GROUP_CAPTION,
+            raster_group_caption=report_constants.raster_group_caption(),
             output_raster_stats_table=output_raster_stats_table,
             input_raster_stats_table=input_raster_stats_table,
-            stats_table_note=report_constants.STATS_TABLE_NOTE,
+            stats_table_note=report_constants.stats_table_note(),
```
emlys: IMO, it would be a bit more intuitive to keep the strings as constants and wrap them directly in `gettext` here:

```suggestion
            stats_table_note=gettext(report_constants.STATS_TABLE_NOTE),
```

src/natcap/invest/reports/report_constants.py
```diff
@@ -1,21 +1,35 @@
...
-    'rasters are available in the output workspace.'
-)
-
-STREAM_CAPTION_APPENDIX = gettext(
-    ' The stream network may look incomplete at this resolution, and '
-    'therefore it may be necessary to view the full-resolution raster '
-    'in GIS to assess its accuracy.'
-)
+# All `gettext`-wrapped strings defined in this module should be returned from
+# a function, _not_ defined on the module object itself, since text defined at
```
emlys: I'm not sure that the wrapper functions are adding anything here. It should be sufficient if we just make sure that `natcap.invest.gettext` is being called after the locale has been set.

src/natcap/invest/__init__.py
emlys: I should have done this originally, but may I suggest renaming the variable to `_LOCALE_CODE` to emphasize that it's not meant to be used outside this module?

src/natcap/invest/cli.py
emlys: The same thing happens in `ui_server.py` - we reload the model module before returning the spec or validation messages.

src/natcap/invest/spec.py
```diff
@@ -31,10 +31,12 @@
 from . import gettext
 from .unit_registry import u
 from . import validation_messages
+import natcap.invest.reports
```
emlys: Since `natcap.invest` is already imported, is this unnecessary?

src/natcap/invest/coastal_vulnerability/reporter.py
```diff
@@ -222,9 +222,17 @@ def report(file_registry, args_dict, model_spec, target_html_filepath):
 
     na_count = exposure_geo.exposure.isna().sum()
     if na_count:
+        # There is some redundancy here to ensure the checkbox label is
+        # localized properly. Note that passing an f-string to gettext will not
+        # work; we must explicitly use .format() to populate variable values.
```
emlys, line 227: Is the issue that "point(s)" might not translate well into other languages?

src/natcap/invest/internationalization/README.md
```diff
@@ -6,6 +6,9 @@ None of the translations files (.pot, .po, .mo) should be manually edited by us.
 ### `messages.pot`
 Message catalog template file. This contains all the strings ("messages") that are translated, without any translations. All the PO files are derived from this.
 
+### `babel_config.ini`
+Mappings file that tells pybabel where to look when extracting messages into the message catalog. By default, pybabel will extract messages from Python source files; we need this mappings file to ensure it also extracts messages from the Jinja templates that are used for HTML reports.
```
emlys, line 10: Thanks for adding this explanation!

src/natcap/invest/internationalization/README.md
```diff
@@ -69,7 +73,8 @@ Then follow the "Process to update translations" instructions above, starting fr
 * Model titles
 * `MODEL_SPEC` `name` and `about` text
 * Validation messages
-* Strings that appear in the UI, such as button labels and tooltip text
+* Strings that appear in the Workbench UI, such as button labels and tooltip text
```
emlys: It's true that workbench UI strings are eventually localized, but there is a different process for that, which is documented in `workbench/readme.md`. The process in this doc doesn't apply to those strings, so I suggest not mentioning them here.

src/natcap/invest/carbon/reporter.py
emilyanndavis: I agree—it doesn't seem quite right to define "constants" as functions—but I think the reason I ended up doing that is because I wanted the constants module itself to handle the localization. While your suggestion makes sense (and works), it shifts the responsibility to each reporter module, and as more reports are added I worry that it would be too easy to forget to wrap those constants in `gettext`, and we don't currently have a mechanism to catch that mistake.

I was trying to avoid reloading the `report_constants` module, but since I'm not thrilled with these alternatives, I'm now thinking it might be best to follow the same pattern we use for localizing validation messages, and just go ahead and reload the `report_constants` module when we need it (in this case, that would be just before reloading the model module when we're about to run a model that has a reporter). Do you see any reason not to do that?

src/natcap/invest/spec.py
emilyanndavis: Maybe. I only did this to appease my code editor's static analysis tools that seemed unconvinced of the existence of `natcap.invest.reports`. On further inspection, it looks like `natcap.invest` is imported only in order to access the `__version__`, so I could probably streamline these imports a bit.

src/natcap/invest/coastal_vulnerability/reporter.py
emilyanndavis: Yep! Some languages use different adjective forms for singular vs. plural nouns; verbs often must change as well (that's true in English, but sometimes we can sidestep that with implied verbs, e.g., "1 point [is] missing" vs. "2 points [are] missing"); and changes in word order are not out of the question. It's typically best to provide the full context for both the singular option and the plural option.

(Shorthand forms that rely on the ability to perceive punctuation, like "point(s)" or "is/are", aren't great for screenreader users either, but that's a bit of an aside—the primary motivation here is support for translations.)

src/natcap/invest/internationalization/README.md
emilyanndavis: This line was already here—I just added the word `Workbench`—but you raise a good point. I think it's still worth mentioning these items here (since they _are_ translated), but I'll clarify that they're handled separately.

src/natcap/invest/carbon/reporter.py
emlys: Hmm, well it is nice to avoid needing to reload the module. Reloading is really a kind of hacky antipattern, and I have wondered before if we should refactor other parts of invest to wrap translated objects in a function, rather than reloading. Since the reference to `stats_table_note` here is already wrapped in `render()`, it is evaluated in the correct language at runtime, so I think I prefer your current solution vs. introducing a need to reload.

I'd be curious what others think about the value of the wrapper functions as a "reminder" mechanism vs. directly calling gettext. If we keep the current design, I'd suggest renaming `report_constants` to `report_messages` or `report_strings`.

emilyanndavis: After mulling this over some more, I've decided to reload the module, at least for now. While it may not be ideal, I'm not satisfied with a wrap-every-string-in-a-function solution, either. It seems more logical to follow what we already do in similar scenarios than to introduce a new paradigm I'm not sold on. I do agree it's worth rethinking this though, so I created [an issue about seeking an alternative to our module-reloading solution](https://github.com/natcap/invest/issues/2473).

PR #2461, opened 2026-03-25
SWY Report Updates

megannissel:
This PR updates the SWY report based on the latest round of feedback from the model maintainers. Specifically:
- Updates the `about` text for the model output `L.tif` to include the note about negative local recharge values. I considered adding this to the report caption only, but it seems like a valuable enough piece of information to include in the output description in general. 
- Swaps colormap for the annual and monthly quickflow rasters from green-blue (GnBu) to Blues
- Swaps the x-axis labels in the QF/B/P chart from month names to indices
- Swaps the titles of the monthly QF rasters from English month names to Month #{x} 

The latter two changes allow us to avoid issues with localization and also make the labels more flexible. (Just because we generally expect monthly inputs starting in January and ending in December doesn't mean that's the 12-month timeframe users always input; nothing in the model would prevent a user from working with a dataset spanning June-May, for example.)

## Checklist

### DIFF
--- a/src/natcap/invest/seasonal_water_yield/reporter.py
+++ b/src/natcap/invest/seasonal_water_yield/reporter.py
@@ -32,14 +32,6 @@ TEMPLATE = jinja_env.get_template('models/seasonal_water_yield.html')
 
 MAP_WIDTH = 450 # pixels
 
-qf_label_month_map = {
-    f"qf_{month_index}": str(month_index) for month_index in range(1, 13)
-}
-
-
-def _label_to_month(row):
-    return qf_label_month_map[row['MonthLabel']]
-
 
 def _create_aggregate_map(geodataframe, xy_ratio, attribute,
                           title):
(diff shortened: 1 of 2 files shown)


src/natcap/invest/seasonal_water_yield/reporter.py
```diff
@@ -97,7 +89,7 @@ def _create_linked_monthly_plots(aoi_vector_path, aggregate_csv_path, xy_ratio):
     bar_chart = base_chart.mark_bar().transform_fold(
         ['baseflow', 'quickflow']
     ).encode(
-        altair.X("month(month):O").title("Month"),
+        altair.X("month", sort=[i for i in range(1, 13)]).title("Month"),
```
davemfish, line 92: Does this still need the "O" for Ordinal datatype? I remember that solved a problem with tick placement at one point. Or maybe now that these are ordered integers Altair is correctly guessing that's what we want?

megannissel: Yep, it seems the "O" is redundant now that we're specifying a sort order (which was necessary to prevent 10, 11, and 12 from appearing between 1 and 2).

PR #2464, opened 2026-03-26
start a keywords vocabulary and apply to model inputs

davemfish:
This PR adds `natcap.invest.keywords` and annotates inputs for AWY, Carbon, Coastal Blue Carbon, Coastal Vulnerability, SDR, Urban Flood Risk Mitigation.

#2467

## Checklist


src/natcap/invest/keywords.py
```diff
@@ -0,0 +1,157 @@
...
+    uuid='dd0b8bc9-90b3-4e7d-a021-e91dc676d622',
+    full_path='EARTH SCIENCE > HUMAN DIMENSIONS > POPULATION > POPULATION SIZE')
+
+PRECIPITATION = GCMDKeyword(
+    value='PRECIPITATION',
+    uuid='1532e590-a62d-46e3-8d03-2351bc48166a',
+    full_path='EARTH SCIENCE > ATMOSPHERE > PRECIPITATION')
+
+PRECIPITATION_RATE = GCMDKeyword(
+    value='PRECIPITATION_RATE',
```
megannissel: I don't think the GCMD keyword has an underscore; I think it's just `PRECIPITATION RATE`.

src/natcap/invest/keywords.py
```diff
@@ -0,0 +1,157 @@
...
+    uuid='ac50c468-df2f-429c-8394-9d63efcc6f9d',
+    full_path='EARTH SCIENCE > ATMOSPHERE > PRECIPITATION > PRECIPITATION_RATE')
+
+REFERENCE_EVAPOTRANSPIRATION = InvestKeyword(value='REFERENCE_EVAPOTRANSPIRATION')
```
megannissel: Can we do away with the underscore in this `value`, for consistency with the other keywords?

src/natcap/invest/keywords.py
```diff
@@ -0,0 +1,200 @@
...
+class GCMDKeyword(Keyword):
+    value: str
+    uuid: str
+    full_path: str
+    vocabulary: str = 'Global Change Master Directory (GCMD) Keywords'
+
+
+class InvestKeyword(Keyword):
+    value: str
+    definition: str = ''
```
megannissel, line 18: I really like the inclusion of definitions for the keywords we're creating; I imagine it will help future-us know what we meant!

src/natcap/invest/spec.py
```diff
@@ -293,6 +300,27 @@ def describe_rst(self):
...
+        """Get list of child Inputs."""
+        if hasattr(self, 'fields'):
+            return self.fields
+        if hasattr(self, 'columns'):
+            return self.columns
+        if hasattr(self, 'contents'):
+            return self.contents
+        return []
+
+    def get_keywords(self, incldue_children=True):
```
megannissel:
```suggestion
    def get_keywords(self, include_children=True):
```

src/natcap/invest/spec.py
```diff
@@ -293,6 +300,27 @@ def describe_rst(self):
...
+        """Get a list of unique keyword strings for this input and children."""
+        keywords = []
+        if self.keywords:
+            keywords += [keyword.value for keyword in self.keywords]
+        if incldue_children:
```
megannissel:
```suggestion
        if include_children:
```

tests/test_spec.py
```diff
@@ -492,6 +492,60 @@ def test_option_string_input_preprocess(self):
...
+            columns=[
+                spec.LULC_TABLE_COLUMN,
+                spec.NumberInput(
+                    id='carbon', units=None, keywords=[keywords.CARBON]),
+            ]
+        )
+        expected_keywords = [keywords.LULC.value, keywords.CARBON.value]
+        self.assertCountEqual(csv_input.get_keywords(), expected_keywords)
+        self.assertCountEqual(
+            csv_input.get_keywords(incldue_children=False), [keywords.LULC.value])
```
megannissel:
```suggestion
            csv_input.get_keywords(include_children=False), [keywords.LULC.value])
```

tests/test_spec.py
```diff
@@ -492,6 +492,60 @@ def test_option_string_input_preprocess(self):
...
+            fields=[
+                spec.LULC_TABLE_COLUMN,
+                spec.NumberInput(
+                    id='carbon', units=None, keywords=[keywords.CARBON]),
+            ]
+        )
+        expected_keywords = [keywords.LULC.value, keywords.CARBON.value]
+        self.assertCountEqual(vector_input.get_keywords(), expected_keywords)
+        self.assertCountEqual(
+            vector_input.get_keywords(incldue_children=False), [keywords.LULC.value])
```
megannissel:
```suggestion
            vector_input.get_keywords(include_children=False), [keywords.LULC.value])
```

tests/test_spec.py
```diff
@@ -492,6 +492,60 @@ def test_option_string_input_preprocess(self):
...
+            keywords=[keywords.LULC],
+            name="foo",
+            contents=[spec.DEM, spec.SOIL_GROUP]
+        )
+        expected_keywords = [
+            keywords.LULC.value, keywords.DEM.value,
+            keywords.HYDROLOGIC_SOIL_GROUPS.value]
+        self.assertCountEqual(directory_input.get_keywords(), expected_keywords)
+        self.assertCountEqual(
+            directory_input.get_keywords(incldue_children=False),
```
megannissel:
```suggestion
            directory_input.get_keywords(include_children=False),
```

src/natcap/invest/keywords.py
```diff
@@ -0,0 +1,200 @@
...
+    uuid='6f6537f5-773f-4df1-862b-d9ab80eb5e04',
+    full_path='EARTH SCIENCE > BIOSPHERE > VEGETATION > CARBON')
+
+CARBON_SEQUESTRATION = GCMDKeyword(
+    value='CARBON SEQUESTRATION',
+    uuid='e58872a8-6104-4ff8-bbca-4b00ba4b38e8',
+    full_path='EARTH SCIENCE > BIOSPHERE > ECOLOGICAL DYNAMICS > ECOSYSTEM FUNCTIONS > CARBON SEQUESTRATION')
+
+CARBON_SOIL = GCMDKeyword(
+    value='CARBON',  # Note this value is not unique in this module
```
megannissel: I'm slightly concerned about having non-unique keyword values. If the keyword that is stored in the Hub is the `value` here (which is likely to be the case), then there will be no way to distinguish between this `CARBON` and the other `CARBON` when searching datasets. 

In this instance specifically, it looks like these keywords have different places within the GCMD hierarchy, but the definitions are the same, "Any measurement of Carbon in the soil, in organic or inorganic fractions." With that in mind, I'm inclined to say we should choose only one of them to include in our keywords vocabulary. 

More generally, if we won't be able to leverage the difference (implied by the GCMD `full_path`) between keywords with non-unique values when searching for relevant inputs, is there a benefit to having them? Please let me know if you can think of any ways in which they might be useful to us beyond input-searching!

src/natcap/invest/carbon/carbon.py
```diff
@@ -91,6 +94,9 @@
             about=gettext(
                 "A table that maps each LULC code to carbon pool data for that LULC type."
             ),
+            keywords=[
+                keywords.BIOPHYSICAL_TABLE,
+                keywords.CARBON, keywords.CARBON_SOIL],  # TODO: Could place on columns instead
```
megannissel: I think it may be helpful to choose a convention and try to stick to it. I don't have a strong opinion on whether the keywords ought to go on the top-level Input itself or the relevant column, but I am in favor of consistency (especially with long-term maintainability in mind) and now feels like the best time to establish a convention.

megannissel (changes requested):
Thanks, @davemfish, I think this is looking really good! I have a couple of small suggestions and a bigger question about the pros and cons of non-unique keyword values for your consideration.

src/natcap/invest/keywords.py
davemfish: I agree, we are not leveraging the difference, so adding both versions of this keyword is less about facilitating our use case (DataHub search) and more about just being thorough with adding all relevant keywords to the InVEST input. In that sense the benefits are only in the _potential_ use someone might have for cross-referencing invest inputs with other catalogs that use the GCMD.

But you make a good point that both of these GCMD keywords have the same definition, so there's probably no value in storing both. If they were more distinct, and both relevant, then I think there's a case to be made for including multiple keywords on an input. It doesn't (or shouldn't) mean the DataHub dataset needs to have them all.

src/natcap/invest/carbon/carbon.py
davemfish: I think there can be good reason to place the keyword on the column/field when the keyword specifically describes the contents of that column. In this case `keywords.CARBON` is not quite specific enough to assign to any individual column so I'll leave it here.

davemfish:
Thanks, @megannissel , I fixed the typo, added some missing definitions, and removed some redundant keywords.

megannissel (approved):
Thanks, @davemfish, I think this is looking really good!

Issue #2315, opened 2026-01-26
Workbench Puppeteer test failing on Mac binary builds


davemfish:
Workbench Puppeteer test is consistently failing on Mac binary builds. I haven't tested this locally, but in GHA I have observed this test failing the same way a bunch of times recently:
https://github.com/natcap/invest/actions/runs/21360397455/job/61477950115?pr=2306

```
FAIL tests/binary_tests/puppet.test.js (224.56 s)
  ✕ Run a real invest model (98783 ms)
  ✓ Check local userguide links (123327 ms)
  ○ skipped Install and run a plugin
  ○ skipped App re-launch will exit and focus on first instance

  ● Run a real invest model

    TimeoutError: Waiting for selector `[name="Cancel Run"][role="button"]` failed: Waiting failed: 15000ms exceeded
```

In the past, we have seen intermittent errors similar to this, but this time it is pretty consistent. The first thing I would try is doing all the software updates related to #1800 .

dcdenu4:
Re: Slack conversation. I agree with both @davemfish and @emilyanndavis that we should skip the test for now in our CI/CD environment but be able to run it locally. @emilyanndavis made a good point that when we update React (#2481) and other dependencies, including puppeteer, it would be good to revisit adding the test back for CI/CD to see if anything worked itself out.

davemfish:
> From what I've read, `playwright` only works by running the electron app in dev mode. So it runs the app's source code, not the binary executable. I think one of the key benefits of our end-to-end test is that it runs the production binary. I don't think we should give that up.

I don't think this is accurate actually. Many examples show launching the app via `main.js`, but we should try launching it from the Workbench binary as well.

I recently tried upgrading puppeteer and it broke our script. Getting it working again would require figuring out how to make it compatible with jest, probably with the help of `jest-puppeteer-preset`. Though I tried and failed to get that working with `puppeteer-core` rather than the full `puppeteer`.

Puppeteer is probably not the right tool for the job. Selenium or Playwright have more direct support for Electron.

PR #2468, opened 2026-04-01
Rec model - reduce redundancy in file transfers

### LINKED ISSUE #1950: Rec server process crashes sometimes
The Rec model server process sometimes ends without any traceback in the log. The end of the log indicates that a job had completed it's PUD and TUD calculations, and the workspaces were zipped. The last message in the logfile is
`TUD_f913a74e-ec67-49d3-a1d0-d21dbdee826a INFO     finished TUD; zipping result`.

But we should also see the next message, and we do not:
`LOGGER.info('all user-day calculations complete; sending binary back')`

I'm going to try to add some more logging and wait to see if it happens again.

The AOI in this case was very large in terms of number of features (over 3 million). In terms of number of photos/tweets in the AOI, it was well below our threshold for allowing the queries.

davemfish:
Instead of sending results from server to client in two different geopackages (PUD_results.gpkg and TUD_results.gpkg), store userday counts in CSVs on the server and send those back to the client instead. The client then joins those tables to the AOI polygons. There are no changes to the structure of files in the user's workspace.

Since this fix requires changes to both the client and server, we need a dedicated server process running for this new client. That process is already live, based on the changes in this branch. We also still need the former server process to support invest clients up to and including the current release (3.18.0). I have updated this wiki to describe the current server infrastructure: https://github.com/natcap/invest/wiki/Recreation-model-cloud-infrastructure

Fixes #1950

## Checklist


davemfish:
GHA failures are due to #2315 and #2470

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -803,32 +811,71 @@ def wrap_calculate_userdays():
...
+
+        # Copy the monthly results table to the output workspace
+        # Results are returned from the server without a results_suffix.
+        monthly_filename = f'{result_key}_monthly_table.csv'
+        pre, post = os.path.splitext(monthly_filename)
+        target_filepath = os.path.join(
+            output_dir, f'{pre}{file_suffix}{post}')
+        shutil.copy(
+            os.path.join(temporary_output_dir, monthly_filename),
+            target_filepath)
```
emlys, line 876: Can these paths be constructed or referenced using the file registry?

src/natcap/invest/recreation/recmodel_server.py
```diff
@@ -675,24 +665,24 @@ def _calc_aggregated_points_in_aoi(
...
-                line += (
-                    ",".join(['%s' % len(ud_monthly_set[header])
-                              for header in table_headers]))
-                line += '\n'  # final newline
-                monthly_table.write(line)
+                poly_feat = aoi_layer.GetFeature(fid)
+                poly_id = poly_feat.GetField(poly_id_field)
+
+                monthly_values_str = ",".join(
+                    [str(len(ud_monthly_set[header])) for header in table_headers]
```
emlys, line 675: I know it was this way before, so feel free to disregard, but might it be worth using pandas/geopandas for this so we're not directly writing out CSVs line by line?

HISTORY.rst
```diff
@@ -64,6 +64,13 @@
 Unreleased Changes
 ------------------
 
+General
+=======
+* Refactored ``make deploy`` into artifact-specific targets (``deploy_dist``,
+  ``deploy_data``, ``deploy_userguide``, ``deploy_workbench``) and updated
+  these targets to fail on missing artifacts instead of silently ignoring
+  errors. (`#831 <https://github.com/natcap/invest/issues/813>`_)
```
emlys, line 72: Did this history note sneak in from another branch somehow?

scripts/recreation_server/launch_recserver.sh
```diff
@@ -2,6 +2,10 @@
 # activate a python environment before calling this script:
 # mamba activate /usr/local/recreation-server/invest_3_15_0/invest/env
 # I think I had a hard time doing the activation within this script, for some reason.
-nohup python -u /usr/local/recreation-server/invest_3_15_0/invest/scripts/recreation_server/execute_recmodel_server.py \
+export PYRO_LOGFILE=/usr/local/recreation-server/invest_3_15_0/server/pyro_invest319.log
+export PYRO_LOGLEVEL=DEBUG
+export PYRO_THREADPOOL_SIZE_MIN=8
+export PYRO_DETAILED_TRACEBACK=true
+nohup python -u /usr/local/recreation-server/invest_3_15_0/invest_3_19_0/scripts/recreation_server/execute_recmodel_server.py \
```
emlys, line 9: Is this path meant to contain both `invest_3_15_0` and `invest_3_19_0`?

.github/workflows/build-and-test.yml
emlys, line 527: Are these changes meant to go in this branch?

emlys (commented):
Thanks, @davemfish, this makes sense to me. I had a few minor suggestions, and also wanted to check if there might be some unrelated changes going in this branch unintentionally?

.github/workflows/build-and-test.yml
davemfish: These changes are already in the target branch. I guess they were not when I first opened this PR, but I don't know why github still shows this diff.

scripts/recreation_server/launch_recserver.sh
davemfish: Yeah, it's an unfortunate bit of poor naming / file management on my part. If you want the details they are in https://github.com/natcap/invest/wiki/Recreation-model-cloud-infrastructure

src/natcap/invest/recreation/recmodel_client.py
davemfish: Oh right, I forgot about that pattern. I started doing it, but I'm not sure it's a win in this case due to extra book-keeping. We need to use `result_key` to figure out if we have PUD or TUD data, so it means building another lookup between those acronyms and the `file_registry` items. Or we can just use `result_key` directly in the filename and skip the registry as it is now.

src/natcap/invest/recreation/recmodel_server.py
davemfish: We're getting these data off a queue so I think writing them line by line as we get them makes sense.

HISTORY.rst
davemfish: The `General` section was out of order so I moved it up to the top here.

davemfish:
> Thanks, @davemfish, this makes sense to me. I had a few minor suggestions, and also wanted to check if there might be some unrelated changes going in this branch unintentionally?

Thanks, @emlys , back to you. I'm not sure how to get GitHub to update its diff, but I'm confident those changes are already in the release branch.

src/natcap/invest/recreation/recmodel_client.py
```diff
@@ -803,32 +811,71 @@ def wrap_calculate_userdays():
...
-                layer = vector.GetLayer()
-                _rename_layer_from_parent(layer)
-                layer = vector = None
+        LOGGER.info('Joining results to AOI polygons')
+        driver = gdal.GetDriverByName('GPKG')
+        aoi_vector = gdal.OpenEx(local_aoi_path, gdal.OF_VECTOR)
+        # Copy the input AOI into the designated output folder
+        # Results are returned from the server without a results_suffix.
+        target_filepath = os.path.join(
+            output_dir, f'{result_key}_results{file_suffix}.gpkg')
```
emlys, line 829: I think this path can be accessed from the file registry?
```suggestion
        target_filepath = file_registry[f'{result_key}_results']
```

src/natcap/invest/recreation/recmodel_client.py
emlys: I might be missing something, but couldn't it be all in one like
`file_registry[f'{result_key}_monthly_table']`? Since there are separate outputs in the spec for `pud_monthly_table` and `tud_monthly_table`. 

I think that's preferable too so that the file registry records that these outputs have actually been produced.

davemfish: We would have to pass the `file_registry` object into this function. It's not currently in scope. And that would introduce a side-effect right, where the function modifies that object? That could be okay, we're probably doing it in other places.

After your first comment I was thinking of the pattern of indexing into `file_registry` in `execute` and passing in the necessary filepaths to this function. That was the path I started going down and that's what prompted the need for extra book-keeping. Something like this where `result_key` is 'PUD' or 'TUD',
```python
'PUD': [pud_monthly_table_path, pud_results_vector_path],
'TUD': [tud_monthly_table_path, tud_results_vector_path]
```
This felt bad because it was like deconstructing the file registry only to build back up another similar lookup. So if anything at this point I would probably prefer to either refactor the function to take file_registry as an arg, or leave it as-is.

>I think that's preferable too so that the file registry records that these outputs have actually been produced.

We are indexing into the registry for these items in the `target_path_list` of the task, so we're at least covered there.

davemfish: One argument for leaving the function definition as it is: It's friendlier to python API users like this, I think. I can call the `_retrieve_user_days` function directly with filepaths as arguments rather than setting up the `file_registry` first. I know no one is really doing stuff like that, but I like the idea of staying open to it with the design of our functions.

emlys: Ah okay, I think that's alright then, since we are indexing the registry for these paths elsewhere.

emlys (approved):
Thanks @davemfish, you've convinced me not to worry about the file registry. The failing tests look unrelated.

PR #2475, opened 2026-04-03
Bugfix: NDR report conditionals

### LINKED ISSUE #2472: Reports: NDR KeyError when the model is run with conditional configurations.
NDR report exception when the model is run with some conditional configurations.

```python
2026-04-01 15:09:46,246 (natcap.invest.spec) spec.execute(2276) INFO Generating report for results
2026-04-01 15:09:46,247 (natcap.invest.utils) utils.prepare_workspace(173) ERROR Exception while executing ndr
  File "natcap\invest\utils.py", line 171, in prepare_workspace
  File "natcap\invest\spec.py", line 2284, in execute
  File "natcap\invest\ndr\reporter.py", line 76, in report
KeyError: 'n_surface_export'
```

In this case, we're not checking whether or not `calc_n` is `True` before referencing this key in the file_registry. The same is probably true for `calc_p` and related outputs.

Part of the args dict:
```python
calc_n                       False
calc_p                       True
```

emilyanndavis:
## Description
Fixes #2472. Specifically: I've moved the affected `RasterPlotConfig` instantiations into the relevant conditional blocks.

I also removed the `_get_nutrient_dependent_list` function. I think I initially used this function to support `output_raster_plot_configs` as well as `results_vector_cols_to_sum`, but since it was dropped from one use case, it seemed logical to drop it from the other as well.

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -81,6 +81,13 @@ Workbench
   (`#2436 <https://github.com/natcap/invest/issues/2436>`_)
 
 
+NDR
+===
+* Fixed a bug that caused the NDR report to fail when the model was run with
+  ``calc_n=False`` and ``calc_p=True``, or vice versa.
+  (`#2472 <https://github.com/natcap/invest/issues/2472>`_)
+
+
 3.18.0 (2026-02-25)
 -------------------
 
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Thank you!

PR #2482, opened 2026-04-07
Workbench bugfix for inaccurate "Model Complete" status

### LINKED ISSUE #1655: Workbench: changing parameters from a previous run does not change "model complete" state

davemfish:
This adds the same fix for updating the job status after batch-loading input args that we already have in place after modifying individual input args. Also added tests to cover both of these cases.

Fixes #1655

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -79,6 +79,9 @@ Workbench
 * The Workbench now filters model args to include only those whose
   corresponding form fields are enabled at the time a model is run.
   (`#2436 <https://github.com/natcap/invest/issues/2436>`_)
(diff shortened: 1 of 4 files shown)


emilyanndavis (approved):
Looks good, @davemfish. Thanks for the update and for adding tests!

Issue #2483, opened 2026-04-07
Consistently handle library redistribution in wheels


emlys:
Our manylinux wheels on PyPI are over 50MB, while our wheels for other systems are <2MB. Is something being bundled in that shouldn't be?

Wheels should generally be kept small, and for reference PyPI limits wheels to 100MB by default, so this is quite large. Looking at a couple of other projects that use manylinux, their manylinux wheels are not larger than other dists. Very likely something isn't right here.

emlys:
Update: the size difference is attributable to a whole bunch of library files being included in the manylinux wheels. Turns out this is what `auditwheel` does (a tool we run in the build process that makes the wheels portable across linux systems). I did not realize that it does that by just copying in all the .so files!

The two concerns I have with this are
- are we including the necessary license files etc. to be redistributing these libraries?
- is 50MB unacceptably large for a wheel?

emlys:
## include libs in our wheels for all platforms
There are equivalent tools like `auditwheel` for macos and windows (`delocate`, `delvewheel`). We could do the same thing on each platform. This would increase complexity of our build process and require that we include all the library licenses in our wheels to be compliant with their license requirements.

If we choose to do that, we'd likely need to keep a copy of each library's license in our repo to make them available during the build. There would be a one-time effort to manually find each license, and it would be good to occasionally check if any libraries have been added or changed licenses. There are ~60 libraries included when we build against the "full" GDAL installation from conda-forge that includes optional drivers.

## or remove libs from our linux wheels
If we stop using `auditwheel`, we'd no longer be redistributing other libraries, avoiding the license issue. This would also make our wheels not `manylinux` compliant. And it's likely that the same `libgdal` version conflict issue that we've seen on macos (https://github.com/natcap/invest/issues/2206) could happen on linux too.

PR #2487, opened 2026-04-09
Constrain GDAL to one minor version for wheel builds

### LINKED ISSUE #2206: GDAL lib version conflicts with natcap.invest wheels
`natcap.invest` 3.16.2 should be compatible with `gdal` 3.10.3. However, it seems it's not guaranteed that we're getting the right version of the gdal libs:
```
mamba create -n test python==3.13.0 gdal==3.10.3
mamba activate test
pip install natcap.invest==3.16.2
python -c "from natcap.invest.delineateit import delineateit_core"
```
fails with
```
ImportError: dlopen(/Users/emily/miniforge3/envs/test/lib/python3.13/site-packages/natcap/invest/delineateit/delineateit_core.cpython-313-darwin.so, 0x0002): Library not loaded: @rpath/libgdal.37.dylib
  Referenced from: <CE9AD72F-0FBD-360A-A858-0A0060BEE043> /Users/emily/miniforge3/envs/test/lib/python3.13/site-packages/natcap/invest/delineateit/delineateit_core.cpython-313-darwin.so
  Reason: tried: '/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.37.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.37.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS@rpath/libgdal.37.dylib' (no such file), '/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/micromamba/envs/env/lib/libgdal.37.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.37.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.37.dylib' (no such file)
```
Perhaps we need to be building wheels against the lowest compatible version of gdal?

emlys:
## Description
Fixes #2206, #2483
Added an extra GDAL constraint to the wheel build environment and package dependencies. Removed the `auditwheel` step from the wheel builds on linux.

## Checklist


setup.py
```diff
@@ -8,13 +8,6 @@
 from setuptools.command.build_py import build_py as _build_py
 from setuptools.extension import Extension
 
-# Read in requirements.txt and populate the python readme with the
-# non-comment, non-environment-specifier contents.
-_REQUIREMENTS = [req.split(';')[0].split('#')[0].strip() for req in
```
emlys, line 13: This is not a necessary change for this PR, but we might as well take advantage of the support for reading a requirements file directly in pyproject.toml.

pyproject.toml
```diff
@@ -27,9 +27,11 @@ classifiers = [
     "Topic :: Scientific/Engineering :: GIS"
 ]
 # the version is provided dynamically by setuptools_scm
-# `dependencies` and `optional-dependencies` are provided by setuptools
-# using the corresponding setup args `install_requires` and `extras_require`
-dynamic = ["version", "dependencies", "optional-dependencies"]
```
emlys, line 32: We're not using `optional-dependencies` (though maybe we should!), so I removed it for now.

setup.py
phargogh: Yep, makes sense to consolidate towards pyproject.toml wherever possible!

.github/workflows/build-and-test.yml
```diff
@@ -94,6 +95,16 @@ jobs:
...
+      # The C++ extensions will be compiled and linked against the libgdal version 
+      # that's in the build environment, and so that same libgdal version has to be 
+      # available in the target environment. Each libgdal version corresponds to a 
+      # minor version of GDAL itself. Adding this requirement to the package metadata
+      # should prevent it from being installed in an env without the necessary libgdal.
+      # Users who need a different GDAL version can use the conda-forge package or 
+      # build their own wheel from source.
+      - name: Add gdal constraint for wheel
+        run: echo "gdal==${{ env.GDAL_VERSION_SPECIFIER_FOR_WHEEL }}" >> requirements.txt
+
```
phargogh, line 107: Won't this direct modification of `requirements.txt` result in a `.d<today>` dirty tag in the version?

Personally, I am hopeful that we might be able to write a new custom command for setup.py since we're still using the setuptools build backend.  Something like this, maybe?  I think doing this should result in the version being unchanged.

```diff
diff --git a/setup.py b/setup.py
index 8dcb559d1..35bca5ca1 100644
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ import subprocess
 import numpy
 from Cython.Build import cythonize
 from setuptools import setup
+from setuptools.command.bdist_wheel import bdist_wheel
 from setuptools.command.build_py import build_py as _build_py
 from setuptools.extension import Extension

@@ -58,6 +59,12 @@ class build_py(_build_py):
         # then execute the original run method
         _build_py.run(self)

+class CustomWheel(bdist_wheel):
+    def run(self):
+        # Dynamically modify requirements before the wheel is packed
+        self.distribution.install_requires.append("gdal==3.10.*")
+        bdist_wheel.run(self)
+

 setup(
     install_requires=_REQUIREMENTS,
@@ -92,6 +99,8 @@ setup(
     ], compiler_directives={'language_level': '3'}),
     include_dirs=[numpy.get_include()],
     cmdclass={
-        'build_py': build_py
+        'build_py': build_py,
+        'build_wheel': CustomWheel,
+        'bdist_wheel': CustomWheel,
     }
 )

phargogh (commented):
Thanks @emlys !  I really like the solution of modifying `requirements.txt`, but wouldn't doing so also cause the version string to be modified since there would now be uncommitted changes?  I offered a possible alternative that _might_ be the kind of thing that would modify the library requirements before the wheel is built, which is the kind of thing we used to have to do when setuptools was in its heyday.  Looking ahead, there's probably a more modern way to do this since setuptools is probably past its prime.

.github/workflows/build-and-test.yml
phargogh: And also, I'm sorry about the deep setuptools stuff!  It was always arcane, and I've spent _way_ too much time mired in setuptools subclasses over the years, so this kind of modification seems like it should work, but there is probably a way better way to to it.

phargogh (commented):
Huh, looks like it didn't post my response as commented, github is saying that I saw the review but didn't actually post my response?  I'll try resubmitting here.

emlys:
@phargogh that's a very good point, thanks for catching that! The custom build class you suggest does seem like a valid solution. A couple other possibilities come to mind:

- We could put the gdal constraint in an optional dependency group called something like `wheel` - then `pip install natcap.invest[wheel]` would require that specific version, but `pip install natcap.invest` would not.
- We could put the gdal constraint in a separate requirements file e.g. `wheel-constraints.txt` and tell users to install as `pip install natcap.invest -c wheel-constraints.txt`. _I think this is worse than the optional-dependency approach because it's not embedded in the package metadata._

I do want to prioritize moving in the direction of using `pyproject.toml`, and I think there's a good case to be made that our requirements should eventually live in `[project.dependencies]` (and generate `requirements.txt` from that as needed, not the other way around). Particularly because `pyproject.toml` is a `setuptools` concept that's specified in multiple PEPs, whereas `requirements.txt` is ultimately a `pip`-specific format. And TOML has been supported in the standard library since python 3.11. With that in mind, I kind of like the optional-dependency approach because it could be fully static in `pyproject.toml`. But, it only works if the user knows to install `natcap.invest[wheel]` and not `natcap.invest`. What do you think?

phargogh:
Of the two options you mention, I too prefer the `pip install natcap.invest[wheel']` approach, but I also worry that it's already far too easy for someone to `pip install natcap.invest`, get a wheel and then wonder why it doesn't work.

On the specific point about requirements.txt being a pip format and not in setuptools, that is a really good point!  And I agree with you that the dependencies should be defined in pyproject.toml instead of in requirments.txt.  But for the case where we are modifying the wheel build config, we can still remove the `install_requires` parameter to `setup()` in setup.py and setuptools still reads the parameters from pyproject.toml and we can still modify them that way for the wheel build.  Would that change things?  Or do you think the extras approach would still be preferable?

emlys:
Ok, you've convinced me that the requirement belongs in the regular wheel without needing a modifier. The custom wheel class you suggest seems like a good solution! I'm just a little hesitant to add more code to `setup.py`.

A possible alternative I found in PEP 517 would be to implement an [in-tree build backend](https://peps.python.org/pep-0517/#in-tree-build-backends) that overrides [`prepare_metadata_for_build_wheel`](https://peps.python.org/pep-0517/#prepare-metadata-for-build-wheel). This approach comes recommended by the [setuptools docs](https://setuptools.pypa.io/en/latest/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks), and it has the advantage of not further tying us to `setup.py`. It also supports passing in custom configuration options to the build frontend. So we could do something like `python -m build --wheel -Cwheel_requires='gdal=3.10.*'`, which seems like a neat alternative to reading from an env variable.

However, this is at least as arcane as the the custom class approach, and requires slightly more code and an additional file. What do you think?

phargogh:
I think this is a great solution!  Definitely still a little arcane, but I have to say that it is _so_ nice to have officially-supported build hooks like this that are well defined, well-documented and unlikely to break in a bugfix release of some distutils derivative that you didn't know your builds depended on.  It's still a little obscure, but SO much better than the blog-based rumors and old stackoverflow posts that we had to rely on for setup.py modifications.

So yes, I think this would be AWESOME and I totally agree that it'd be way better to move this away from setup.py.  hopefully the new file won't be a big burden to write, but I don't imagine it would be for this particular case.

emlys:
@phargogh I redid this using the same approach that we merged in pygeoprocessing, unpacking and modifying the wheel after it's built.

phargogh (approved):
Changes look good, thanks @emlys !  The two build/test failures appear to be consistent with flaky environment management I've seen before.  I'm re-running just in case, but adding my approval and I'll merge anyways as they look unrelated to this content.

PR #2503, opened 2026-04-15
remove ManagedRaster and build against pygeoprocessing's ManagedRaster

### LINKED ISSUE #2483: Consistently handle library redistribution in wheels

### LINKED ISSUE #2502: Build against pygeoprocessing's `ManagedRaster`
`pygeoprocessing` has its own copy of `ManagedRaster`, and the header files are already being included in the wheel and sdist. invest should build against those, and we can remove the duplicate files from the invest source.

emlys:
## Description
Fixes #2502
Fixes #2483 
- Deletes natcap.invest's copy of the ManagedRaster source files
- Import these from pygeoprocessing instead
- This requires those source files from pygeoprocessing to be available to the natcap.invest build. pygeoprocessing is now a build-time dependency, and the `pygeoprocessing.extensions` source directory is now passed to the compiler as an `include` path.

With `pygeoprocessing` as a build dependency, `gdal` is also a build dependency (it already was implicitly, but we didn't have it in `build-system.requires`). By default, that means the build process will try to install `pygeoprocessing` and `gdal` wheels into an isolated build environment, and the `gdal` install will fail because `libgdal` is missing. To work around this, we can disable build isolation and instead manage the build environment ourselves. 

It seems that disabling build isolation is not unheard of for packages that have C++ dependencies: see discussion [here](https://discuss.python.org/t/build-isolation-passthrough-allowing-a-specific-package-to-always-passthrough-the-build-isolation/77284) and [here](https://pypackaging-native.github.io/key-issues/abi/); the `uv` package manager provides an [option](https://docs.astral.sh/uv/reference/settings/#no-build-isolation-package) to disable build isolation for specific packages only.

Build isolation is still good to have, so the new process looks like
1. Read `build-system.requires` requirements from `pyproject.toml` and write them to a requirements file
2. Use our existing script to convert those pip requirements to conda requirements
3. Create and activate a build environment with conda
4. Build the wheel or sdist in the build environment
5. Switch to the test environment
6. Test the package in the test environment

Unrelatedly, I noticed that our github actions steps weren't failing as I'd expect. When one line in a step fails, it would continue on to the next line. This is the default behavior of a bash script, and we need to set the `-e` flag in order to exit on error. Setting the `-e` flag revealed a couple of undefined variables that the `check-sytanx-errors` job should've been catching.

## Checklist


setup.py
```diff
@@ -15,7 +16,8 @@
                  if (not req.startswith(('#', 'hg+', 'git+'))
                      and len(req.strip()) > 0)]
 
-include_dirs = [numpy.get_include(), 'src/natcap/invest/managed_raster']
+include_dirs = [numpy.get_include(),
+                os.path.join(pygeoprocessing.__path__[0], 'extensions')]
```
emlys, line 20: It would be nice for pygeoprocessing to provide its own `get_include` function. https://github.com/natcap/pygeoprocessing/issues/480

.github/workflows/build-and-test.yml
```diff
@@ -125,29 +146,15 @@ jobs:
...
+          NATCAP_INVEST_GDAL_LIB_PATH="$CONDA_PREFIX/Library" python -m build --wheel --no-isolation
           ls -la dist
 
-      # This produces a wheel that should work on any distro with glibc>=2.39.
-      # This is a very recent version. If we want to support older versions, I
-      # suspect we would need to build GDAL from source on an appropriate
-      # system (such as a manylinux docker container) to ensure compatibility.
-      # Symbols used in libgdal are the cause of the high minimum version,
-      # possibly because of installing with conda.
-      - name: Audit and repair wheel for manylinux
```
emlys, line 139: Stop packaging binary dependencies into the linux wheels. #2483

src/natcap/invest/spec.py
```diff
@@ -1008,11 +1008,11 @@ def check_value(value):
                 df[col][df[col].notna()].apply(check_value)
 
         if any(df.columns.duplicated()):
-            duplicated_columns = df.columns[df.columns.duplicated]
+            first_duplicated_column = df.columns[df.columns.duplicated][0]
             raise ValueError(validation_messages.DUPLICATE_HEADER.format(
-                header=header_type,
```
emlys, line 1013: Undefined variables that weren't being caught by the check-syntax job

.github/workflows/build-and-test.yml
```diff
@@ -314,10 +353,20 @@ jobs:
         with:
           fetch-depth: 0 # Fetch complete history for accurate versioning
 
+      - name: Create build requirements file
```
emlys, line 356: I was undecided about repeating this step in each job vs. tracking a new `requirements-build.txt` file. It's redundant to repeat the step each time, but I wanted to avoid having another file that's redundant to `pyproject.toml`.

.github/workflows/build-and-test.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
 env:
   # build: dependency of make install
   # setuptools_scm: needed for versioning to work
-  CONDA_DEFAULT_DEPENDENCIES: python-build setuptools_scm
+  CONDA_DEFAULT_DEPENDENCIES: python-build setuptools_scm gdal==3.10.3
```
emlys, line 22: Pin gdal here for now to ensure that the same `libgdal` is used in the build env and test env.

docker/Dockerfile
```diff
@@ -1,38 +1,36 @@
 # syntax=docker/dockerfile:1
 
-# Build the InVEST wheel in a separate container stage
-FROM debian:13 AS build
-RUN apt update && apt install -y python3 python3-dev python3-pip python3-build build-essential git python3.13-venv libgdal-dev
+# Create the container for distribution that has runtime dependencies.
+FROM mambaorg/micromamba:2.5-debian13-slim
```
emlys, line 4: I made some pretty significant changes to this Dockerfile - which could totally be going in the wrong direction. Since the wheel build now requires the `gdal` python package, I thought it made sense to consolidate the build into the same environment that the container runs. As a result there is now only one build stage, and everything is installed by micromamba. There could be a downside to this that I'm not aware of...

.github/workflows/build-and-test.yml
```diff
@@ -94,9 +94,29 @@ jobs:
           path: data/invest-test-data
           key: git-lfs-testdata-${{ hashfiles('Makefile') }}
 
-      - name: Set up python environment
+      - name: Create build requirements file
+        shell: python
+        run: |
+          import tomllib
+          with open('pyproject.toml', 'rb') as f:
```
phargogh: Not that we're likely to have anything other than text data in pyproject.toml, but is the `b` needed?  Wouldn't that interfere with text parsing/loading if we have, say, unicode data?

.github/workflows/build-and-test.yml
```diff
@@ -314,10 +353,20 @@ jobs:
...
+      - name: Create build requirements file
+        shell: python
+        run: |
+          import tomllib
+          with open('pyproject.toml', 'rb') as f:
+              pyproject_toml = tomllib.load(f)
+          with open('requirements-build.txt', 'w') as f:
+            for requirement in pyproject_toml['build-system']['requires']:
+              f.write(f'{requirement}\n')
+
```
phargogh: Since we have this same step written out 6 times across several of our files, would it be worth abstracting this into a separate step?  It's a pretty simple block of code, so I'm a little on the fence about it myself, but definitely curious what you think.

docker/Dockerfile
```diff
@@ -1,38 +1,36 @@
...
+
+ARG PYTHON_VERSION="3.13"
+
+ARG GDAL_VERSION="3.10"
```
phargogh, line 8: Is this gdal version defined here coupled to anything?  Previously the GDAL version was determined by the version in the debian repos,  but it seems like that constraint is lifted here and might even be able to be user-configured at container build time.  Is that true, or are there any constraints here we should consider when we eventually need to update the GDAL version we're compiling against?

phargogh (commented):
Thanks @emlys !  This really feels like a momentous day when pygeoprocessing is a compile-time dependency.  I just had a couple of questions that I don't think should get in the way of merging (hence why I didn't request changes), and curious what you think about them.

.github/workflows/build-and-test.yml
emlys: `tomllib` actually requires files to be read in binary mode! Apparently so it can enforce UTF-8 itself.

.github/workflows/build-and-test.yml
emlys: Oh good point. I didn't want to write a separate script for it, but it's pretty trivial to abstract it out as a reusable action!

docker/Dockerfile
emlys: I think it's independent of anything else, as long as the GDAL and python versions are compatible. The user could set a different value for `GDAL_VERSION` when they build the container.

phargogh (approved):
Thanks @emlys !  The changes look good, just waiting for the builds to finish and pass before merging.

PR #2521, opened 2026-04-23
Bugfix: plugin run logging

### LINKED ISSUE #2515: Workbench: plugin runs are not being logged
This has been a problem for a while. InVEST usage stats include 0 plugin runs in 2026 and just 5 in 2025.

Initial debugging indicates an attempt to log a plugin run fails on an uncaught KeyError when `ui_server.log_model_start` tries to look up the plugin's model ID in `models.model_id_to_pyname`. 

Further inspection shows `models.model_id_to_pyname` does not include any plugins. (Meanwhile, the Workbench's `settingsStore` does hold a list of all currently installed plugins.) Something must be going wrong in the construction of `models.model_id_to_pyname`.

I suspect the root cause here could be related to #2489 as well.

emilyanndavis:
## Description
Fixes #2515.

Specifically: version info is stripped from a plugin ID before attempting to log a model run.

I also added/updated tests and did some minor cleanup (removing unused imports, removing unused variables, adding/removing whitespace) in the files I was working in.

Note that further changes to how we keep track of plugins (e.g., see #2525) may be coming in the not-too-distant future. The goal of this PR is only to get plugin usage logging working again.

## Checklist

### DIFF
--- a/src/natcap/invest/datastack.py
+++ b/src/natcap/invest/datastack.py
@@ -30,14 +30,11 @@ import re
 import shutil
 import tarfile
 import tempfile
-import warnings
 
-from osgeo import gdal
-
-from . import spec
-from . import utils
-from . import models
-
+from natcap.invest import models
+from natcap.invest import spec
+from natcap.invest import utils
+from natcap.invest.utils import base_model_id
 
 LOGGER = logging.getLogger(__name__)
 DATASTACK_EXTENSION = '.invest.tar.gz'
@@ -142,7 +139,7 @@ def build_datastack_archive(args, model_id, datastack_path):
     """
     module = importlib.import_module(
         # For plugins, use the model id before the '@'
-        name=models.model_id_to_pyname[model_id.split('@')[0]])
+        name=models.model_id_to_pyname[base_model_id(model_id)])
 
     args = args.copy()
     temp_workspace = tempfile.mkdtemp(prefix='datastack_')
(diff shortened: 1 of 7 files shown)


src/natcap/invest/datastack.py
```diff
@@ -142,7 +139,7 @@ def build_datastack_archive(args, model_id, datastack_path):
     """
     module = importlib.import_module(
         # For plugins, use the model id before the '@'
-        name=models.model_id_to_pyname[model_id.split('@')[0]])
+        name=models.model_id_to_pyname[base_model_id(model_id)])
```
emilyanndavis, line 142: Replaced the explicit string conversion that was here with a call to the `base_model_id` function I added to `utils`.

All other changes in this file are just tidying up.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -201,7 +201,7 @@ function storePluginMetadataSync(
 
   // Write plugin metadata to the workbench's config.json
   logger.info('writing plugin info to settings store');
-  // Uniquely identify plugin by a hash of its ID and version
+  // Uniquely identify plugin by its ID and version
```
emilyanndavis, line 204: Updated this comment to reflect current reality (it was a hash initially, but it isn't anymore).

src/natcap/invest/ui_server.py
```diff
@@ -269,8 +269,9 @@ def build_datastack_archive():
 @app.route(f'/{PREFIX}/log_model_start', methods=['POST'])
 def log_model_start():
     payload = request.get_json()
+    modelID = base_model_id(payload['model_id'])
     usage._log_model(
-        pyname=models.model_id_to_pyname[payload['model_id']],
+        pyname=models.model_id_to_pyname[modelID],
```
emilyanndavis, line 274: This (together with `utils.base_model_id`) is the core of the bugfix!

davemfish (commented):
This looks good, @emilyanndavis , just wonder if it's useful for us to record this fix in HISTORY. Even though it's not a user-facing change, it could be nice for us to know which version of invest started properly logging plugin runs. And also just to know that something was fixed and needs to be released, for when we're deciding whether or not to do a bugfix release in May, or whenever.

davemfish (approved):

PR #2524, opened 2026-04-23
Use `open` instead of (deprecated) `codecs.open`

### LINKED ISSUE #2520: Replace `codecs.open` (deprecated) with `open`
While working in a file that uses `codecs.open`, I stumbled across a deprecation warning: 
```
The function "open" is deprecated
  Deprecated since Python 3.14. Use `open()` instead.
```

This is likely a simple find-and-replace-all task, but I'm not familiar enough with `codecs` to know why we might have been using `codecs.open` instead of `open` in the first place, so I figured I'd make an issue to make sure I'm not missing something!

emilyanndavis:
## Description
Fixes #2520 by replacing calls to `codecs.open` (deprecated in Python 3.14) with calls to `open`, and remove `codecs` import where no longer needed.

## Checklist

### DIFF
--- a/scripts/update-citation-cff.py
+++ b/scripts/update-citation-cff.py
@@ -1,5 +1,4 @@
 import argparse
-import codecs
 import os
 
 RELEASES_BUCKET = 'http://releases.naturalcapitalproject.org'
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Makes sense to me, thanks @emilyanndavis !  And I see that all checks are passing except for the one mac failure which looks to be unrelated.

PR #2526, opened 2026-04-23
Urban Mental Health report localization and testing

### LINKED ISSUE #2510: UMH report localization
Set up Urban Mental Health report for localization before we send out requests for translation

claire-simpson:
## Description
Localization for UMH report. Follows https://github.com/natcap/invest/pull/2452
This PR also adds a test module for the UMH report

Fixes #2510

## Checklist

### DIFF
--- a/tests/reports/test_urban_mental_health_template.py
+++ b/tests/reports/test_urban_mental_health_template.py
@@ -1,0 +1,111 @@
+import unittest
+from bs4 import BeautifulSoup
+
+from natcap.invest.urban_mental_health import MODEL_SPEC
+from natcap.invest.reports import jinja_env
+
+TEMPLATE = jinja_env.get_template('models/urban_mental_health.html')
+
+BSOUP_HTML_PARSER = 'html.parser'
+
+
+def _get_render_args(model_spec):
+    locale = 'en'
+    report_filepath = 'carbon_report_test.html'
+    invest_version = '987.65.0'
+    timestamp = '1970-01-01'
+    args_dict = {'suffix': 'test', 'lulc_base': ''}
+    img_src = 'bAse64eNcoDEdIMagE'
+    output_stats_table = '<table class="test__output-stats-table"></table>'
+    input_stats_table = '<table class="test__input-stats-table"></table>'
+    stats_table_note = 'This is a test!'
+    inputs_caption = ['input.tif:Input map.']
+    lulc_pre_caption = 'This is a test of the LULC broadcasting system!'
+    outputs_caption = ['results.tif:Results map.']
+    vegalite_json = '{}'
+    test_caption = 'This is another test!'
+    agg_results_table = '<table class="test__agg-results-table"></table>'
+    agg_map_source_list = ['/source/file.shp']
+
+    return {
+        'locale': locale,
+        'report_script': model_spec.reporter,
+        'invest_version': invest_version,
+        'report_filepath': report_filepath,
+        'model_id': model_spec.model_id,
+        'model_name': model_spec.model_title,
+        'model_description': model_spec.about,
+        'userguide_page': model_spec.userguide,
+        'timestamp': timestamp,
+        'args_dict': args_dict,
+        'agg_results_table': agg_results_table,
+        'inputs_img_src': img_src,
+        'inputs_caption': inputs_caption,
+        'outputs_img_src': img_src,
+        'outputs_caption': outputs_caption,
+        'intermediate_baseline_img_src': img_src,
+        'intermediate_baseline_img_caption': test_caption,
+        'intermediate_delta_ndvi_img_src': img_src,
+        'intermediate_delta_ndvi_img_caption': test_caption,
+        'intermediate_reclass_lulc_img_src': img_src,
+        'intermediate_reclass_lulc_img_caption': test_caption,
+        'raster_group_caption': test_caption,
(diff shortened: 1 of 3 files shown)


claire-simpson:
Tests are failing because `'locale' is undefined` in the SWY reports tests. I could go ahead and make that change here or wait for Megan's upcoming localization PR to be merged

tests/reports/test_urban_mental_health_template.py
```diff
@@ -0,0 +1,111 @@
...
+from natcap.invest.reports import jinja_env
+
+TEMPLATE = jinja_env.get_template('models/urban_mental_health.html')
+
+BSOUP_HTML_PARSER = 'html.parser'
+
+
+def _get_render_args(model_spec):
+    locale = 'en'
+    report_filepath = 'carbon_report_test.html'
```
emilyanndavis: Obviously it makes no real difference to the test itself, but I might rename this just to minimize confusion.

```suggestion
    report_filepath = 'urban_mental_health_report_test.html'
```

Maybe when we get around to solving the "too many repeated render args" problem, we'll come up with a shared solution for testing as well. 😃

tests/reports/test_urban_mental_health_template.py
```diff
@@ -0,0 +1,111 @@
...
+    invest_version = '987.65.0'
+    timestamp = '1970-01-01'
+    args_dict = {'suffix': 'test', 'lulc_base': ''}
+    img_src = 'bAse64eNcoDEdIMagE'
+    output_stats_table = '<table class="test__output-stats-table"></table>'
+    input_stats_table = '<table class="test__input-stats-table"></table>'
+    stats_table_note = 'This is a test!'
+    inputs_caption = ['input.tif:Input map.']
+    lulc_pre_caption = 'This is a test of the LULC broadcasting system!'
```
emilyanndavis, line 23: Haha, I'm such a dork! 🙃

emilyanndavis (changes requested):
Looks good, @claire-simpson, thanks!

I think the only things that remain untranslated are the `sum_cases` and `sum_cost` legend/tooltip labels in the AOI maps, but I don't see a way to make those translatable without some refactoring. If you can come up with a solution without too much trouble, then go for it; but if not, I think there's enough context otherwise (i.e., localized plot titles and captions) that it would be OK to leave `sum_cases` and `sum_cost` as they are.

Other than that, just one tiny (and oh so inconsequential) recommendation.

emilyanndavis:
Also, I just merged Megan's PR, so if you could pull in the latest, then we can double-check that all the tests pass.

tests/reports/test_urban_mental_health_template.py
claire-simpson: Yes great idea. It would be nice to make this a little less redundant!

tests/reports/test_urban_mental_health_template.py
claire-simpson: The exclamation points make it exciting 😆

claire-simpson:
> I think the only things that remain untranslated are the sum_cases and sum_cost legend/tooltip labels in the AOI maps, but I don't see a way to make those translatable without some refactoring. If you can come up with a solution without too much trouble, then go for it; but if not, I think there's enough context otherwise (i.e., localized plot titles and captions) that it would be OK to leave sum_cases and sum_cost as they are.

Hmm I'm not sure I understand why wrapping `attribute` in `gettext` like this wouldn't work?
```python
tooltip=[
            altair.Tooltip(f'{attribute}:Q', title=gettext(attribute), format=',.2f')
        ]
```

emilyanndavis:
> Hmm I'm not sure I understand why wrapping `attribute` in `gettext` like this wouldn't work?

I think the problem is that `gettext` does two things: first, it identifies specific strings that should be extracted into the message catalog; then, at runtime, it looks up strings (or references to strings) in the message catalog.

So, `gettext(attribute)` would work at runtime, if `attribute` referred to a string that already existed in the message catalog. But that string wouldn't exist in the message catalog, unless we also called `gettext` with the string literal, e.g., `gettext('sum_cases')`.

In this case, we need to be careful about wrapping `'sum_cases'` or `'sum_cost'` in `gettext`, since those are the names of attributes in the data (and we wouldn't want the code to attempt to access some nonexistent attribute because the attribute name has been translated at runtime). There's also perhaps a question of how one would translate `'sum_cases'` (or `'sum_cost'`), since it's sort of a shorthand form, and it's unclear whether the translation should be similarly abbreviated, needs to contain underscores, etc.

What I might do is map `'sum_cases'` and '`sum_cost'` to display names, define those display names with gettext, then pass the display names to the altair config. Something along these lines, perhaps (with actual display names and other details up to you, of course):

```
display_names = { 'sum_cases': gettext('Number of cases'), 'sum_cost': gettext('Total cost') }
…
altair.Tooltip(f'{attribute}:Q', title=display_names[attribute], format=',.2f')
```

emilyanndavis (approved):
Looks good, @claire-simpson, thank you!

claire-simpson:
Thanks for the suggestion and clarification regarding `gettext`! I took the suggestion of using `display_names`, and this brought up a bug which I made an issue for (I didn't fix it in this PR as its a bit unrelated to localization and brought up some other questions around reports formatting). Also, setting short legend titles makes the plots look fine: 
Screenshot 2026-04-27 at 4 37 26 PM
orter legend titles:

PR #2530, opened 2026-04-24
Align raster subplots by adding top padding even if there isn't a units subheading

### LINKED ISSUE #2471: Reports: adjacent raster subplots may appear different sizes if some rasters don't have `units`
## Background/Motivation
Side-by-side raster plots for rasters that have the same extent/resolution/dimensions/pixel size can appear different sizes if one raster has units metadata and the other doesn't. This is because a unit subtitle takes up space in the raster plot, effectively shrinking raster plots that have defined units. This occurs in `raster_utils.plot_raster_list`, where units text is added and axis y-limit is modified:

```python
units = _get_raster_units(raster_path)
if units:
        (ylim_kwargs,
         text_kwargs) = _get_units_text_kwargs(units, len(arr))
        ax.set_ylim(**ylim_kwargs)
        ax.text(**text_kwargs)
```

`_get_units_text_kwargs` extends the y-axis beyond the raster extent to make room for the units subtitle, so rasters without units metadata appear larger than raster _with_ units.

## Description
Side-by-side rasters should appear the same size regardless of whether one has units. There are a few options to solve this, including adding a `Units: None` subtitle if no units (either by default or as an option), or (optionally?) adjusting a subplot without units to match the dimensions/offset of subplots with a units subtitle, i.e., allowing us to set `ylim` top `subtitle_offset` to add some padding above the plot even if there aren't units.

## Additional context

Image

claire-simpson:
## Description
Adds padding to the top of a raster plot if _any_ raster in the `raster_list` has units (and more specifically, a units subheading). This ensures that rasters in adjacent subplots are the same size even if one raster in the list has a units subheading and another does not. If no raster in a group has `units`, no padding will be added to any of the subplots.

Fixes #2471

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -66,6 +66,12 @@
   Unreleased Changes
   ------------------
 
+General
+=======
+* To ensure consistent raster plot sizing in reports, padding is now added to
+  the top of raster plots if an adjacent raster plot has a units subheading.
+  (`#2471 <https://github.com/natcap/invest/issues/2471>`_)
+
 3.19.0 (2026-04-16)
 -------------------
 
--- a/src/natcap/invest/reports/raster_utils.py
+++ b/src/natcap/invest/reports/raster_utils.py
@@ -355,7 +355,7 @@ def _get_title_kwargs(title: str, resampled: bool, line_width: int, facets=False
     }
 
 
-def _get_units_text_kwargs(units: str, subtitle_offset: int):
+def _get_units_text_kwargs(units: str, subtitle_offset: float):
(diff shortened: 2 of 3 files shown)


davemfish (changes requested):
Thanks for addressing this, @claire-simpson . Would you mind adding (or modifying) a test to exercise this case of a mix of rasters with and without units? This fix is the sort of thing that we could accidentally "unfix" during some future refactor, so a test would help guard against that.

davemfish (approved):
Thanks for adding that test!

PR #2537, opened 2026-04-30
Make Input, Output, ModelSpec instances immutable.

### LINKED ISSUE #2228: ``spec.Input`` should be immutable (pydantic "frozen")
I managed to break invest by installing a plugin into my dev environment with the following bug:

Setting up the model_spec for my plugin:
```python
# I'd rather not support use of a results suffix in this plugin,
# but invest requires it be in the MODEL_SPEC
RESULTS_SUFFIX = spec.SUFFIX
RESULTS_SUFFIX.hidden = True

MODEL_SPEC = spec.ModelSpec(
    ....
    input_field_order=[
        ['workspace_dir', 'scenarios']],
    inputs=[
        spec.WORKSPACE,
        RESULTS_SUFFIX,
```

And then subsequent imports of any model, such as opening the model in the Workbench, raise this:
```
[1] pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelSpec
[1]   Value error, Input results_suffix is hidden but appears in input_field_order [type=value_error, input_value={'model_id': 'coastal_vul...h_cache/taskgraph.db')]}, input_type=dict]
```

davemfish:
Added a `Parent` class for shared config.

There's an interesting case during validation where we do intentionally modify the `required` attribute because it needs to be evaluated dynamically, before a subsequent call to validation. It seemed cleaner to override the immutability than to create copies of all the nested specs, update them, and copy the parent spec and update it with the new list of nested specs, before calling the parent's `validate` method.

This only came up for `Inputs` with nested specs, because for the `Input` itself, while it can have a `required` expression that needs evaluating, there is no need to update the original `required` value after evaluating the expression.

Fixes #2228

## Checklist


src/natcap/invest/spec.py
```diff
@@ -180,17 +180,28 @@ def validate_permissions_string(permissions):
     return permissions
 
 
-class Input(BaseModel):
+class Parent(BaseModel):
```
emlys: Could this have a more descriptive name like `ImmutableBaseModel`?

src/natcap/invest/validation.py
```diff
@@ -221,22 +215,25 @@ def validate(args, model_spec):
         # Extra args that don't exist in the MODEL_SPEC are okay
         # we don't need to try to validate them
         try:
-            # Using deepcopy to make sure we don't modify the original spec
-            parameter_spec = copy.deepcopy(model_spec.get_input(key))
+            # Using a deep copy to make sure we don't modify the original spec
+            # spec or original nested specs.
```
emlys: Is "spec" repeated an extra time in this sentence?

src/natcap/invest/spec.py
```diff
@@ -973,7 +971,9 @@ def get_validated_dataframe(self, csv_path: str, read_csv_kwargs={}, args=None):
         columns = copy.deepcopy(self.columns)
         for col_spec in columns:
             if isinstance(col_spec.required, str):
-                col_spec.required = bool(utils.evaluate_expression(
+                # attributes have faux immutability; we cannot assign
+                # to them, but we can assign to the underlying dict.
+                col_spec.__dict__['required'] = bool(utils.evaluate_expression(
```
emlys: Nice to know that we can get around immutability like this if needed! Would it be worth implementing some sort of copy-with-update function to create a new instance with the changed attribute? Given that it's only used in one place, maybe it's not worth it, but it would be nice to respect immutability if possible.

emlys (commented):
Thanks for tackling this @davemfish! I had avoided the immutability question myself when first developing this because of the difficulty with evaluating the `required` strings. Just a few comments

src/natcap/invest/spec.py
davemfish: Pydantic provides `model_copy(update={})` so it's probably worth using in this spot. I made that update. In the similar, but more complicated, situation in `validation.py` it did not seem worth it to copy and replace all the objects because of how they are nested.

src/natcap/invest/spec.py
```diff
@@ -1008,11 +1008,11 @@ def check_value(value):
...
         if any(df.columns.duplicated()):
-            duplicated_columns = df.columns[df.columns.duplicated]
+            duplicated_columns = df.columns[df.columns.duplicated()]
             raise ValueError(validation_messages.DUPLICATE_HEADER.format(
-                header=header_type,
-                header_name=expected,
-                number=count))
+                header=f'{self.orientation}(s)',
+                header_name=','.join(duplicated_columns),
+                number='multiple'))
```
davemfish, line 1018: This block is unrelated to the rest of the PR, but I noticed these variables were not defined. I think this code basically never runs because if pandas reads a CSV with duplicate column names, it adds a suffix to the names as needed. I think the only way to get here is when the orientation is "rows", but then we transpose the table and rows become columns, then there can be duplicate columns at that point. 

There can be multiple duplicate columns, so it makes for a complicated message to try to describe the count of each duplicate. I think it's more helpful to just have the message read:

`'Expected the row(s) "1, 2" only once but found it multiple times'`
or
`'Expected the row(s) "1" only once but found it multiple times'`

Still not ideal, but this message is also used elsewhere so I didn't want to change the template.

emlys (approved):

PR #2539, opened 2026-04-30
Add user feedback on attempt to load datastack with unrecognized model ID

### LINKED ISSUE #2489: Workbench: uncaught error when attempting to load a datastack with unrecognized model_id
## Description
This issue came up while updating the SDR USLE C Raster plugin: [Loading params from sample datastack fails silently](https://github.com/natcap/invest-sdr-usle-c-raster/issues/13). The `model_id` in the plugin's sample datastack did not include version information; therefore, it could not be loaded and triggered an uncaught error when the Workbench attempted to generate an error message. (See [SetupTab.loadParametersFromFile](https://github.com/natcap/invest/blob/main/workbench/src/renderer/components/SetupTab/index.jsx#L289).) While this particular scenario is an edge case, it's not hard to imagine a plugin user inadvertently attempting to load a datastack with a plugin version that is different from the version they have installed.

At minimum, we should add a fallback so that the Workbench doesn't attempt to access the `modelTitle` of an object that does not exist. I think we could easily go a step further and make the error message more explicit, perhaps including both model IDs, so it's clear where the mismatch is and how it might be resolved.

emilyanndavis:
## Description
Fixes #2489. Specifically, attempting to load a datastack/logfile with an unrecognized model ID (via any Workbench-provided method) now presents an error message instead of failing silently.

### Plugin ID missing version info, via setup tab
datastack_error-plugin_id_without_version-2026_04_30

### Plugin ID missing version info, via home tab
datastack_error-plugin_id_without_version-home_tab-2026_04_30

### Malformed model ID (kebab case instead of snake case), via setup tab
datastack_error-hyphenated_model_id-2026_04_30

### Malformed model ID (kebab case instead of snake case), via home tab
datastack_error-hyphenated_model_id-home_tab-2026_04_30

### Other unrecognized model ID, via setup tab
datastack_error-unrecognized_model-2026_04_30

### Other unrecognized model ID, via home tab
datastack_error-unrecognized_model-home_tab-2026_04_30

## Checklist


davemfish (approved):
Great fix!

PR #2540, opened 2026-05-01
Reports: styling & other usability improvements to Vega-Lite visualizations

### LINKED ISSUE #2532: Reports: vector plot overflows container and format doesn't match raster plot
## Description
While working on #2526, I tested using a longer legend title in a vector plot, which overflowed its container:

Image
 
While debugging this, I noticed other semi-related formatting inconsistencies, where our vector plot behavior doesn't match the raster plot behavior:
1. Rasters shrink proportionally upon resizing the report window in the workbench vs. vector plots have a scroll bar. I think the raster behavior is better, but I couldn't get the default vega `canvas` type to replicate this behavior. Along with using the `svg` format (i.e., setting `vegaEmbed` to have `renderer: 'svg'`), the following code allowed the vector plots to shrink appropriately + fixed the legend title overflow issue, _however_ a long legend title made the plot much smaller (see image below).
```
.vega-fit-horizontal {
    width: 100%;
    max-width: 100%;
    min-width: 0;
    overflow-x: auto;
    text-align: center;
  }
  .vega-fit-horizontal .vega-embed svg {
    max-width: 100%;
    height: auto;
    display: block;
  }
```

Image

My CSS styling might need improvement, but conceptually how do we want vector plots to scale and flex with different window sizes and different legend/legend title widths? If there was a way to force the legend to maintain a consistent width across subplots that could be a good way to go. 

2. Our raster plots are usually (though not always) titled with their filename whereas the vector plots are titled with more human readable name. 
3. A "legend" in a raster plot is simply the colorbar; the vector plot legends are titled. Do vector plot legends need a title (especially given that the tooltips we're using describe the same thing)?
4. The captions for rasters are stacked vertically vs. vector captions are directly underneath the plot they describe (constrained horizontally). Was this an intentional decision (because the vector plots aren't titled with their filename so there isn't an easy way to reference which figure the caption refers to?) or should we make this consistent?
Image

Image

Aside from the initial legend title overflow bug I noted, I don't think any of this formatting is necessarily bad, but I did wonder if we want to make the vector and raster plots a bit more consistent (and maybe add some of this to the report author guide wiki?).

emilyanndavis:
## Description
Fixes #2532. Specifically, Vega-Lite visualizations (e.g., vector plots) are now styled to prevent them from overflowing their containers or overlapping one another. With this change, the `.vega-fit-horizontal` styles no longer appear to be needed, so I removed them.

I also snuck in a few other small improvements, namely:
- All vegaEmbed elements now use the SVG renderer instead of the default canvas renderer.
- The default filename for "Save as PNG" or "Save as SVG" downloads via the Vega actions menu is now `InVEST-{elementId}`, where the `elementId` is the `id` of the HTML element that contains the visualization. This ID is defined by the report author and is typically a short name that identifies the purpose of the visualization, e.g., `cases_map`, `habitat_map`, or `exposure_histogram`.
- The Vega actions menu button now has slightly higher contrast to make it easier for people to discover it.

### Vector plots scale down at smaller viewport widths
vector_plots_responsive-2026_05_01

### New default filename when saving an SVG or PNG
vega_download_filename-2026_05_01

### Vega actions menu button, before
vega_actions_button-before-2026_05_01

### Vega actions menu button, after
vega_actions_button-after-2026_05_01

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -78,6 +78,9 @@ General
 * To ensure consistent raster plot sizing in reports, padding is now added to
   the top of raster plots if an adjacent raster plot has a units subheading.
   (`#2471 <https://github.com/natcap/invest/issues/2471>`_)
+* Made some styling and usability improvements to the Vega-Lite visualizations
+  (e.g., vector plots, histograms) that appear in some InVEST reports.
+  (`#2532 <https://github.com/natcap/invest/issues/2532>`_)
 
 3.19.0 (2026-04-16)
 -------------------
--- a/src/natcap/invest/reports/templates/models/urban_mental_health.html
+++ b/src/natcap/invest/reports/templates/models/urban_mental_health.html
@@ -32,25 +32,25 @@
   )}}
 
   {% set cases_map_block = content_grid([
-    ('<div class="vega-fit-horizontal"><div id="cases_map"></div></div>', 100),
+    ('<div id="cases_map"></div>', 100),
     (caption(cases_map_caption, aggregate_map_source_list), 100)
   ]) %}
-  
+
   {% set cases_cost_items = [
     (cases_map_block, 100 if not cost_map_json else 50)
   ] %}
-  
+
   {% if cost_map_json %}
     {% set cost_map_block = content_grid([
-      ('<div class="vega-fit-horizontal"><div id="cost_map"></div></div>', 100),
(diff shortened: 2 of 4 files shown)


claire-simpson (approved):
Thanks @emilyanndavis! I tested this locally and it looks good

PR #2541, opened 2026-05-01
Consolidate wheel and sdist build jobs

### LINKED ISSUE #2514: Can we consolidate the wheel and sdist build steps?
I noticed that in our sdist build step, when we install the sdist, it just builds a wheel from that sdist and then installs the wheel. This is the standard behavior. Can we then consolidate the sdist and wheel build into one job? 

The default behavior of `python -m build` is to build an sdist, and then build a wheel from that sdist. Uploading both artifacts from one job would eliminate some redundant steps and save on Github Actions runtime.

emlys:
## Description
Fixes #2514

This PR removes the separate sdist build job and consolidates it into the test job, where we already build and upload a wheel. The failing puppeteer test seems unrelated.

## Checklist


.github/workflows/build-and-test.yml
```diff
@@ -148,14 +149,25 @@ jobs:
...
           make test
 
+      - name: Check long description with twine
+        run: twine check $(find dist -name "natcap[._-]invest*")
+
+      - name: Upload sdist artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: Source distribution for ${{ matrix.os }} ${{ matrix.python-version }}
+          path: dist/*.tar.gz
```
dcdenu4: Does this allow us to upload multiple source distributions without having to worry if they're being overwritten? Re: a comment we had below
```
# Only upload sdist in one of the matrix cases so we don't
# overwrite artifacts or have duplicates (mac/windows sdists have
# different extensions)
```

dcdenu4 (changes requested):
Thanks @emlys ! Looks like we have a merge conflict and otherwise I just had a question about now uploading multiple source distributions and whether we need to worry about them being overwritten.

.github/workflows/build-and-test.yml
emlys: Oh, good point. So we've been uploading just one sdist to the GCP bucket, but _all_ the sdists as github actions artifacts. Sdists should work the same on any OS or python version, so I think we can drop the extra artifact uploads as well.

emlys: I changed this to upload only one sdist to GCP and github.

emlys: This simplifies the release workflow a little, too, since we don't have to choose an arbitrary one of multiple sdist artifacts to use.

dcdenu4 (approved):
Thanks @emlys !

PR #2544, opened 2026-05-06
Workbench: Include docs link in plugin modal

### LINKED ISSUE #2145: Improve the visibility of the Plugin feature
From a Slack thread:

> I'd like to include a link to the Workbench plugin feature in a response to someone on the forum. But i'm finding out how hard it is to actually encounter plugin information. It's not in the User Guide, nor anywhere obvious on our web site (like in any of the Software pages). Eventually i found [this Developer's Guide link](https://invest.readthedocs.io/en/latest/plugins.html), but only by clicking on API documentation on the web site, then seeing the Plugins link on the Readthedocs page. We really need to get this info more easily accessible!

> If i search our website, i do find the related news release and video, but personally news releases are not the first (or second) place i'd go to look for technical information. 

> Looking at the news release again, there's a lot of good info in there. Could we link to it in an obvious way from the software/invest page?

Will the news release be there in perpetuity, or does it fade quickly? If the latter, one idea could be to turn it into a non-news page for plugins, since the info in there is so good as is.

I'd also recommend adding a section to the User Guide, even if it just provides a few-sentence intro to plugins then points to the Developer's Guide for now. Heck, i could probably do that, lifting text from the news release and/or Developer's Guide.

~ Stacie

megannissel:
## Description
Fixes #2145

Link the plugin developer docs from the Workbench "manage plugins" modal, to make it easier for folks to find more information. 

I debated whether to put this link at the top or bottom of the modal. I landed on the bottom, but really don't have a strong opinion and would be fine with switching it. Screenshots for comparison:
Screenshot 2026-05-06 at 2 44 12 PM
Screenshot 2026-05-06 at 2 43 12 PM 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -252,6 +252,7 @@ export default function PluginModal(props) {
     );
   }
 
+  const pluginsURL = "https://invest.readthedocs.io/en/latest/plugins.html";
   let modalBody = (
     <Modal.Body>
       <Form aria-labelledby="add-plugin-form-title">
@@ -398,7 +399,9 @@ export default function PluginModal(props) {
         <p>
           {t('For more information about plugins, read our ')}
           <a
-            href="https://invest.readthedocs.io/en/latest/plugins.html"
+            href={pluginsURL}
+            title={pluginsURL}
+            aria-label={t("Plugins Developer's Guide (opens in web browser)")}
             onClick={openLinkInBrowser}
           >{t("Developer's Guide")}</a>.
         </p>
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -9,6 +9,7 @@ import Spinner from 'react-bootstrap/Spinner';
 import { useTranslation } from 'react-i18next';
 import { MdCheckCircleOutline, MdClose } from 'react-icons/md';
 
+import { openLinkInBrowser } from '../../utils';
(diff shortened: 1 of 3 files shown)


workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -392,6 +393,16 @@ export default function PluginModal(props) {
           }
         </div>
       </Form>
+      <hr />
+      <div>
+        <p>
+          {t('For more information about plugins, read our ')}
+          <a
+            href="https://invest.readthedocs.io/en/latest/plugins.html"
+            onClick={openLinkInBrowser}
```
davemfish, line 405: Thanks, @megannissel , just for thoroughness, would you mind adding a `title` and `aria-label` attributes as well? I think those enable tooltips and screenreaders, respectively. Here's an example from our link to the Forum:

```javascript
      <a
        href={forumURL}
        title={forumURL}
        aria-label={t('Frequently Asked Questions (opens in web browser)')}
        onClick={openLinkInBrowser}
      >
        <MdOpenInNew className="mr-1" />
        {t('Frequently Asked Questions')}
      </a>
```

davemfish (changes requested):

workbench/src/renderer/components/PluginModal/index.jsx
megannissel: Ah, good call, thanks!

I'd been looking at the GeoMetaMaker link as a reference, which didn't include these attributes. For the sake of completeness, I went ahead and added them there as well.

davemfish (approved):
Thanks for these updates!

PR #2552, opened 2026-05-19
SDR: handle topology exception in `_geometries_overlap`

### LINKED ISSUE #2386: handle exceptions from `_geometries_overlap`
`urban_nature_access._geometries_overlap` is a utility for checking whether or not geometries within a polygon vector overlap with one another. Knowing that can help optimize rasterization during zonal_stats calculations, for example.

But there could be cases where imperfect geometries raise a `TopologyException` from this function that would not otherwise be a problem during rasterization. In cases like SDR, where `_geometries_overlap` is used for optimization, I think we should handle such exceptions and at least try to do the zonal_stats without knowing whether geometries overlap or not.

https://community.naturalcapitalalliance.org/t/error-in-sdr-invest-3-17/6333

```python
2026-01-26 10:30:36,584 (natcap.invest.utils) utils.prepare_workspace(173) ERROR Exception while executing sdr
  File "natcap\invest\utils.py", line 171, in prepare_workspace
  File "natcap\invest\spec.py", line 2194, in execute
  File "natcap\invest\sdr\sdr.py", line 1005, in execute
  File "taskgraph\Task.py", line 684, in add_task
  File "taskgraph\Task.py", line 1103, in _call
  File "natcap\invest\sdr\sdr.py", line 1574, in _generate_report
  File "natcap\invest\urban_nature_access.py", line 1578, in _geometries_overlap
  File "shapely\ops.py", line 117, in unary_union
  File "shapely\decorators.py", line 173, in wrapper
  File "shapely\decorators.py", line 88, in wrapped
  File "shapely\set_operations.py", line 553, in union_all
shapely.errors.GEOSException: TopologyException: side location conflict at 105958.42278252472 1827072.958573573. This can occur if the input geometry is invalid.
```

davemfish:
The `urban_nature_access._geometries_overlap` is a utility that can raise an exception on invalid geometries. The way it is used by SDR, we don't need to exit the program if there are invalid geometries.

I considered adding the exception-handling within the function itself, but depending on how the function is used, it may not always make sense to handle the exception in the same way.

In `urban_nature_access` I updated a couple places where the raw AOI input file was read for checking geometries, and for getting a list of fields, instead of reading from the reprojected version, which is the version actually used in subsequent geoprocessing.

Fixes #2386

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -92,6 +92,14 @@ Workbench
 * Added a link to the InVEST Plugin Developer's Guide to the Workbench Manage
   Plugins modal. (`#2145 <https://github.com/natcap/invest/issues/2145>`_)
 
+SDR
+===
+* Added exception-handling when checking if watershed geoemtries overlap.
+  Invalid geometries will no longer raise an exception during this check
+  because they can still be rasterized and used in zonal_statistics.
+  (`#2386 <https://github.com/natcap/invest/issues/2386>`_)
+
+
 3.19.0 (2026-04-16)
 -------------------
 
--- a/src/natcap/invest/sdr/sdr.py
+++ b/src/natcap/invest/sdr/sdr.py
@@ -15,6 +15,7 @@ import pygeoprocessing
 import pygeoprocessing.routing
 from osgeo import gdal
 from osgeo import ogr
+from shapely.errors import GEOSException
 
 from natcap.invest import gettext
 from natcap.invest import spec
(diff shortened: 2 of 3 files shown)


davemfish:
I'm not sure what the deal is with the Windows binaries. It is failing to download or move a micromamba.exe. It's unrelated to these changes.

claire-simpson (commented):
Thanks @davemfish, this looks good. Would we want to consider moving `_geometries_overlap` to `utils.py` since it is being used outside of UNA?

davemfish:
> Thanks @davemfish, this looks good. Would we want to consider moving `_geometries_overlap` to `utils.py` since it is being used outside of UNA?

Yes, it's worth considering. I imagine whoever used it in SDR might have considered that question at the time. This PR isn't adding any additional uses, so I don't necessarily want to override that previous decision. As a rule of thumb, if we were using a function in three different places, then it would be worth putting it in `utils`. So far it has two uses.

claire-simpson (approved):

PR #2554, opened 2026-05-20
Add a starter .gitignore

### LINKED ISSUE #258: Do we still need to .hgignore?
Are we done with mercurial and this tracked ignore file?

phargogh:
This PR adds a `.gitignore` file with common InVEST-created files/folders excluded, as well as things related to development environments (vscode, sublime, vim/nvim).

This also adds a `.dockerignore` that has helped me speed up docker builds locally that I had forgotten about in the sea of untracked files in `git status`.

Fixes #258 

## Checklist


.gitignore
```diff
@@ -0,0 +1,113 @@
...
+*.tmlanguage.cache
+*.tmPreferences.cache
+*.stTheme.cache
+
+# Workspace files are user-specific
+*.sublime-workspace
+
+# Project files should be checked into the repository, unless a significant
+# proportion of contributors will probably not be using Sublime Text
+# *.sublime-project
```
davemfish: I think we should remove this block lest a contributor thinks it's a good idea to check-in a `.sublime-project` file.

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+!.vscode/launch.json
+!.vscode/extensions.json
+!.vscode/*.code-snippets
+!*.code-workspace
+
+# Built Visual Studio Code Extensions
+*.vsix
+
+# Files related to Sublime Text
+# Adapted from  https://github.com/github/gitignore/blob/main/Global/SublimeText.gitignore
```
davemfish: There's a lot of Sublime stuff here that I don't have. Perhaps the locations of these files are dependent on installation preferences, but the only files that end up my repo directory are `*.sublime-workspace` and `*.sublime-project`.

I'm in favor of a comprehensive ignore file in general, but given how few people are using Sublime, maybe we don't need to be so comprehensive here, at first. What do you think @emlys ?

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+!.vscode/*.code-snippets
+!*.code-workspace
+
+# Built Visual Studio Code Extensions
+*.vsix
+
+# Files related to Sublime Text
```
dcdenu4: For OCD reasons, should there be a line of `###` above here signifying a new section from VScode to Sublime?

dcdenu4 (changes requested):
Thanks @phargogh! just one nitpick.

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+doc/api-docs/api
+build/
+dist/
+coverage.xml
+cover/
+*.egg-info/
+env
+requirements-all.yml  # Generated in make env
+workbench/build
+workbench/node_modules/
```
davemfish, line 26: I also have,
* `workbench/coverage/`  - from jest coverage reports
* `workbench/resources/storage_token.txt` - created by `yarn run token` as part of the workbench build process. The other files in `resources` are tracked in the repo
* `workbench/dist/` - though oddly, the files I have here are already being ignored 🤔

davemfish (changes requested):

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+.pypirc
+_build/
+.eggs/
+.ipynb_checkpoints/
+.pytest_cache/
+.coverage
+__pycache__/
+*.so
+*.dylib
+
```
davemfish, line 58: Also `*.pyd`? `

`.cpp` is interesting in our case where they are generated by Cython, but it would be weird to ignore all `.cpp` files, right?

.gitignore
emlys: I've actually never seen any Sublime files come up in my repo... it must be saving them somewhere else. In favor of removing this section

.gitignore
emlys: Yeah, we could ignore `.cpp` in specific directories, but maybe not the whole project. We have one C++ file (`ManagedRaster.h`) that _should_ be a `.cpp`, but for some reason, it was significantly slower when split into a separate header and source file. So everything is in the header file for now... but I should probably revisit that.

.gitignore
phargogh: Yes!  This was an oversight.  Patched in 82de95555

.gitignore
phargogh: Definitely seems like this sublime text content is opinionated!

@emlys to confirm, you're in favor of removing the entire sublime text section?

.gitignore
phargogh: Uffda, thanks for catching this.  Removed in 3642a81b0

.gitignore
phargogh: Added in 289427417.

Not sure about why `workbench/dist` is being ignored, except maybe the `dist/` pattern applies to nested subdirs of the same name?

emilyanndavis: Yes, `build` and `dist` will catch any `build` or `dist` directories, regardless of where they are in the tree. It doesn't hurt to explicitly list `workbench/build` and `workbench/dist`, but they are technically redundant.

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+#################################################################################
+# Files related to VS Code
+# Adapted from https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+!.vscode/*.code-snippets
+!*.code-workspace
```
emilyanndavis: Since our project doesn't rely on VSCode-specific tools or workflows, I think we can safely ignore everything in the `.vscode` directory.

```suggestion
.vscode
```

.gitignore
phargogh: Yes, we should also be ignoring `*.pyd`, and I ignored C++ files in the directories where we have cython files.  Patched 2d7617f30

.gitignore
```diff
@@ -0,0 +1,113 @@
...
+
+# Built Visual Studio Code Extensions
+*.vsix
```
emilyanndavis, line 65: I've never seen this—I wonder if this type of file is generated when you're developing an extension? We may not need it, but I don't think it hurts to leave it, just in case.

.gitignore
davemfish: I'm fine with removing the entire sublime section too

phargogh: Removed in 4e4ea6ce8

.gitignore
```diff
@@ -0,0 +1,87 @@
+#################################################################################
+# InVEST .gitignore
+#
+# This file isn't intended to match all possible files.  Please consider using
+# a personal .gitignore to augment this based on your own environment.
+# Personal, custom .gitignore files can be placed at a few different locations
+# for example:
+#    * ~/.config/git/ignore  # for your global .gitignore
+#    * ~/.gitignore_global   * alternate location for global .gitignore
+#    * ./.git/info/exclude   # for your personal ignore rules for this repo
```
emilyanndavis: It doesn't matter, since this is all commented out, but I just noticed there's a `*` where I think you meant `#`. And while I'm at it, I'll recommend a couple of other punctuation updates:

```suggestion
# This file isn't intended to match all possible files. Please consider using
# a personal .gitignore to augment this based on your own environment.
# Personal, custom .gitignore files can be placed at a few different locations,
# for example:
#    * ~/.config/git/ignore  # for your global .gitignore
#    * ~/.gitignore_global   # alternate location for global .gitignore
#    * ./.git/info/exclude   # for your personal ignore rules for this repo
```

davemfish (approved):

.gitignore
phargogh: Thanks for catching these, @emilyanndavis !

emilyanndavis (approved):
Thanks for spearheading this, @phargogh !

emlys (approved):

dcdenu4 (approved):

dkav:
Some further suggestions :

- Exclude `messages.mo` (build artifact of `messages.po` source files).
- Instead of having multiple entries for `.cpp` files just exclude them all with `src/natcap/invest/**/*.cpp`.
- Use `env/` instead of `env`. Explicitly ignore the env folder, not an env folder or file.
- Remove non-project related ignores, i.e. tooling and OS specific. Leave them for the global ignore.

PR #2558, opened 2026-05-20
Make sure GDAL_DATA env var is set

### LINKED ISSUE #2555: pyinstaller binaries produce warnings about `GDAL_DATA`
On Windows, pyinstaller binaries produce warnings about `GDAL_DATA`:
```
osgeo              WARNING  [errno 3] Cannot find tms_NZTM2000.json (GDAL_DATA is not defined)
```

There are a couple problems with pyinstaller hooks:

1. Our custom `exe/hooks/hooks-osgeo.py` overrides the `pyinstaller-hooks-contrib/stdhooks/hook-osgeo.py` such that the data that should be collected by the contrib hook does not get included in the bundle.
2. Removing our custom hook solves the first problem, but there appears to be a bug in `pyinstaller-hooks-contrib/rthooks/pyi_rth_osgeo.py` such that `GDAL_DATA` never gets set. There is an explicit check for `gcs.csv` in the data folder, to see if the correct folder was found. [But `gcs.csv` is no longer shipped by GDAL since the PROJ6 integration](https://github.com/OSGeo/gdal/issues/1471#issuecomment-486078468).

I will file a bug about the second problem: https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/1020

Regarding the first problem, I'm not sure if need the custom hook or not. If we do, then I guess we will need to add logic to it collect the data.

davemfish:
This env var should be set by the osgeo runtime hook from `pyinstaller-hooks-contrib`, but there is a bug that was leaving it unset on Windows. We can set it in our own runtime hook until that upstream issue is fixed.

The path to data folder itself is determined by `pyinstaller-hooks-contrib/std/hook-osgeo.py`, and while there are couple different places where that folder could end up in the binary bundle, I think we only ever build and package based on conda/mamba environments, so that is the `os.path.join('Library', 'share', 'gdal')` path.

I also removed our custom `hook-osgeo.py` because its presence seems to prevent `pyinstaller-hooks-contrib/std/hook-osgeo.py` from running at all, which meant the GDAL data was never included in the bundle _and_ the env var was not being set.
Removing our hook did not seem to have any negative consequences, all the `*.pyd` files were still collected and included.

Fixes #2555 

I built the binaries locally and tested the `invest` CLI.

## Checklist


phargogh (approved):
Thanks @davemfish !  The changes make sense to me, and the one test failure appears unrelated to the changes (and is consistent with other recent mac build failures).

PR #2566, opened 2026-05-22
GHA maintenance: fix `deploy_reports`; replace `gsutil`

### LINKED ISSUE #2508: GHA: Some model reports from `invest-autotest` are missing
Lately when we run `invest_autotest` on `main`, only two reports are being deployed to GCS (Carbon and Coastal Vulnerability). A few weeks ago we were also getting the NDR and SDR reports. And currently, there should be 6 reports (SWY and UMH were added).

The autotest script does not show an exceptions from any models. But the subsequent deploy step only finds the two reports in the model workspaces. When I run the autotest locally and use the same `find` command as used in `make deploy_autotest_reports`, I see all 6 reports exist.

https://github.com/natcap/invest/actions/runs/24508128442/job/71631874365

### LINKED ISSUE #2562: remove `make sampledata_single` from our build workflow
I don't think we have a need for `make sampledata_single`, which collects all the sample data into a single zip file. We currently build that zip as part of `build-sampledata` in `build-and-test.yml`, but the zipfile never gets deployed anywhere.

The zip file is written to `dist/`, but since #2448 and #2561 we do not deploy everything in `dist/*`, and `make deploy_data` deploys only what is in `dist/data/`.

We no longer need to create this zipfile. It was formerly used by the NSIS installer, but since the Workbench we do not use it, even if a user selects to download all sample data.

davemfish:
This PR replaces all uses of `gsutil` in our `Makefile` and Actions workflows with `gcloud storage`. `gsutil` is considered legacy.

That change fixed the problem with deploying reports after running the invest-autotest.

All of these changes are running "on-push" on my fork for testing because many steps will not run in the PR context. And I temporarily set the autotest steps to run there as well:
https://github.com/davemfish/invest/actions/runs/26300658559

Fixes #2508
Fixes #2562 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ PIP = $(PYTHON) -m pip
 VERSION := $(shell $(PYTHON) -m setuptools_scm)
 PYTHON_ARCH := $(shell $(PYTHON) -c "import sys; print('x86' if sys.maxsize <= 2**32 else 'x64')")
 
-GSUTIL := gsutil
+GSUTIL := gcloud storage
 SIGNTOOL := SignTool
 RST2HTML5 := rst2html5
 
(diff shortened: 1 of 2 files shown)


davemfish:
Let's squash this one on merge because there a bunch of messy debugging commits in there.

Makefile
```diff
@@ -186,7 +185,7 @@ invest_autotest: $(GIT_SAMPLE_DATA_REPO_PATH) $(INVEST_BINARIES_DIR)
 	$(INVEST_AUTOTESTER)
 
 deploy_autotest_reports:
-	find $(AUTOTEST_DIR) -name "*report*.html" | $(GSUTIL) -m cp -I $(REPORTS_BASE_URL)
+	find $(AUTOTEST_DIR) -name "*report*.html" | gcloud storage cp -I $(REPORTS_BASE_URL)
```
phargogh: Should this use the makefile var?

```suggestion
	find $(AUTOTEST_DIR) -name "*report*.html" | $(GCLOUD_STORAGE) cp -I $(REPORTS_BASE_URL)
```

phargogh (commented):
Thanks @davemfish !  Just one suggestion about using the new Makefile variable, but looks great otherwise!

Makefile
davemfish: Yes, good catch, thank you.

phargogh (approved):
Thanks @davemfish !

phargogh:
I figured I'd wait until the tests finished to confirm everything else is working

PR #2569, opened 2026-05-22
Urban Nature Access Report

### LINKED ISSUE #2534: Report for Urban Nature Access

claire-simpson:
## Description
Add report for UNA

Report Examples:
1. [With radius per population group](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_nature_access_report_POP_052826.html)
2. [With radius per urban nature class](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_nature_access_report_LULC_052826.html)
3. [Uniform radius](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_nature_access_report_UNIFORM_052826.html)

Note: localization not done as this will be PR'ed separately into the `feature/reports-localization` branch

Fixes #2534

## Checklist
- [ ] Updated the user's guide (if needed) --> Will do after report is finalized and can be uploaded to final GCS bucket


src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+
+def _create_vector_maps(
+        geodataframe,
+        xy_ratio):
+    """Create choropleth maps and return one Vega JSON spec."""
+
+    display_names = {
+        'SUP_DEMadm_cap': (
+            gettext('Balance'),
+            gettext('Per Person Average Urban Nature Supply/Demand Balance by Admin Unit')),
```
davemfish, line 54: Is any of the text in the model spec useful for these descriptions? If the spec is lacking, is it worth updating it with the text you have here and using it as the source?

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+            raster_path=file_registry['urban_nature_balance_totalpop'],
+            spec=model_spec.get_output('urban_nature_balance_totalpop'),
+            datatype=RasterDatatype.divergent,
+            transform=RasterTransform.linear,
+            special_values=SpecialValueConfig(
+                extend="both",
+                threshold=(-1000, 1000),
+                label=('<-1000', '>1000'),
+                color=('#8C4300', '#1F0737')
+            )
```
davemfish: The `balance` maps here have a different transform than the demographic-specific `balance` maps (log transformed). I would expect their distributions to be similar so I would think they should get the same transformation. What do you think?

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+                datatype=RasterDatatype.continuous,
+                spec=model_spec.get_output('urban_nature_population_ratio')
+            ),
+        ]
+        intermediates_img_src = raster_utils.plot_and_base64_encode_rasters(
+            intermediate_raster_config_list)
+        intermediates_caption = raster_utils.caption_raster_list(
+            intermediate_raster_config_list)
+    else:
+        intermediates_img_src = intermediates_caption = None
```
davemfish: I think it's slightly safer and more readable to initialize variables like this to `None` at the top level, before the `if` statement. Then there is one fewer code block, and it's clearer that these variables can never go undeclared.

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+        RasterPlotConfig(
+            raster_path=file_registry['urban_nature_balance_percapita'],
+            spec=model_spec.get_output('urban_nature_balance_percapita'),
+            datatype=RasterDatatype.divergent,
+            transform=RasterTransform.linear,
+            special_values=SpecialValueConfig(
+                extend="both",
+                threshold=(-1000, 1000),
+                label=('<-1000', '>1000'),
+                color=('#8C4300', '#1F0737')
```
davemfish: Is there a convenient way to get these colors from the actual colormap that will be used in this plot? If not, let's at least comment where these come from.

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+                        spec=model_spec.get_output(
+                            'accessible_urban_nature_lucode_[LUCODE]'),
+                    )
+                )
+        output_accessible_raster_caption = raster_utils.caption_raster_list(
+            output_accessible_list)[0].replace("lucode_1", "lucode_LUCODE")
+        output_balance_img_src = None
+        output_balance_raster_caption = None
+        output_balance_percapita_img_src = None
+        output_balance_percapita_raster_caption = None
```
davemfish: Same comment here, I think we should declare these variables before the `if` statement.

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+
+        output_balance_percapita_img_src = raster_utils.plot_and_base64_encode_rasters(
+            output_balance_percapita_list)
+        output_balance_percapita_raster_caption = re.sub(
+            r'(pop_).*?(\.tif)', r'\1POP_GROUP\2',
+            output_balance_percapita_list[0].caption)
+
+    elif args_dict['search_radius_mode'] == 'radius per urban nature class':
+        nature_class_list = pandas.read_csv(args_dict[
+            'lulc_attribute_table'])['lucode']
```
davemfish: We may want to use our `natcap.invest.utils.read_csv_to_dataframe` function to be safe. I think we say that column names are not case-sensitive, for example.

src/natcap/invest/urban_nature_access/urban_nature_access.py
```diff
@@ -34,10 +34,16 @@
 RADIUS_OPT_POP_GROUP = 'radius per population group'
 POP_FIELD_REGEX = '^pop_'
 ID_FIELDNAME = 'adm_unit_id'
+
+_model_description = """
+Nature in urban areas provides numerous benefits to people–from urban cooling to flood mitigation to opportunities for recreation, physical activity, and renewal. The InVEST Urban Nature Access model allows for examination of access to the recreational, social, psychological, and physical health benefits of nature. Assessing nature-based recreation and its health benefits first requires an understanding of both the “supply” of urban nature and “demand” for urban nature. The supply is determined by the type, size, proximity, and quality of urban nature, while the demand depends on people’s preferences or local policy requirements/urban planning standards. The InVEST Urban Nature Access model provides a measure of both supply and demand for urban nature, as well as the supply-demand balance, which quantifies the extent to which supply meets demand at the individual, administrative, and city levels."""
```
davemfish: I think we can do without this first sentence since it doesn't specifically pertain to this model. We should wrap this in `gettext` too.

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+            spec=model_spec.get_input('lulc_raster_path')
+        ),
+        RasterPlotConfig(
+            raster_path=args_dict['population_raster_path'],
+            datatype=RasterDatatype.continuous,
+            spec=model_spec.get_input('population_raster_path')
+        ),
+    ]
+
+    if args_dict['search_radius_mode'] != "radius per urban nature class":
```
davemfish: It looks like we could import these "radius mode" strings from `urban_nature_access.py`

tests/reports/test_urban_nature_access_template.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class UrbanNatureAccessTemplateTests(unittest.TestCase):
+    """Unit tests for Urban Nature Access template."""
+
+    def test_render_search_radius_by_lulc_option(self):
+        """Test report rendering with LULC option without health_cost_rate"""
+
+        render_args = _get_render_args(MODEL_SPEC)
+        render_args['args_dict']['search_radius_mode'] = 'radius per urban nature class'
```
davemfish: Looks like we could use the radius string constants here as well.

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
+import logging
+import time
+
+import altair
+import geopandas
+import geometamaker
```
davemfish: This is an unused import

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,376 @@
...
+from natcap.invest.reports.raster_utils import RasterDatatype, \
+    RasterPlotConfig, RasterTransform, SpecialValueConfig
+
+LOGGER = logging.getLogger(__name__)
+
+TEMPLATE = jinja_env.get_template('models/urban_nature_access.html')
+
+MAP_WIDTH = 450  # pixels
+
+NEAR_ZERO_RANGE = (-0.01, 0.01)
```
davemfish: I don't think this variable is used.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -424,13 +424,36 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
         imshow_kwargs['norm'] = transform
         imshow_kwargs['interpolation'] = 'none'
         cmap = config.colormap
-        if dtype == 'divergent':
-            if transform == 'log':
-                transform = matplotlib.colors.SymLogNorm(linthresh=0.03)
-            else:
-                transform = matplotlib.colors.CenteredNorm()
+        if dtype in ['divergent', 'continuous']:
```
davemfish: Do we actually need to check for these dtypes before parsing the `config.special_values`? I know the attributes only apply in these two cases, but the check for `if config.special_values` could be enough. If we really don't want `special_values` used with other dtypes, then I think the right approach would be to add custom validation to the pydantic model.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -424,13 +424,36 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
...
+            vmin = vmax = halfrange = None
+            if config.special_values:
+                if config.special_values.extend == 'min':
+                    vmin = config.special_values.threshold
+                    halfrange = abs(vmin)
```
davemfish: How does overriding the default `halfrange` affect plots with a single-sided threshold? I'm not sure that this report includes any examples so just want to make sure we have reviewed the effects of this.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -154,14 +154,14 @@ class SpecialValueConfig:
     values beyond the `threshold` will be colored with `color` and the end
     of the colorbar will be labeled with `label`.
     """
-    extend: Literal["min", "max"]
+    extend: Literal["min", "max", "both"]
```
davemfish: It's possible that some of the logic in the plotting function would get slightly simpler if we didn't have this `extend` attribute, but instead required each of `threshold`, `label` and `color` to always be a 2-tuple. And single-sided thresholding could be achieved if one side of those tuples was always `None`. It would require some extra checking for `None`, but less checking for the value of `extend`.

But perhaps more critically, right now there is some overlap in the information stored in `extend` and those other attributes, which means it's theoretically possible to mis-configure this with `extends=both` and `threshold=0.5`, for example. I think trying to eliminate `extend` would be a nice way to handle that. Otherwise we may want a custom validator for this class.

src/natcap/invest/reports/templates/caption.html
```diff
@@ -28,11 +28,11 @@
...
-      {% if source_list %}
-        <p class="sources">
-          Sources: {{ source_list | join('; ') }}
-        </p>
-      {% endif %}
+    {% endif %}
+    {% if source_list %}
+      <p class="sources">
+        Sources: {{ source_list | join('; ') }}
+      </p>
```
davemfish, line 35: This seems better!

davemfish (changes requested):
Thanks for taking on this report, @claire-simpson . You did a great job with some challenging data.

I left mostly minor comments but also wanted to get your take on the vector map scales. I think it would be easiest to just talk about that.

src/natcap/invest/urban_nature_access/reporter.py
claire-simpson: These colors are intentionally not part of the `purpleorange` color ramp so they would be distinct. Do you think this would look better if they naturally extended the min and max color values exactly?

davemfish: Oh okay. I was mainly thinking about documenting the "magic hexcodes" somehow. But now that you mention it, it might look better to tone down the value a little bit so the images are not so dark. I think it could be okay to use the same color as is used for 1000 since I don't think we want to convey any important differences between 1000 and 1001, for example.

src/natcap/invest/reports/raster_utils.py
claire-simpson: Hm great question, currently `halfrange` forces the colorbar to limit on both sides to whichever threshold is specified, which is probably not what we want. This also raised other concerns. Namely, it doesn't seem possible with this norm to have `divergent` rasters with `0` correctly centered at the midway point color-wise and also specify upper and lower thresholds that don't meet the condition `abs(vmin) = vmax`. If `transform=linear`, we can use `TwoSlopeNorm`, but in the case of `log` transformed data, we may have to either develop a custom norm or enforce this `abs(vmin) = vmax` relationship between the 2 threshold values if `datatype=divergent`

davemfish: >Namely, it doesn't seem possible with this norm to have divergent rasters with 0 correctly centered at the midway point color-wise and also specify upper and lower thresholds that don't meet the condition abs(vmin) = vmax

Do we have any use-cases yet where the threshold values are not equidistant from zero? If not then maybe we don't need to worry about this yet.

If we can get away with not defining a custom `halfrange` that would be nice. Because I don't think we needed a `halfrange` when using a single-sided threshold, before this PR. And I think the "both" case here ends up setting the `halfrange` to the value it would get by default anyway, so long as `abs(vmin) = vmax`.

src/natcap/invest/reports/raster_utils.py
claire-simpson: Ok I've refactored this to remove `extend` and added validation as well

src/natcap/invest/reports/raster_utils.py
claire-simpson: > Do we have any use-cases yet where the threshold values are not equidistant from zero? If not then maybe we don't need to worry about this yet.

There are no use cases with 2 non-equidistant threshold values, so for now I've just added a warning if this case does come up in the future (should this be an error?). I've removed `halfrange`

src/natcap/invest/urban_nature_access/reporter.py
claire-simpson: Sounds good, we now get the hex codes dynamically from the color ramp defined for `divergent` data

src/natcap/invest/urban_nature_access/reporter.py
claire-simpson: I don't think the model spec text would be useful, but not because its lacking. The spec is actually more detailed than we'd want for just a title, so I opted not to use it

claire-simpson:
Thanks @davemfish for all the feedback. A few notes:
1. I'd like to add `raster_utils` tests relating to the special value configuration, including adding new test images. Would it be helpful to add these tests in this PR or fine to proceed with this PR as is and to do that work elsewhere?
2. As you mentioned alongside the San Antonio report, I looked into using `log` transform on the first demand raster map, but I think given that 0 is a common value in this raster (and log(0) is infinite), _just_ switching to `log` may not be the best choice since it looks like there are a bunch of NoData values in the raster. However, we could consider using a log transform and setting a `special_value` min threshold at 0.01. What do you think?

Screenshot 2026-06-03 at 12 28 05 PM

src/natcap/invest/reports/raster_utils.py
```diff
@@ -150,19 +150,29 @@ class RasterTransform(str, Enum):
...
-    label: str | tuple[str, str]
+    thresholds: tuple[float | None, float | None]
+    """Boundary value for the special upper and lower bounds, respectively"""
+    labels: tuple[str | None, str | None]
     """Label(s) to show on the colorbar for the special region"""
-    color: str | tuple[str, str]
+    colors: tuple[str | None, str | None]
     """Color(s) used for the special values"""
 
+    def __post_init__(self):
```
davemfish: I think this `__post_init__` is almost functionally the same, but the `Pydantic` way to do this would be to decorate an instance method with `@model_validator(mode='after')`. I think the only thing that will add beyond what you have here is to present the exception back to the user as a `ValidationError` (even if you still raise a `ValueError`). And it allows us to give the method a more descriptive name, no need to call it `__post_init__` if you use the decorator. https://pydantic.dev/docs/validation/latest/concepts/validators/#model-validators

src/natcap/invest/reports/raster_utils.py
```diff
@@ -183,6 +193,12 @@ class RasterPlotConfig:
     """Will customize the color and labeling of a special range of values"""
 
     def __post_init__(self):
+        if (self.special_values and self.datatype not in [
+                RasterDatatype.continuous, RasterDatatype.divergent]):
+            raise ValueError(
+                "`special_values` may only be defined for raster configs with "
+                "`datatype` of `continuous` or `divergent`."
+            )
```
davemfish: Similar here, this could be pulled out to a separate `check_special_values_and_datatype` validator. I really don't mind it as-is, but I do think there's slight benefit in getting back the `ValidationError` type.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -389,6 +405,34 @@ def _get_legend_kwargs(num_patches: int, n_plots: int, xy_ratio: float):
     }
 
 
+def _configure_special_values(cmap, special_values):
```
davemfish: A docstring or type annotations describing the types of these arguments would be a friendly addition.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -421,43 +465,34 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
...
-                        vmin=vmin, vmax=vmax)
-            imshow_kwargs['norm'] = transform
-        elif dtype.startswith('binary'):
-            transform = matplotlib.colors.BoundaryNorm([0, 0.5, 1], cmap.N)
-            imshow_kwargs['vmin'] = -0.5
-            imshow_kwargs['vmax'] = 1.5
+        vmin = vmax = None
+        if config.special_values:
+            vmin = config.special_values.thresholds[0]
+            vmax = config.special_values.thresholds[1]
```
davemfish: This could be condensed to a one-liner:

`vmin, vmax = config.special_values.thresholds`

src/natcap/invest/reports/raster_utils.py
```diff
@@ -421,43 +465,34 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
...
+        if dtype == 'divergent':
+            if transform == 'log':
+                if None not in [vmin, vmax] and abs(vmin) != vmax:
+                    raise UserWarning(
```
davemfish, line 474: I think this makes sense as a warning.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -421,43 +465,34 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
...
+                transform = matplotlib.colors.TwoSlopeNorm(
+                    vmin=vmin, vcenter=0, vmax=vmax)
+        elif dtype == 'continuous':
+            if transform == 'log':
+                transform = matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
+            else:
+                transform = matplotlib.colors.Normalize(
+                    vmin=vmin, vmax=vmax)
+        if dtype.startswith('binary'):
+            transform = matplotlib.colors.NoNorm()
```
davemfish, line 490: Neat, I suppose we should have been using this `NoNorm` all along.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -421,43 +465,34 @@ def plot_raster_list(raster_list: list[RasterPlotConfig]):
...
+                        "the divergent color ramp, vmax should equal "
+                        f"abs(vmin). Actual vmin: {vmin}, vmax: {vmax}")
+                transform = matplotlib.colors.SymLogNorm(
+                    linthresh=0.03, vmin=vmin, vmax=vmax)
+            else:
+                transform = matplotlib.colors.TwoSlopeNorm(
+                    vmin=vmin, vcenter=0, vmax=vmax)
+        elif dtype == 'continuous':
+            if transform == 'log':
+                transform = matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
```
davemfish, line 485: I suppose we could make sure `vmin > 0` before doing a log transform. Like you commented earlier, we should not be log-transforming datasets with 0s in them. And I think we were not guarding against that anywhere yet. Here's a kind of arbitrary check we do in `plot_raster_facets` to handle this same case:
```
    if transform == 'log':
        if numpy.isclose(vmin, 0.0):
            vmin = 1e-6
        normalizer = matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
```

But this effectively treats the 0s like nodata in the image, and I don't think that's a good idea either. For now, maybe we just don't choose a log-transform for any datasets that can have 0s. And we can make a separate issue to deal with this. When in doubt, a linear transformation is safest/least confusing to viewers.

src/natcap/invest/urban_mental_health/reporter.py
```diff
@@ -223,10 +223,9 @@ def report(file_registry: dict, args_dict: dict, model_spec: ModelSpec,
     }
     if is_continuous:
         common_kwargs["special_values"] = SpecialValueConfig(
-            extend="min",
-            threshold=NEAR_ZERO_RANGE[1],
-            label=str(NEAR_ZERO_RANGE[0]),
-            color='#FEE0B6'  # light orange
+            thresholds=(NEAR_ZERO_RANGE[1], None),
+            labels=(str(NEAR_ZERO_RANGE[0]), None),
```
davemfish: `NEAR_ZERO_RANGE = (-0.01, 0.01)`
This looks like the label will not match the value, right? The min value is 0.01, but the label is -0.01.

davemfish (changes requested):
Thanks @claire-simpson , maybe a few small changes left here. I fully support working on more raster tests, but I think it's okay to tackle that as a separate task if you want to. In the meantime, we will all just keep looking at the latest sample reports from all the models to make sure things are still looking good, and I believe they are with these changes.

As for the log-transformations, maybe best for us to file that as an issue for future improvement and to use that transformation sparingly for now, even if our plots are a bit uniform.

src/natcap/invest/reports/raster_utils.py
claire-simpson: Thanks! This seems like an improvement.

src/natcap/invest/urban_mental_health/reporter.py
claire-simpson: You're right, thanks!

src/natcap/invest/reports/raster_utils.py
claire-simpson: > For now, maybe we just don't choose a log-transform for any datasets that can have 0s. And we can make a separate issue to deal with this.

Sounds good, I will made an issue for this

src/natcap/invest/urban_nature_access/reporter.py
```diff
@@ -0,0 +1,370 @@
...
+        )
+    ]
+
+    inputs_img_src = raster_utils.plot_and_base64_encode_rasters(
+        input_raster_config_list)
+    inputs_raster_caption = raster_utils.caption_raster_list(
+        input_raster_config_list)
+
+    lulc_img_src, lulc_legend_html = raster_utils.plot_categorical_raster_with_table(
+        [args_dict['lulc_raster_path']])
```
claire-simpson, line 297: Is there currently a way to use info from `lulc_attribute_table` input csv to populate the legend? If not, is that something we might want to add to `plot_categorical_raster_with_table`?

davemfish: Not currently, but we can certainly make an issue for it. I don't think we have a way of knowing which table is likely to be more useful. Names/Description columns may or may not exist in the input CSV, same as with the RAT. I think the RAT is slightly more likely to have an accurate set of pixel values, the CSV may not, though that would probably have raised an exception somewhere in the model. And the RAT is more likely to have the count for each value, which I think is useful.

davemfish (approved):
Updates look good, thanks @claire-simpson !

PR #2570, opened 2026-05-26
Replace DirectoryInput with WorkspaceInput

### LINKED ISSUE #2179: Consider removing the `DirectoryInput` class
For reasons described in past issues ( ), we've moved away from using directory inputs to organize groups of input files. Using a CSV to point explicitly to each input filepath is preferred. Now the only remaining use for the `DirectoryInput` class is workspace directory for each model, which is fundamentally different in some ways - it's not expected to contain any files to begin with, except for possibly the taskgraph cache. We should reevaluate the `DirectoryInput` class and see if there's a better design to describe the workspace directory. Perhaps a more specialized input class, or it could be treated separately from the other model inputs.

emlys:
## Description
Fixes #2179 

This PR takes a minimal approach to removing `DirectoryInput` - I renamed the class to `WorkspaceInput` and removed the custom attributes, so now each model has an identical `WorkspaceInput` instance. 

Because the `WorkspaceInput`s are now all identical and each model has one, potentially we could go farther - should it have to be included in the model inputs list, or be added automatically? Should it even be considered an "input"? Would a different abstraction better represent the workspace? 

## Checklist


HISTORY.rst
```diff
@@ -75,6 +75,11 @@ General
 * ``Input, Output, and ModelSpec`` classes now create immutable objects.
   Use ``model_copy(update=dict(...))`` to copy an object and update attributes.
   (`#2228 <https://github.com/natcap/invest/issues/2228>`_)
+* Removed the ``DirectoryInput`` class and replaced it with ``WorkspaceInput``
```
dcdenu4: Could this be more descriptive to better convey what `DirectoryInput` is even a part of? Maybe `natcap.invest.spec.DirectoryInput` or if that's too verbose, "Removed the Model Specs DirectoryInput Class"...

dcdenu4 (changes requested):
Thanks @emlys,

> should it have to be included in the model inputs list, or be added automatically? Should it even be considered an "input"? Would a different abstraction better represent the workspace?

Leaving how it is makes sense to me, at least for now. I think it's nice to have it as an intentional part of the Model Spec, even if all InVEST models share this class. Since a user ultimately is inputting the directory for the workspace, I think it's fine to live at home in the inputs attribute.

megannissel:
This change makes sense to me. I fall on the side wanting to leave `workspace_dir` as part of the MODEL_SPEC, personally. I think it's helpful for the MODEL_SPEC to fully encapsulate the inputs a user provides to the model. 

The TNC Calibration Tool plugin is no longer using `DirectoryInput`, so no concerns on that front. They are modifying `WORKSPACE` (via `model_copy`) in order to provide custom `about` text, though; I think it's nice to continue allowing plugin authors to customize in this way.

HISTORY.rst
emlys: I revised it to use the full import names.

dcdenu4 (commented):
Thanks @emlys , I did want to check on the failing windows test which may or may not be related?

https://github.com/natcap/invest/actions/runs/26782533210/job/78950341247?pr=2570#step:24:83

src/natcap/invest/spec.py
```diff
@@ -1066,57 +1066,23 @@ def describe_rst(self):
...
-
-    def model_post_init(self, context):
-        self._contents_dict = {x.id: x for x in self.contents}
+    id: typing.ClassVar[str] = 'workspace_dir'
+    name: typing.ClassVar[str] = 'workspace'
+    about: typing.ClassVar[str] = gettext(
+        "The folder where all the model's output files will be written."
+        " If this folder does not exist, it will be created. If data"
+        " already exists in the folder, it will be overwritten.")
+    type: typing.ClassVar[str] = 'workspace'
```
davemfish, line 1080: @emlys The Workbench relies on this type to build the input form. It is expecting 'directory' here. I believe that is why the pptr test failed. A bunch of other Workbench tests did not fail because they use mocked spec that still has the old 'directory' type. So those tests should be updated as well. Such as in `setuptab.test.js`, `investtab.test.js`, `invest_subprocess.test.js`, and `app.test.js`. Thanks!

dcdenu4 (approved):
Thanks @emlys and @davemfish for iterating on this!

PR #231, opened 2020-07-16
Fix UFRM nodata handling

dcdenu4:
This PR fixes a bug where input rasters with nodata of 0 were not properly being masked and causing downstream side-effects. 

Closes issue #185

Issue #44, opened 2020-03-02
Carbon model casts lucode integers to floats


dcdenu4:
A user experienced an issue where her pools CSV had integer *lucode* values but floating point column values. When calling `utils.build_lookup_from_csv()`, the function returns a dictionary with the *lucode* values as floats. Setting all column values to integers, then *lucode* is also returned as an integer in the dictionary.

https://community.naturalcapitalproject.org/t/keyerror-invest-3-8-0-carbon-storage-model/850

dcdenu4:
Looks like when using `pandas` the `dtype` arg could be set to `object` which to preserve and not interpret type.

phargogh:
Hey @dcdenu4 , I couldn't help but notice this and thought it sounded unexpected.  Doesn't python hash `4101` and `4101.0` to the same key in a dict?

```python
>>> a = {4101: 'foo'}
>>> a[4101.0]
'foo'
```

dcdenu4:
That is something I didn't know, the more you learn! I'll look back into this user issue.

dcdenu4:
Alright, so after looking into it a bit more the real issue the user was having was the **NA** row, where the *lucode* ID was *NA*. This caused a `KeyError` to be reported on the first ID that it came across. Just another example of the need to address how this type of thing is handled in InVEST.

PR #2574, opened 2026-05-29
Urban Cooling: Add report and move `CC` and `T_air` out of intermediate outputs

### LINKED ISSUE #2322: Reports for Urban Cooling, Urban Nature Access
Develop reports for Urban Cooling and Urban Nature Access. Both models have a primary output which is a polygon with zonal statistics results in the table as well as primary and secondary raster outputs.

Report author guidance: https://github.com/natcap/invest/wiki/InVEST-Reports:-Author's-Guide

Model Supporters to review a draft of the reports, or to collaborate with on the design.
Urban Cooling: Yingjie, Chris Nootenboom, Jesse
Urban Nature Access: Yingjie, Lisa

### LINKED ISSUE #2564: Urban Cooling: `T_air` and `cc` should land in main output directory, not `intermediate_outputs`
The Urban Cooling model places `T_air` in `intermediate_outputs`, but `T_air` is in fact the final output generated by the model. More importantly, it's clear from people who use and teach the model that `T_air` is a primary output, and having to go to `intermediate_outputs` to find it is counterintuitive and mildly inconvenient.

Similarly, `cc`—while not the final output generated by the model—is a key output that would be more easily discovered in the main output directory.

- [] Move `T_air` out of `intermediate_outputs`.
- [] Move `cc` out of `intermediate_outputs`.
- [] Update User Guide to reflect new output locations.

emilyanndavis:
## Description
Resolves #2322. Resolves #2564.

Sample reports:
- [Factors method, with valuation and multi-feature AOI](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_cooling_model_report_multi_feature_aoi_factors_yelloworangered.html) (This report offers a yellow-orange-red color scheme as an option for temperature maps. Currently awaiting feedback from model experts. In the meantime, the code uses a monochromatic red color scheme for temperature maps.) 
- [Intensity method, with valuation and multi-feature AOI](https://storage.googleapis.com/data.naturalcapitalproject.org/invest-reports/urban_cooling_model_report_multi_feature_aoi_intensity_reds.html) (This report uses a monochromatic red color scheme for temperature maps.)

## Checklist
- [x] Updated the user's guide (see [UG PR #231](https://github.com/natcap/invest.users-guide/pull/231))
- [x] Updated test data (some raster plot snapshots needed updates, see [invest-test-data PR #44](https://bitbucket.org/natcap/invest-test-data/pull-requests/44))


src/natcap/invest/reports/templates/caption.html
```diff
@@ -28,11 +28,11 @@
...
-      {% if source_list %}
-        <p class="sources">
-          Sources: {{ source_list | join('; ') }}
-        </p>
-      {% endif %}
+    {% endif %}
+    {% if source_list %}
+      <p class="sources">
+        Sources: {{ source_list | join('; ') }}
+      </p>
```
emilyanndavis, line 35: Claire made this same change for the UNA report.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -191,35 +191,6 @@ def __post_init__(self):
                                      else COLORMAPS[self.datatype])
 
 
-def build_raster_plot_configs(id_lookup_table, raster_plot_tuples):
```
emilyanndavis, line 196: This function was no longer being used anywhere.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -318,7 +289,7 @@ def _figure_subplots(xy_ratio, n_plots, small_plots=False):
 
     fig, axs = plt.subplots(
         n_rows, n_cols, figsize=(figure_width, figure_height),
-        layout='compressed')
+        layout='compressed', gridspec_kw={'hspace': 0.1})
```
emilyanndavis, line 294: This adds a bit more space between rows of subplots, to help minimize confusion about which titles refer to which subplots (which was a real issue in the UC report using sample data). `0.1` is a relative unit, meaning "0.1 of the average Axes height", so the actual spacing will depend on the data. 0.1 is large enough to solve the UC problem but small enough that it's unlikely to create comically large gaps in other scenarios.

Test data had to be updated since this change affected all snapshots containing multiple rows of subplots.

src/natcap/invest/reports/vector_utils.py
```diff
@@ -44,3 +48,108 @@ def get_geojson_bbox(geodataframe):
...
+        fields = [vector_spec.get_field(field_name)
+                  for field_name in fields_to_include]
+    else:
+        fields = vector_spec.fields
+
+    return [f'{field.id}:{field.about}{_get_units_text(field)}'
+            for field in fields]
+
+
+def generate_results_table_from_vector(
```
emilyanndavis, line 100: This function was the last thing remaining in `sdr_ndr_utils`. I moved it into `vector_utils` and deleted `sdr_ndr_utils`. I also refactored the logic & outputs and added a docstring.

.gitignore
```diff
@@ -39,7 +39,7 @@ src/natcap/invest/seasonal_water_yield/*.cpp
 
 #################################################################################
 # Random frustrating files
-.DS_Store  # oh mac OS
+.DS_Store
```
emilyanndavis, line 42: Removed the inline comment that was preventing `.DS_Store` from being ignored.

src/natcap/invest/reports/vector_utils.py
```diff
@@ -44,3 +48,108 @@ def get_geojson_bbox(geodataframe):
         "properties": {}
     }
     return extent_feature, xy_ratio
+
+
+def get_vector_attr_table_caption(
```
megannissel, line 54: I implemented something similar for the AWY report, though using a CSVOutput rather than a VectorOutput to generate the list of fields. Probably not worth trying to do before this release, but it might be cool to unify these implementations in the future!

src/natcap/invest/urban_cooling_model/urban_cooling_model.py
```diff
@@ -324,10 +345,13 @@
...
                     about=gettext(
-                        "Avoided energy consumption (kWh or $ if optional energy cost"
+                        "Avoided energy consumption (kWh or currency units if optional energy cost"
                         " input column was provided in the Energy Consumption Table)."
                     ),
-                    units=u.none
+                    # Units intentionally set to `None` to avoid printing
+                    # 'unitless' in reports (this output is not unitless;
+                    # its units vary depending on inputs).
+                    units=None
```
megannissel: Would this be a good use case for `u.other`? Claire pointed it out to me recently. From `unit_definitions.txt`:
```
# for cases where no unit label should be displayed
# such as inputs that can have any or multiple different units
other = []
```

src/natcap/invest/urban_cooling_model/urban_cooling_model.py
```diff
@@ -359,7 +383,10 @@
                         " contains NO natural areas nor green spaces; where CC = 0 for"
                         " all LULC classes."
                     ),
-                    units=u.none
+                    # Units intentionally set to `None` to avoid printing
+                    # 'unitless' in reports (this output is not unitless;
+                    # its units vary depending on inputs).
+                    units=None
```
megannissel: Same suggestion as above, re: `u.other`

src/natcap/invest/spec.py
```diff
@@ -1856,7 +1856,8 @@ class NumberOutput(Output):
 
 class IntegerOutput(Output):
     """An integer output, or result, of an invest model."""
-    pass
+    units: typing.Union[pint.Unit, None] = None
```
megannissel: Rather than adding `units` to all three of these, I wonder if `IntegerOutput`, `RatioOutput` and `PercentOutput` all ought to inherit from `NumberOutput` (which already has `units`), much in the same way that `IntegerInput`, `RatioInput`, and `PercentInput` all inherit from `NumberInput`.

src/natcap/invest/urban_cooling_model/reporter.py
```diff
@@ -0,0 +1,419 @@
...
+        ) = _create_bldg_maps(model_spec, file_registry, args_dict)
+    else:
+        bldg_table = None
+        bldg_totals_table = None
+        bldg_table_caption = None
+        bldg_map_json = None
+        bldg_map_caption = None
+        bldg_map_source_list = None
+
+    # Secondary raster outputs: cooling capacity (always);
```
megannissel: This is 100% a nitpick, but should this comment list green area alongside cooling capacity as "always included"?

src/natcap/invest/urban_cooling_model/reporter.py
```diff
@@ -0,0 +1,419 @@
...
+            bldg_map_caption,
+            bldg_map_source_list
+        ) = _create_bldg_maps(model_spec, file_registry, args_dict)
+    else:
+        bldg_table = None
+        bldg_totals_table = None
+        bldg_table_caption = None
+        bldg_map_json = None
+        bldg_map_caption = None
+        bldg_map_source_list = None
```
megannissel: This may just be stylistic preference, but I think it may be a bit clearer to instantiate these variables before the `if` statement -- makes it really clear that they all always need to be defined.

src/natcap/invest/urban_cooling_model/reporter.py
```diff
@@ -0,0 +1,419 @@
...
+            altair.Scale(scheme='reds')
+        )
+    }
+
+    agg_bldg_results = geopandas.read_file(
+        file_registry['buildings_with_stats'], engine='fiona')
+    _, xy_ratio = vector_utils.get_geojson_bbox(agg_bldg_results)
+
+    def _create_bldg_map(attribute: str, title: str, units: Unit,
+                        scale: altair.Scale) -> altair.Chart:
```
megannissel:
```suggestion
    def _create_bldg_map(attribute: str, title: str, units: Unit,
                         scale: altair.Scale) -> altair.Chart:
```

src/natcap/invest/reports/templates/models/urban_cooling.html
```diff
@@ -0,0 +1,161 @@
...
+          font_size_px=15
+        ), 100),
+        (caption(bldg_table_caption, definition_list=True), 100)
+      ]) %}
+    {% else %}
+      {% set bldg_agg_results = content_grid([
+        wide_table(
+          bldg_table | safe,
+          font_size_px=15
+        ),
```
megannissel:
```suggestion
        (wide_table(
          bldg_table | safe,
          font_size_px=15
        ), 100),
```

megannissel (changes requested):
This is looking great, @emilyanndavis! Just a few suggestions for your consideration.

src/natcap/invest/reports/vector_utils.py
emilyanndavis: Yes, I peeked at an AWY report and figured you'd done something very similar. Definitely a good idea to consolidate things when we get the chance!

src/natcap/invest/urban_cooling_model/urban_cooling_model.py
emilyanndavis: Yes, good idea. And a good opportunity to update my caption logic to handle `u.other` as well (which it probably should have been doing already).

src/natcap/invest/spec.py
emilyanndavis: Ah, yes, I bet they were meant to inherit from `NumberOutput` all along, thank you!

src/natcap/invest/urban_cooling_model/reporter.py
emilyanndavis: Yes, thank you! (I updated this comment a few times during development and still managed to miss green area somehow.)

src/natcap/invest/reports/templates/models/urban_cooling.html
emilyanndavis: Whoops, great catch, thank you!

src/natcap/invest/spec.py
emilyanndavis: I updated these so they inherit from `NumberOutput` now, but I thought `PercentOutput` was a special enough case that it's worth defining the default units as `u.percent` (rather than `None`). Can you think of a scenario where we _wouldn't_ want that?

src/natcap/invest/reports/vector_utils.py
megannissel: For what it's worth, I like your implementation more 😄

src/natcap/invest/spec.py
megannissel: That makes sense to me; if a model is specifying that an output will be a `PercentOutput`, it stands to reason that the units should be `u.percent` by default.

src/natcap/invest/reports/templates/models/urban_cooling.html
```diff
@@ -0,0 +1,161 @@
...
+{% endblock content %}
+
+{% from 'vegalite-plot.html' import embed_vega %}
+{% block scripts %}
+  {{ super() }}
+  {% include 'datatable-js.html' %}
+  {% include 'vega-dependencies.html' %}
+  {% include 'vega-embed-js.html' %}
+  {% set chart_spec_id_list = [
+    (aoi_map_json, 'aoi_map'),
```
megannissel: I noticed when running the report in the Workbench that I was seeing `Uncaught ReferenceError: None is not defined` where it was trying to call `embedVega(None, "aoi_map")`. 

Since `_create_aoi_maps` doesn't return map JSON if the AOI has only one feature, I think it would be a good idea to only add `aoi_map_json` to `chart_spec_id_list` if `aoi_map_json` isn't `None`.

The error doesn't break anything, but probably best to avoid it anyway!

megannissel (changes requested):
Sorry, @emilyanndavis, one more minor thing that I noticed when running the report from the Workbench.

The rest is looking great, though!

src/natcap/invest/reports/templates/models/urban_cooling.html
emilyanndavis: Argh. Thanks for catching that… and for having an example in your AWY template I could follow instead of having to stumble onto the right syntax!

megannissel (approved):
Thanks, @emilyanndavis, this looks great!

PR #2577, opened 2026-06-01
Reports: display raster attribute table as the LULC legend

### LINKED ISSUE #2333: Reports: Display Raster Attribute Tables for LULC maps, when available
## Background/Motivation
When a LULC raster appears in a report, it is accompanied by a legend; however, since the legend only maps colors to integer values (not named categories), it is difficult to interpret such a map directly. It might be worthwhile to check for a raster attribute table (RAT) and, if one is available, render it in the report.

## Description
For each LULC raster that is rendered in the report and has a RAT, include that RAT as an HTML table in the report. Also include a note (i.e., a "pre-caption") accompanying the LULC maps indicating that RATs (when available) can be found in the report.

## Benefits
Including RATs can help users interpret LULC maps in reports. While we neither expect nor encourage users to perform in-depth analysis of rasters as they are rendered in a report, including RATs will provide some context that can help them spot-check results or inputs as they seek to verify a model run or debug an issue.

## Alternatives
In an ideal world, we could cross-reference a RAT with the raster itself in order to produce a legend that maps colors to LULC categories (not just their integer values). However, this is impractical in the general case, since the name of the column that stores LULC category names can vary. We could define a specific RAT format that we support (or even go so far as to require that LULC inputs include RATs that adhere to that format), but that seems like a separate issue.

## Additional context
This feature is relevant to the Carbon, NDR, and SDR reports, as well as to future reports where LULC rasters are primary inputs.

I had implemented this feature on a branch in the invest-reports repo, but after migrating that branch to the invest repo, the feature stopped working. I haven't yet had the chance to figure out why, so in the interest of getting a dev build ready for internal testing, I decided to de-prioritize this for now. Ideally, it will make it into the 3.18.0 release, but it's OK if it becomes an enhancement in a future release instead.

davemfish:
This PR adds a `reports.raster_utils` function for plotting categorical rasters and building a table to use as the legend.

The table is based on the Raster Attribute Table, if it exists. Often a RAT will include names/descriptions of values, and that makes for a better legend than the numbers alone. But there are no guarantees what columns exist in a RAT. If a RAT does not already exist, a frequency table is calculated and used instead.

5 existing reports plot LULC rasters, and they were each updated. If a report includes a single LULC map, then the table is displayed in the same row. If the report includes multiple LULC maps in the panel, then the table is always displayed below the image on its own row.

- [Carbon](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev849%2Bg1a367fca5/reports/carbon_report_willamette.html)
- [NDR](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev849%2Bg1a367fca5/reports/ndr_report_gura.html)
- [SDR](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev849%2Bg1a367fca5/reports/sdr_report_gura.html)
- [SWY](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev849%2Bg1a367fca5/reports/seasonal_water_yield_report_gura.html)
- [UMH](https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev849%2Bg1a367fca5/reports/urban_mental_health_report.html)

Fixes #2333 
Also fixes issue with requesting too many colors from `distinctipy`.

## Checklist


src/natcap/invest/reports/raster_utils.py
```diff
@@ -710,13 +840,29 @@ def _get_raster_metadata(filepath):
...
+def _get_raster_attribute_table(filepath):
+    rat = None
+    resource = _get_raster_metadata(filepath)
+    if resource is None:
+        # if no metadata already exists, generate it
+        resource = geometamaker.describe(filepath)
+        if isinstance(resource, geometamaker.models.RasterResource):
+            rat = resource.get_rat(1)
+    else:
+        rat = resource.get_rat(1)
```
emilyanndavis: I realize it's unlikely `_get_raster_attribute_table` will be called with a non-raster filepath, but it probably wouldn't hurt to check the `resource` type in all cases, just to be sure, right?

```suggestion
    if isinstance(resource, geometamaker.models.RasterResource):
        rat = resource.get_rat(1)
```

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+    rat_list = []
+    for raster_path in raster_path_list:
+        rat_list.append(_get_raster_attribute_table(raster_path))
+    if None not in rat_list:    
+        rat_df_list = []
+        rat_value_names = set()
+        for rat in rat_list:
+            value_col = None
+            for col in rat.columns:
+                if col.usage == 'MinMax':
```
emilyanndavis, line 712: Might be worth directly referencing the GDAL enum (`gdal.GFU_MinMax`) instead of a string literal, to make it clear we are looking to a standard defined outside of InVEST. Maybe also a short comment explaining that MinMax denotes a class value (since min=max)?

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+                    value_col = col.name
+            rat_value_names.add(value_col)
+            df = pandas.DataFrame(rat.rows)
+            # Some RAT include color tables. This is not useful to display in
```
emilyanndavis:
```suggestion
            # Some RATs include color tables. This is not useful to display in
```

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+    if len(rat_df_list) > 1:
+        rat_dataframe = functools.reduce(
+            lambda left, right: pandas.merge(
+                left, right, on=[value_col_name], how='outer'),
+            rat_df_list)
+    elif len(rat_df_list) == 1:
+        rat_dataframe = rat_df_list[0]
+
+    colors = get_categorical_colors(rat_dataframe.shape[0])
+    # sort pixel values so colors are assigned the same here as in imshow.
```
emilyanndavis: I found this comment a little confusing (what the code does isn't quite what I expected from the comment). I've suggested an update—does it sound right to you?
```suggestion
    # Sort values before matching them with colors to ensure the mapping is the same here as in imshow.
```

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+    colors_dict = dict(zip(sorted(rat_dataframe[value_col_name]), colors))
+    legend_col_name = ''  # the color swatch column needs no label
```
emilyanndavis: In practice, you're probably right, but it feels wrong to me to leave a column unlabeled when there is an obvious label for it, and it somehow seems especially strange since this column ends up being sortable. I don't think there's any harm in labeling it.
```suggestion
    legend_col_name = 'Color'
```

src/natcap/invest/reports/templates/styles.html
```diff
@@ -173,6 +171,16 @@
   button {
     cursor: pointer;
   }
+  .legend-table {
+    font-size: 0.8rem;
```
emilyanndavis: When the root font size is 16px (browser default), 0.8rem ends up being 12.8px, and it's generally best to avoid partial pixel values. Can I convince you make this 0.875rem (14px) instead?
```suggestion
    font-size: 0.875rem;
```

src/natcap/invest/reports/templates/models/sdr-ndr-report.html
```diff
@@ -77,6 +77,16 @@ <h2 class="section-header">Inputs</h2>
     args_table(args_dict)
   )}}
 
+  {{ accordion_section(
+    'Landuse / Landcover Input',
```
emilyanndavis: When we spell this out, e.g., in the Workbench UI, we use "Land Use/Land Cover". I'd do the same here, for consistency.
```suggestion
    'Land Use/Land Cover Input',
```

src/natcap/invest/reports/templates/models/sdr-ndr-report.html
```diff
@@ -77,6 +77,16 @@ <h2 class="section-header">Inputs</h2>
...
+    content_grid([
+      (caption(raster_group_caption, pre_caption=True), 100),
+      (raster_plot_img(lulc_img_src, 'LULC Input'), 60),
+      (lulc_legend_html | safe, 40),
+      (caption(lulc_caption, definition_list=True), 100),
+    ])
+  )}}
+
   {{ accordion_section(
     'Raster Inputs',
```
emilyanndavis: I noticed you renamed this to 'Other Inputs' in the SWY report, and I think that's a good pattern.
```suggestion
    'Other Inputs',
```

src/natcap/invest/reports/templates/models/seasonal_water_yield.html
```diff
@@ -82,7 +87,17 @@ <h2 class="section-header">Inputs</h2>
   )}}
 
   {{ accordion_section(
-    'Raster Inputs',
+    'Landuse / Landcover Input',
```
emilyanndavis:
```suggestion
    'Land Use/Land Cover Input',
```

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -109,17 +109,29 @@ <h2 class="section-header">Inputs</h2>
...
 
+  {% set img_width = 100 %}
+  {% set table_width = 100 %}
+  {% if args_dict['model_option'] == 'ndvi' %}
+    {% set img_width = 60 %}
+    {% set table_width = 40 %}
+  {% endif %}
+  {% if lulc_img_src != None %}
+    {{ accordion_section(
+      'LULC Maps',
```
emilyanndavis: Update this heading for consistency with NDR/SDR/SWY?
```suggestion
      'Land Use/Land Cover Inputs',
```

src/natcap/invest/reports/templates/models/urban_mental_health.html
```diff
@@ -109,17 +109,29 @@ <h2 class="section-header">Inputs</h2>
...
+        (caption(raster_group_caption, pre_caption=True), 100),
+        (raster_plot_img(lulc_img_src, 'LULC Inputs'), img_width),
+        (wide_table(lulc_legend_html | safe), table_width),
+        (caption(lulc_caption, definition_list=True), 100)
+      ])
+    )}}
+  {% endif %}
+
   {{ accordion_section(
     'Input Maps',
```
emilyanndavis:
```suggestion
    'Other Inputs',
```

src/natcap/invest/reports/templates/models/carbon.html
```diff
@@ -63,13 +63,19 @@ <h2 class="section-header">Inputs</h2>
...
   {% endif %}
 
+  {% set img_width = 100 %}
+  {% set table_width = 100 %}
+  {% if not args_dict['calc_sequestration'] %}
+    {% set img_width = 60 %}
+    {% set table_width = 40 %}
+  {% endif %}
   {{ accordion_section(
     'LULC Maps',
```
emilyanndavis: Update this heading for consistency with NDR/SDR/SWY?
```suggestion
      'Land Use/Land Cover Inputs',
```

HISTORY.rst
```diff
@@ -83,6 +83,8 @@ General
 * Fixed an issue with Windows binary builds where users would see runtime
   warnings saying ``GDAL_DATA`` is not defined.
   (`#2555 <https://github.com/natcap/invest/issues/2555>`_)
+* Reports display a Raster Attribute Table for Landuse / Landcover
+  input rasters. (`#2333 <https://github.com/natcap/invest/issues/2333>`_)
```
emilyanndavis:
```suggestion
* Reports now display a Raster Attribute Table for Land Use/Land Cover
  input rasters. (`#2333 <https://github.com/natcap/invest/issues/2333>`_)
```

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+            LOGGER.info(
+                f'Calculating frequency table for classes in {raster_path}')
+            table = pygeoprocessing.raster_reduce(
+                _count_frequency, (raster_path, 1), collections.Counter())
+            rat_df_list.append(pandas.DataFrame(
+                table.items(), columns=[value_col_name, count_col_name]))
+    if len(rat_df_list) > 1:
+        rat_dataframe = functools.reduce(
+            lambda left, right: pandas.merge(
+                left, right, on=[value_col_name], how='outer'),
```
emilyanndavis: I noticed the sample reports with two LULC inputs ended up with `_x` and `_y` suffixes on their same-named columns, and I wonder if we could make those a little more intuitive. It looks like we can customize those suffixes with the `suffixes` keyword argument to `pandas.merge`, e.g.,
```suggestion
            lambda left, right: pandas.merge(
                left, right, on=[value_col_name], how='outer', suffixes=('_left', '_right')),
```
For our use cases thus far (Carbon, UMH), we could even get away with `('_baseline', '_alternate')` (or `('_base', '_alt')` for brevity), but if we don't want to lock in that level of specificity right now, `('_left', '_right')` would still be an improvement.

src/natcap/invest/reports/raster_utils.py
```diff
@@ -665,6 +668,133 @@ def plot_raster_facets(tif_list, datatype, transform=None, title_list=None,
...
+        subset=[legend_col_name])
+    classes = 'datatable legend-table'
+    if rat_dataframe.shape[0] >= 20:
+        classes += ' paginate'
+    if rat_dataframe.shape[1] >= 5:
+        # 3 or 4 columns will be typical (color, value, count, name/desc)
+        # But we have no idea how many there might be.
+        classes += ' colvis'
+    lulc_rat_html = styler.hide(axis='index').to_html(
+        table_attributes=f'class="{classes}"', na_rep='')
```
emilyanndavis: I don't think `na_rep=''` works in `Styler.to_html` like it does in `DataFrame.to_html`. In the sample Carbon report, I am seeing `nan` values in the `COUNT_y`, `DESCRIPTIO_y`, and `lulc_group` columns. (Sort the table by `COUNT_y` descending.) Maybe this is a task for the next bugfix release, but it would be good to track down where exactly those `nan`s are being introduced, and overwrite them (ideally with `0` in count columns, `''` elsewhere).

src/natcap/invest/reports/templates/datatable-js.html
emilyanndavis, line 50: Maybe not crucial right now, since the DataTables implementation allows users to sort the table however they want, but it would be cool to customize the default sort order so that the table is initially sorted by value.

Assuming the `value` column is always immediately after the color column, I think we could do something like this:
```suggestion
    }
    if (table.classList.contains('lulc-rat') {
      options = {
        ...options,
        order: [[1, 'asc']]
      }
    }
```
and make sure `'lulc-rat'` is included in the CSS classes when the HTML table is generated.

If you try that out and it works, great! But if it's not that simple, I think it can wait.

emilyanndavis (changes requested):
Very cool, @davemfish! I know a lot of people will be thrilled with this update. I have a handful of suggestions, as usual.

src/natcap/invest/reports/raster_utils.py
davemfish: Yes, good idea. Though gdal constants are integers so there a couple more steps to get the string. We could do `geometamaker.utils._GFU_INT_TO_STR[gdal.GFU_MinMax]`, which is what we do in `geometamaker`. But I'm not sure that adding this additional dependency on the `geometamaker` API is better than hardcoding the string.

src/natcap/invest/reports/raster_utils.py
davemfish: It's so self-explanatory that I didn't want to label it. It's like adding the word "legend" above a map's legend. It feels like extra clutter. But I agree that table columns should have headers, and it's possible to sort, or copy/paste this table into a spreadsheet, so we'll give it a name.

The other tricky thing is we have no control over what columns already exist in someone's RAT. But it's unlikely they have a "color" column, and if they do, it's probably best to hide it anyway otherwise it could be confusing.

src/natcap/invest/reports/raster_utils.py
davemfish: Okay, I thought I was actually being clever and supporting an arbitrary number of tables with this reducer, but the default suffixes don't make sense with 3 tables (the 3rd table gets no suffix) and they start colliding with 4 or more. I limited this function to handling 1 or 2 rasters for now and I updated the suffixes.

src/natcap/invest/reports/raster_utils.py
davemfish: For now I updated this to pass the `na_rep=''` format in the appropriate place.

src/natcap/invest/reports/templates/datatable-js.html
davemfish: It works! Thanks for the suggestion.

davemfish:
Thanks for the review @emilyanndavis , I believe I addressed all your suggestions. Here are some updated samples:

- https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev860%2Bg0d9dcf569/reports/carbon_report_willamette.html
- https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev860%2Bg0d9dcf569/reports/ndr_report_gura.html
- https://storage.googleapis.com/natcap-dev-build-artifacts/invest/davemfish/3.17.3.dev860%2Bg0d9dcf569/reports/urban_mental_health_report.html

emilyanndavis (approved):
Looks great, @davemfish. Thanks for making those updates!

PR #2578, opened 2026-06-01
Datastack: copy CSV inputs into argument-named subdirectories and preserve original filenames

### LINKED ISSUE #2442: CSV inputs renamed in datastack cause metadata filename mismatch
## Description
In the Workbench, when you save "Parameters and data", most inputs are added to their own subfolder (named via the input's `id` and type) in the output tgz, and the filename remains the same (e.g., `farms.shp` is copied into subfolder `farm_vector_path_vector`). 

In contrast, CSV tables are _not_ moved into their own subfolder, and instead the filename itself changes to match the `id` and type (e.g., for `id=lulc_attr_csv`, the table is renamed to `lulc_attr_csv_csv.csv`). If that CSV has geometamaker metadata, the yaml is correctly copied into the tgz, but the metadata filename matches the original CSV filename and therefore is now out of date and won't be recognized as describing that table. If no metadata initially exists when saving the datastack, it is created but doesn't match the output CSV filename in the tgz either.

## Observed behavior
CSV filename changes when saving a datastack cause metadata to become incompatible with the CSV filename.

## Expected behavior
Should CSVs also be copied into their own subfolder? Should we rename any existing metadata to match the target filename? At minimum, we should ensure that any metadata generated during datastack creation is compatible with the files in the tgz.

## Screenshots

Image

claire-simpson:
## Description
When saving a datastack, CSV inputs are now treated like other (multipart) spatial files and are copied into their own argument-named subfolders within the output datastack directory. Original CSV filenames are now preserved, which means that sidecar geometamaker metadata matches (as it is named by the original filename).

Fixes #2442

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -83,6 +83,9 @@ General
 * Fixed an issue with Windows binary builds where users would see runtime
   warnings saying ``GDAL_DATA`` is not defined.
   (`#2555 <https://github.com/natcap/invest/issues/2555>`_)
+* When saving a datastack, CSVs are now copied into argument-named
+  subfolders, which preserves original filenames and ensures metadata
+  compatibility. (`#2442 <https://github.com/natcap/invest/issues/2442>`_)
 
 Workbench
 =========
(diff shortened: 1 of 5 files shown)


dcdenu4 (approved):
Thanks @claire-simpson, this looks good to me, thanks!

PR #2580, opened 2026-06-02
Update vega altair versions for compatibility

### LINKED ISSUE #2579: Reports: upgrade the vega dependency
There are warnings about a mismatch between the vegalite scheme version that altair exports and the vegalite.js version included in the documents.

davemfish:
Fixes #2579 

Also, for some reason the https://cdn.jsdelivr.net/npm/vega@5 script from the CDN was tagged with a `text/plain` content-type instead of `text/javascript` in the response header and the Workbench failed to load that resource in one case. Oddly, it did not fail for all reports I tried. Anyway, the version 6 file is fetched with the correct content-type.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ geometamaker>=0.3.1
 pydantic
 certifi
 docutils
-altair
+altair>=6.0.0
 fiona
 geopandas
 jinja2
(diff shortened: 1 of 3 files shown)


claire-simpson (approved):
Thanks @davemfish!

PR #2581, opened 2026-06-02
AWY: allow MULTIPOLYGON geometry type on input.

### LINKED ISSUE #2513: AWY watershed input should support MULTIPOLYGON
I haven't used AWY in a long while, and am just now noticing that it does not take .gpkg as a watershed input. It gives a Workbench UI error "Geometry type must be one of {'POLYGON'}". I tried drag/dropping it in, as well as navigating to the .gpkg through the file explorer, same error. So that means that i can't create watersheds in DelineateIt and directly use the result in AWY.  But gpkg does work in the watershed input for SDR, NDR and SWY, (and, we've had so much discussion about standardizing on gpkg in InVEST) so it should work in AWY as well.

Workbench 3.18.0

Dave says in Slack: the problem is not the file format (GPKG vs SHP). The problem is POLYGON geometry vs. MULTIPOLYGON geometry.

davemfish:
The subwatershed input already allows `MULTIPOLYGON`, the watershed input should too. They are used exactly the same way by the model.

Fixes #2513 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -97,6 +97,11 @@ Workbench
 * Added a link to the InVEST Plugin Developer's Guide to the Workbench Manage
   Plugins modal. (`#2145 <https://github.com/natcap/invest/issues/2145>`_)
 
+Annual Water Yield
+==================
+* Allow watershed input to have a POLYGON or MULTIPOLYGON geometry type.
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Looks straightforward to me, thanks!

PR #2582, opened 2026-06-02
Tolerate `None` or empty dict returned from a model's `execute` function

### LINKED ISSUE #2519: Relax the warning that emanates from not returning a dict from `execute`
InVEST model `execute` functions return a dictionary representing the file registry. This is used in metadata generation and in reports. But if a plugin does not return anything from `execute`, we should tolerate this. Right now it results in a warning, which includes a caught exception, 

```python
2026-04-20 14:02:52  natcap.invest.spec WARNING  Something went wrong while generating metadata
  File "/Users/<>/Library/Application Support/invest-workbench/micromamba_envs/plugin_1776689697943/lib/python3.13/site-packages/natcap/invest/spec.py", line 2260, in execute
    self.generate_metadata_for_outputs(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        registry, preprocessed_args)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/<>/Library/Application Support/invest-workbench/micromamba_envs/plugin_1776689697943/lib/python3.13/site-packages/natcap/invest/spec.py", line 2115, in generate_metadata_for_outputs
    for key, value in file_registry.items():
                      ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

This is a bit heavy-handed. It's okay for plugins to return `None` from `execute` so we shouldn't issue a message that makes people think something went wrong.

## Benefits
The bar to getting a plugin working will be lowered because developers will not necessarily need to create an output specification or use a `file_registry` object before getting a functional plugin. 

## Mitigation of side effects
Fewer developers may opt in to using these features (file_registry and metadata) which is not a good thing. We can provide documentation and examples of plugins that do use these features in order to encourage these practices.

davemfish:
This PR removes the scary warning that included a full traceback in the event that a model (or plugin) returned `None` from `execute`. We can still log a warning in that case, since returning `None` is incompatible with other `ModelSpec.execute` options. And no warning is logged if `execute` returns an empty dictionary.

Fixes #2519 

## Checklist


src/natcap/invest/spec.py
```diff
@@ -2235,6 +2235,22 @@ def execute(self, args, create_logfile=False, log_level=logging.NOTSET,
             model_module = importlib.import_module(self.module_name)
             registry = model_module.execute(args)
 
+            if not isinstance(registry, dict):
+                msg = "`execute` was called with {kwargs}, but no" \
+                      " file registry dictionary was returned from `execute`."
+                kwargs_used = []
```
emlys: Just a suggestion to condense these if-statements, something like
```
kwargs_used = [f'{kwarg}=True' for kwarg in [
    'check_outputs', 'generate_metadata', 'save_file_registry', 'generate_report'
] if locals()[kwarg]]
```

src/natcap/invest/spec.py
```diff
@@ -2235,6 +2235,22 @@ def execute(self, args, create_logfile=False, log_level=logging.NOTSET,
...
+                if check_outputs:
+                    kwargs_used.append('check_outputs=True')
+                if generate_metadata:
+                    kwargs_used.append('generate_metadata=True')
+                if save_file_registry:
+                    kwargs_used.append('save_file_registry=True')
+                if generate_report:
+                    kwargs_used.append('generate_report=True')
+                if len(kwargs_used):
```
emlys: I think we can skip the `len` check since `bool([]) == False`
```suggestion
                if kwargs_used:
```

emlys (commented):
Thanks @davemfish, just a couple of minor suggestions

davemfish:
Thanks for the suggestions, @emlys

src/natcap/invest/spec.py
davemfish: It turns out `locals` within a list comprehension is only scoped to that list comprehension, so I ended up doing this with a `for` loop instead. At least that appears to be true before Python 3.12. https://peps.python.org/pep-0709/

emlys (approved):

PR #2594, opened 2026-06-05
Include missing files in datastack; object-oriented datastack archiving

### LINKED ISSUE #2589: Crop Production datastack archive is built incompletely
## Description
Crop Production inputs include tables that reference other input data. When building a datastack archive, I observed that not all the necessary data was included. The crop-specific percentile yield tables referenced from the `percentile_yield_csv_table` input are left out.

It appears that this is because `datastack.build_datastack_archive` only pulls in files referenced in a table if they are spatial files. It should include any type of file.

emlys:
## Description
This fixes #2589 by updating the datastack-archiving logic to include any file-based inputs referenced in a CSV, not just spatial inputs.

I also decided to refactor `build_datastack_archive` into an object-oriented design, where each `Input` type owns its method for archiving a value. I think this is overall clearer and more extensible than the previous design. In the case of the HRA criteria table, the custom archiving logic now lives in a `HRACriteriaTableInput` subclass.

`write_metadata_file`, which is called in the datastack archiving process, also had some conditional logic that applied to different input/output types. I refactored that into class methods on those `Input`s/`Output`s as well.

## Checklist


tests/test_ui_server.py
```diff
@@ -159,7 +159,7 @@ def test_build_datastack_archive(self):
         target_filepath = os.path.join(self.workspace_dir, 'data.tgz')
         data_path = os.path.join(self.workspace_dir, 'data.csv')
         with open(data_path, 'w') as file:
-            file.write('hello')
+            file.write('lucode,c_above,c_below,c_soil,c_dead')
```
emlys, line 162: A side effect is that we're now calling `get_validated_dataframe` on CSVs in order to archive them, which means that they have to pass validation. Trying to archive a CSV that's invalid for the particular model will raise an error.

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
...
 class IOModel(ImmutableBaseModel):
     """Base class for both `Input` and `Output`."""
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
-    """Allow fields to have arbitrary types that don't inherit from BaseModel
-    (needed for pint.Unit)."""
+    id: str
+    """Input/output identifier that should be unique within a model"""
+
+    about: typing.Union[str, None] = None
```
emlys, line 236: Moved these attributes into the base class because they are used by both `Input` and `Output`.

src/natcap/invest/datastack.py
```diff
@@ -123,6 +123,14 @@ def get_datastack_info(filepath, extract_path=None):
     return 'logfile', extract_parameters_from_logfile(filepath)
 
 
+class Datastack():
```
claire-simpson: Should this live at the beginning or end of the module?

tests/test_datastack.py
```diff
@@ -406,7 +397,20 @@ def test_archive_extraction(self):
...
+                f.write('{"foo": 1}')
+            # copy existing DEM
+            # copy existing watersheds
+            # new raster
+            # new vector
+            spatial_csv.write('ID,spatial_path,nonspatial_path\n')
+            spatial_csv.write(f"1,{params['raster']},{nonspatial_path}\n")
+            spatial_csv.write(f"2,{params['vector']},{nonspatial_path}\n")
+            spatial_csv.write(f'3,{target_csv_raster_path},{nonspatial_path}\n')
+            spatial_csv.write(f'4,{target_csv_vector_path},{nonspatial_path}\n')
```
claire-simpson: It might be helpful for one of these `nonspatial_path`s to be a relative path instead of absolute

HISTORY.rst
```diff
@@ -72,6 +72,9 @@ General
   architectures, particularly on conda-forge.  This should also speed up
   the operation of ``setup.py``.
   (`#2630 <https://github.com/natcap/invest/issues/2630>`_)
+* Fixed a bug in datastack archiving that affected HRA. Refactored some
```
claire-simpson: What about something a bit more specific, like:
```suggestion
* Fixed a bug that prevented non-spatial files referenced in CSVs from being added to datastack archives. Refactored some
```
And then if we want to mention HRA specifically, should that be in the HRA section?

src/natcap/invest/datastack.py
```diff
@@ -154,210 +162,47 @@ def build_datastack_archive(args, model_id, datastack_path):
...
-
-        else:
-            LOGGER.debug(
-                f"Type {type(input_spec)} is not filesystem-based; "
-                "recording value directly")
-            # not a filesystem-based type
-            # Record the value directly
-            target_arg_value = args[key]
-        rewritten_args[key] = target_arg_value
+        LOGGER.info(f'Starting to archive arg "{key}": {args[key]}')
```
claire-simpson: Looks like this is duplicated
```suggestion
       
```

src/natcap/invest/datastack.py
```diff
@@ -154,210 +162,47 @@ def build_datastack_archive(args, model_id, datastack_path):
...
-            #     useful to include in the logfile.
-            rewritten_args[key] = getattr(module, override_funcname)(
-                args[key], data_dir, files_found)
-            continue
-
         LOGGER.info(f'Starting to archive arg "{key}": {args[key]}')
         # Possible that a user might pass an args key that doesn't belong to
         # this model.  Skip if so.
         if key not in module.MODEL_SPEC.inputs:
             LOGGER.info(f'Skipping arg {key}; not in model MODEL_SPEC')
```
claire-simpson, line 175: What happens if the `key` isn't in the `MODEL_SPEC` `inputs`? Should we `continue` here?

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
...
+        Create metadata for invest model inputs or outputs, taking care to
+        preserve existing human-modified attributes.
+
+        Note: We do not want to overwrite any existing metadata so if there is
+        invalid metadata for the datasource (i.e., doesn't pass geometamaker
+        validation in ``describe``), this function will NOT create new metadata.
+
+        Args:
+            datasource_path (str) - filepath to the data to describe
+            keywords_list (list) - sequence of strings
```
claire-simpson:
```suggestion
            keywords_list (list[str]): sequence of keywords
```

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
...
+
+        Create metadata for invest model inputs or outputs, taking care to
+        preserve existing human-modified attributes.
+
+        Note: We do not want to overwrite any existing metadata so if there is
+        invalid metadata for the datasource (i.e., doesn't pass geometamaker
+        validation in ``describe``), this function will NOT create new metadata.
+
+        Args:
+            datasource_path (str) - filepath to the data to describe
```
claire-simpson:
```suggestion
            datasource_path (str): filepath to the data to describe
```

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
...
+            keywords_list (list) - sequence of strings
+            lineage_statement (str, optional) - string to describe origin of
```
claire-simpson:
```suggestion
            lineage_statement (str): (optional) string to describe origin of
```

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
...
+                the dataset
+            out_workspace (str, optional) - where to write metadata if different
```
claire-simpson:
```suggestion
            out_workspace (str): (optional) where to write metadata if different
```

src/natcap/invest/spec.py
```diff
@@ -1065,6 +1272,121 @@ def describe_rst(self):
...
+
+        # If we already know about the parameter, then we can just reuse it
+        # and skip the file copying.
+        if source_path in datastack.files_found:
+            LOGGER.debug(
+                f'Key {self.id} is known: using {datastack.files_found[source_path]}')
+            datastack.args[self.id] = datastack.files_found[source_path]
+            return
+
+        # check the CSV for columns that may be spatial.
```
claire-simpson:
```suggestion
        # check the CSV for columns that may reference files.
```

src/natcap/invest/spec.py
```diff
@@ -1065,6 +1272,121 @@ def describe_rst(self):
...
+        # But also, the columns specification might not be listed, so don't
+        # require that 'columns' exists in the MODEL_SPEC.
+
+        csv_dir = os.path.join(datastack.target_dir, f'{self.id}_csv')
+        os.makedirs(csv_dir)
+        target_csv_path = os.path.join(
+            csv_dir, os.path.basename(source_path))
+
+        if self.columns:
```
claire-simpson, line 1324: Wouldn't this `if/else` statement mean that a file that has `columns` but none of them are file-based, doesn't get copied to the datastack archive?

src/natcap/invest/spec.py
```diff
@@ -1065,6 +1272,121 @@ def describe_rst(self):
...
+            datastack.args[self.id] = ''
+            return
+
+        # Python can't handle mixed file separators, so let's just
+        # standardize on linux filepaths.
+        source_path = value.replace('\\', '/')
+
+        # If we already know about the parameter, then we can just reuse it
+        # and skip the file copying.
+        if source_path in datastack.files_found:
```
claire-simpson, line 1309: I believe this is used 3x; worth abstracting out?

src/natcap/invest/spec.py
```diff
@@ -180,19 +181,107 @@ def validate_permissions_string(permissions):
     return permissions
 
 
+def set_metadata_field_descriptions(field_specs, resource):
+    """Set field descriptions on a geometamaker resource.
```
claire-simpson:
```suggestion
    """Set field or column descriptions on a geometamaker resource.
```

src/natcap/invest/spec.py
claire-simpson, line 643: Since `SingleBandRasterInput` explicitly sets the units on band 1., should `RasterInput`,  `RasterOutput`, and `RasterOrVectorInput` have a specialized `configure_metadata` implementation to write units for individual bands?

src/natcap/invest/spec.py
```diff
@@ -1065,6 +1272,121 @@ def describe_rst(self):
...
+        # Python can't handle mixed file separators, so let's just
+        # standardize on linux filepaths.
+        source_path = value.replace('\\', '/')
+
+        # If we already know about the parameter, then we can just reuse it
+        # and skip the file copying.
+        if source_path in datastack.files_found:
+            LOGGER.debug(
+                f'Key {self.id} is known: using {datastack.files_found[source_path]}')
+            datastack.args[self.id] = datastack.files_found[source_path]
```
claire-simpson, line 1312: Is `source_path` absolute and is this intended to be stored in the CSV or should we prefer to copy a path that is relative to CSV?

src/natcap/invest/spec.py
```diff
@@ -1065,6 +1272,121 @@ def describe_rst(self):
...
+                    LOGGER.debug(
+                        'File referenced in CSV copied from '
+                        f'{source_filepath} --> {target_filepath}')
+                    dataframe.at[
+                        row_index, column_spec.id] = target_filepath
+                    datastack.files_found[source_filepath] = target_filepath
+
+                LOGGER.debug(
+                    f'Rewritten spatial CSV written to {target_csv_path}')
+                dataframe.to_csv(target_csv_path)
```
claire-simpson:
```suggestion
                dataframe.to_csv(target_csv_path, index=self.index_col is not None)
```
To correctly preserve or avoid adding an index

claire-simpson (changes requested):
Thanks @emlys! This refactor makes a lot of sense. One other note, which doesn't have to be addressed in this PR, but it could be nice to generate metadata for the files referenced by CSVs when we copy them into the datastack archive

src/natcap/invest/datastack.py
emlys: Moved to the top of the module

src/natcap/invest/datastack.py
emlys: Looks like this was the case before, but you're right, a `continue` is needed to prevent raising an error on the next line. I do wonder if it might be better to let an error be raised in that case, since I don't think we would ever do that on purpose? But I suppose we can keep the original intent and let it `continue`.

src/natcap/invest/spec.py
emlys: Good catch! The `.to_csv` call should be one indentation level up, so that it happens whether or not there are any file-based columns.

src/natcap/invest/spec.py
emlys: Hmm, I see what you mean, but I'm not seeing a great way to abstract it out. Since the repeated code has a couple of conditional `return` statements, the calling function would have to check the return value or something... and at that point it seems about as complex as just repeating the code. So I'd vote to leave it as-is

src/natcap/invest/spec.py
emlys: Sure! I don't think those types were being handled before, but we may as well add it.

emlys: I added `configure_metadata` to `RasterInput` and `RasterOutput`, but I realized I'm not sure if/how it would work for `RasterOrVectorInput`, since the metadata depends on whether a raster or vector was supplied.

src/natcap/invest/spec.py
emlys: I think this `source_path` is the path to the CSV itself and it should always be absolute, since model inputs can't be relative.

emlys:
Thanks for the thoughtful feedback @claire-simpson, I think I've addressed everything. The failing check looks unrelated.

src/natcap/invest/spec.py
claire-simpson: Yeah, maybe we should add a method `get_spatial_type` or something for the `RasterOrVectorInput` class to tell us what was actually input (and this could be reused in `validation`). Seems like this is getting a bit out of scope for this PR though

claire-simpson (approved):
Thanks @emlys, this all looks good!

PR #2601, opened 2026-06-08
Clean up usage logging code

### LINKED ISSUE #2600: Clean up InVEST usage logging code
Remove code related to usage logging that is no longer relevant (e.g., attributes that are ignored by the API endpoint).

emilyanndavis:
## Description
Closes #2600.

Specifically, this PR removes the following:
- `node_hash` and `session_id` (we are no longer logging these attributes; the API endpoint ignores them),
- model exit logging (we no longer track model exit details; the API endpoint ignores the request body), and
- the `log_run` context manager (was used only by the Qt app, which is no longer actively maintained).

I tested this locally (by temporarily enabling logging in dev mode) and confirmed there were no errors and my model run arrived in the DB as expected.

## Checklist

### DIFF
--- a/src/natcap/invest/usage.py
+++ b/src/natcap/invest/usage.py
@@ -119,33 +119,6 @@ def _calculate_args_bounding_box(args, model_spec):
     return bb_intersection, bb_union
 
 
-def _log_exit_status(session_id, status):
-    """Log the completion of a model with the given status.
-
-    Args:
-        session_id (string): a unique string that can be used to identify
-            the current session between the model initial start and exit.
-        status (string): a string describing the exit status of the model,
-            'success' would indicate the successful completion while an
-            exception string could indicate a failure.
-
-    Returns:
-        None
-    """
-    logger = logging.getLogger('natcap.invest.usage._log_exit_status')
-
-    try:
-        log_finish_url = requests.get(_ENDPOINTS_INDEX_URL).json()['FINISH']
-        requests.post(log_finish_url, data={
-            'session_id': session_id,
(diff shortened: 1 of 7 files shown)


src/natcap/invest/usage.py
```diff
@@ -232,13 +147,11 @@ def _node_hash():
...
             'invest_release': natcap.invest.__version__,
             'invest_interface': invest_interface,
-            'node_hash': _node_hash(),
             'system_full_platform_string': platform.platform(),
             'system_preferred_encoding': locale.getdefaultlocale()[1],
             'system_default_language': locale.getdefaultlocale()[0],
             'bounding_box_intersection': str(bounding_box_intersection),
             'bounding_box_union': str(bounding_box_union),
-            'session_id': session_id,
             'type': type,
```
phargogh: I realize this isn't a change in this PR, but since `type` is a builtin function, would it maybe be worth renaming the variable to `type_` in keeping with python convention?

phargogh (approved):
One minor comment, but no reason why that should hold up the PR!  I'll add my approval and if you'd like to address that one minor thing, but feel free to merge if not.

Good to see this cleaned up, thanks @emilyanndavis!

src/natcap/invest/usage.py
emilyanndavis: Yes, good idea, thanks! I took this a step further and changed `type` to `model_type` and `source` to `plugin_source`.

phargogh (approved):
Thanks @emilyanndavis !

PR #2602, opened 2026-06-08
Refine .gitignore to improve file tracking and maintenance

dkav:
Based on https://github.com/natcap/invest/pull/2554#issuecomment-4530593286.

## Checklist


davemfish (changes requested):
Thanks for these changes @dkav . As discussed in #2554, we do not wish to ignore all `.cpp` files. It's possible we will want to add some in the future that are not Cython-generated.

dkav:
@davemfish I reverted the Cython change. That said, I would have kept it myself and amended .gitignore at the time when handwritten C++ files were added to the project. An override ignore (`!handwritten.cpp`) could be used then instead of the existing patterns.

davemfish (approved):
Thank you for these changes @dkav

PR #2622, opened 2026-06-11
Make the big sampledata archive for the github release object

### LINKED ISSUE #2619: Release Part 2: Publishes unsigned MacOS binary?
Our release publishing workflow (Release Part 2) downloads artifacts and publishes them in a GitHub release and PyPI release. It downloads the Windows Workbench exe from GCS, so that it gets the code-signed version. But it downloads the MacOS binary from the GitHub artifacts, so that one is _not_ codesigned, right? Shouldn't we be downloading the Mac binary from GCS as well?

### LINKED ISSUE #2620: Release Part 2: Tries to publish a sampledata archive that no longer exists.
As of #2566 we stopped building a single sampledata zip archive. We zip and deploy each model's data separately. I removed the single big archive because as far as I know, we don't use it anywhere.

But the Release workflow still expects that artifact to exist and it fails when it tries to download it.
https://github.com/natcap/invest/actions/runs/27355164894/job/80831945209

Should we revert the change and start building this archive again? Or should we update our release workflow to stop looking for it?

davemfish:
Fixes #2620 
Fixes #2619 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -301,7 +301,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-15-intel, windows-latest]
+        os: [windows-latest, macos-15-intel]
         include:
           - os: macos-15-intel
             puppeteer-log: ~/Library/Logs/invest-workbench/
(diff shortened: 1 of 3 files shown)


claire-simpson (approved):
Thanks!

PR #2647, opened 2026-07-02
Upgrade Bootstrap & React Bootstrap

### LINKED ISSUE #1798: Workbench: upgrade to react-bootstrap 2 and Bootstrap 5

emilyanndavis:
## Description
Closes #1798.

Specifically, this PR includes the following:
- Bootstrap upgraded to latest stable version (5.3.8).
- React Bootstrap upgraded to latest stable version (2.10.10).
- Markup adjusted as needed to ensure all elements are functional (some interactions were broken by Bootstrap/React Bootstrap API changes).
- Styles adjusted as needed. (For the most part, I have restored previous styling. I've also made some small changes where I thought improvements could be made.)
- Tests updated as needed.

## Checklist


emilyanndavis:
The only lingering test failure is our favorite thorn-in-our-side puppeteer test on macOS. The same test passes on the Windows runner and locally on my Mac.

workbench/src/renderer/components/SaveAsModal/index.jsx
```diff
@@ -112,77 +112,88 @@ class SaveAsModal extends React.Component {
...
-              </ToggleButton>
-            </ButtonGroup>
-            <Button onClick={this.browseSaveFile}>
-              <MdSave className="mr-1" />
-              {t('Save')}
-            </Button>
+            <Form>
+              <Form.Check className="save-as-option">
+                <Form.Check.Label>
+                  <Form.Check.Input
```
davemfish, line 124: The radio buttons here could use a little more padding on the left.

Image

davemfish: Also the background color darkens for the selected option, but it remains dark for the top option when the "use relative paths" box is checked, even after I click a different radio.

Image

workbench/src/renderer/components/SettingsModal/index.jsx
```diff
@@ -147,101 +145,92 @@ class SettingsModal extends React.Component {
...
+                              {t('threaded task management: tasks execute only ' +
+                                  'in the main process, using multiple threads.')}
+                            </li>
+                            <li>
+                              {t('n CPUs: depending on the InVEST model, tasks ' +
+                                  'may execute in parallel using up to this many processes.')}
+                            </li>
+                          </ul>
+                        </Accordion.Body>
+                      </Accordion.Item>
```
davemfish: This is minor, but I wonder if it looks different on Mac. The accordion looks very similar to a dropdown menu. Maybe expanding it to the full width of the modal would help? Or removing the border so it's just text next to the 🔽 , which is how it looked previously.

Image

workbench/src/renderer/styles/style.css
```diff
@@ -418,47 +507,46 @@ exceed 100% of window.*/
...
-
-.invest-sidebar-col .nav-link.active {
-  color: #000;
-  background-color: ghostwhite;
-  border-right: 5px solid var(--invest-green);
-  border-radius: 0;
+  &.active {
+    color: #000;
+    background-color: ghostwhite;
+    border-left-color: var(--invest-green);
```
davemfish: Previously this border was on the right. Was it intentional to move it the left? Should the gap on the left be closed?

Image

After toggling the tabs, the focus outline appears and the right-side border ends up slightly mis-aligned with the top navbar's tab outline.

Image

Here's how it used to look for comparison.

Image

davemfish (commented):
Thanks for updating us, @emilyanndavis ! I tested this out locally and it looks good and functional. I found a few minor style details you might wish to address.

workbench/src/renderer/styles/style.css
emilyanndavis: Yes, I intentionally moved the active tab indicator to the left, because it slightly bugged me to see that intersection of green borders that were almost—but not quite—the same width!

I took your suggestion to remove the gap on the left, and I think it looks much better that way. I also widened the left border, to make it stand out a bit more. I also adjusted the sidebar buttons & links so they're better aligned with the setup/log tabs, and I adjusted the horizontal position of the top-nav model tabs so that the tab background—not the tab border—aligns precisely with the background of the tab content below.

sidebar_tab_active-2026_07_09


Finally, I changed the sidebar tab focus styles. Now, on keyboard focus (i.e., `:focus-visible`), there will be no outline; instead, the background and text colors will change. On click, there will be no visible focus indicator, because the active tab styling is sufficient in that case.

sidebar_tab_keyboard_focus-2026_07_09

workbench/src/renderer/components/SaveAsModal/index.jsx
emilyanndavis: Whoa, not sure how I missed this. I'm sure I had the alignment correct at some point, but I must have changed something else that threw it off. Thanks!

workbench/src/renderer/components/SettingsModal/index.jsx
emilyanndavis: No, it looks the same on Mac, but you're right, it does look quite different from before. Our custom accordion toggle no longer worked due to API changes, and I thought it would simplest to implement a standard accordion, even though it looked different. Let me play around with this some more to see what I can come up with.

workbench/src/renderer/components/SaveAsModal/index.jsx
emilyanndavis: Fixed!

datastack_options-2026_07-10

workbench/src/renderer/components/SettingsModal/index.jsx
emilyanndavis: After trying out a few styling options and being thrilled with none of them, I realized I couldn't convince myself of any reason the n_workers info shouldn't just be visible all the time, so I decided to remove the accordion mechanism entirely. I also made the logging/taskgraph form fields full-width (with labels stacked on top). This keeps the full-width detail text from looking awkward, and I think it also helps clarify that the detail text is relevant to the n_workers field specifically. Finally, I made the modal narrower so the form fields wouldn't look too comically large. Let me know what you think!

settings_modal-persistent_details-stacked_labels-2026_07_10

davemfish: I think this looks great, thanks!

davemfish (approved):
Thanks for the updates!

PR #2654, opened 2026-07-08
add fields to configure plugin conda executable and env paths

### LINKED ISSUE #2645: Workbench option to use a user-provided conda installation
## Background/Motivation
We ship micromamba with the workbench to allow plugin environment management. While we try to keep it isolated from other conda things on the system, we have seen issues come up on some users' laptops where it appears to be conflicting with pre-existing settings.

There is also a long-standing issue where the micromamba installation cannot install dependencies that invoke the C compiler, because a space in the path of mamba and/or the python interpreter breaks the command.

## Description
Add a workbench option where the user can point to their own conda/mamba installation. This would serve as a backup option that can potentially get around the space-in-path issue as well as other unexpected issues that may come up.

### LINKED ISSUE #2646: Workbench option to set a user-defined conda env for a plugin
Allow the user to configure the path to the conda env that a plugin will be run in. 

Currently, the env is created and managed by the workbench, and not directly exposed to the user. The development process includes finding the env (by looking in the workbench config.json), activating it, and editable installing the plugin to enable faster iteration.

We could make this easier by allowing the user to point to their existing dev environment. More generally, this seems like a logical thing to be able to configure, and adds flexibility that could be helpful in future situations.

emlys:
## Description
Fixes #2645, fixes #2646 
Add an interface in the "Manage Plugins" modal to change the configured conda executable and plugin environment paths (equivalent to the `micromamba` and `plugins.<plugin ID>.env` properties in `config.json`).

I added `defaultMicromamba` and `plugins.<plugin ID>.defaultEnv` properties to the `config.json` to track the original values so that they can be reset if needed.

Needed to mock the plugin data returned from `GET_SETTING` in a couple more tests.

## Checklist


workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -76,7 +76,7 @@ async function createBaseEnv(baseEnvPrefix, micromamba) {
   logger.info(`creating base environment from file:\n${baseEnvYMLContents}`);
   fs.writeFileSync(baseEnvYMLPath, baseEnvYMLContents);
   await spawnWithLogging(micromamba, [
-    'create', '--yes', '--prefix', `"${baseEnvPrefix}"`,
+    'env', 'create', '--yes', '--prefix', `"${baseEnvPrefix}"`,
```
emlys, line 79: It seems that "env create" rather than just "create" is necessary for compatibility with conda.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
           }
         </div>
       </Form>
+      <hr />
+      <Form aria-labelledby="configure-conda-form-title">
```
emilyanndavis: Thanks for including the `aria-labelledby` attribute here! Might be a good idea to link the form to its descriptive text as well, via `aria-describedby`.
```suggestion
      <Form aria-labelledby="configure-conda-form-title" aria-describedby="conda-executable-description">
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+        <Form.Group>
+          <h5 id="configure-conda-form-title" className="mb-3">{t('Configure conda executable (Advanced)')}</h5>
+          <Form.Text
+            as="span"
+            id="conda-executable-desciption"
```
emilyanndavis:
```suggestion
            id="conda-executable-description"
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+              id="condaPath"
+              type="text"
+              value={condaPath}
+              onChange={(event) => setCondaPath(event.target.value)}
+              className="me-1"
+            />
+            <Button
+              aria-label="browse for conda executable"
+              className="browse-button ms-1 me-1"
+              id="browse-conda-button"
```
emilyanndavis: No need for an `id` here unless you are referencing it elsewhere (and it doesn't look like you are). If you want a handle for testing, it can be convenient to add a `data-` attribute that is clearly there for testing purposes. (For example, `data-testid`, which we use in a couple of components, automatically works with the DOM Testing Library's `findByTestId` query.)

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+              {t('Reset')}
+            </Button>
+          </div>
+          <Button onClick={saveCondaPath} className="text-nowrap mt-3">
+            {t('Save')}
+          </Button>
+        </Form.Group>
+      </Form>
+      <hr />
+      <Form aria-labelledby="configure-plugin-envs-form-title">
```
emilyanndavis:
```suggestion
      <Form aria-labelledby="configure-plugin-envs-form-title" aria-describedby="plugin-env-description">
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+        <Form.Group>
+        <h5 id="configure-plugin-envs-form-title" className="mb-3">{t('Configure plugin environments (Advanced)')}</h5>
+        <Form.Text
+            as="span"
+            id="plugin-env-desciption"
```
emilyanndavis:
```suggestion
            id="plugin-env-description"
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+                value={pluginEnvs[pluginID]}
+                onChange={(event) => setPluginEnvs(
+                  {...pluginEnvs, [pluginID]: event.target.value}
+                )}
+                className="me-1"
+              />
+              <Button
+                aria-label="browse for env"
+                className="browse-button ms-1 me-2"
+                id="browse-env-button"
```
emilyanndavis: No need for an `id` here either.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +479,106 @@ export default function PluginModal(props) {
...
+                variant="outline-dark"
+                onClick={selectDirectory}
```
emilyanndavis: I couldn't figure out why the `browse` buttons didn't seem to be working for me, until I looked more closely at the code. Every `browse` button in the modal is calling `selectDirectory` on click… and `selectDirectory` always populates the `path` field in the 'Add a plugin' form! So when I use a `browse` button to choose a conda path or a plugin env, my selection lands in the wrong form. You may want to write separate functions, or it might make sense to refactor `selectDirectory` to take another argument or perhaps parse data attached to the `event`.

emilyanndavis (changes requested):
Thanks, @emlys. I found a bug in the `browse` button behavior, and I left some other (minor) suggestions as well.

Beyond those code-specific comments, a couple of other things came to mind when I was trying out the "Configure conda executable" UI:
- It's not immediately clear that the path to the executable is actually the path to the directory that contains the executable, not the full path including the name of the executable (at least I'm assuming that's the case, since I was able to select only folders, not actual executable files). I wonder if there's a way to elegantly include this detail in the description, and/or add an example? I expect many people who use this option will use the `browse` button and figure it out from context (like I did), but I wonder what would happen if someone directly entered the full path including the name of the executable.
- The default value, `micromamba`, could be misleading, since it's not obviously a path. I can imagine someone entering simply `conda` and ending up disappointed. Is the default value in fact a path? If so, is there a way to make that more obvious (such as expanding it to its full path)? If not, could we display the path? Or perhaps expand the display text to something like `InVEST-bundled micromamba (default)` so it's abundantly clear it's not merely the name of the package manager?

workbench/src/renderer/components/PluginModal/index.jsx
emilyanndavis: Some automated tests might be helpful here, too! 😃

workbench/src/renderer/components/PluginModal/index.jsx
emilyanndavis: …unless that `id` ends up being referenced by the click handler, in which case, make sure each instance of this button has a unique `id` (e.g., including the relevant plugin ID).

emlys:
Clearly I forgot to try out the file chooser dialog for the conda executable! It is supposed to select a file, not a directory. When the binaries are built, the default executable will look like an actual path, something like `/Users/emily/Applications/InVEST Workbench.app/Contents/Resources/micromamba`. In dev mode, it defaults to `micromamba` because it assumes you have `micromamba` installed and on the `PATH`.

I don't know that the file chooser is very useful here, I imagine the typical use would be replacing the micromamba path with `conda` or `mamba` which the user has already installed, so providing the full path is unnecessary. I just figured I'd include the file chooser for consistency with the other form fields.

emlys:
Thanks @emilyanndavis, I fixed the file chooser dialog, added some unit tests, and hopefully clarified what the conda executable input is supposed to be!

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -149,15 +166,53 @@ export default function PluginModal(props) {
...
+
+  const saveCondaPath = () => {
+    ipcRenderer.send(
+      ipcMainChannels.SET_SETTING, 'micromamba', condaPath
+    );
+  };
+
+  const resetPluginEnv = (pluginID) => {
+    ipcRenderer.invoke(
+      ipcMainChannels.GET_SETTING, `plugins.${pluginID}.defaultEnv`
```
emilyanndavis: In testing the "Configure plugin environments" feature, I ran into the somewhat obtuse error: `A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component.`

I ultimately traced it back to this `defaultEnv` property: since the plugin was already installed before you added the `defaultEnv` property to the code, the result of this lookup is `undefined`… causing the "Reset" operation to fail, because the value of the input can't be changed to `undefined` (and of course, we don't actually want it to be `undefined`—we want it to be the true default env).

Seems like we should do an initial check of each installed plugin and define `defaultEnv` wherever it doesn't already exist—maybe a good place for this would be immediately after fetching the plugins list from the settings store (line 72)?

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +487,108 @@ export default function PluginModal(props) {
...
+            id="conda-executable-description"
+            className="plugin-form-text mb-3"
+          >
+            {t('InVEST is distributed with a copy of micromamba, a conda-like '
+              + 'package manager that is used to manage plugin environments. '
+              + 'If you have conda or mamba installed elsewhere on the system, '
+              + 'you can configure InVEST to use that executable instead. This '
+              + 'may be useful if you run into limitations of the included '
+              + 'micromamba distribution. You can enter an absolute path, or '
+              + 'the name of an executable that is on the system PATH.')}
```
emilyanndavis, line 508: This is much clearer, thanks! The only change I might suggest is to add examples of executable names:
```suggestion
              + 'the name of an executable (e.g., "conda", "mamba", or "micromamba", without quotes) that is on the system PATH.')}
```
…though, as I write this, I realize it's possible those all require the `.exe` extension (or are otherwise different) on Windows, so maybe it's better not to risk adding confusion here?

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -433,6 +487,108 @@ export default function PluginModal(props) {
...
+              </Button>
+            </div>
+          </Form.Group>
+        ))}
+        <Button
+          onClick={savePluginEnvs}
+          className="text-nowrap mt-3"
+          disabled={!Object.keys(pluginEnvs).length}>
+            {t('Save')}
+        </Button>
```
emilyanndavis: Rather than disabling this button when there are no plugins to configure, why not hide it instead? In fact, we could take it a step further and add some short explanatory text like "No plugins to configure."

```suggestion
        {Object.keys(pluginEnvs).length
          ? <Button
              onClick={savePluginEnvs}
              className="text-nowrap mt-3">
                {t('Save')}
            </Button>
          : <p>{t('No plugins to configure.')}</p>
        }
```

emilyanndavis (changes requested):
Thanks, @emlys; much improved! Just a few lingering items to address.

workbench/src/renderer/components/PluginModal/index.jsx
emlys: Sure, thanks for the suggestion!

workbench/src/renderer/components/PluginModal/index.jsx
emlys: Yeah, it could be different on different systems. Hopefully anyone who's configuring this knows what is meant by an executable on the system path.

workbench/src/renderer/components/PluginModal/index.jsx
emlys: Good catch. I think we can avoid this problem by keeping the `env` setting as the default value, and creating a new setting called `userDefinedEnv` for the new value - that way we don't have to edit previously installed plugin settings.

emlys:
Thanks @emilyanndavis! I fixed another bug I found where the user-defined micromamba setting was being overwritten on startup. ~~And I moved the definition of the default micromamba path into the preload script, so now it's exposed as `window.Workbench.DEFAULT_MICROMAMBA`, rather than as a setting. I think this makes more sense because it is not really a "setting", it will always be constant for one workbench installation. The value of having it in the settings store was really just to expose it to both the renderer and main process.~~

The above got complicated in the tests and built binaries, so I'm keeping them all as settings. But now the defaults will use the existing settings (`micromamba` and `env`), and user-defined values will use new settings (`userDefinedMicromamba` and `userDefinedEnv`). This should avoid needing to update the settings from previously-installed plugins.

emilyanndavis (approved):
Looks good, @emlys. Thanks for your patience with all the back-and-forth!

PR #2662, opened 2026-07-15
Workbench plugin test with Selenium

emlys:
## Description
Trying out selenium as an alternative to puppeteer to implement end-to-end testing of plugins in the workbench.

This re-implements the puppeteer test for plugins (which we've been skipping) in python using selenium. It installs the `invest-demo-plugin` via the workbench, launches it, runs it, and waits for it to complete.

## Checklist


workbench/src/renderer/components/SetupTab/index.jsx
```diff
@@ -637,6 +637,7 @@ class SetupTab extends React.Component {
           </Portal>
           <Portal elId={sidebarFooterElementId}>
             <Button
+              name="Run"
```
emlys, line 640: Makes this button a lot easier to find in the tests, and also probably more accessible?

.github/workflows/build-and-test.yml
```diff
@@ -331,6 +333,7 @@ jobs:
             ${{ env.CONDA_DEFAULT_DEPENDENCIES }}
             python=${{ env.LATEST_SUPPORTED_PYTHON_VERSION }}
             pandoc
+            selenium
```
davemfish, line 336: I guess it may not make sense to add selenium to `requirements-dev.txt`. But for someone wanting to run this test locally, this is a weird limbo where it also won't get installed by someone setting up their Workbench dev environment.

.github/workflows/build-and-test.yml
```diff
@@ -460,6 +463,25 @@ jobs:
           DEST=gs://releases.naturalcapitalproject.org/invest-reports/latest
           gcloud storage rsync -r ${{ steps.invest-autotest-deploy.outputs.REPORTS_BASE_URL }} $DEST
 
+      - name: Install chromedriver
+        working-directory: workbench
+        env:
+          CHROMEDRIVER_URL_PREFIX: http://storage.googleapis.com/chrome-for-testing-public
+        run: |
+          ELECTRON_VERSION=$(jq -r '.build.electronVersion' package.json)
```
davemfish, line 471: It looks like electron-builder may leave behind some data on which electron version it used when building. Check out `workbench/dist/builder-effective-config.yaml`. I would hope the chromium version might discoverable somewhere too, but I haven't found it.

workbench/package.json
```diff
@@ -100,11 +105,12 @@
...
     "jest-environment-jsdom": "^29.5.0",
     "puppeteer-core": "^24.38.0",
-    "react": "^19.0.0",
+    "react": "^18.2.0",
     "react-bootstrap": "^2.0.0",
-    "react-dom": "^19.0.0",
+    "react-dom": "^18.2.0",
     "react-icons": "^4.9.0",
     "rimraf": "^6.1.2",
+    "selenium-webdriver": "^4.46.0",
```
davemfish, line 113: Is this package necessary if we are using Python's selenium? And is `electron-chromedriver` needed if we are manually downloading chromedriver?

.github/workflows/build-and-test.yml
```diff
@@ -460,6 +463,25 @@ jobs:
...
+          echo "corresponding chromium version: $CHROMIUM_VERSION"
+          curl -L $CHROMEDRIVER_URL_PREFIX/$CHROMIUM_VERSION/${{ matrix.chromedriver-url-suffix }} \
+            -o chromedriver.zip
+          unzip chromedriver.zip -d ./chromedriver
+
+      - name: Test electron app with selenium
+        working-directory: workbench
+        run: |
+          ./chromedriver/chromedriver*/chromedriver* --port=9515 &
+          pytest tests/binary_tests/plugins.py
```
davemfish, line 483: Do you want to add this to the package.json scripts to make it easier for people to run it locally? Or maybe that gets complicated because of the need to install chromedriver after building the app.

davemfish (commented):
Hey @emlys I just have one suggestion for getting the electron version that might let us avoid hardcoding it in package.json.

I'm also a little unsure about the chromedriver/webdriver javascript dependencies, and if they are needed. 

I think it will be really nice to have plugin installation under test. My guess is that the main difference now between this new test and `puppet.test.js` is really the timeout. Given enough time, I wonder if that test would consistently pass also. Because they both appear to take a similar amount of time to start the app. And I'm not sure if there are any material differences in the approach of A) chromedriver launching the app with an open port for selenium to connect to vs. B) launching the workbench executable with the `--remote-debugging-port` chromium flag and using the puppeteer API to connect to it.

Finally, I think there are some decisions we should make for future -- not right now -- to avoid some extra context-switching:
1. Stick with either selenium or puppeteer for all the e2e tests.
2. Stick with either python or javascript for all these tests

.github/workflows/build-and-test.yml
emlys: If we end up liking the selenium tests, I'd expect to add this to `requirements-dev.txt`. I'm thinking this is kind of in beta - I don't know for sure yet if the selenium tests will ultimately be more reliable, and we won't really know until it's been on github actions for a while.

emlys:
> My guess is that the main difference now between this new test and puppet.test.js is really the timeout. Given enough time, I wonder if that test would consistently pass also.

Which timeout are you referring to? 

I tried to get the puppeteer test up and running for comparison, and it works for me locally, but it fails to find the "workspace directory" input field when running on Github Actions. I don't want to spend a lot of time debugging it, given that this looks like more flaky behavior to me, and that was the reason we disabled the test in the first place.

davemfish:
> Which timeout are you referring to?

Sorry, didn't mean to lead you down a path of extending timeouts. This is the timeout we see basically every time the MacOS `puppet.test.js` has failed in recent history
```
FAIL tests/binary_tests/puppet.test.js (266.701 s)
  ✕ Run a real invest model (111206 ms)
  ✓ Open each model and each local userguide (153604 ms)
  ○ skipped Install and run a plugin
  ○ skipped App re-launch will exit and focus on first instance

  ● Run a real invest model

    TimeoutError: Waiting for selector `[name="Cancel Run"][role="button"]` failed

      at Function.waitFor (node_modules/puppeteer-core/src/common/QueryHandler.ts:213:36)
      at CdpElementHandle.waitForSelector (node_modules/puppeteer-core/src/api/ElementHandle.ts:632:13)
      at CdpElementHandle.<anonymous> (node_modules/puppeteer-core/src/api/ElementHandle.ts:149:20)

    Cause:
    TimeoutError: Waiting failed: 30000ms exceeded
```

> I tried to get the puppeteer test up and running for comparison, and it works for me locally, but it fails to find the "workspace directory" input field when running on Github Actions. I don't want to spend a lot of time debugging it, given that this looks like more flaky behavior to me, and that was the reason we disabled the test in the first place.

That input label text changed relatively recently. The "Run a real model" test selects it like this:
```
const workspace = await argsForm.waitForSelector(
    'aria/[name="Workspace Directory"][role="textbox"]');
```
whereas the plugin test is still using `[name="Workspace (directory)"][role="textbox"]`. 

The puppeteer test has been very consistent in the sense that the Windows test always passes and the Mac test always fails in the same way. We have unfortunately labeled `puppet.test.js` as flaky when in fact it's the MacOS Github Actions runners that are flaky.

emlys:
> That input label text changed relatively recently.

Thanks, that fixed it. It behaves inconsistently because the plugin requires `natcap.invest>=3.19.0`, and so the label is different depending on whether 3.19 or 3.20 gets installed.

Haven't we previously tried increasing the timeouts and eventually run into the same issue, though?

At least for the "run an invest model" test, it waits for the log tab to become active before looking for the Cancel button. And it looks like those things should happen at basically the same time. They're both triggered by job state changes that happen in the same callback in response to stdout logging from the invest process starting. So I don't think there should be a 30+ second gap there.

davemfish:
> Haven't we previously tried increasing the timeouts and eventually run into the same issue, though?

Yeah, maybe more than once. I agree that it's a problem that it takes so long to run on GHA, and just extending timeouts is not an ideal solution.

> At least for the "run an invest model" test, it waits for the log tab to become active before looking for the Cancel button. And it looks like those things should happen at basically the same time. They're both triggered by job state changes that happen in the same callback in response to stdout logging from the invest process starting. So I don't think there should be a 30+ second gap there.

I think the Log tab is active as soon as Run is clicked, at least that's how it effectively behaves when I run a model. But then there's a lag of 1 - 2 seconds before anything shows up in the log. And the Cancel button does not appear until after that because it's based on the invest process stdout. Still definitely should not be a 30 second gap! This has something in common with the slowness of the Workbench startup in general: in both cases nodejs is launching a subprocess.

emlys:
Well, it switches to displaying the blank log as soon as you click Run. But I see a lag before the "Log" tab in the sidebar is no longer grayed out. That is what the test checks for before the Cancel button (`invest-tab-tab-log` is the id of the sidebar tab): 
https://github.com/natcap/invest/blob/6fb5df9aae0b273baa168787b78f0d7215e31064/workbench/tests/binary_tests/puppet.test.js#L236-L242

And the sidebar "Log" tab status change appears to be triggered by the same callback that causes the Cancel button to appear.

dcdenu4:
@emlys , is this still active or something we want to move to draft / close?

emlys:
I think we still want to merge some version of this. I'll revisit it this week

Issue #2467, opened 2026-03-27
Annotate invest inputs with keywords


davemfish:
## Background/Motivation
We're working on InVEST - DataHub Integration and we want to enable accurate, semantic searching of the DataHub from the Workbench. Search will be triggered from a specific model input, and results should be a list of datasets from the DataHub that would be suitable for that input. Having a controlled vocabulary of keywords attached to invest inputs and to DataHub datasets will facilitate this.

## Description
`spec.Input` types will have an optional `keywords` attribute. It will be a list of strings (or more likely a new `Keyword` type). The new types could look like this, where they include the keyword `value` as a string. As well as other metadata to reference the source of the keyword.

```python
class GCMDKeyword(BaseModel):
    value: str
    uuid: str
    full_path: str
    vocabulary: str = 'Global Change Master Directory (GCMD) Keywords'


class InvestKeyword(BaseModel):
    value: str
    vocabulary: str = 'InVEST Keywords'

# Examples:

PAWC = InvestKeyword(value='PLANT AVAILABLE WATER CONTENT')

PRECIPITATION = GCMDKeyword(
    value='PRECIPITATION',
    uuid='1532e590-a62d-46e3-8d03-2351bc48166a',
    full_path='EARTH SCIENCE > ATMOSPHERE > PRECIPITATION')
```

We want the keywords to be standardized, so when possible we will use keywords defined in the NASA Global Change Master Directory of Keywords. https://gcmd.earthdata.nasa.gov/KeywordViewer

If there are no suitable matches there, we can define our own keyword, or reference another controlled vocabulary.

## Benefits
In addition to enabling DataHub searches, using keywords from an established vocabulary like the GCMD could enable integration with other data catalogs or models that are annotated with that same vocabulary.

## Mitigation of side effects
InVEST models pertain to several different scientific domains. While the GCMD is somewhat comprehensive -- and the community can request additions to it -- there is not complete agreement across domains about a standard vocabulary. Our approach can allow inclusions of keywords from any vocabulary, along with a reference to their source. Though maintaining all that within InVEST might get out of hand. 
 
## Alternatives
The "id", "name" and "about" attributes of inputs already have some text that could conceivably by used to do text-based searches on the DataHub, but some basic attempts yielded poor search results. In general, results were too broad and many items were not a good match for the specific input.

davemfish:
We decided that due to issues related to special characters in CKAN search, we want to limit the characters that we're using in keywords. 
[...]

PR #2664, opened 2026-07-16
Another batch of invest model keywords

davemfish:
This PR adds keywords for another batch of models. It also adds the `alias` attribute to `Keyword` in order to address #2675 as well as the problem of special characters discussed in #2467 .

Both of these issues can remain open until we complete the full set of keywords for invest. I'm also leaving out History notes until this task is complete in order to avoid inevitable conflicts.

## Checklist


src/natcap/invest/keywords.py
```diff
@@ -40,6 +56,12 @@ class InvestKeyword(Keyword):
     uuid='6f6537f5-773f-4df1-862b-d9ab80eb5e04',
     full_path='EARTH SCIENCE > BIOSPHERE > VEGETATION > CARBON')
 
+CARBON_REGRESSION_PARAMETERS = InvestKeyword(
+    value='CARBON REGRESSION PARAMETERS',
+    definition=(
+        'Parameters used by the regression model in the InVEST Forest Carbon'
+        ' Edge Effects model.'))
```
megannissel: Super nitpicky, but "effect" is singular in the model name
```suggestion
        ' Edge Effect model.'))
```

src/natcap/invest/ndr/ndr.py
```diff
@@ -104,6 +87,7 @@
                 " Calculate Phosphorus is selected. All LULC codes in the LULC raster"
                 " must have corresponding entries in this table."
             ),
+            keywords=[keywords.PHOSPHOROUS, keywords.NITROGEN],
```
megannissel, line 90: Related to https://github.com/natcap/data.naturalcapitalproject.stanford.edu/issues/233, I think we ought to include `keywords.BIOPHYSICAL_TABLE` on biophysical table inputs (since we do have that keyword and may need to leverage it to help address the unique confluence of needs between the Hub's UI and API-based search).

src/natcap/invest/pollination/pollination.py
```diff
@@ -113,6 +110,8 @@
                 " values in the LULC raster must have corresponding entries in this"
                 " table."
             ),
+            keywords=[keywords.POLLINATOR_NESTING_SUITABILITY,
+                      keywords.POLLINATOR_FLORAL_RESOURCES],
```
megannissel, line 113: Likewise, should we include `keywords.BIOPHYSICAL_TABLE` here as well?

src/natcap/invest/keywords.py
```diff
@@ -97,17 +149,85 @@ class InvestKeyword(Keyword):
...
+    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER',
+    alias=['LAND USE LAND COVER'])
 
+# TODO: this is not currently in use. Instead biophysical tables are
+# tagged with LULC. Should we use this tag instead?
 LULC_CLASSES = GCMDKeyword(
     value='LAND USE/LAND COVER CLASSES',
     uuid='e5815f58-8232-4c7f-b50d-ea71d73891a9',
-    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE/LAND COVER CLASSES')   
+    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE/LAND COVER CLASSES',
```
megannissel: I think this keyword might be incorrect. Looking at the GCMD, it looks like there's a `LAND USE CLASSES` keyword, with UUID `fe2f8240-4d8e-4b1f-b869-29fee59692f7` and path `EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE CLASSES`. The UUID here matches that of the `LAND USE/LAND COVER` keyword.

src/natcap/invest/keywords.py
```diff
@@ -143,11 +272,12 @@ class InvestKeyword(Keyword):
...
     value='SHORELINES',
     uuid='1d3b4eb7-9931-44bf-8457-26847051b7a8',
-    full_path='EARTH SCIENCE > OCEANS > COASTAL PROCESSES > SHORELINES')
+    full_path='EARTH SCIENCE > OCEANS > COASTAL PROCESSES > SHORELINES',
+    aliases=['COASTLINE'])
 
 SOIL_ERODIBILITY = InvestKeyword(
     value='SOIL ERODIBILITY',
-    definition='The susceptibility of soil to erosion as a function of soil properties.')
+    definition='The susceptibility of soil to erosion as a function of soil properties.',)
```
megannissel:
```suggestion
    definition='The susceptibility of soil to erosion as a function of soil properties.')
```

megannissel (changes requested):
This is looking good, Dave! Just a few questions / suggestions.

Also, related to our conversation today: To help avoid future human error, can we add a test to ensure we don't accidentally assign the same alias to multiple keywords?

src/natcap/invest/ndr/ndr.py
davemfish: Wouldn't adding `BIOPHYSICAL_TABLE` make it more likely that a query for this input would find a table for a different model? I tested some things out and put my thoughts down in #2681 . We should definitely discuss further! I know it's inconsistent thus far and I imagine after we figure this out I'll come back and do a sweep through all the models again to add/subtract that keyword as needed.

src/natcap/invest/keywords.py
davemfish: Aha! Good eye, thank you! I also have a note there that we aren't using this keyword anywhere. Maybe I added it thinking we would use it on biophysical tables instead of the plain `keywords.LULC`, but the tables currently on the DH are tagged with "LAND USE LAND COVER", so I think we should stick with that on invest too. I'll leave this in the dictionary for now though.

src/natcap/invest/keywords.py
```diff
@@ -2,23 +2,39 @@
 
 
 class Keyword(BaseModel):
```
emilyanndavis, line 4: Is there any scenario in which we'd want to define a `Keyword` that is not a `GCMDKeyword` or an `InvestKeyword` … or some other vocabulary-specific subclass? If not, I wonder if we might want to throw an error on any attempt to instantiate a `Keyword` directly, to enforce the use of well-defined subclasses instead?

Maybe I'm overthinking it, and/or jumping too far ahead. I just looked at the `vocabulary` property on the `Keyword` class and imagined some future where, for example, we had a collection of keywords that are supposed to point back to the GCMD but don't really, because they were created as one-off `Keyword` instances and don't have the `uuid` and `full_path` attributes we would expect of a proper `GCMDKeyword`.

src/natcap/invest/keywords.py
```diff
@@ -60,12 +82,29 @@ class InvestKeyword(Keyword):
...
+    definition=(
+        'An index describing the suitability of a location for a specific crop,'
+        ' based on relevant climate parameters'))
+
 CROP_COEFFICIENT = InvestKeyword(value='CROP COEFFICIENT')
 
+CROP_PLANT_YIELDS = GCMDKeyword(
+    value='CROP/PLANT YIELDS',
+    uuid='f12d8026-f24a-4413-91d0-4704c243c9e7',
+    full_path='EARTH SCIENCE > AGRICULTURE > AGRICULTURAL PLANT SCIENCE > CROP/PLANT YIELDS')
```
emilyanndavis: Would it be useful to define `aliases` here, since the `value` has a `/` in it? Or is that not relevant unless/until there is a known use case affecting CKAN search?

src/natcap/invest/keywords.py
```diff
@@ -77,10 +116,23 @@ class InvestKeyword(Keyword):
...
     value='FLOOD DAMAGE VALUE',
     definition='The economic value of flood damage to a property.')
 
+HABITAT_SENSITIVITY = InvestKeyword(
+    value='HABITAT SENSITIVITY',
+    definition='Parameters related to the sensitivity of habitat to specific threats.')
+
+HABITAT_THREATS = InvestKeyword(
+    value='HABITAT THREATS',
+    definition='Geographic distribution and parameters related to threats to habitat.')
```
emilyanndavis, line 146: I'm probably getting way ahead of myself, but I can't help but wonder: do you think we will want to localize keyword definitions? I imagine they'll be mostly for internal use, at least initially, but should we anticipate any user-facing scenarios as well?

src/natcap/invest/keywords.py
```diff
@@ -97,17 +149,86 @@ class InvestKeyword(Keyword):
 LULC = GCMDKeyword(
     value='LAND USE/LAND COVER',
     uuid='e5815f58-8232-4c7f-b50d-ea71d73891a9',
-    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER')
+    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER',
+    alias=['LAND USE LAND COVER'])
```
emilyanndavis: This should be `aliases`, right?
```suggestion
    aliases=['LAND USE LAND COVER'])
```

emilyanndavis: Also, are there any aliases missing here? In a quick search on the Data Hub I am seeing `LULC` and `LAND COVER` in addition to `LAND USE LAND COVER`.

src/natcap/invest/keywords.py
```diff
@@ -97,17 +149,86 @@ class InvestKeyword(Keyword):
...
+
+NITROGEN = GCMDKeyword(
+    value='NITROGEN',
+    uuid='bf03dba8-2881-44ac-abfc-ba3353f67a24',
+    full_path='EARTH SCIENCE > TERRESTRIAL HYDROSPHERE > WATER QUALITY/WATER CHEMISTRY > NUTRIENTS > NITROGEN')
+
+NUTRIENTS = GCMDKeyword(
+    value='NUTRIENTS',
+    uuid='9bcb805c-718e-42c3-913d-174bdf06d4c1',
+    full_path='EARTH SCIENCE > BIOSPHERE > VEGETATION > NUTRIENTS')
```
emilyanndavis, line 184: It seems like there are multiple GCMD keywords called `NUTRIENTS`. Considering the NDR context, and looking at the keywords you've created for `NITROGEN` and `PHOSPHORUS`, I wonder if `NUTRIENTS` should reference the `EARTH SCIENCE > TERRESTRIAL HYDROSPHERE > WATER QUALITY/WATER CHEMISTRY > NUTRIENTS` path (and the uuid `1459a39c-4781-4481-8bd9-510762865efd`) instead?

Or… do you think there might be a case to be made for having multiple `NUTRIENTS` keywords in InVEST as well?

tests/test_keywords.py
```diff
@@ -0,0 +1,27 @@
...
+
+from natcap.invest import keywords
+
+
+class TestKeywords(unittest.TestCase):
+    """Test and validate characteristics of the keywords vocabulary."""
+
+    def test_unique(self):
+        """Test values, aliases, and uuids are all unique."""
+        values = set()  # values and aliases teogether must be unique
```
emilyanndavis:
```suggestion
        values = set()  # values and aliases together must be unique
```

src/natcap/invest/spec.py
```diff
@@ -316,22 +316,59 @@ def get_child_inputs(self):
...
         return []
 
-    def get_keywords(self, include_children=True):
-        """Get a list of unique keyword strings for this input and children."""
+    def get_keywords(self, include_children=True, include_aliases=True):
+        """Get a list of unique keyword strings for this input and children.
+
+        Args:
+            include_children (bool): include the keywords of nested Input or
+                Output specs.
```
emilyanndavis: Can you clarify why this docstring—and the one for `get_keyword_objects`—references "nested Input or Output specs"? Can an input contain a nested output? Or are these methods intended to be used for outputs as well?

src/natcap/invest/spec.py
```diff
@@ -316,22 +316,59 @@ def get_child_inputs(self):
...
+
+        Returns:
+            list of dicts
+
+        """
+        keywords = []
+        if self.keywords:
+            keywords += [keyword.model_dump() for keyword in self.keywords]
```
emilyanndavis: Just curious why you are dumping each object to a dict rather than retaining the object's own structure? I'm sure there's a reason; it's just not obvious to me!

tests/test_spec.py
```diff
@@ -547,6 +547,20 @@ def test_get_keywords_csv_input(self):
         self.assertCountEqual(csv_input.get_keywords(), expected_keywords)
         self.assertCountEqual(
             csv_input.get_keywords(include_children=False), [keywords.LULC.value])
+        self.assertCountEqual(
+            csv_input.get_keywords(include_aliases=True),
+            expected_keywords + keywords.LULC.aliases)
+
+    def test_get_keywords_irregular_csv_input(self):
+        """Test that CSVInput.get_keywords works when columns None."""
```
emilyanndavis:
```suggestion
        """Test that CSVInput.get_keywords works when columns=None."""
```

src/natcap/invest/pollination/pollination.py
```diff
@@ -1,4 +1,5 @@
 """Pollinator service model for InVEST."""
+from ast import keyword
```
emilyanndavis: Accidental import?

src/natcap/invest/keywords.py
emilyanndavis: Ah, I see now that this keyword is referenced by the Crop Production models, not by NDR. In that context, I agree this is the appropriate GCMD keyword. Still, I am left wondering if we might want to call this `CROP_NUTRIENTS`, lest someone be tempted to use `NUTRIENTS` to tag an NDR input?

emilyanndavis (changes requested):
Very cool to see this coming together! I found a few typos but mostly just had a bunch of thinking-out-loud sorts of questions.

src/natcap/invest/keywords.py
davemfish: I think this is a good idea. I haven't done enough OO design to know for sure, but I think it makes sense to enforce the use of subclasses. Update in 64171767a

src/natcap/invest/keywords.py
davemfish: I think it's a good idea. Even if it is not in use on the DataHub now, the alias can serve as a suggestion for what the DH tag should be if there ever is a need for it.

src/natcap/invest/keywords.py
davemfish: I don't know about any user-facing applications yet.

src/natcap/invest/keywords.py
davemfish: Yes, `aliases`, thanks! I added some Pydantic config to `Keyword` to catch this and do some other generally useful things.

>Also, are there any aliases missing here? In a quick search on the Data Hub I am seeing LULC and LAND COVER in addition to LAND USE LAND COVER

When I searched the DH I concluded that `LULC` and `LAND COVER` were redundant so adding those other aliases to queries would not find any new datasets. Going forward, I think there's a case to be made for removing the redundant tags on the DH. See https://github.com/natcap/data.naturalcapitalproject.stanford.edu/issues/241

src/natcap/invest/keywords.py
davemfish: Hmm, I'm going back and forth. Given the meaning of `NUTRIENTS` I think it is also relevant to NDR. There are more specific keywords that we use in NDR instead though. Even if it was named `CROP NUTRIENTS` or `PLANT NUTRIENTS`, those meanings are still relevant to the NDR biophysical table with it's parameters related to nutrient loading.

src/natcap/invest/spec.py
davemfish: `Inputs` can contain nested `Inputs`, such as `fields: list[Input]`. And likewise for `Output`. I guess I was designing these methods for both, but they only exist on `Input` right now. I guess I'll wait until we need them for outputs to refactor these methods (and `get_child_inputs`) onto the parent `IOModel`. But I definitely expect we will add keywords to `Outputs` as well.

src/natcap/invest/spec.py
davemfish: I was using this in a script and wanting something serializable. But we're not using this method anywhere in invest yet, so I think I will just remove it for now.

src/natcap/invest/keywords.py
```diff
@@ -1,24 +1,51 @@
...
+        return super().__new__(cls)
 
 
 class GCMDKeyword(Keyword):
-    value: str
+    """A Keyword from the NASA Global Change Master Directory (GCMD)"""
     uuid: str
+    """The unique id from the GCMD."""
     full_path: str
+    """The full hierarchy of they keyword within the GCMD."""
```
megannissel:
```suggestion
    """The full hierarchy of the keyword within the GCMD."""
```

src/natcap/invest/keywords.py
```diff
@@ -97,17 +161,86 @@ class InvestKeyword(Keyword):
...
 
+# TODO: this is not currently in use. Instead biophysical tables are
+# tagged with LULC. Should we use this tag instead? It would not currently
+# match any tags in use on the DH.
 LULC_CLASSES = GCMDKeyword(
     value='LAND USE/LAND COVER CLASSES',
-    uuid='e5815f58-8232-4c7f-b50d-ea71d73891a9',
-    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE/LAND COVER CLASSES')   
+    uuid='fe2f8240-4d8e-4b1f-b869-29fee59692f7',
+    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE/LAND COVER CLASSES',
```
megannissel: I think the [full path](https://gcmd.earthdata.nasa.gov/KeywordViewer/scheme/all/fe2f8240-4d8e-4b1f-b869-29fee59692f7?gtm_keyword=LAND%20USE%20CLASSES&gtm_scheme=sciencekeywords) is still incorrect and should be:
```suggestion
    full_path='EARTH SCIENCE > LAND SURFACE > LAND USE/LAND COVER > LAND USE CLASSES',
```

megannissel (changes requested):
Thanks, Dave, I like the updates you've made to the `Keyword` class and testing. Just a couple of minor things I noticed.

I'm also fine with leaving `BIOPHYSICAL_TABLE` out for now; we can always revisit later if necessary. Addressing those two issues related to accurate search and sibling datasets should help inform things!

emilyanndavis (approved):
Changes look good, thanks!

davemfish:
> Thanks, Dave, I like the updates you've made to the `Keyword` class and testing. Just a couple of minor things I noticed.
> 
> I'm also fine with leaving `BIOPHYSICAL_TABLE` out for now; we can always revisit later if necessary. Addressing those two issues related to accurate search and sibling datasets should help inform things!

Thanks for those last two fixes!

megannissel (approved):
Thanks, @davemfish!

Issue #2630, opened 2026-06-16
Missing 3.20.0 conda-forge builds


davemfish:
The 3.20.0 conda-forge builds are missing for the these platforms/architectures.

    - linux_aarch64
    - linux_ppc64le
    - osx_arm64

We explicitly skipped those builds in the feedstock recipe due to what seems like an upstream problem with `rtree` where it sometimes fails to find `libspatialindex` libraries and cannot import. It's possible this PR could address that: https://github.com/Toblerity/rtree/pull/404

phargogh:
Some progress on this front!  It turns out that all of the problematic platforms are cross-compiled, which is where conda-forge's build process fools the python process into thinking that it's actually aarch64 or ppc64le when it is, in fact, just using those compilers but the python process is still x64.  This causes problems in `setup.py` because we require pygeoprocessing, which requires rtree, and then rtree doesn't match the architecture that the process is looking for.

I've worked up a patch for this fix in https://github.com/conda-forge/natcap.invest-feedstock/pull/60 which uses package metadata to retrieve the required information from pygeoprocessing instead of importing the whole package.  A nice side effect of this is that the compilation should be noticeably faster since we're no longer needing to import pygeoprocessing as part of running `setup.py`.  A PR will be coming shortly with these changes in this repo.

phargogh:
This will be addressed for 3.20 by https://github.com/conda-forge/natcap.invest-feedstock/pull/60 and resolved in future builds thanks to #2668

davemfish:
3.20.0 builds are now available

PR #2668, opened 2026-07-22
Fix cross-compilation of `natcap.invest` for other architectures on conda-forge

phargogh:
This PR updates `setup.py` to read the pygeoprocessing package location from metadata instead of from the imported module object, which avoids a cross-compilation issue in conda-forge.

It turns out that the problem described in #2630 was due to cross-compilation: the aarch64 (mac and linux) and ppc64le (linux) builds are actually built on x64 hardware at this time.

When we go to build a cross-compiled package, `setup.py` is being executed by a different interpreter, `cross-python` inside the `build` environment on conda-forge, which is an x64 environment.  Everything installed into the build environment (including pygeoprocessing and its dependencies like rtree/libspatialindex) is x64.  The binaries that _result_ from the execution of `setup.py` by `cross-python` is the target architecture (e.g. aarch64).  However, this means that if we import `pygeoprocessing` inside of `setup.py`, it's trying to load an x64 binary from an "aarch64" (spoofed) python, which is why it couldn't find a compatible `libspatialindex_c` binary and the cross-compiled builds were failing.

Fortunately, we only need the location of the `pygeoprocessing` package, which we can get from the package metadata!  And that's what this change does.

Plus, not importing the entirety of pygeoprocessing and its dependencies should make `setup.py` executions a whole lot faster.

Fixes:#2630

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -66,6 +66,13 @@
 Unreleased Changes
 ------------------
 
+General
+=======
+* Fixed an issue with cross-compiling ``natcap.invest`` for non-native
+  architectures, particularly on conda-forge.  This should also speed up
+  the operation of ``setup.py``.
+  (`#2630 <https://github.com/natcap/invest/issues/2630>`_)
+
 Workbench
 =========
 * Added a file browse button to the input field for selecting a
--- a/setup.py
+++ b/setup.py
@@ -1,16 +1,28 @@
+import importlib.util
 import os
 import platform
 import subprocess
 
 import numpy
-import pygeoprocessing
 from Cython.Build import cythonize
 from setuptools import setup
 from setuptools.command.build_py import build_py as _build_py
 from setuptools.extension import Extension
 
-include_dirs = [numpy.get_include(),
-                os.path.join(pygeoprocessing.__path__[0], 'extensions')]
+include_dirs = [numpy.get_include()]
+
+pygeoprocessing_spec = importlib.util.find_spec("pygeoprocessing")
+if pygeoprocessing_spec is not None and pygeoprocessing_spec.submodule_search_locations:
+    # Don't import all of pygeoprocessing, just get the include dir from
+    # package metadata.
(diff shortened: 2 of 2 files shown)


davemfish:
The failures here are all related to #2670 and can be ignored if we don't want to wait for that fix.

emlys (approved):
Wow, this is a great workaround and it also might save a lot of trouble I've had with the build environment! We might not need to disable build isolation any more - if this means we don't have to worry about libgdal version compatibility because gdal isn't being imported?

As something to think about for the future, it would be nice if pygeoprocessing could provide a method or attribute for other packages to easily discover its include dir location (https://github.com/natcap/pygeoprocessing/issues/480). That's not currently possible without requiring all the dependencies, because pygeoprocessing's `__init__.py` imports all its submodules.

emlys:
The test failures do look unrelated, so I'll merge

phargogh:
>if this means we don't have to worry about libgdal version compatibility because gdal isn't being imported?

I haven't tested this yet, but I _think_ we still need to worry about the libgdal version available because we still need to compile against libgdal version X and then run the compiled binaries with libgdal version X installed.  So I think we still need a way to get the gdal development headers into the environment or else we might still be stuck with disabling build isolation for now.

>As something to think about for the future, it would be nice if pygeoprocessing could provide a method or attribute for other packages to easily discover its include dir location (https://github.com/natcap/pygeoprocessing/issues/480). That's not currently possible without requiring all the dependencies, because pygeoprocessing's __init__.py imports all its submodules.

Yeah, I totally agree this would be great to have!  And I have an idea for a workaround until [PEP 810 (explicit lazy imports)](https://peps.python.org/pep-0810/) is released in python 3.15, although lazy imports would be so so great to have for so many reasons aside from this.  I'll try to put together a proof-of-concept and PR for this in pygeoprocessing and tag you on it. (Update: implemented in https://github.com/natcap/pygeoprocessing/pull/487)

PR #2672, opened 2026-07-22
Refactor server startup logic to better handle errors

### LINKED ISSUE #2637: Workbench can get into an inconsistent state when reopening a plugin
## Description
When developing a plugin in the workbench, a standard process is:
- Install the plugin initially via the workbench interface
- Activate the plugin env in your terminal (path to the env found in the workbench's config.json)
- Editable install the plugin into that env
- Quit and reopen the workbench (reloading a package that has switched to an editable install doesn't work)
- Now, changes to the plugin source code should show up in the workbench when you close and reopen the plugin.

Sometimes however, an error on plugin startup isn't being caught by the workbench main process. The workbench main process log shows the traceback, but continues to retry connecting to the server that failed to start. The workbench UI will show "Starting up plugin" until the retries run out, and then subsequently closing and reopening the plugin does nothing - you immediately get the "Failed to launch plugin" screen. 

Closing and reopening the workbench clears this state, which is not too much extra trouble, but the workbench should be catching all plugin errors.

Note that reinstalling the plugin is still not necessary.

emlys:
## Description
Fixes #2637 
Refactored the server startup process to solve a couple of issues:
- 'exit' events from the server process were not be handled other than to log them. Now an exit with non-zero code will cause the "Failed to launch plugin" screen to be displayed.
- This logic is now integrated with the server startup polling, so that it does not continue to retry connecting to the server after an exit.

Also improved the plugin 'failed to launch' screen, to direct users to find their logs.

## Checklist

### DIFF
--- a/workbench/src/main/createPythonFlaskProcess.js
+++ b/workbench/src/main/createPythonFlaskProcess.js
@@ -35,6 +35,8 @@ export async function handleServerStartup(pythonServerProcess, url, maxRetries=5
   pythonServerProcess.stderr.on('data', (data) => {
     logger.debug(`${data}`);
   });
+  // The 'error' event happens when the child process fails to spawn.
+  // This should be rare.
   pythonServerProcess.on('error', (err) => {
     logger.error(pythonServerProcess.spawnargs);
     logger.error(err.stack);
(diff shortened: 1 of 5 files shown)


davemfish (changes requested):
@emlys It looks like there's a bug somewhere in the server startup. Workbench tests that start the flask app are failing.

emlys:
@davemfish the workbench tests are passing now.

workbench/src/main/createPythonFlaskProcess.js
```diff
@@ -20,45 +20,55 @@ async function getFreePort() {
...
-  try {
-    await fetch(`${HOSTNAME}:${port}/api/ready`, {
-      method: 'get',
-    });
-  } catch (error) {
-    if (error.code === 'ECONNREFUSED') {
-      while (i < retries) {
-        i++;
-        // Try every X ms, usually takes a couple seconds to startup.
+async function getFlaskIsReady(url, signal, maxRetries = 500) {
```
davemfish, line 31: I see the default number of retries changed a lot, but I guess the beauty of this new pattern is we never have to wait until we exhaust the retries to find out something went wrong.

workbench/src/main/createPythonFlaskProcess.js
```diff
@@ -20,45 +20,55 @@ async function getFreePort() {
...
+  let retries = 0;
+  while (!signal.aborted && retries < maxRetries) {
```
davemfish, line 33: thanks for refactoring this so it is no longer a recursive function.

davemfish (approved):
This looks like a nice improvement overall to how this startup process is managed.

workbench/src/main/createPythonFlaskProcess.js
emlys: it was 500 retries before - the function defaulted to 41, but was overridden to 500 where it was called.

PR #2676, opened 2026-07-24
Updated snapshot test for Matplotlib 3.11.1 compatibility

### LINKED ISSUE #2670: A change in matplotlib 3.11.1 causes `test_plot_raster_list_wide_nominal` to fail
## Description

A change in `matplotlib` `3.11.1` is causing one of the `raster_utils` test images (`plot_raster_list_wide_nominal_fail.png`) to again be slightly incorrect in size and raise this error:

`FAILED tests/reports/test_raster_utils.py::RasterPlotLegendTests::test_plot_raster_list_wide_nominal - AssertionError: ('Image sizes do not match expected size: (311, 804, 3) actual size (311, 805, 3)', 'actual image saved to data\\invest-test-data\\reports\\snapshots\\plot_raster_list_wide_nominal_fail.png')
`

I tested this locally to confirm: `test_plot_raster_list_wide_nominal` passes when I use `matplotlib` `3.11.0` but fails if `matplotlib=3.11.1.` 

This is likely due to this change: https://github.com/matplotlib/matplotlib/pull/32038, which changed canvas dimensions to avoid losing a pixel due to weird floating point values. More specifically, in `get_width_height` "This method normally truncates the height/width to remove any fractional pixel. However, if the height/width is extremely close to the integer pixel (within 1e-8 pixel), the height/width is instead rounded up to account for floating-point precision effects."

To avoid having to regularly update snapshots, our plan is to unit test `_get_legend_kwargs` instead and remove all tests that do image comparisons for plots with legends, since we now plot categorical rasters with separate attribute tables instead of legends.

davemfish:
The latest matplotlib release broke one of our snapshot tests. I updated the snapshot in the test data repo. I considered refactoring some of the legend tests do use assertions based on something other than a snapshot so we can avoid having to update snapshots for stuff like this, but ultimately I wasn't very pleased with those options.

Making assertions based on attributes of a `Figure` object and the bounding box and anchor position of the legend felt like it would be based on too many implementation details of matplotlib. And unit-testing `_get_legend_kwargs` felt like just testing "magic numbers" that we came up with by experiment. The snapshot tests have value during development because they give us a convenient way to view the plots we create, in addition to guarding against regressions. Those other tests would only guard against regressions.

If matplotlib releases continue to break our snapshot tests, we can explore those options again.

Fixes #2670

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GIT_SAMPLE_DATA_REPO_REV    := 098183806c0c715e6422d4d0570db91321c0f4e8
 
 GIT_TEST_DATA_REPO          := https://bitbucket.org/natcap/invest-test-data.git
 GIT_TEST_DATA_REPO_PATH     := $(DATA_DIR)/invest-test-data
-GIT_TEST_DATA_REPO_REV      := e5541ca176a3ac1c6cdca9c0159b0f1b94c47e69
+GIT_TEST_DATA_REPO_REV      := 44a4fdcc550c094050c9d4247d5f8f1ca1c2488e
 
 GIT_UG_REPO                 := https://github.com/natcap/invest.users-guide
 GIT_UG_REPO_PATH            := doc/users-guide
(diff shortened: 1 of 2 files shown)


claire-simpson (approved):
Thanks for looking into these options @davemfish!

PR #2680, opened 2026-07-29
no need to check the micromamba exe on Workbench startup.

### LINKED ISSUE #2656: Workbench: the check for `micromamba` passes when it shouldn't
We execute the `micromamba` exe to check if it works. In my case, it does not return an `error` object, but it does return some `stderr`: `The system cannot find the path specified.`

```javascript
/**
 * Return the available micromamba executable.
 *
 * @param {boolean} isDevMode - a boolean designating dev mode or not.
 * @returns {string} micromamba executable.
 */
export function findMicromambaExecutable(isDevMode) {
  let micromambaExe;
  if (isDevMode) {
    micromambaExe = 'micromamba'; // assume that micromamba is available
  } else {
    micromambaExe = `"${upath.join(process.resourcesPath, 'micromamba')}"`;
  }
  // Check that the executable is working
  const { stderr, error } = spawnSync(micromambaExe, ['--help'], { shell: true });
  if (error) {
    logger.error(stderr.toString());
    logger.error('micromamba executable is not where we expected it.');
    throw error;
  }
  logger.info(`using micromamba executable '${micromambaExe}'`);
  return micromambaExe;
}
```

davemfish:
I decided that there is not much value in running a `micromamba` command every time the Workbench launches, just to test the exe. #2656 describes a bug in that function.

In production, we point to the `micromamba` packaged with the Workbench. If for some reason that exe doesn't work, the new Advanced Plugins options are the user's main recourse. Many Workbench users will not end up using plugins or `micromamba`, so it seems okay to wait until they do to discover if there is a problem, rather than alert users to a problem that may not effect them.

In dev mode it is arguably more useful to test that the local `micromamba` works, since it's not using the one packaged with the Workbench. But in this case as well, I think it's okay to wait until the app tries to use `micromamba` to discover if there's a problem, and then suggest the user re-configure things via "Manage Plugins".

The case on my system in dev mode was weird: the check using `spawnSync(micromamba --help)` failed to find the `micromamba` command. But opening a plugin with `spawn(micromamba ...)` _did_ work. Something to do with the shell? I do have `micromamba.bat` on the PATH, but that's not usable by my `git-bash` shell. Anyway, the checks we were doing were buggy and unnecessary.

Fixes #2656

Most of the changes to `plugin.test.js` were just to resolve warnings from React about uncontrolled components changing to controlled components.

## Checklist

### DIFF
--- a/workbench/src/main/createPythonFlaskProcess.js
+++ b/workbench/src/main/createPythonFlaskProcess.js
@@ -52,14 +52,16 @@ async function getFlaskIsReady(url, signal, maxRetries = 500) {
 }
 
 /**
- * Wait for a invest server process to start up, handling any errors.
+ * Wait for an invest server process to start up, handling any errors.
  * @param {ChildProcess} pythonServerProcess - server process instance.
+ * @param {string} modelID - the plugin key used in the settings store,
+ *        or 'core' if the model is not a plugin.
  * @param {string} url - URL to poll for server status
  * @returns {number} PID of the started process, or undefined if it fails to launch
  */
-export async function handleServerStartup(pythonServerProcess, url) {
+export async function handleServerStartup(pythonServerProcess, modelID, url) {
   const controller = new AbortController();
-  const signal = controller.signal;
+  const { signal } = controller;
 
   pythonServerProcess.stdout.on('data', (data) => {
     logger.debug(`${data}`);
@@ -82,12 +84,21 @@ export async function handleServerStartup(pythonServerProcess, url) {
   // typo in the micromamba command or a plugin failing to import
   pythonServerProcess.on('exit', (code) => {
     logger.debug(`Flask process exited with code ${code}`);
-    if (code != 0) {
+    if (code !== 0) {
       controller.abort();
     }
+    if (modelID !== 'core') {
+      // Clear these because the process is no longer running and these
+      // values signal to the renderer if it needs to launch the process
+      // or not when the model tab is opened.
+      // If it was the core model process, the user must close and re-open
+      // the Workbench anyway, which will clear these values.
+      settingsStore.set(`plugins.${modelID}.pid`, '');
+      settingsStore.set(`plugins.${modelID}.port`, '');
+    }
   });
-  pythonServerProcess.on('close', (code, signal) => {
-    logger.debug(`Flask process closed with code ${code} and signal ${signal}`);
+  pythonServerProcess.on('close', (code, sig) => {
+    logger.debug(`Flask process closed with code ${code} and signal ${sig}`);
   });
   pythonServerProcess.on('disconnect', () => {
     logger.debug('Flask process disconnected');
@@ -125,7 +136,7 @@ export async function createCoreServerProcess(_port = undefined) {
 
   logger.debug(`Started python process as PID ${pythonServerProcess.pid}`);
   const pid = await handleServerStartup(
-    pythonServerProcess, `${HOSTNAME}:${port}/api/ready`);
+    pythonServerProcess, 'core', `${HOSTNAME}:${port}/api/ready`);
   return pid;
 }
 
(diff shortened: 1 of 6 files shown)


workbench/src/main/createPythonFlaskProcess.js
```diff
@@ -82,12 +84,21 @@ export async function handleServerStartup(pythonServerProcess, url) {
...
+    if (code !== 0) {
       controller.abort();
     }
+    if (modelID !== 'core') {
+      // Clear these because the process is no longer running and these
+      // values signal to the renderer if it needs to launch the process
+      // or not when the model tab is opened.
+      // If it was the core model process, the user must close and re-open
+      // the Workbench anyway, which will clear these values.
+      settingsStore.set(`plugins.${modelID}.pid`, '');
```
emlys, line 96: I was thinking recently would it make sense to stop tracking the plugin server port and PID in the settings? They could probably be tracked in a javascript object instead, then we wouldn't have to worry about clearing the values. 

It could also simplify things to let each plugin tab create its own server process, rather than checking if one already exists for that plugin. I doubt we'd ever have enough tabs open that it would be a problem.

emlys: (Just a thought for a future issue, nothing you need to do here)

workbench/src/renderer/components/InvestTab/index.jsx
```diff
@@ -228,16 +228,21 @@ class InvestTab extends React.Component {
...
-            {t('Find My Logs')}
-          </Button>
+          <p>{t('Failed to launch plugin. Check the workbench logs.')}</p>
+          <div className="text-center mt-3">
+            <Button onClick={handleClickFindLogfiles}>
+              {t('Find My Logs')}
+            </Button>
+          </div>
+          <p>
+            {t(`You may wish to manually configure the conda executeable
```
emlys: I'm not sure if I agree that it makes sense to prompt the user to configure the conda settings... a plugin can fail to launch due to any sort of error, and I wouldn't want to send users down a rabbit hole thinking they can fix it by changing the environment when that's not the case.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -512,7 +512,7 @@ export default function PluginModal(props) {
             <Form.Control
               id="condaPath"
               type="text"
-              value={condaPath}
+              value={condaPath || ''}
```
emlys, line 515: In what case is this needed (`condaPath` becomes undefined?)

workbench/tests/renderer/plugin.test.js
```diff
@@ -17,7 +17,16 @@ import App from '../../src/renderer/app';
 
 jest.mock('../../src/renderer/server_requests');
 
-describe('Add plugin modal', () => {
+const PLUGIN_SETTING_ITEM = {
```
emlys, line 20: Nice to reuse this!

workbench/src/main/createPythonFlaskProcess.js
davemfish: Yeah, I like that idea. But I'm torn because I also like that the second time you open a plugin tab the server is already running and there's no waiting for it to startup. I guess that could still be achieved just by storing the data in a parent component.

workbench/src/renderer/components/InvestTab/index.jsx
davemfish: Yeah, that's true. Okay I'll get rid of it.

workbench/src/renderer/components/PluginModal/index.jsx
davemfish: Only in some tests because neither the `microcmamba` or `userDefinedMicromamba` properties are present in the mocked settings. We could mock it, but this was just a much easier fix and generally not a bad practice either, I don't think.

emlys (approved):
Thanks @davemfish!

PR #2682, opened 2026-07-29
Reports: support "userguide" link for plugins

### LINKED ISSUE #2640: Reports: UG link in base template doesn't make sense for plugins
The base template for reports includes a programmatically-generated link to the User Guide. This works great for core models, but it doesn't make sense for plugins, since a plugin doesn't have an entry in the InVEST User Guide.

It would be nice to determine whether the calling reporter is defined in a core model or in a plugin. If it's a plugin, we should skip the UG link, but if the plugin's model spec specifies a "userguide" URL, we could render a link to the plugin documentation (analogous to the link rendered in the Workbench).

emilyanndavis:
## Description
Fixes #2640.

### Plugin with non-empty `userguide_page` links to the specified URL
report_ug_link-plugin-2026_07_29

### Core model links to InVEST User Guide (no change)
report_ug_link-core_model-2026_07_29

### Report with undefined `userguide_page` renders no link (mocked by temporarily removing that arg from the carbon reporter)
report_ug_link-empty_userguide-2026_07_29

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -78,21 +78,24 @@ Workbench
 * Added a file browse button to the input field for selecting a
   local plugin directory to install.
   (`#2641 <https://github.com/natcap/invest/issues/2641>`_)
-* Bootstrap and React Bootstrap dependencies have been updated to their latest
-  stable versions (Bootstrap 5.3.8, React Bootstrap 2.10.10).
-  (`#1798 <https://github.com/natcap/invest/issues/1798>`_)
 * Added an option in the Manage Plugins modal to configure which
   conda executable is used to run plugins.
   (`#2645 <https://github.com/natcap/invest/issues/2645>`_)
 * Added an option in the Manage Plugins modal to configure which
   conda environment is used to run each plugin.
   (`#2646 <https://github.com/natcap/invest/issues/2646>`_)
-* React dependency has been updated to its latest stable version (19.2.7).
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -78,21 +78,24 @@ Workbench
...
-* React dependency has been updated to its latest stable version (19.2.7).
-  (`#2481 <https://github.com/natcap/invest/issues/2481>`_)
-* Improved handling of errors on plugin server startup
+* Improved handling of errors on plugin server startup.
   (`#2637 https://github.com/natcap/invest/issues/2637`)
 * The "View Results" button now supports plugins.
   (`#2638 <https://github.com/natcap/invest/issues/2638>`_)
+* The base report template now better supports plugins by including a link to
+  plugin-specific documentation instead of a link to the InVEST User Guide.
+  (`#2640 <https://github.com/natcap/invest/issues/2640>`_)
```
emilyanndavis, line 93: This is the new entry. Existing entries had landed in a slightly strange order, so I rearranged them a bit.

megannissel (approved):
Thanks, @emilyanndavis!

I wondered if we could leverage jinja globals along with some of the logic in `models.py` to programmatically detect the IDs of core InVEST models so that we wouldn't need a hard-coded list. But the same logic that makes plugin discovery work so smoothly makes distinguishing between the two less straightforward within the plugin environment. As you point out, the list should only require updates if we add or remove a core model, so maintenance shouldn't be burdensome.

PR #2683, opened 2026-07-30
Support drag-and-drop in plugins modal

### LINKED ISSUE #2642: Dragging-and-dropping a plugin directory into the plugin local file path input creates a new window
To reproduce:

1. Open the workbench plugin dialog (I'm working on the official mac binaries from InVEST 3.20.0)
3. In the Add a Plugin section, select "Local Path"
4. Drag a directory from the finder in to the local absolute path input
5. Observe issue

https://github.com/user-attachments/assets/399b146a-fb86-43ed-ab40-cbc415ef7ce1

claire-simpson:
## Description
Fixes a bug where dragging-and-dropping a plugin directory into the local absolute path would open a new window. Now this filepath populates correctly. I also added drag-and-drop support for the conda executable and plugin environments inputs. Finally, if the user intends to install a plugin from a git URL, dragging-and-dropping a directory does nothing (vs. previously this would also open a new blank window).
 
Fixes #2642 

## Checklist

### DIFF
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -17,7 +17,7 @@ import {
 import { openLinkInBrowser } from '../../utils';
 import { ipcMainChannels } from '../../../main/ipcMainChannels';
 
-const { ipcRenderer } = window.Workbench.electron;
+const { getFilePath, ipcRenderer } = window.Workbench.electron;
 
 export default function PluginModal(props) {
   const {
@@ -208,6 +208,22 @@ export default function PluginModal(props) {
     }
   };
 
+  const handlePluginPathDragOver = (event) => {
+    event.preventDefault();
+    event.stopPropagation();
+    event.dataTransfer.dropEffect = 'copy';
+  };
+  
+  const handlePluginPathDrop = (event) => {
+    event.preventDefault();
+    event.stopPropagation();
+  
+    const droppedFile = event.dataTransfer.files[0];
+    if (droppedFile) {
+      setPath(getFilePath(droppedFile));
+    }
+  };
+
   const selectFile = async (event) => {
     const data = await ipcRenderer.invoke(
       ipcMainChannels.SHOW_OPEN_DIALOG, { properties: ['openFile'] }
@@ -309,6 +325,8 @@ export default function PluginModal(props) {
               : 'C:\\Documents\\path\\to\\plugin\\'}
             value={path}
             onChange={(event) => setPath(event.currentTarget.value)}
+            onDragOver={handlePluginPathDragOver}
+            onDrop={handlePluginPathDrop}
             aria-describedby={pluginSourceMissingError ? 'path-error' : ''}
           />
           <Button
--- a/workbench/src/renderer/components/PluginModal/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/index.jsx
@@ -208,20 +208,23 @@ export default function PluginModal(props) {
     }
   };
 
-  const handlePluginPathDragOver = (event) => {
+  const dragOverHandler = (event) => {
     event.preventDefault();
     event.stopPropagation();
     event.dataTransfer.dropEffect = 'copy';
   };
   
-  const handlePluginPathDrop = (event) => {
+  const getDroppedFilePath = (event) => {
     event.preventDefault();
     event.stopPropagation();
   
     const droppedFile = event.dataTransfer.files[0];
-    if (droppedFile) {
-      setPath(getFilePath(droppedFile));
-    }
+    return droppedFile ? getFilePath(droppedFile) : undefined;
+  };
+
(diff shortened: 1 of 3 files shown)


workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -309,6 +330,13 @@ export default function PluginModal(props) {
               : 'C:\\Documents\\path\\to\\plugin\\'}
             value={path}
             onChange={(event) => setPath(event.currentTarget.value)}
+            onDragOver={dragOverHandler}
```
dcdenu4, line 375: For our other drag and drop functionality it looks like we handle `onDragEnter` and `onDragLeave`. I think we use these in the `SetupTab/ArgInput/index.jsx` for InVEST model inputs to better signal to the user when you can drag and drop something over an input. Maybe it'd be worth adding the same type of functionality? Then we could use the `event.currentTarget.classList.add('input-dragging')` styling?

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -208,6 +208,25 @@ export default function PluginModal(props) {
     }
   };
 
+  const dragOverHandler = (event) => {
+    event.preventDefault();
+    event.stopPropagation();
+    event.dataTransfer.dropEffect = 'copy';
+  };
+  
+  const getDroppedFilePath = (event) => {
```
dcdenu4: It could be worth taking a look at our other implementation of this `inputDropHandler` to adopt the precaution we take there around more than one file being dropped and some light error handling if something happens to go wrong.

dcdenu4: We expect this to be a directory for the local plugin path right? Do we currently do any validation or checking if someone were to enter a path / file that wasn't a directory?

dcdenu4 (changes requested):
Thanks @claire-simpson. I had a few suggestions and questions. I was also wondering if we wanted to add any tests related to this? Do we have some from our other drag and drop functionality we could steal from?

workbench/src/renderer/components/PluginModal/index.jsx
claire-simpson: As soon as the user clicks `Add` plugin, a `not a directory` error is raised, but there isn't error handling up front

workbench/src/renderer/components/PluginModal/index.jsx
claire-simpson: Ok I've that functionality, and I agree this could be abstracted out if another use case crops up!

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -208,6 +208,67 @@ export default function PluginModal(props) {
...
 
+  /**
+   * Prevent the default case for onDragOver so onDrop event will be fired.
+   *
+   * @param {Event} event - dragover event
+   */
+  function dragOverHandler(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    if (event.currentTarget.disabled) {
```
dcdenu4, line 219: I think it's fine to leave this disabled check in, but are we disabling it under any condition?

workbench/tests/renderer/plugin.test.js
```diff
@@ -346,6 +346,59 @@ describe('Manage Plugins modal', () => {
...
+    const {
+      findByText, findByRole, findByLabelText,
+    } = render(<App />);
+
+    const spy = jest.spyOn(ipcRenderer, 'send');
+    await userEvent.click(await findByRole('button', { name: 'menu' }));
+    const managePluginsButton = await findByText(/Manage plugins/i);
+    await userEvent.click(managePluginsButton);
+
+    const input = await findByLabelText('Conda or mamba executable');
```
dcdenu4: I'm not quite sure why there's an error showing here given you're building off a previous test.

dcdenu4 (dismissed):
Hey @claire-simpson thanks for those changes. There's one test failure showing that I can't explain and then just had one comment / question about needing to check against `disabled`, which I ultimately think is fine to leave in.

workbench/src/renderer/components/PluginModal/index.jsx
claire-simpson: No, as of now there is not a scenario where these [executable, environment, or plugin path] fields would become disabled. I left these in as more of a future-proof safeguard (as I could see us doing something like disabling the environment fields unless the user explicitly clicks an "advanced options" button or something)

davemfish (approved):

davemfish:
I tested this out; it works great. Tests look good. @claire-simpson , in the future if you find yourself doing a lot of repetitive setup in a series of tests (e.g. mocking the same function in the same way) that can be a good reason to nest a group of tests within a `describe` block and put all that common setup into the `dsecribe's` `beforeEach` block.

PR #2684, opened 2026-08-03
Revise auto release process

### LINKED ISSUE #2504: Autorelease PRs trigger duplicate CI runs
## Background/Motivation
We have a lot of seemingly duplicate tests run during the release process, and specifically when merging a release branch into main and in the PR that is opened when you trigger release part 1 (i.e., to merge `autorelease/*` into `main`). This is because our `build-and-test` workflow is triggered on both `push` and `pull_request`, and there are multiple pushes. This leads to redundant execution of time consuming test matrices and build steps, which prolongs the release process.

Using `3.19.0` as an example, this is what occurs:
1. on push to `autorelease/3.19.0` --> triggers full push workflow
2. push of tag 3.19.0 --> triggers another full push workflow (this one is important for release artifacts, as we note in the auto PR)
3. pull request from `autorelease/3.19.0` into `main` --> triggers full pull_request workflow (also probably necessary since this is checking the merge)

## Description
It would be nice if we could reduce the number of expensive tests and potentially decrease time it takes to complete a release. We could ignore the tests we don't care about, but it would be better not to run them at all.

Could we skip tests triggered by push to `autorelease/*`? These tests seem redundant given that the PR run tests the merged result and the tag push run produces the release artifacts. In addition: are there other tests we might be able to skip?

## Mitigation of side effects
- Are any push-only side effects (e.g., deploys, artifact uploads, and codesigning are only run on `push`, not on `pull_request`) currently relying on the `autorelease/*` push run?
- Any part of release parts 1 or 2 depend on artifacts produced specifically from that run?

emlys:
## Description
Fixes #2504 
This updates the automated release workflows to remove the intermediate PR step. The whole release will now happen on the `autorelease/X.Y.Z` branch, and that branch will be merged into `main` after the release is completed. This has some benefits:
- The release process only has to wait for 2 sets of checks to run, rather than 3, because the PR checks happen after the release is done (and don't really matter at that point anyway)
- It's easier to roll back changes if there's a problem in the second release workflow, because nothing has been merged into `main` yet

I also moved the rollback process into a separate manually triggered workflow, so that won't happen automatically on failure, but we can decide when to run it.

## Checklist


dcdenu4 (approved):
Thanks @emlys , this all looks good to me. Do you mind updating our release checklist wiki to reflect the changes in the manual steps needed?

Issue #856, opened 2021-11-02
address dependabot security alerts


davemfish:
At or before the beta-release, we should clear the queue of suggested security updates. And also figure out the right way to manage this. The alerts are nearly always about packages that we don't directly specify in `package.json`. 

So is the correct approach to directly edit the `yarn.lock` file? We never need to manually touch that file for any other reason, and that seems like a good thing. Before resorting to that, we could try updating everything listed in `yarn outdated` and see if that addresses everything.

phargogh:
We discussed in our team meeting today that we would reenable dependabot PRs.

phargogh:
Dependabot will now alert us for malware detected in dependencies and will now create PRs for security updates.  I've also enabled "grouped security updates" to try to keep the PR numbers down. Hopefully this won't cause problems for us as we're catching up!

Non-security updates like package version bumping are disabled.

PR #2690, opened 2026-08-04
Add dependabot configuration for non-security updates

phargogh:
This PR adds supplementary dependabot configuration to trigger version bumps for github actions and docker builds.  I'm hoping that this will help us keep on top of action version updates (in GHA) and updated debian releases (in docker).

Note that this does not affect security PRs, which are controlled through repo settings and not through this configuration file.

RE:#856


davemfish (approved):
This is a very welcome addition, thank you.

PR #2696, opened 2026-08-04
Speedup for Workbench tests that sometimes timed out

### LINKED ISSUE #2618: GHA: `autotest` during a PR tries and fails to deploy reports
The `invest autotest` step normally doesn't run in the context of a PR, except if the head is an `autorelease` branch. In that case the test runs, but then the following step to deploy the reports to GCS fails, probably because we did not initialize `gcloud` earlier on. Normally we only deploy to GCS when `github.event_name != 'pull_request'`.

Either we should avoid trying to deploy reports in this case, or we should avoid running the autotest altogether. I'm not sure there's any benefit to running the test in a PR like this, because there's also a parallel Actions workflow running on the `push` to the `autorelease` branch. Though currently the autotest does not run there:

```
      - name: Run invest-autotest with binaries
        id: invest-autotest
        if : |
          (github.event_name == 'push' &&
          (startsWith(github.ref, 'refs/heads/release') || github.ref == 'refs/heads/main')) ||
          (github.event_name == 'pull_request' && startsWith(github.head_ref, 'autorelease'))
        run: |
          mkdir autotest_dir
          make invest_autotest AUTOTEST_DIR=autotest_dir
```

If we stop running the test in the PR, we probably want to start running it on pushes to `autorelease`

https://github.com/natcap/invest/actions/runs/27349971166/job/80808888529?pr=2617

### LINKED ISSUE #2695: Workbench tests timed out on MacOS runners
` tests/renderer/invest_subprocess.test.js` tests have recently started failing somewhat regularly on MacOS runners. Each test has Jest's default 5 second timeout period, and individual tests were sometimes taking longer than that. In addition, sometimes the workspace input field that the test was typing into did not end up with expected value.

These two things led me here: https://github.com/testing-library/user-event/discussions/977

After replacing calls to `userEvent.type` with `userEvent.paste`, I noticed the tests running noticeably faster for me locally. For example, 8 seconds to run that test file instead of 15. With lots of variance.

davemfish:
Pasting text is a lot faster than typing. I also re-structured the use of the temp workspaces to make extra sure that each test was isolated from each other. See more details in #2695 

As a bonus I also fixed #2618 here instead of opening another PR and adding to the backlog of queued jobs. 

Fixes #2695 

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -441,8 +441,7 @@ jobs:
         id: invest-autotest
         if : |
           (github.event_name == 'push' &&
-          (startsWith(github.ref, 'refs/heads/release') || github.ref == 'refs/heads/main')) ||
-          (github.event_name == 'pull_request' && startsWith(github.head_ref, 'autorelease'))
+          (startsWith(github.ref, 'refs/heads/release') || startsWith(github.head_ref, 'autorelease') || github.ref == 'refs/heads/main'))
         run: |
           mkdir autotest_dir
           make invest_autotest AUTOTEST_DIR=autotest_dir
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks @davemfish!

PR #2698, opened 2026-08-04
Draft a security policy

### LINKED ISSUE #2697: Add a SECURITY.md
We should have a defined security policy, which github recognizes as SECURITY.md.

phargogh:
This PR drafts a security policy, which I have tried to customize for the InVEST use case with feedback from Claude.  Any and all feedback welcome!

Fixes #2697 

## Checklist


davemfish:
I have attempted to cancel the workflows because they are not relevant to this PR and we are totally bogged down by the number of open PRs right now.

davemfish (approved):

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Out of Scope
+
+InVEST is distributed as desktop, CLI, and Python-library software that
+users run locally against data they supply themselves. Given this, the
+following are considered out of scope:
+
+* Theoretical attacks without a realistic proof of concept.
+* Standard automated scanner outputs without actionable validation.
+* Issues that assume a level of local access equal to or greater than what
```
claire-simpson: Why would assuming a level of access equal to what InVEST requires be considered out of scope?

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+  running InVEST itself would grant.
+* Denial-of-service or resource-exhaustion from a user's own oversized or
+  malformed input data on their own local run. (Note: a *crafted* file
+  that causes code execution or data exfiltration when opened by another,
+  unsuspecting user — e.g., a shared malicious raster/vector/plugin file —
```
claire-simpson: Is "plugin file" here referring to an InVEST plugin? If so, I think we should be more explicit and verbose about this, as plugins are the biggest question mark for me related to this

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+Please **do not open public GitHub issues** for security vulnerabilities.
+
+To report a vulnerability, please use one of the following private channels
+(GitHub Private Vulnerability Reporting preferred):
+
+1. **GitHub Private Vulnerability Reporting**: Navigate to the "Security" tab of
+   this repository, click "Vulnerability reporting", and select "Report a vulnerability".
+2. **Email**: Send a detailed report to <natcap-software@lists.stanford.edu>.
+
+Please include the following details in your report:
```
claire-simpson, line 31: Should we also request the version(s) of InVEST? (Or do we assume the version is just the latest?)

claire-simpson (changes requested):
Thanks @phargogh, I had a few clarifying questions

SECURITY.md
phargogh: Well, the thinking is that if some already has the level of access that InVEST has and can run an exploit, then why go through the trouble of compromising InVEST when they could just run the exploit?  But, at the same time, it does seem reasonable that someone _could_ exploit InVEST to do something malicious, so I'm inclined to remove the `equal to or` from the sentence.  I've patched this in 7d0b86d60 in case you agree.

SECURITY.md
phargogh: Yeah, I agree, and I also think this is better covered by the [security policy in the plugins registry](https://github.com/natcap/invest-plugin-registry/blob/main/SECURITY.md) (even if we could stand to spend some time on that file).  I've modified this sentence to specifically link to the plugin registry's security policy for further instructions in d1836ee58.  Seems more orthogonal that way than try to make this one SECURITY file all-inclusive across the InVEST ecosystem.

SECURITY.md
phargogh: Yeah, I think this would be wise.  One might infer this from the environment/configuration request, but I think it's helpful to ask specifically for this.  Patched in 6f322a071

SECURITY.md
claire-simpson: Great, thanks! I will note that I do not see the "Report a vulnerability" button in the registry repo. Perhaps Private Vulnerability Reporting has not been enabled?

claire-simpson (approved):
Looks good, thanks @phargogh! Note: If this gets merged after the next release, we should make sure that the history note is still in the correct place

SECURITY.md
phargogh: It's definitely enabled!  But based on the docs it sounds like maybe since you're an admin on this repo, maybe their recommendation is to create a security advisory directly instead of filing a report?  I'm able to see a button for "Report a vulnerability" when I visit https://github.com/natcap/invest/security in an incognito window, so I bet this is what's happening

claire-simpson: Oh interesting, would not have guessed being an admin meant I _wouldn't_ see a button 🤔

phargogh: Same here!  I found that to be very surprising behavior.

emlys (approved):
This all makes sense to me, thanks!

SECURITY.md
```diff
@@ -0,0 +1,70 @@
...
+
+* A description of the issue and its potential impact.
+* Steps to reproduce the bug (including proof-of-concept code if available).
+* Any specific configurations or environments required to trigger the issue.
+* The InVEST version or versions known to exhibit this issue.
+
+## Our Response Process
+
+* **Acknowledgment**: You will receive an acknowledgment within 72 hours.
```
lmandle: Does the 72 hours specification put someone on the hook for responding over 3 days weekends or over the Stanford winter close, and are we capable of/prepared to do that? Or should there be exceptions for those periods?

SECURITY.md
```diff
@@ -0,0 +1,70 @@
...
+* **Triage**: Our team will evaluate the report and update you on the status
+  within 7 days.
```
lmandle: Should there be an exception for over the Stanford winter close?

lmandle (commented):
Cool to see this! I added a couple of comments/questions about the specified response timeframes.

SECURITY.md
phargogh: Ooo yes, this is a great point.  I've added a caveat about university closures (Winter closure and otherwise) and that we'll do our best to respond but cannot commit to specified timelines during closures.  Do you think this will be sufficient?  Patched in 003a80840

SECURITY.md
phargogh: I think so, yes!  Same as above, I proposed a change for this exception in  003a80840

lmandle: Looks good to me!

lmandle (commented):
The updates re: university closures look good to me!

dcdenu4 (approved):
Thanks @phargogh ! Looks good to me.

megannissel (approved):
Looks good to me, thanks!

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Reporting a Vulnerability
+
+Please **do not open public GitHub issues** for security vulnerabilities.
+
+To report a vulnerability, please use one of the following private channels
+(GitHub Private Vulnerability Reporting preferred):
+
+1. **GitHub Private Vulnerability Reporting**: Navigate to the "Security" tab of
+   this repository, click "Vulnerability reporting", and select "Report a vulnerability".
```
emilyanndavis: I think things may have changed slightly since the time you first drafted this? The "Security" tab is now called "Security and quality", and the "Report a vulnerability" button seems to be available at the top of the tab—no intermediate step needed. (At least, that's what I'm seeing in the Plugin Registry repo while logged out, and I'm assuming it'll look the same for the InVEST repo once this document is published.) I've also taken the liberty of adding a hyperlink to the "Security and quality" tab.

```suggestion
1. **GitHub Private Vulnerability Reporting**: Navigate to the ["Security and quality" tab](https://github.com/natcap/invest/security) of
   this repository and select "Report a vulnerability".
```

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+* **Fix**: If verified, we will coordinate a patch and appropriate notification
+  across relevant channels and coordinate with you if an embargo is necessary.
+  Fixed vulnerabilities will be disclosed via a GitHub Security Advisory, with
+  credit to the reporter unless anonymity is requested.
+
+\*Please be aware that our software team adheres to the Stanford University
+administrative calendar (<https://cardinalatwork.stanford.edu/benefits-rewards/time-leaves/university-holiday-schedule>)
+which includes occasional holidays and a two-week winter closure.  During
+University closures, we will do our best to respond as soon as we can,
+but cannot guarantee adherence to specific timelines.
```
emilyanndavis: It's generally better to use more human-friendly language as the visible link text, especially when the URL is long like this one!

```suggestion
\*Please be aware that our software team adheres to the [Stanford University
administrative calendar](https://cardinalatwork.stanford.edu/benefits-rewards/time-leaves/university-holiday-schedule),
which includes occasional holidays and a two-week winter closure. During
University closures, we will do our best to respond as soon as we can,
but cannot guarantee adherence to specific timelines.
```

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+* Standard automated scanner outputs without actionable validation.
+* Issues that assume a level of local access greater than what
+  running InVEST itself requires.
+* Denial-of-service or resource-exhaustion from a user's own oversized or
+  malformed input data on their own local run. (Note: a crafted file
+  that causes code execution or data exfiltration when opened by another,
+  unsuspecting user, such as a shared malicious raster/vector file,
+  is in scope.  Security issues with plugins should be reported via the
+  instructions in the Plugin Registry repo:
+  <https://github.com/natcap/invest-plugin-registry>)
```
emilyanndavis:
```suggestion
  is in scope. Security issues with plugins should be reported via the
  instructions in the Plugin Registry repo:
  <https://github.com/natcap/invest-plugin-registry>.)
```

SECURITY.md
```diff
@@ -0,0 +1,76 @@
+# Security Policy
+
+This policy covers code and artifacts generated by the `natcap/invest`
+repository, including the InVEST models, command-line interface,
+workbench, and packaging related to distributing the `natcap.invest`
+python package on the Python Package Index.
```
emilyanndavis:
```suggestion
Workbench, and packaging related to distributing the `natcap.invest`
Python package on the Python Package Index.
```

emilyanndavis (changes requested):
Thanks for doing this, @phargogh! I have just some small suggested changes, the most notable of which aims to sync the instructions with what seems to be the current state of the GitHub UI. (The others are minor formatting/punctuation/capitalization suggestions I brought along for the ride.)

phargogh:
Thanks @emilyanndavis !  Yes, I think GitHub might have changed some things in the UI a bit in the last couple weeks, so thank you for the updates/edits, and for the good points about having links be readable.  My thought with having the links be rendered in plan text (albeit hyperlinked) was to make sure that it was clear where the link was going to ... but of course the whole document is open-source, so if someone really wants to confirm where a link goes, they can just grab the source code, so it makes lots of sense to have it be readable.

emilyanndavis (approved):

Issue #2699, opened 2026-08-05
Performance regression in Workbench input form fields


davemfish:
In the latest `main`, model input text fields are noticeably less responsive than they were in `3.20.0`. Enter text in a field by holding down a keypress and observe a lag.

I believe the regression is due to the React or Bootstrap updates we did recently. I tested the source code from `main` with the library versions prior to those upgrades and the performance was good, as in `3.20.0`. So I don't think any of our source changes are to blame, but perhaps we needed to refactor something in order to get the optimal performance out of React 19?

@emilyanndavis any thoughts?

davemfish:
Here are the profiles of a keystroke to a model input field.

### on `main`
input#workspace_dir-_r_3_.form-control.is-invalid **144 ms**
Phase Local duration (ms)
Input delay 1
Processing duration 136
Presentation delay 7

### on `3.20.0`
input#workspace_dir-:r0:.form-control.is-invalid **32 ms**
Phase Local duration (ms)
Input delay 0
Processing duration 19
Presentation delay 13

There is lots of variance around these, but the overall difference is consistent. The problem is in "processing duration", which refers to everything that happens within the event handler. In our case that is `updateArgValues`.

davemfish:
I still think the performance difference is a result of the React 19 upgrade, but I also think it reveals issues in our app that we should deal with. I'm not sure why React 18 seems to perform better for our app, but here's some things I've learned.

Here's the meaningful part of the event handler
```javascript
  updateArgValues(key, value) {
    const { argsValues } = this.state;
    argsValues[key].value = value;
    this.setState({
      argsValues: argsValues,
    }, () => {
      this.props.updateJobProperties(this.props.tabID, {
        status: undefined, // Clear job status to hide model status indicator.
      });
      this.debouncedValidate();
      this.debouncedArgsEnabled();
      this.debouncedDropdownFunctions();
    });
  }
```

The entire component tree, from `App` on down re-renders on every keystroke in one of these input fields! That's due to the `updateJobProperties` call.

Skipping the call to `updateJobProperties` and the `SetupTab` component still renders 4 times on a single keystroke. Probably due to the state updates in each of `this.debouncedValidate(); this.debouncedArgsEnabled(); this.debouncedDropdownFunctions();`. Those are debounced so the idea is that will _not_ be called on every keypress in a series of presses. But they do each have their own `setState` call which likely triggers a re-render.

[...]

PR #2701, opened 2026-08-06
Performance enhancements for React app; keystroke responsiveness

davemfish:
This partially addresses #2699. It checks whether or not `job` properties have actually changed before updating the list of jobs stored in `App` state. This improves the responsiveness of the App on changes to model input form fields because the `App` component is not re-rendered in the most common case. Single keystroke performance is down around ~45ms instead of ~200ms. There is still some slightly noticeable lag when holding down a key.

Also,
* an optimization for the `RecentInvestJobs` component because that was the worst performing component during renders.
* react-devtools is loaded during dev mode.

## Checklist


workbench/src/main/main.js
```diff
@@ -213,6 +217,9 @@ export function main() {
   }
 
   app.on('ready', async () => {
+    // installExtension([REACT_DEVELOPER_TOOLS])
+    //   .then(([react]) => console.log(`Added Extensions: ${react.name}`))
+    //   .catch((err) => console.log('An error occurred: ', err));
```
emilyanndavis: I assume it's OK to remove these commented-out lines now?

workbench/src/renderer/app.jsx
```diff
@@ -134,8 +134,17 @@ export default function App({isFirstRun = false, isNewVersion = false, nCPU = 1}
    * @param {boolean} save - if true, save the updated job to persistent store
    */
   async function updateJobProperties(tabID, jobObj, save = false) {
+    const existingJob = openJobs.get(tabID);
+    // It's worth checking before updating state because this component
+    // has so many children and most of the time the job properties have
+    // not changed.
+    const skipUpdate = Object.keys(jobObj).every(
+      (key) => existingJob[key] === jobObj[key]
+    );
```
emilyanndavis: This should catch and filter out most (maybe all) unnecessary updates, but if we want something more robust (now or later), I think we'd be better off writing an `equals` method on the `InvestJob` class (for example, so we can accurately compare non-primitives, such as `argsValues`).

workbench/src/renderer/components/HomeTab/index.jsx
```diff
@@ -173,54 +171,15 @@ function RecentInvestJobs(props) {
...
     if (job && job.argsValues && job.modelTitle) {
-      let badge;
-      if (job.type === 'plugin') {
-        badge = <Badge className="me-1" bg="secondary">Plugin</Badge>;
-      }
-      recentButtons.push(
-        <Card
-          className="col-12 text-start recent-job-card me-2 w-100"
+      const card = (
+        <RecentJobCard
```
emilyanndavis, line 175: Making this a separate component is a great readability improvement as well, thanks!

emilyanndavis (changes requested):
Thanks for taking this on, @davemfish! One minor thing, plus a recommendation for refactoring, but I don't think that's crucial right now. From what you showed us, these changes should provide noticeable improvements.

workbench/src/renderer/app.jsx
davemfish: That's a good point about `argsValues`. I like the idea of an `InvestJob` method for this, but in these cases the object passed to `updateJobProperties` is not an instance of `InvestJob`, it's just an object with a subset of the properties. So the new method might be something like `existingJob.contains(jobObj)`? Or `.isSuperSet`?

workbench/src/main/main.js
davemfish: Whoops forgot to push that commit; thanks!

davemfish:
I discovered that the `electron` or `electron-builder` upgrade here includes a breaking change where `ico` format files are no longer supported in the Mac DMG builds. So I replaced the `ico` files with our official InVEST `png` logo. I also updated a couple other places where we were already using an invest png logo in the app.

workbench/src/renderer/menubar/about.jsx
```diff
@@ -1,13 +1,12 @@
...
-import {createRoot} from 'react-dom/client';
+import { createRoot } from 'react-dom/client';
 import { Translation } from 'react-i18next';
 
 import i18n from '../i18n/i18n';
 
 import { handleClickExternalURL } from './handlers';
 import { ipcMainChannels } from '../../main/ipcMainChannels';
-import investLogo from '../static/invest-logo.png';
+import investLogo from '../static/InVEST-Logo-registeredtm_202305_150dpi.png';
```
emilyanndavis, line 9: It's a minor thing, but is there any reason we can't rename the image file rather than update these references? As long as we're using the same version of the logo everywhere, it seems most intuitive to me to call it `invest-logo`.

workbench/src/renderer/InvestJob.js
```diff
@@ -88,6 +88,25 @@ export default class InvestJob {
...
+    let isSuperSet;
+    isSuperSet = Object.keys(obj).every(
+      (key) => this[key] === obj[key]
+    );
+    if (!isSuperSet) { return false; }
+    if (Object.keys(obj).includes('argsValues')) {
+      isSuperSet = Object.keys(obj.argsValues).every(
+        (key) => this.argsValues[key] === obj.argsValues[key]
+      );
+    }
```
emilyanndavis: I think we need to check the argsValues first, before checking the other keys. If we don't, the check on line 99 will fail even if all the `argsValues` match, because when comparing objects, `===` will only return true when both objects reference the same location in memory. If `const a = {fruit: 'banana'}` and `const b = {fruit: 'banana'}`, `a !== b`, even though `a` and `b` contain identical information.

I was stumped at first, but I realized your test isn't catching this because the test case is referencing the same `baseArgsValues` object that is referenced in the comparison `job`. Change that to `{...baseArgsValues}`, and I think you'll see a failure there.

workbench/tests/renderer/investjob.test.js
```diff
@@ -133,4 +133,20 @@ describe('InvestJob', () => {
...
+  test('isSuperSet returns correct booleans', async () => {
+    const job = new InvestJob({
+      modelID: 'foo',
+      modelTitle: 'Foo',
+      argsValues: baseArgsValues,
+      status: 'success',
+    });
+
+    expect(job.isSuperSet({ status: undefined })).toBe(false);
+    expect(job.isSuperSet({ argsValues: baseArgsValues })).toBe(true);
```
emilyanndavis: The test case I referred to in my previous comment. Try changing `baseArgsValues` to `{...baseArgsValues}`.

emilyanndavis (changes requested):
Sorry, @davemfish, I didn't mean to overcomplicate this! I appreciate your adding an `isSuperSet` method, but I think there's a flaw in the logic there.

workbench/src/renderer/menubar/about.jsx
davemfish: I didn't really think about it, but I got the new file from our official Comms team google drive and that's what it is named. Using that filename is at least a signal to ourselves that we're using the official logo. The former `invest-logo.png` was smaller than recommended by electron-builder.

workbench/src/renderer/InvestJob.js
davemfish: Right you are, thank you! I updated the function and the test. By the way, before and after this fix I was deliberately not trying to write a function for the most general case of two arbitrary objects. I figure since it's a method of `InvestJob` we should take advantage of what we know about that object and avoid a recursive function.

emilyanndavis (approved):
Thanks for the updates!

workbench/src/renderer/InvestJob.js
emilyanndavis: Yep, that makes sense to me too!

PR #2709, opened 2026-08-12
Workbench revert icon changes

davemfish:
This PR mostly just reverts some poorly conceived changes that I tacked onto #2701 . That PR updated electron and electron-builder, at which point we discovered we cannot use an `ico` file as the Mac icon.

The only changes here with respect to the state prior to #2701 are the addition of the `icns` file as the Mac icon, and the fix to the link in HISTORY.rst.

## Checklist


claire-simpson (approved):
Thanks!

PR #2714, opened 2026-08-14
Wind Energy: Remove redundant vector reprojection

### LINKED ISSUE #2707: Wind Energy: `_clip_and_reproject_vector` function error if input vectors have same projection
## Description
In the `_clip_and_reproject_vector` function, `reprojected_clip_path` is never created if the input `base_vector_path` and `clip_vector_path` have the same projections. This causes an error where, in a later call to `_clip_vector_by_vector`, the `reprojected_clip_path` is used as in an input but that file hasn't actually been created.

More specifically, if the land polygon  and AOI vectors, which represent the `base_vector_path` and `clip_vector_path`, respectively, have different `projection_wkt`s, the model will raise a `RuntimeError` error. There are no constraints that mean these 2 vectors would _necessarily_ be in different projections. The AOI must be projected in meters, while the land polygon doesn't _have_ to be projected (and in the sample data, it has a GCS, which is likely why we haven't observed this error).
 
Here is a forum post that reported this: https://community.naturalcapitalalliance.org/t/errors-in-processing-wind-energy-production/6487

## To replicate
 Run the Wind Energy model where the input AOI vector and land polygon vector have the same projection WKT.

## Logfile(s)
Here is the relevant part of the logfile:

> 2026-08-12 15:37:20,559 (natcap.invest.wind_energy.wind_energy) wind_energy.execute(1005) INFO Clip and project land polygon to AOI
2026-08-12 15:37:20,561 (natcap.invest.wind_energy.wind_energy) wind_energy._clip_and_reproject_vector(2563) INFO Entering _clip_and_reproject_vector
2026-08-12 15:37:20,576 (natcap.invest.wind_energy.wind_energy) wind_energy._clip_vector_by_vector(2618) INFO Entering _clip_vector_by_vector
2026-08-12 15:37:20,583 (taskgraph.Task) Task.add_task(716) ERROR Something went wrong when adding task clip_and_reproject_land_poly_to_aoi (4), terminating taskgraph.
  File "taskgraph\Task.py", line 684, in add_task
  File "taskgraph\Task.py", line 1103, in _call
  File "natcap\invest\wind_energy\wind_energy.py", line 2587, in _clip_and_reproject_vector
  File "natcap\invest\wind_energy\wind_energy.py", line 2625, in _clip_vector_by_vector
  File "pygeoprocessing\utils.py", line 32, in wrapper
  File "pygeoprocessing\geoprocessing.py", line 2059, in get_vector_info
  File "osgeo\gdal.py", line 9231, in OpenEx
RuntimeError: D:\InVEST\clip-reproject-t4ujj1c5\reprojected_clip_vector.shp: No such file or directory
2026-08-12 15:37:20,587 (natcap.invest.utils) utils.prepare_workspace(171) ERROR Exception while executing wind_energy
  File "natcap\invest\utils.py", line 169, in prepare_workspace
  File "natcap\invest\spec.py", line 2201, in execute
  File "natcap\invest\wind_energy\wind_energy.py", line 1006, in execute
  File "taskgraph\Task.py", line 684, in add_task
  File "taskgraph\Task.py", line 1103, in _call
  File "natcap\invest\wind_energy\wind_energy.py", line 2587, in _clip_and_reproject_vector
  File "natcap\invest\wind_energy\wind_energy.py", line 2625, in _clip_vector_by_vector
  File "pygeoprocessing\utils.py", line 32, in wrapper
  File "pygeoprocessing\geoprocessing.py", line 2059, in get_vector_info
  File "osgeo\gdal.py", line 9231, in OpenEx
RuntimeError: D:\InVEST\clip-reproject-t4ujj1c5\reprojected_clip_vector.shp: No such file or directory

megannissel:
## Description
Fixes #2707

This PR fixes a bug where the model would raise a `RuntimeError` if the input land polygon and AOI already had the same projection. 

Specifically, the issue was in `_clip_and_reproject_vector`: If the base vector (land polygon) and clip vector (AOI) had the same projection, `reprojected_clip_path` was never created -- which would then cause `_clip_vector_by_vector` to fail because one of the inputs didn't exist.

I've removed the SRS check and associated reprojection in `_clip_and_reproject_vector` as it was redundant. `_clip_and_reproject_vector` always calls `_clip_vector_by_vector`, which already checks for matching SRS and reprojects the clip vector (in this case, AOI) to the spatial reference of the base vector (in this case, land polygon) if necessary prior to clipping. 

I also renamed the temp file `clipped_vector_path` created by `_clip_and_reproject_vector` to `clipped_base_vector_path`, for increased clarity. (I found having both `clip_vector_path` and `clipped_vector_path` -- which represent two completely different vectors -- to be unnecessarily confusing.)

Finally, I've updated the docstring of `_clip_vector_by_vector`, which incorrectly stated that the function "assumes all data are in the same projection."

I considered adding a test that uses an AOI and land polygon with the same projection, but decided that it probably wasn't worth the overhead / additional test data storage (as this model still uses files in the test data repo rather than data generated during testing). I did double-check that this fix addresses the issue in a local run with same-projection inputs.

## Checklist

### DIFF
--- a/src/natcap/invest/wind_energy/wind_energy.py
+++ b/src/natcap/invest/wind_energy/wind_energy.py
@@ -2568,19 +2568,19 @@ def _clip_and_reproject_vector(base_vector_path, clip_vector_path,
     target_sr_wkt = pygeoprocessing.get_vector_info(clip_vector_path)[
         'projection_wkt']
 
-    # Create path for the reprojected shapefile
-    clipped_vector_path = os.path.join(
+    # Create path for the clipped base vector
+    clipped_base_vector_path = os.path.join(
         temp_dir, 'clipped_vector' + file_ext)
 
     # Clip the base vector to the AOI.
     # If the projections don't match, _clip_vector_by_vector will first
     # reproject the clip vector to the base vector's SRS
     _clip_vector_by_vector(
-        base_vector_path, clip_vector_path, clipped_vector_path, temp_dir)
+        base_vector_path, clip_vector_path, clipped_base_vector_path, temp_dir)
 
     # Reproject the clipped base vector to the spatial reference of clip vector
     pygeoprocessing.reproject_vector(
-        clipped_vector_path, target_sr_wkt, target_vector_path,
+        clipped_base_vector_path, target_sr_wkt, target_vector_path,
         driver_name=driver_name)
 
     shutil.rmtree(temp_dir, ignore_errors=True)
--- a/src/natcap/invest/wind_energy/wind_energy.py
+++ b/src/natcap/invest/wind_energy/wind_energy.py
@@ -2574,7 +2574,9 @@ def _clip_and_reproject_vector(base_vector_path, clip_vector_path,
 
     # Clip the base vector to the AOI.
     # If the projections don't match, _clip_vector_by_vector will first
-    # reproject the clip vector to the base vector's SRS
+    # reproject the clip vector to the base vector's SRS. Reprojecting
+    # the base vector to match the AOI's SRS first could be computationally
+    # expensive if the base vector is very large.
     _clip_vector_by_vector(
         base_vector_path, clip_vector_path, clipped_base_vector_path, temp_dir)
 
@@ -2614,10 +2616,14 @@ def _clip_vector_by_vector(
     # Get the base and target spatial reference in Well Known Text
     base_sr_wkt = pygeoprocessing.get_vector_info(base_vector_path)[
         'projection_wkt']
+    base_sr = osr.SpatialReference()
+    base_sr.ImportFromWkt(base_sr_wkt)
     target_sr_wkt = pygeoprocessing.get_vector_info(clip_vector_path)[
         'projection_wkt']
+    target_sr = osr.SpatialReference()
+    target_sr.ImportFromWkt(target_sr_wkt)
 
-    if base_sr_wkt != target_sr_wkt:
+    if not base_sr.IsSame(target_sr):
         # Reproject clip vector to the spatial reference of the base vector.
         # Note: reproject_vector can be expensive if vector has many features.
         temp_dir = tempfile.mkdtemp(dir=work_dir, prefix='clip-')
@@ -2655,11 +2661,11 @@ def _clip_vector_by_vector(
     target_layer = None
     target_vector = None
     clip_vector = None
-    clip_vector = None
+    clip_layer = None
     base_layer = None
     base_vector = None
 
-    if base_sr_wkt != target_sr_wkt:
+    if not base_sr.IsSame(target_sr):
         shutil.rmtree(temp_dir, ignore_errors=True)
 
     if empty_clip:
--- a/src/natcap/invest/wind_energy/wind_energy.py
+++ b/src/natcap/invest/wind_energy/wind_energy.py
@@ -2564,28 +2564,19 @@ def _clip_and_reproject_vector(base_vector_path, clip_vector_path,
     temp_dir = tempfile.mkdtemp(dir=work_dir, prefix='clip-reproject-')
     file_ext, driver_name = _get_file_ext_and_driver_name(target_vector_path)
 
-    # Get the base and target spatial reference in Well Known Text
-    base_sr_wkt = pygeoprocessing.get_vector_info(base_vector_path)[
-        'projection_wkt']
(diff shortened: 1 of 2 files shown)


src/natcap/invest/wind_energy/wind_energy.py
```diff
@@ -2564,32 +2564,23 @@ def _clip_and_reproject_vector(base_vector_path, clip_vector_path,
...
+    # If the projections don't match, _clip_vector_by_vector will first
+    # reproject the clip vector to the base vector's SRS
     _clip_vector_by_vector(
-        base_vector_path, reprojected_clip_path, clipped_vector_path, temp_dir)
+        base_vector_path, clip_vector_path, clipped_base_vector_path, temp_dir)
 
     # Reproject the clipped base vector to the spatial reference of clip vector
     pygeoprocessing.reproject_vector(
-        clipped_vector_path, target_sr_wkt, target_vector_path,
+        clipped_base_vector_path, target_sr_wkt, target_vector_path,
```
claire-simpson, line 2585: To clarify the order of operations: 
1. we grab the target projection from the `clip_vector` (`target_sr_wkt`)
2. run `_clip_vector_by_vector` which (a) reprojects the `clip_vector` to match the projection of the `base_vector` before (b) clipping and producing the output `clipped_base_vector_path`, which has the same spatial reference as `base_vector`,
3. reproject the  `clipped_base_vector_path` to match `target_sr_wkt`

Is this correct and if so, why not just reproject the `base_vector` to match the projection of the `clip_vector` in `_clip_vector_by_vector` instead of the other way around? This would prevent the need to reproject `clipped_base_vector_path` at the end of the function, right? I am guessing the answer to this question is "it would be more expensive/time consuming to reproject `base_vector` because this can be a global vector with many vertices" but wanted to check! If this is the case, I think it would be helpful to add a comment to explain the need for this somewhat circular logic. (I also realize `_clip_vector_by_vector` is used once elsewhere which would also require special accommodation if we changed the function.)

src/natcap/invest/wind_energy/wind_energy.py
claire-simpson: I realize this is not related to your PR but should this second `clip_vector = None` actually be `clip_layer = None`?

claire-simpson: We want to be using `osgeo.osr.IsSame()` for wkt comparison! Doug mentioned this to me during the UMH review so I'll pass it on:

> Unfortunately we can't rely on comparing the WKT strings directly for projection equality. But I think gdal has an osgeo.osr.IsSame function that hopefully does the trick. Here's a post about equality checking that Rich, former NatCap software lead, wrote over a decade ago!
> https://gis.stackexchange.com/questions/50312/testing-if-two-gdal-datasets-are-in-same-projection

(Also not directly related to your PR, sorry!)

claire-simpson (changes requested):
Thanks @megannissel! I had a few small/quick notes, mostly related to things that you did not introduce here, so I apologize, but do think they're worth fixing!

src/natcap/invest/wind_energy/wind_energy.py
megannissel: Yeah, I believe that the concern is related to how computationally expensive it could be to reproject the entire base vector. Looking back through the git history from when this function was written (8 years ago! Turns out I didn't touch any of this logic during my refactor last year), there's a [commit message](https://github.com/natcap/invest/commit/42b1dbb29e06b25de810530b527a6f9a4fe6eed8) to that effect: `reproject clip_vector and then reproject clipped_vector back when it's clipped to save computations. see #BITBUCKET-3804`. I'm inclined to say we ought to leave the logic as-is, even though it seems circular. But agree that a comment explaining this would be helpful!

src/natcap/invest/wind_energy/wind_energy.py
megannissel: Oh, good catch! May as well fix this, too, while we're in here.

megannissel: Good to know, thanks for passing along the insight!

src/natcap/invest/wind_energy/wind_energy.py
claire-simpson: Yep, I agree!

claire-simpson (approved):
Looks good!

PR #2728, opened 2026-09-01
Avoid deadlock during validation when using `vsicurl`

### LINKED ISSUE #2724: validation is not thread-safe when using `vsicurl`
When using the Workbench to validate a remotely-sourced input raster, I noticed that the invest flask application would sometimes stop responding to subsequent requests.

## To Reproduce (at least on Windows)
* Run the invest flask application in a Python 3.14 environment. (e.g. Workbench version 3.20.1)
* Enter a remote filepath that does not exist into a model's raster input field: (e.g. `https://storage.googleapis.com/data.naturalcapitalproject.org/foo.tif`)
* Observe that either a) the correct message "File could not be opened as a GDAL raster" never made it back from the server, or b) the correct validation response did come back, but no subsequent requests will respond.
* Try triggering more validation, or opening a new invest model.

## Some findings
The invest server process never crashes, but it does not respond. I can resolve the problem by doing _both_ of these things:
1. run the flask app without multithreading: `ui_server.app.run(port=args.port, threaded=False)`
2. Do not use the `@timeout` decorator around `validate`, which normally uses a thread to call the validate function.

I'm guessing there is something not thread safe about GDAL. Either in the `gdal.OpenEx` call, or maybe in our `GDALUseExceptions` context manager.

Here's the function from `natcap.invest.spec.SingleBandRasterInput`, with lots of extra debugging that I added to confirm that the fail case only occurs when we encounter that `RuntimeError` from `gdal.OpenEx`.

```python
    @timeout
    def validate(self, filepath: str):
        """Validate a raster file against the requirements for this input.

        Args:
            filepath (string): The filepath to validate.

        Returns:
            A string error message if an error was found.  ``None`` otherwise.
        """
        LOGGER.debug('*** validate SingleBandRasterInput')
        with GDALUseExceptions():
            LOGGER.debug('*** using exceptions')
            gdal_path = utils._GDALPath.from_uri(filepath)
            LOGGER.debug(gdal_path.to_normalized_path())
            if gdal_path.is_local:
                LOGGER.debug('*** is_local')
                file_warning = super().validate(filepath)
                LOGGER.debug(file_warning)
                if file_warning:
                    return file_warning

            try:
                LOGGER.debug('*** opening')
                gdal_dataset = gdal.OpenEx(
                    gdal_path.to_normalized_path(), gdal.OF_RASTER)
                LOGGER.debug('*** opened')
            except RuntimeError as e:
                LOGGER.exception(e)
                return validation_messages.NOT_GDAL_RASTER

            # Check that an overview .ovr file wasn't opened.
            if os.path.splitext(filepath)[1] == '.ovr':
                return validation_messages.OVR_FILE
            LOGGER.debug('*** Getting Spatial Ref')
            srs = gdal_dataset.GetSpatialRef()
            LOGGER.debug('*** got spatial ref')
            projection_warning = _check_projection(
                srs, self.projected, self.projection_units)
            LOGGER.debug('*** checked proj')
            gdal_dataset = None
            LOGGER.debug('*** closed file')
            if projection_warning:
                LOGGER.debug('*** got proj warning')
                return projection_warning
```

davemfish:
Fixes #2724 

## Checklist

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -72,6 +72,9 @@ General
   added to datastack archives. Refactored some datastack creation logic into
   class methods on input and output specs.
   (`#2589 <https://github.com/natcap/invest/issues/2589>`_)
+# Fixed a bug where an error state while reading a remote GDAL dataset
+  could result in a deadlock during invest validation.
+  (`#2724 <https://github.com/natcap/invest/issues/2724>`_)
 
 Wind Energy
 ===========
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -72,7 +72,7 @@ General
   added to datastack archives. Refactored some datastack creation logic into
   class methods on input and output specs.
   (`#2589 <https://github.com/natcap/invest/issues/2589>`_)
-# Fixed a bug where an error state while reading a remote GDAL dataset
+* Fixed a bug where an error state while reading a remote GDAL dataset
   could result in a deadlock during invest validation.
   (`#2724 <https://github.com/natcap/invest/issues/2724>`_)
 
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -753,6 +753,11 @@ class SingleBandRasterInput(SpatialFileInput):
                 gdal_dataset = gdal.OpenEx(
                     gdal_path.to_normalized_path(), gdal.OF_RASTER)
             except RuntimeError:
+                # vsicurl is not currently safe for multithreaded use.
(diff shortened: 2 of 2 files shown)


phargogh (approved):
Looks like all `gdal.OpenEx` calls in this file are in try/except blocks, with the new `gdal.VSICurlClearCache()` call in the exception handler for all 3 of them.  Looks good to me!  Just waiting for tests to pass before merging.

phargogh:
Build failure appears unrelated; merging!

Issue #1586, opened 2024-05-28
How should we handle x64/arm64 compatibility for mac binaries?


phargogh:
With GitHub Actions' move to M1 runners on `macos-14` (which is the current `macos-latest`), our `electron-builder` builds are now all `arm64` as a result.  This is unfortunately a problem for mac users of InVEST with intel CPUs, as Mac OS will not allow `arm64` binaries to run on intel chips.  The emulation available does allow x64 binaries to run on ARM chips on mac.

In the short term, the safest thing to do is to downgrade our runners to x64 until we can find the right path forward.

Longer-term, we may want to figure out an appropriate solution to this problem that does not alienate InVEST users who are still using intel CPUs ... it may yet be a while before mac on Intel is unreasonable to support.

Implementation-wise, questions remain.  Should we provide 2 separate mac builds?  Is it better to build a universal binary?  Should we run our tests on both intel and M1 chips?  I'm sure there are others as well.

davemfish:
It looks like `electron-builder` has some facility to help with the electron app side of this problem. Not easy to find, but Find on "universal" on this page: https://www.electron.build/docs/architecture

On a different page they refer to the `arch` property of the config, which can apparently take the "universal" value for mac targets. https://www.electron.build/cli#target

phargogh:
I just today fired up InVEST 3.19.0 and got a notification from Mac OS X (I'm running OSX 26.4.1 "Tahoe") that "Support Ending for Intel-based Apps" and that the InVEST workbench will not open in a future release of MacOS (see screenshot):

<img width="341" height="101" alt="Image" src="https://github.com/user-attachments/assets/6cff9164-4f3f-4738-8f2a-88f1068677d6" />


The notification linked to a section on this page in the apple docs: https://support.apple.com/en-us/102527?cid=mc-ols-rosetta-article_102527-macos_finder-52526201#update

So, I guess it's approaching time to address our build architecture?  Or at least to create Universal binaries on mac, in case we want to retain compatibility with intel chips.

emilyanndavis:
According to that Apple Support article, Rosetta (i.e., support for Intel-based apps on Apple silicon) will be dropped in macOS 28. Judging by the release schedule over the past few years, we should expect macOS 28 to be released in September 2027.

phargogh:
We decided on the team call today to
* Add a new build for arm64 mac binaries as 1 PR
* Later on, we'll add arm64 support for wheels as another PR.

[...]

PR #2733, opened 2026-09-03
Add apple silicon builds for the workbench.

phargogh:
This PR adds Apple Silicon builds for the workbench.  This does, unfortunately require the usage of another runner in GHA until we decide to drop intel builds.  Per #1586 we are not building or testing any wheels at this time, which will come later on.

This does not close the linked issue, only updates it until the wheels are also addressed.

A built apple silicon workbench is/will be available here: https://storage.googleapis.com/natcap-dev-build-artifacts/invest/phargogh/3.20.3.dev4+g16e6facd8/workbench/invest_3.20.3.dev4+g16e6facd8_workbench_darwin_arm64.dmg

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -248,7 +248,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        os: [windows-latest, macos-15-intel]
+        os: [windows-latest, macos-15-intel, macos-15]
 
     steps:
       - name: Check out repo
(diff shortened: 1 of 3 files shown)


emlys (commented):
Thanks @phargogh, this is good to have and while it does require more GHA runners, it's hopefully less error-prone than cross compiling. My only question is whether we may want to move to `macos-latest` (which is now `macos-26`) at some point?

emlys (approved):

phargogh:
>My only question is whether we may want to move to `macos-latest` (which is now `macos-26`) at some point?

Yeah, that's an interesting question!  I think yes, we should move to `macos-latest`.  It turns out that in the [pyinstaller docs](https://pyinstaller.org/en/stable/usage.html#making-macos-apps-forward-compatible) they talk about compiling on the earliest possible version of mac os in order to have the best compatibility with the most OS versions.  I think (but haven't tested) that we are able to largely sidestep this concern with our setup given that 1) we use conda-forge python, which should be linked against its own ecosystem for the most part and 2) mac os upgrades are free these days so it shouldn't be a big issue to ask someone to update their OS.  There is still a possibility that someone would be trying to run InVEST on a mac that isn't supported any longer, but I suspect that'll be an edge case that is unlikely to come up.

dcdenu4:
@phargogh, friendly nudge

phargogh:
@emlys the apple silicon builds are in place!  And although the workbench tests appear to be failing, the binaries are still uploading to GCS.  What do you think about these test failures w/r/t this PR?  Might you have any insights into why the flask application might be failing right away (after only 1 attempt to contact it)?

Issue #2736, opened 2026-09-04
Integrate the Plugin Registry into the Workbench


megannissel:
## Background / Motivation
Given our recent work through NSF POSE; collaboration with others (Peter Hawthorne, TNC, etc.); the focus of the Contributor Track at the Symposium; and general interest at the Alliance level, plugins are a core priority and will be a core feature of InVEST going forward. 

As such, plugins should be easily findable. Users shouldn't need to search for how to engage with them. And the process for adding a plugin should be as streamlined as possible.

We now have a [Plugin Registry](https://natcap.github.io/invest-plugin-registry/index.html) for tracking InVEST plugins (both those contributed by community members and those developed internally). We want to make it easier for InVEST Workbench users to:

- Browse available plugins;
- Search for and evaluate plugins that may meet their specific needs;
- Install a plugin listed on the Registry (specifically, install the version tracked by the Registry) 

without needing to navigate away from the Workbench or manually input plugin URL and version information.

### Plugin Manager Modal Updates

The existing Plugin Manager modal can serve as a starting point. Once updated, it ought to:
- Foreground the Plugin Registry for easy perusal and installation. 
  - Plugin listings should include:
    - Name (display name provided for the Registry)
    - Version
    - Description
    - Authors / Maintainers
    - License
    - Link to Registry listing
    - Link to source code
    - Link to Documentation
    - Link to issue tracker
    - Last updated date
    - Tags (`plugin_type` and keywords)
    - Number of installs
  - If a plugin matching the package name and version is already installed, the user should see some indication that the plugin is already installed rather than the standard "Install Plugin" button.
- Maintain the existing functionality: 
  - Allow a user to remove a plugin
  - Allow a user to manually install a plugin via git host URL or local path
  - Allow a user to access conda environment configuration Advanced Settings

### Data Retrieval and Caching

We ought to cache plugin info from the Registry so that we don't need to fetch it every time the modal is opened. 

The Registry site serves a [JSON file](https://natcap.github.io/invest-plugin-registry/metadata.json) of plugin metadata which is used to generate the display pages on the Registry site itself. While it would be possible to use this file for the Workbench as well, a purpose-built version including just the metadata needed by the Workbench, with a simpler data structure, is in the works; see https://github.com/natcap/invest-plugin-registry/issues/66.

### Tracking Download Statistics

[...]

PR #2737, opened 2026-09-09
Workbench Plugin Registry Integration

megannissel:
## Description
Addresses #2736

This PR overhauls the Plugin Manager modal to integrate the Plugin Registry.

There are now separate tabs for: Plugin Registry, Installed Plugins, Manual Install, Advanced Settings, and About. The content of each tab has been moved to its own component, which I think makes it easier to track the flow of information (at the cost of needing to pass more pieces of state around). The parent `PluginModal` component needs to track state for things like what plugin is actively being (un)installed and if there are any errors, so that clicking away to a different part of the modal while an (un)installation is occurring doesn't lose all information about that process.

Two new attributes are being tracked in the `settingsStore`: 
- `sourceType`: How the plugin was installed. Possible values are `registry`, `git_url`, or `local_path`
- `packageName`: The value of `tool.natcap.invest.package_name` in the `pyproject.toml`. This serves as an important key for cross-referencing with plugins available on the Registry, when determining if a Registry plugin is already installed.

A few outstanding questions:
- I've added a caching mechanism for the plugin metadata retrieved from the Registry site, using `localStorage` and a 24-hour expiration. Does this implementation seem reasonable? Should the expiration time be longer (or shorter)? 
  - I may also need to investigate adding in a browser cache clear when the `localStorage` cache expires. I think there's the possibility a user could still end up with outdated content due to this additional layer of caching.
- Currently, I'm preventing multiple installations or uninstallations from occurring simultaneously (which makes it easier to keep track of which plugin is being installed and its success or error state). Does this seem reasonable?
- State regarding active installations is still being managed by the `PluginModal` component, not the main workbench `HomeTab`, which means that closing the modal during an installation would lose the state of that process. To address this, I've made it so that the modal can't be closed while a plugin is actively being installed or uninstalled. How do folks feel about that?  

Incomplete items:
- Add testing
- Hook up the connection to a Cloud Run function that will log plugin installations in a database (WIP; I've created a new database, `plugin_downloads`, in the `software-usage-stats` instance and added a table called `plugin_downloads_table`. Working on the Cloud Run function currently.)
- Build out a mechanism for retrieving download stats (these likely will be provided by the Plugin Registry site, but won't be included in the same `workbench_metadata.json` file (unless we rethink how / when this file is built, which is a potential option).
- Any sort of type checking. I admittedly have not made use of PropTypes, given that it sounds like typechecking in this way is no longer supported. Since the DH/Workbench integration may be adding in TypeScript support, it may be worth making use of that here as well.
- Finalize content for the About tab. I quickly threw together some copy based on preexisting parts of the Plugin Registry and Developer Docs, but this could definitely use an additional pass / contributions from others. I dumped this content into the [Plugin Registry / Workbench Integration design doc](https://docs.google.com/document/d/1V-5c_yDCthr6KFDHEkRUk4li0Ch8AjC3m6m9qXMHKxo/edit?tab=t.0) for easier collaboration and editing.

## Checklist

### DIFF
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -175,13 +175,14 @@ async function installPlugin(
 /**
  * Get plugin metadata from the installed Python package,
  * and store that metadata in the Workbench's settingsStore.
- * @param  {string} micromamba      path to the micromamba executable
- * @param  {string} pluginEnvPrefix location of the plugin's micromamba env
- * @param  {string} packageName     the plugin's python package name
- * @param  {string} installString   pip install argument for the plugin
+ * @param  {string} micromamba       path to the micromamba executable
+ * @param  {string} pluginEnvPrefix  location of the plugin's micromamba env
+ * @param  {string} packageName      the plugin's python package name
+ * @param  {string} installString    pip install argument for the plugin
+ * @param  {string} pluginSourceType source (local vs registry vs non-registry)
  */
 function storePluginMetadataSync(
-  micromamba, pluginEnvPrefix, packageName, installString
+  micromamba, pluginEnvPrefix, packageName, installString, pluginSourceType
 ) {
   const modelID = execSync(
     `${micromamba} run --prefix "${pluginEnvPrefix}" ` +
@@ -210,7 +211,9 @@ function storePluginMetadataSync(
       modelTitle: modelTitle,
       type: 'plugin',
       source: installString,
+      sourceType: pluginSourceType,
       env: pluginEnvPrefix,
+      packageName: packageName,
       version: version,
     }
   );
@@ -219,7 +222,7 @@ function storePluginMetadataSync(
 export function setupAddPlugin(i18n) {
   ipcMain.handle(
     ipcMainChannels.ADD_PLUGIN,
-    async (event, url, revision, localPath) => {
+    async (event, url, revision, localPath, pluginSourceType) => {
       try {
         let pyprojectTOML;
         let installString;
@@ -294,7 +297,7 @@ export function setupAddPlugin(i18n) {
           );
           event.sender.send('plugin-install-status', i18n.t('Importing plugin...'));
           storePluginMetadataSync(
-            micromamba, pluginEnvPrefix, packageName, installString
+            micromamba, pluginEnvPrefix, packageName, installString, pluginSourceType
           );
           logger.info('successfully added plugin');
         } catch (error) {
--- a/workbench/src/main/setupAddRemovePlugin.js
+++ b/workbench/src/main/setupAddRemovePlugin.js
@@ -9,6 +9,7 @@ import { Downloader } from 'nodejs-file-downloader';
 
 import { logger } from './logger';
 import { ipcMainChannels } from './ipcMainChannels';
+import ELECTRON_DEV_MODE from './isDevMode';
 import { settingsStore } from './settingsStore';
 import { shutdownPythonProcess } from './createPythonFlaskProcess';
 
@@ -219,6 +220,38 @@ function storePluginMetadataSync(
   );
 }
 
+/**
+ * Log installation of plugin from the Plugin Registry.
+ * @param  {string} projectName project.name in pyproject.toml (used as ID by Registry)
+ * @param  {string} packageName tool.natcap.invest.package_name (used by InVEST)
+ * @param  {string} version     the version of the plugin
+ */
+async function logRegistryPluginInstall(projectName, packageName, version) {
+  logger.info('logging Plugin Registry plugin installation');
+  // START DEBUGGING:
+  console.log(`project_name: ${projectName}`);
+  console.log(`invest_package_name: ${packageName}`);
+  console.log(`version: ${version}`);
+  // END DEBUGGING
+
+  // TO DO: Plug in Cloud Run function endpoint
+  try {
+    const response = await fetch("loggingURLGoesHere", {
+      method: "POST",
+      body: JSON.stringify({
+        project_name: projectName,
+        invest_package_name: packageName,
+        version: version
+      })
+    });
+    if (!response.ok) {
+      throw new Error(`Logging failed with non-OK status: ${response.status}`);
+    }
+  } catch (error) {
+    console.log(error.message);
+  }
+}
+
 export function setupAddPlugin(i18n) {
   ipcMain.handle(
     ipcMainChannels.ADD_PLUGIN,
@@ -283,6 +316,7 @@ export function setupAddPlugin(i18n) {
         // Access plugin metadata from the pyproject.toml
         const condaDeps = pyprojectTOML.tool.natcap.invest.conda_dependencies;
         const packageName = pyprojectTOML.tool.natcap.invest.package_name;
+        const projectName = pyprojectTOML.project.name;
 
         // Create a conda env containing the plugin and its dependencies
         // use timestamp to ensure a unique path
@@ -300,6 +334,18 @@ export function setupAddPlugin(i18n) {
             micromamba, pluginEnvPrefix, packageName, installString, pluginSourceType
           );
           logger.info('successfully added plugin');
+
+          // TO DO: Log the plugin installation!
+          if (pluginSourceType === 'registry') {
+            logRegistryPluginInstall(projectName, packageName, revision);
+          }
+          // Commented out for testing purposes; ultimately we'll only want to log
+          // installations from the production Workbench
+          // if (!ELECTRON_DEV_MODE && !process.env.PUPPETEER
+          //     && pluginSourceType === 'registry') {
+          //   // revision will always equate to version for registry plugins
+          //   logRegistryPluginInstall(projectName, packageName, revision);
+          // }
         } catch (error) {
           logger.info('Cleaning up failed installation environment.');
           await spawnWithLogging(
--- a/workbench/src/renderer/components/HomeTab/index.jsx
+++ b/workbench/src/renderer/components/HomeTab/index.jsx
@@ -9,6 +9,7 @@ import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 import Button from 'react-bootstrap/Button';
 import { useTranslation } from 'react-i18next';
+import { FaPlugCirclePlus } from "react-icons/fa6";
 
 import OpenButton from '../OpenButton';
 import InvestJob from '../../InvestJob';
@@ -42,7 +43,8 @@ export default class HomeTab extends React.Component {
       investList,
       openInvestModel,
       deleteJob,
-      clearRecentJobs
+      clearRecentJobs,
+      showPluginModal
     } = this.props;
     let sortedModelIds = {};
     if (investList) {
@@ -102,6 +104,16 @@ export default class HomeTab extends React.Component {
         <Col md={6} className="invest-list-container">
           <ListGroup className="invest-list-group">
             {investButtons}
+            <ListGroup.Item
+              key="add-plugin"
+            >
+              <Button
+                className="open-button plugin-modal-button text-wrap"
+                onClick={showPluginModal}
+              >
+                <span className="pe-2"><FaPlugCirclePlus /></span>Add a Plugin
+              </Button>
+            </ListGroup.Item>
             <ListGroup.Item
               key="browse"
             >
--- a/workbench/src/renderer/components/PluginModal/AboutTab/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/AboutTab/index.jsx
@@ -1,0 +1,179 @@
+import React from 'react';
+
+import Accordion from 'react-bootstrap/Accordion';
+import { useTranslation } from 'react-i18next';
+
+import { openLinkInBrowser } from '../../../utils';
+
+export default function AboutTab() {
+  const { t } = useTranslation();
+
+  const pluginRegistryURL = "https://natcap.github.io/invest-plugin-registry/";
+  const pluginSubmissionURL = "https://natcap.github.io/invest-plugin-registry/docs/submission_process.html";
+  const registryGitHubURL = "https://github.com/natcap/invest-plugin-registry/";
+  const securityReportURL = "https://github.com/natcap/invest-plugin-registry/security";
+  const pluginDocsURL = "https://invest.readthedocs.io/en/latest/plugins.html";
+
+  return (
+    <>
+      <h5>{t(`About Plugins:`)}</h5>
+      <Accordion>
+        <Accordion.Item eventKey="0">
+          <Accordion.Header>{t(`What Is a Plugin?`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+            {t(`Conceptually, an InVEST plugin is an ecosystem services model or performs `
+                + `functions related to ecosystem services modeling. A plugin could be a new `
+                + `stand-alone model, a variation on a model that already exists in core InVEST, `
+                + `or a workflow that composes multiple models to solve a domain-specific problem. `
+                + `A plugin could also perform pre-processing steps to prepare data for use in `
+                + `modeling, or post-processing steps on the outputs from another model.`)}
+            </p>
+            <p className="plugin-small-text">
+              {t(`Like the core InVEST models, it takes in data of various formats (usually `
+                  + `including some geospatial data), processes that data, and produces output files `
+                  + `that contain the results. Unlike the core models, a plugin is not 'official', `
+                  + `i.e., not reviewed or maintained by NatCap. Plugins may be developed, used, and `
+                  + `distributed totally independently of the natcap/invest repo and the Natural `
+                  + `Capital Alliance.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="1">
+          <Accordion.Header>{t(`Installing Plugins`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+              {t(`You can use the Plugin Registry tab to browse for plugins included on the InVEST `
+                + `Plugin Registry, a listing of community-contributed plugins. If you find one you `
+                + `would like to install, you can do so directly from the Workbench using the "Install" `
+                + `button.`)}
+            </p>
+            <p className="plugin-small-text">
+              {t(` Installation will take a few minutes. Once it completes, the "Install" form will `
+                + `be replaced by a message saying "This plugin is installed! You can then close the `
+                + `modal and launch the plugin from the list of models.`)}
+            </p>
+            <p className="plugin-small-text">
+              {t(`If you want to install a plugin that isn't listed on the Registry, and either have the `
+                + `link to a git repo where it is hosted or have the code on your local machine, you can `
+                + `install it using the Manual Install tab.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="2">
+          <Accordion.Header>{t(`Reporting Issues`)}</Accordion.Header>
+          <Accordion.Body>
+            <h6>General Plugin Bugs:</h6>
+            <p className="plugin-small-text">
+              {t(`If you encounter an issue while running a plugin, please report it to the maintainers `
+                + `of the plugin. If you have installed a plugin from the Registry, its Registry listing `
+                + `should include a link to the plugin's Issue Tracker. This is the recommended avenue `
+                + `for reporting bugs.`)}
+            </p>
+            <h6>Security Vulnerabilities:</h6>
+            <p className="plugin-small-text">
+              {t(`The inclusion of a plugin in the InVEST Plugin Registry does not imply or guarantee `
+                + `anything about the plugin's quality, suitability, or security, and the Natural `
+                + `Capital Alliance reserves the right to remove any plugin from the Registry that is `
+                + `believed to be a security risk. If you believe a plugin poses a security risk, please `
+                + `go to the `)}
+              <a
+                href={securityReportURL}
+                title={securityReportURL}
+                aria-label={t(`Security and Quality tab of Registry GitHub repo (opens in web browser)`)}
+                onClick={openLinkInBrowser}
+              >{t(`Security and Quality tab`)}</a>
+              {t(` of the Registry GitHub Repo and use the "Report a vulnerability" button to file a report.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="3">
+          <Accordion.Header>{t(`Developing a Plugin`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+              {t(`Have an idea for a plugin or an existing script that you'd like to turn into `
+                + `one? Take a look at the `)}
+              <a
+                href={pluginDocsURL}
+                title={pluginDocsURL}
+                aria-label={t(`Plugins Developer's Guide (opens in web browser)`)}
+                onClick={openLinkInBrowser}
+              >{t(`Plugin Developer Docs`)}</a>
+              {t(` to learn more.`)}
+            </p>
+            <p className="plugin-small-text">
+              {t(`If you've created an InVEST plugin and would like to make it available to `
+                + `others in the community, the best way to make it discoverable is to submit it `
+                + `for inclusion in the Registry. Take a look at the `)}
+              <a
+                href={pluginSubmissionURL}
+                title={pluginSubmissionURL}
+                aria-label={t(`Plugin Submission Docs (opens in web browser)`)}
+                onClick={openLinkInBrowser}
+              >{t(`Plugin Submission Docs`)}</a>
+              {t(` for a detailed guide.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+      </Accordion>
+      <h5 className="mt-4 mb-3">{t(`About the Tabs in This Modal:`)}</h5>
+      <Accordion>
+        <Accordion.Item eventKey="0">
+          <Accordion.Header>{t(`Plugin Registry`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+              {t(`You can use the Plugin Registry tab to browse and install plugins listed on the `
+                + ` InVEST Plugin Registry from directly within the Workbench.`)}
+            </p>
+            <p className="plugin-small-text">
+              {t(`The `)}
+              <a
+                href={pluginRegistryURL}
+                title={pluginRegistryURL}
+                aria-label={t(`Plugin Registry (opens in web browser)`)}
+                onClick={openLinkInBrowser}
+              >{t(`Plugin Registry`)}</a>
+              {t(` is a hub for community-contributed plugins. The goal of the Plugin Registry is `
+                + `to make it easy for members of the InVEST community to both share the plugins `
+                + `they have created and discover plugins created by others.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="1">
+          <Accordion.Header>{t(`Installed Plugins`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+            {t(`This Installed Plugins tab provides a list of all plugins you have installed `
+              + `in the Workbench. The listing includes the plugin version, as well as information `
+              + `about the installation source. You can use this tab to uninstall plugins from the `
+              + `Workbench.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="2">
+          <Accordion.Header>{t(`Manual Install`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+              {t(`The Manual Install tab allows you to install a plugin via a git URL or local file `
+                + `path. This is useful if you are working on developing a plugin, or if you want to `
+                + `install a plugin (or a version of a plugin) that is not available on the Plugin `
+                + `Registry.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+        <Accordion.Item eventKey="3">
+          <Accordion.Header>{t(`Advanced Settings`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+              {t(`The Advanced Settings tab allows you to manually configure the conda executable `
+                + `used by InVEST for plugin management, as well as each plugin's conda environment. `
+                + `These options are primarily geared towards plugin developers; unless you encounter `
+                + `a problem with micromamba when installing plugins, you likely won't need to adjust `
+                + `these settings.`)}
+            </p>
+          </Accordion.Body>
+        </Accordion.Item>
+      </Accordion>
+    </>
+  )
+}
--- a/workbench/src/renderer/components/PluginModal/AdvancedSettingsTab/index.jsx
+++ b/workbench/src/renderer/components/PluginModal/AdvancedSettingsTab/index.jsx
@@ -1,0 +1,207 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+
+import Button from 'react-bootstrap/Button';
+import Form from 'react-bootstrap/Form';
+import { useTranslation } from 'react-i18next';
+import { MdFolderOpen } from 'react-icons/md';
+
+import { ipcMainChannels } from '../../../../main/ipcMainChannels';
+
+const { ipcRenderer } = window.Workbench.electron;
+
+export default function AdvancedSettingsTab(props) {
+  const {
+    plugins,
+    dragOverHandler,
+    dragEnterHandler,
+    dragLeavingHandler,
+    selectFile,
+    selectDirectory,
+    getDroppedFilePath,
+  } = props;
+  const [condaPath, setCondaPath] = useState('');
+  const [pluginEnvs, setPluginEnvs] = useState({});
+
+  useEffect(() => {
+    Promise.all([
+      ipcRenderer.invoke(ipcMainChannels.GET_SETTING, 'micromamba'),
+      ipcRenderer.invoke(ipcMainChannels.GET_SETTING, 'userDefinedMicromamba')
+    ]).then(([micromamba, userDefinedMicromamba]) => {
+      setCondaPath(userDefinedMicromamba || micromamba);
+    });
+    ipcRenderer.invoke(
+      ipcMainChannels.GET_SETTING, 'plugins'
+    ).then((data) => setPluginEnvs(
+      Object.fromEntries(
+        Object.keys(data).map(
+          (pluginID) => [pluginID, data[pluginID].userDefinedEnv || data[pluginID].env]
+        )
+      )
+    ))
+  }, []);
+
+  const resetCondaPath = () => {
+    ipcRenderer.invoke(
+      ipcMainChannels.GET_SETTING, 'micromamba'
+    ).then((data) => {
+      setCondaPath(data);
+    });
+  };
+
+  const saveCondaPath = () => {
+    ipcRenderer.send(
+      ipcMainChannels.SET_SETTING, 'userDefinedMicromamba', condaPath
+    );
+  };
+
+  const resetPluginEnv = (pluginID) => {
+    ipcRenderer.invoke(
+      ipcMainChannels.GET_SETTING, `plugins.${pluginID}.env`
+    ).then((value) => {
+      setPluginEnvs({...pluginEnvs, [pluginID]: value});
+    });
+  };
+
+  const savePluginEnvs = () => {
+    Object.entries(pluginEnvs).forEach(([pluginID, envPath]) => {
+      ipcRenderer.send(
+        ipcMainChannels.SET_SETTING, `plugins.${pluginID}.userDefinedEnv`, envPath
+      );
+    });
+  };
+
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <Form aria-labelledby="configure-conda-form-title" aria-describedby="conda-executable-description">
+        <Form.Group>
+          <h5 id="configure-conda-form-title" className="mb-3">{t('Configure conda executable (Advanced)')}</h5>
+          <Form.Text
+            as="span"
+            id="conda-executable-description"
+            className="plugin-form-text mb-3"
+          >
+            {t('InVEST is distributed with a copy of micromamba, a conda-like '
+              + 'package manager that is used to manage plugin environments. '
+              + 'If you have conda or mamba installed elsewhere on the system, '
+              + 'you can configure InVEST to use that executable instead. This '
+              + 'may be useful if you run into limitations of the included '
+              + 'micromamba distribution. You can enter an absolute path, or '
+              + 'the name of an executable that is on the system PATH.')}
+          </Form.Text>
+          <Form.Label htmlFor="condaPath">{t('Conda or mamba executable')}</Form.Label>
+          <div className="d-flex flex-nowrap w-100">
+            <Form.Control
+              id="condaPath"
+              type="text"
+              value={condaPath || ''}
+              onChange={(event) => setCondaPath(event.target.value)}
+              onDragOver={dragOverHandler}
+              onDragEnter={dragEnterHandler}
+              onDragLeave={dragLeavingHandler}
+              onDrop={(event) => {
+                const droppedPath = getDroppedFilePath(event);
+                if (droppedPath) {
+                  setCondaPath(droppedPath);
+                }
+              }}
+              className="me-1"
+            />
+            <Button
+              aria-label="browse for conda executable"
+              className="browse-button ms-1 me-1"
+              variant="outline-dark"
+              onClick={async (event) => setCondaPath(await selectFile(event) || condaPath)}
+            >
+              <MdFolderOpen />
+            </Button>
+            <Button
+              className="text-nowrap ms-1"
+              onClick={resetCondaPath}
+            >
+              {t('Reset')}
+            </Button>
+          </div>
+          <Button onClick={saveCondaPath} className="text-nowrap mt-3">
+            {t('Save')}
+          </Button>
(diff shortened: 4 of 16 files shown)


workbench/src/renderer/components/HomeTab/index.jsx
```diff
@@ -102,6 +104,16 @@ export default class HomeTab extends React.Component {
...
           <ListGroup className="invest-list-group">
             {investButtons}
+            <ListGroup.Item
+              key="add-plugin"
+            >
+              <Button
+                className="open-button plugin-modal-button text-wrap"
+                onClick={showPluginModal}
+              >
+                <span className="pe-2"><FaPlugCirclePlus /></span>Add a Plugin
```
emilyanndavis, line 114:
```suggestion
                <span className="pe-2"><FaPlugCirclePlus /></span>{t('Add a Plugin')}
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -28,124 +33,151 @@ export default function PluginModal(props) {
...
+  const [fetchError, setFetchError] = useState(false);
+
+  const registryMetadataURL = "https://natcap.github.io/invest-plugin-registry/workbench_metadata.json";
+  const dataCacheKey = "registryData";
+  const cacheTimeout = 1000 * 60 * 60 * 24; // 24 hours
+
   const handleModalClose = () => {
-    setURL('');
-    setRevision('');
+    if (!installLoading) {
```
emilyanndavis, line 60: I think it's OK to prevent the modal from closing while an installation is in progress (especially when I start thinking about what the alternative would entail), with a caveat:

- We need to provide feedback, e.g., "Please wait until installation is complete." Otherwise, people won't know why the modal won't close—especially if they have navigated away from the installation tab—and may conclude the Workbench is broken.

…and two questions:

1.  If I start installing a plugin, then navigate away, how will I know when the process is complete—or, perhaps more importantly, if/when the process failed? Say I start browsing other plugins and become so engrossed I forget I was installing something, then after a while I close the modal. I find myself on the home tab, facing the list of models, and that's when I remember: I installed the Jabberwocky plugin… why don't I see it here? I guess I'd figure it out eventually: navigating to the Installed Plugins list and finding a dearth of Jabberwocky there, I'd try installing it again, and this time I would patiently wait and watch to see what happens. But it would have been nice to have been notified of the first failure so I could have skipped all those other steps. What do you think about adding a persistent modal footer and rendering the real-time installation status there?

2. Do you think we should prevent the modal from closing while a plugin is being _uninstalled_?Uninstallation is typically both quick and successful, but if I manage to close the modal during that split-second process by accidentally hitting the escape key immediately after pressing the Uninstall button, and the process fails, I could find myself in a mildly bewildering situation. (That said, I don't think uninstallation status merits persistence like the installation status does.)

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -28,124 +33,151 @@ export default function PluginModal(props) {
...
-    let formValid = true;
-    if ((installFrom === 'url' && !url)
-        || (installFrom === 'path' && !path)
-    ) {
-      formValid = false;
-      setPluginSourceMissingError(true);
+        setRegistryData(sortedPlugins);
+        setFetchError(false);
+      } catch (error) {
+        console.log(error.message);
```
emilyanndavis, line 129: It might be worth logging this in the Workbench logs, to help with debugging if/when there's a persistent problem.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -295,461 +297,134 @@ export default function PluginModal(props) {
...
-            {t('Error: Path is required.')}
-          </Form.Text>
-        }
-      </Form.Group>
-    );
-  }
-
-  const pluginDocsURL = "https://invest.readthedocs.io/en/latest/plugins.html";
-  const pluginRegistryURL = "https://natcap.github.io/invest-plugin-registry/";
   let modalBody = (
```
emilyanndavis, line 307: It looks like there's now only one possible definition of `modalBody`, so we can change this to `const`.

```suggestion
  const modalBody = (
```

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -295,461 +297,134 @@ export default function PluginModal(props) {
...
+            </Nav>
+          </Col>
+          <Col sm={10}>
+            <Tab.Content>
+              <Tab.Pane eventKey="registry">
+                {fetchError ? (
+                  <div className="registry-fetch-error">
+                    <IconContext.Provider value={{ className: 'registry-warning-icon' }}>
+                      <MdOutlineWarningAmber />
+                    </IconContext.Provider>
```
emilyanndavis, line 337: I'm not familiar with `IconContext`—what does it provide here? Is this equivalent to adding `className="registry-warning-icon"` to the icon element itself (or to a `<div>` or other element wrapping the icon)?

workbench/src/renderer/components/PluginModal/PluginRegistryTab/index.jsx
```diff
@@ -0,0 +1,109 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
```
emilyanndavis, line 2: No need to import `PropTypes` if you're not using it (and I agree there's no reason to use it anymore).

workbench/src/renderer/components/PluginModal/PluginRegistryTab/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+    let pluginID = pluginInfo.invest_package_name;
+    const listItem = (
+      <button
+        key={pluginID}
+        size="lg"
+        className={`registry-list-group-item plugin-registry-button ${activePluginKey === pluginID ? 'active' : ''}`}
+        onClick={(e) => handlePluginClick(pluginID)}
+      >
+        {pluginInfo.plugin_name}
+      </button>
```
emilyanndavis, line 59: These would be better marked up as `Tab`s, both for semantics and for keyboard operability. Is there a reason you chose to use plain buttons instead?

workbench/src/renderer/components/PluginModal/PluginRegistryTab/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  function getInstallStatus(plugin) {
+    if (installedPluginNamesVersions.includes(plugin.invest_package_name + '@' + plugin.version)) {
+      return "thisVersionInstalled";
+    } else if (installedPluginNames.includes(plugin.invest_package_name)) {
+      return "anotherVersionInstalled";
+    } else {
+      return "notInstalled"
+    }
+  }
```
emilyanndavis, line 72: It might be helpful to define these strings as constants (e.g., `const THIS_VERSION_INSTALLED = "thisVersionInstalled") and export them so they can be imported by the other module that needs them. Referencing a named constant is less error-prone than replicating a specific string literal.

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+  const [userAcknowledgmentError, setUserAcknowledgmentError] = useState(false);
+
+  const pluginTypes = {
+    "preprocessing": "Preprocessing",
+    "postprocessing": "Postprocessing",
+    "workflow": "Workflow",
+    "invest_model_variant": "InVEST Model Variant",
+    "new_model": "New Model",
+    "other": "Other"
+  }
```
emilyanndavis, line 44: I know this is probably an artifact of context-switching from Python (or perhaps from JSON), but in any case you shouldn't need to wrap these key names in quotation marks.

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+
+  const handleAddPluginClick = () => {
+    clearFormErrors();
+    if (validateAddPluginForm()) {
+      addPlugin(
+        pluginID,
+        plugin.repository_url, // url
+        plugin.version,        // revision
+        undefined,             // path, used for manual install
+        'registry'             // sourceType
```
emilyanndavis, line 68: Probably a good idea to define the allowed/expected `sourceType` values somewhere, so we don't end up with unexpected values by mistake. An enum (made easy with TypeScript) would be great for this, but in the meantime, a few string constants will work well.

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+            <Form.Text
+              as="p"
+              id={`${pluginID}-user-acknowledgment-error`}
+              className="plugin-error plugin-user-acknowledgment-error mb-1"
+            >
+              {t('Error: Before installing a plugin, you must agree to the terms by selecting the checkbox.')}
+            </Form.Text>
+          }
+          <Button
+            disabled={installLoading || installDisabled}
```
emilyanndavis, line 138: Disabled buttons can be confusing for many users, and they're completely hidden from screen readers. Better to allow interaction with the button, use the click handler to control its effects, and provide feedback when the action normally performed by the button has been blocked. You're providing an on-screen message, which is great! Let's take it a step further by removing this `disabled` attribute, updating `handleAddPluginClick`, and updating the `aria-describedby` value to programmatically associate the button with any relevant on-screen message.

(And in case you're wondering: yes, updating the behavior of the disabled-until-form-validation-passes `Run` button is on my wishlist!)

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+            onClick={handleAddPluginClick}
+            aria-describedby="plugin-installation-duration-notice"
```
emilyanndavis, line 140:
```suggestion
            aria-describedby=`${pluginID}-plugin-installation-disabled-notice ${pluginID}-plugin-installation-duration-notice`
```

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+  );
+
+  return (
+    <>
+      <div className="plugin-pane">
+        <h5>{plugin.plugin_name}</h5>
+        <p className="plugin-small-text">
+          {plugin.description}
+        </p>
+        <Table borderless size="sm" className="plugin-small-text plugin-table">
```
emilyanndavis, line 237: A table is best for truly tabular data, i.e., when each row represents a set of data points corresponding to named columns. A [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl) (sometimes colloquially known as a definition list) would be a better fit here. You can find some examples on the `datahub-integration` branch, specifically in the [`DataHubSearchParams`](https://github.com/natcap/invest/blob/feature/datahub-integration/workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/index.tsx) and [`DataHubSearchResultCard`](https://github.com/natcap/invest/blob/feature/datahub-integration/workbench/src/renderer/components/DataHubSearchModal/DataHubSearchResultCard/index.tsx) components. And I'll happily help with styling if needed!

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+          <tbody>
+            <tr>
+              <td className="text-end"><b>Downloads:</b></td>
```
emilyanndavis, line 240: Probably better to apply bold font weight via CSS than to use `<b>` tags for these elements (which should become `<dt>` elements in a `<dl>`).

Let's also make sure they're wrapped in `t()`!

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+            </tr>
+            <tr>
+              <td className="text-end"><b>More Info:</b></td>
+              <td>
+                <a
+                  href={plugin.registry_url}
+                  title={plugin.registry_url}
+                  aria-label={t("View on Plugin Registry (opens in web browser)")}
+                  onClick={openLinkInBrowser}
+                >View on Registry</a> | <a
```
emilyanndavis, line 281: A few things here:
1. It's OK for an aria-label to include language beyond the visible text (e.g., `(opens in web browser)`), but it's important that it start with exactly the visible text (i.e., `View on Registry`, not `View on Plugin Registry`). There are various reasons for this; one of them is to support speech control.
2. I suspect you've omitted `(opens in web browser)` from the visible text to save space, but screen reader users aren't the only people who might want/need to know that a link will open in the browser. Maybe we could make use of the "external link" icon we use in the changelog? You could apply the `link-external` CSS class, or—what I'd recommend—directly embed the `MdOpenInNew` icon, giving it the aria-label `(opens in web browser)`. Using the icon with its own aria-label will allow us to remove the aria-label from the `<a>`, thus side-stepping the problem of aria-label text diverging from the visible text that I described a few sentences ago.
3. Let's make sure the visible text for all these links gets wrapped in `t()`.

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+            <li>
+              <b>{t('Installed via: ')}</b>
+              {pluginDetails.sourceType}
+            </li>
+          )}
+          <li>
+            <b>{t('Source: ')}</b>
+            {pluginDetails.source}
+          </li>
+        </ul>
```
emilyanndavis, line 116: A `ul` isn't wrong here, but this is another good use case for a `<dl>`, should you feel so inclined! Also, better to use CSS than `<b>` tags for the "Installed via" and "Source" text.

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+        {uninstallErr &&
+          (
+            <>
+              <h5>{t('Error removing plugin:')}</h5>
+              <div className="plugin-error plugin-install-remove-error">{uninstallErrMsg}</div>
+              <Button
+                onClick={() => ipcRenderer.send(
+                  ipcMainChannels.SHOW_ITEM_IN_FOLDER,
+                  window.Workbench.ELECTRON_LOG_PATH,
+                )}
```
emilyanndavis, line 126: You could make this simpler by importing `handleClickFindLogfiles` from `menubar/handlers` and then referencing `handleClickFindLogfiles` instead of sending an IPC message. (That goes for this button, and a few others you are handling in the same way.)

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+              >
+                {t('Find workbench logs')}
+              </Button>
+            </>
+          )
+        }
+      </Col>
+      <Col sm={3}>
+        <Button
+          disabled={uninstallLoading || uninstallDisabled}
```
emilyanndavis, line 136: Again, better to always allow interaction with the button, and control its effects in the click handler.

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+          onClick={handleRemovePluginClick}
+        >
+          {uninstallLoading
+            ? (
+              <div className="adding-button">
+                <Spinner animation="border" role="status" size="sm" className="plugin-spinner">
+                  <span className="visually-hidden">{t('Removing...')}</span>
```
emilyanndavis, line 143: No need for this `<span>`, since it is redundant to the visible text.
```suggestion
```

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+          )}
+          <li>
+            <b>{t('Source: ')}</b>
+            {pluginDetails.source}
+          </li>
+        </ul>
+        {uninstallErr &&
+          (
+            <>
+              <h5>{t('Error removing plugin:')}</h5>
```
emilyanndavis, line 120: There are some funky things going on with heading order, probably an artifact of certain headings being used for styling purposes that were then carried over in the refactor. Not the most crucial thing, but I'll want to dig into it later!

workbench/src/renderer/components/PluginModal/ManualInstallTab/index.jsx
```diff
@@ -0,0 +1,382 @@
...
+        <h5 id="add-plugin-form-title" className="mb-3">{t('Manually Install a Plugin')}</h5>
+        <p>
+          {t('For more information about creating a plugin, read our ')}
+          <a
+            href={pluginDocsURL}
+            title={pluginDocsURL}
+            aria-label={t("Plugins Developer's Guide (opens in web browser)")}
+            onClick={openLinkInBrowser}
+          >{t("Developer's Guide")}</a>.
+        </p>
```
emilyanndavis, line 225: I know this is existing content that was moved in the refactor, but while I have the opportunity, I have two suggestions:
1. Rather than using an aria-label, let's either include `(opens in web browser)` in the visible text, or use the `MdOpenInNew` icon with `aria-label="(opens in web browser)"`.
2. Could we try restructuring this for better translation support? Splitting the link text from the rest of the sentence in this way works for English, and (mostly) works for Spanish, but once we move beyond Latin-based languages, all bets are off when it comes to sentence structure. One option could be to flag the entire sentence as one translatable unit, but I don't know if we can/how cumbersome it might be to use `t` to wrap markup (I'd have to dig into the docs and/or test it). Another option is to update the wording so the two units function as independent (though related) units, e.g., something along the lines of "More information about creating a plugin:" and "Plugin Developer's Guide". Slightly more robotic, but less likely to be untranslatable. (This was already on my brain because I did something similar in the Data Hub integration, opting for `[…] the following input: {inputName}` rather than `[…] the {inputName} input`.)

workbench/src/renderer/components/PluginModal/ManualInstallTab/index.jsx
```diff
@@ -0,0 +1,382 @@
...
+            onChange={(event) => setInstallFrom("url")}
+          />
+          <Form.Check
+            type="radio"
+            id="installFromLocal"
+            name="installFrom"
+            label={t('local path')}
+            checked={installFrom === "path"}
+            onChange={(event) => setInstallFrom("path")}
+          />
```
emilyanndavis, line 246: Good improvement, thanks!

workbench/src/renderer/components/PluginModal/ManualInstallTab/index.jsx
```diff
@@ -0,0 +1,382 @@
...
+            className="plugin-error plugin-user-acknowledgment-error"
+          >
+            {t('Error: Before installing a plugin, you must agree to the terms by selecting the checkbox.')}
+          </Form.Text>
+        }
+        <Button
+          disabled={installLoading}
+          onClick={handleAddPluginClick}
+          aria-describedby="plugin-installation-duration-notice"
+        >
```
emilyanndavis, line 284: I imagine this button and the associated text elements are very similar to what appears on the `PluginDetailPane`. Would it be too much trouble to extract this markup & logic into a reusable component?

workbench/src/renderer/components/PluginModal/AboutTab/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const pluginRegistryURL = "https://natcap.github.io/invest-plugin-registry/";
+  const pluginSubmissionURL = "https://natcap.github.io/invest-plugin-registry/docs/submission_process.html";
+  const registryGitHubURL = "https://github.com/natcap/invest-plugin-registry/";
+  const securityReportURL = "https://github.com/natcap/invest-plugin-registry/security";
+  const pluginDocsURL = "https://invest.readthedocs.io/en/latest/plugins.html";
+
+  return (
+    <>
+      <h5>{t(`About Plugins:`)}</h5>
```
emilyanndavis, line 19: Probably more of a stylistic preference than anything else, but I tend to think a colon isn't necessary on text that's obviously a heading.

workbench/src/renderer/components/PluginModal/AboutTab/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+      <Accordion>
```
emilyanndavis, line 20: My (perhaps controversial) opinion is that accordions should allow multiple items to be open at a time. It looks like we can adopt that behavior by using the [`alwaysOpen`](https://react-bootstrap.github.io/docs/components/accordion/#always-open) prop on the `Accordion`.
```suggestion
      <Accordion alwaysOpen>
```

workbench/src/renderer/styles/style.css
```diff
@@ -926,3 +921,146 @@ code {
...
+  &:hover {
+    color: var(--invest-green);
+    background-color: white;
+    border-color: var(--invest-green);
+  }
+  &.active {
+    color: var(--invest-green);
+    background-color: white;
+    border-color: var(--pinvest-green);
+  }
```
emilyanndavis, line 938: Let's apply the same styles on focus, and customize the focus ring color as well.
(The `&:first-child:active` selector overrides Bootstrap's `.btn:first-child:active` styling—a less specific selector, e.g., `&:active`, doesn't do the trick.)
```suggestion
  &:hover, &:focus, &:focus-visible, &:first-child:active {
    color: var(--invest-green);
    background-color: white;
    border-color: var(--invest-green);
  }
  &:focus, &:focus-visible {
    box-shadow: 0 0 0 .25rem var(--invest-green-opacity-50);
  }
```

workbench/src/renderer/styles/style.css
```diff
@@ -860,9 +860,8 @@ input[type=text]::placeholder {
   }
 }
 
-/* Manage Plugins modal */
+/* Plugin Manager modal */
 .plugin-modal {
-  width: 40rem;
```
emilyanndavis, line 865: Bootstrap modals automatically resize in response to window size, which is generally a good thing, but because this modal has a complex layout, we may find we want to enforce a fixed width or a minimum width.

workbench/src/renderer/styles/style.css
```diff
@@ -926,3 +921,146 @@ code {
...
+    border-color: var(--invest-green);
+  }
+  &.active {
+    color: var(--invest-green);
+    background-color: white;
+    border-color: var(--pinvest-green);
+  }
+}
+
+.react-icons {
```
emilyanndavis, line 941: Could we name this something more specific? Right now it's generic enough that I worry about namespace collisions, i.e., accidentally styling something we don't mean to, or accidentally breaking existing styling if/when these style rules are changed. (It's also mildly inconvenient to find its uses, since searching the codebase for `react-icons` also matches all the `react-icons` imports.)

workbench/src/renderer/styles/style.css
```diff
@@ -926,3 +921,146 @@ code {
...
+  height: 1.25em;
+  width: 1.25em;
+  margin-right: 0.5em;
+  &.react-icons-white {
+    color: white;
+  }
+}
+
+.plugin-small-text {
+  font-size: 0.875rem;
```
emilyanndavis, line 953: Any chance I can convince you to limit (or eliminate) use of this font size? I'm admittedly biased toward larger font sizes as a means of supporting better legibility by default (especially as my own age-related farsightedness creeps in), but since most of the text in the Workbench is 16px (1rem) or larger, I don't see why the plugin names, descriptions, and metadata should be any smaller.

workbench/src/renderer/styles/style.css
```diff
@@ -926,3 +921,146 @@ code {
...
+}
+
+.plugin-modal-nav {
+  padding: 0;
+  border-right: 1px solid var(--bs-modal-header-border-color);
+}
+
+.plugin-modal-nav-item > a.nav-link {
```
emilyanndavis, line 961: Just heads up, I will probably play around with these styles and recommend some adjustments related to hover/active/focus styles, including perhaps ways to distinguish the active tab here from the active tab in the registry pane (assuming such a thing might interest you).

emilyanndavis (changes requested):
This is looking really good, @megannissel! I have a bunch of small suggestions (such as wrapping a text in `t()`, defining string constants for easier reference elsewhere, or combining duplicate style rules), and some larger ones (such as updating semantics, e.g., table -> description list, buttons -> tabs; avoiding disabled buttons; or extracting repeated markup into a shared component).

As for your specific questions:
1. Caching: honestly, I don't have much real-world experience with caching data like this, so I don't have much to draw on other than gut instinct. I think we'll want to do some more testing to make sure we've covered all the scenarios, but your approach seems reasonable to me. Some possible alternatives to busting the cache every 24 hours could include setting the expiry to a specific, predictable time each day (say, midnight UTC) rather than precisely every 24 hours after you happen to have made your first request; and/or occasionally pinging GitHub to see if the JSON file has changed, and retrieving it only if it has (I believe that's a thing offered by the GitHub API; I don't know if it would make much of a practical difference in this case, but it could once the registry is larger).

2. Allowing just one installation/uninstallation at a time: I'm on board with this. I can imagine perhaps wanting to change this at some point, but I don't think it's worth the trouble right now.

3. Preventing the modal from closing during installation/uninstallation: I'm on board with this, with a caveat (user feedback!) and a suggestion to render real-time installation status in a persistent location in the modal. I describe this in (a lot) more detail in an inline comment.

emilyanndavis:
Oh, one more thing—[columbo.gif]—I think we should probably target a feature branch rather than `main`, since there are some outstanding tasks and therefore this is likely just the first in a series of PRs. What do you think?

megannissel:
> I think we should probably target a feature branch rather than `main`

Oh, yeah, we absolutely don't want to merge this into `main`; when I last spoke with Doug, we didn't come to a conclusion about a feature vs release branch and what this PR should target / if it was premature to create either of those branches. I'll revisit that topic with him tomorrow and update the target branch accordingly!

workbench/src/renderer/components/HomeTab/index.jsx
megannissel: Good catch! This change was interesting to make; `HomeTab` is still a class component, so I needed to follow a different pattern for getting `t` in there. I followed the pattern in `InvestTab` and used `withTranslation` rather than `useTranslation` (which threw an `invalid hook call` error). I _think_ I did it right? (Or, at least, nothing is obviously broken!)

workbench/src/renderer/components/PluginModal/InstalledPluginsTab/index.jsx
megannissel: Ah, I missed the existence of `handleClickFindLogfiles`. That's much cleaner, thanks!

workbench/src/renderer/components/PluginModal/ManualInstallTab/index.jsx
megannissel: Great ideas. Also, this is making me think that translation for the About tab will be a nightmare unless I seriously rethink how to include links... But that's a problem for Future Me!

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
megannissel: Absolutely a carry-over from where we have this same mapping defined in the plugin registry repo. I think I just copy-pasted that dict wholesale 😅

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
megannissel: Point taken about disabled buttons being a potential source of confusion, and I'm all for handling it in a clearer way. That said, I do think there's utility in a button having a different appearance when it's disabled (even if we're not using the `disabled` attribute). At least, I find it clearer when buttons that are disabled look different from ones that are enabled. Is that considered bad practice? 

In terms of providing user feedback on click, how would you suggest doing so? I'm not sure what that ought to look like!

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
megannissel: I think this is a great idea and while working on implementing it I learned something new! 

I had assumed that defining a constant in a "parent" component to import in a "child" (that is itself being imported by the parent) would result in a circular import. But that's not how that works; it's not that the _file_ is being imported, it's the specific exported component (or constant) that's being imported. So it does work. (And thank you for doing something very similar with constants on the DH branch; that's what showed me this can be done.) Fascinating!

workbench/src/renderer/components/PluginModal/PluginRegistryTab/index.jsx
megannissel: Whoops, yep, good catch; removed this import from all of the new files that aren't using it!

workbench/src/renderer/components/PluginModal/index.jsx
megannissel: I think when I fiiiirst started working on this, `className=` didn't seem to be working as I expected it to. And the [react-icons docs](https://github.com/react-icons/react-icons#configuration) used `IconContext`, which made me think that must be the "right" way to do it. But I just tried dropping `IconContext` and using `className` directly again and it's working; I must've been doing something wrong before!

workbench/src/renderer/styles/style.css
megannissel: Thank you, this is exactly what I wanted to achieve but my CSS skills weren't quite up to the task!

workbench/src/renderer/styles/style.css
megannissel: Yes, please, if you feel inspired to play around with it, I would LOVE suggestions! I also don't like the way the two sets of tabs sort of blend together, but I couldn't figure out a way to distinguish them from one another that looked good.

workbench/src/renderer/components/PluginModal/index.jsx
emlys: What if we prevent the user from navigating away from the plugin tab that shows the installation status? If we've accepted that users can't access any workbench features outside of the modal while a plugin is installing, limiting them to the plugin tab doesn't seem too much worse.

I'd be in favor of doing the same for the uninstall as well, if it eliminates some possibly confusing error state.

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -175,13 +176,14 @@ async function installPlugin(
 /**
  * Get plugin metadata from the installed Python package,
  * and store that metadata in the Workbench's settingsStore.
- * @param  {string} micromamba      path to the micromamba executable
- * @param  {string} pluginEnvPrefix location of the plugin's micromamba env
- * @param  {string} packageName     the plugin's python package name
- * @param  {string} installString   pip install argument for the plugin
+ * @param  {string} micromamba       path to the micromamba executable
+ * @param  {string} pluginEnvPrefix  location of the plugin's micromamba env
+ * @param  {string} packageName      the plugin's python package name
```
emlys:
```suggestion
 * @param  {string} packageName      the plugin's importable python package name
```

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -210,16 +212,50 @@ function storePluginMetadataSync(
       modelTitle: modelTitle,
       type: 'plugin',
       source: installString,
+      sourceType: pluginSourceType,
       env: pluginEnvPrefix,
+      packageName: packageName,
```
emlys: It's always bothered me that the importable package name can't be derived from the name of the package installed... but such is the state of python packaging.

emlys: Technically, top-level names can be found by inspecting the package metadata, but in theory there could be more than one. I suppose this attribute makes it possible to define a plugin that's a subpackage, which is maybe a feature?

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -210,16 +212,50 @@ function storePluginMetadataSync(
...
 
+/**
+ * Log installation of plugin from the Plugin Registry.
+ * @param  {string} projectName project.name in pyproject.toml (used as ID by Registry)
+ * @param  {string} packageName tool.natcap.invest.package_name (used by InVEST)
+ * @param  {string} version     the version of the plugin
+ */
+async function logRegistryPluginInstall(projectName, packageName, version) {
+  logger.info('logging Plugin Registry plugin installation');
+  // START DEBUGGING:
```
emlys: Should this debugging section be removed?

workbench/src/main/setupAddRemovePlugin.js
```diff
@@ -210,16 +212,50 @@ function storePluginMetadataSync(
...
   );
 }
 
+/**
+ * Log installation of plugin from the Plugin Registry.
+ * @param  {string} projectName project.name in pyproject.toml (used as ID by Registry)
+ * @param  {string} packageName tool.natcap.invest.package_name (used by InVEST)
+ * @param  {string} version     the version of the plugin
+ */
+async function logRegistryPluginInstall(projectName, packageName, version) {
```
emlys: I'm curious about the choice of which info to log. We have available to us the model ID, the version, the `project.name`, the package name as defined in `tool.natcap.invest.package_name`, the "plugin ID" which is a combo of the model ID and version, and also the source path/URL. 

Usage logging currently logs the model ID, the plugin source, and the workbench version (perhaps it should be logging the plugin version as well). For consistency I wonder if we should log the model ID here instead/in addition?

I'm not familiar with how/if the plugin registry maintains a unique identifier for each plugin, but unfortunately I don't think that any of `projectName`, `packageName`, or `version` are normally guaranteed to be unique, and so their combination is possibly not unique either. Is the registry enforcing that somehow?

workbench/src/renderer/components/PluginModal/AboutTab/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+        <Accordion.Item eventKey="0">
+          <Accordion.Header>{t(`What Is a Plugin?`)}</Accordion.Header>
+          <Accordion.Body>
+            <p className="plugin-small-text">
+            {t(`Conceptually, an InVEST plugin is an ecosystem services model or performs `
```
emlys: I appreciate that some documentation probably needs to exist in the workbench, but I wonder if we could just link to the plugin docs web page and avoid the possibility of redundant or incomplete information living there vs. here in the modal? 

Particularly for the "About the Tabs in This Modal" section, I wonder if there's a way to put that info more concisely in the tabs themselves? For instance, could the "Manual Install" tab have some sort of help text or tagline that says "Install a plugin via a git URL or local file path (for plugins not available via the registry)".

workbench/src/renderer/components/PluginModal/AdvancedSettingsTab/index.jsx
```diff
@@ -0,0 +1,207 @@
...
+    });
+  };
+
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <Form aria-labelledby="configure-conda-form-title" aria-describedby="conda-executable-description">
+        <Form.Group>
+          <h5 id="configure-conda-form-title" className="mb-3">{t('Configure conda executable (Advanced)')}</h5>
```
emlys: I like having these settings in the separate Advanced tab! Perhaps it's no longer necessary to put "(Advanced)" in the setting header any more?

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+      </div>
+      <div className="install-pane registry-install-form">
+        {(installStatus === "thisVersionInstalled")
+          ? alreadyInstalledPane
+          : installPane
+        }
+      </div>
+    </>
+  );
+}
```
emlys: Super minor annoying thing, but the Github UI is pointing out to me that some of these files are missing a final newline

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -28,124 +33,151 @@ export default function PluginModal(props) {
...
+      }
     }
-  }, [url, path]);
 
-  useEffect(() => {
-    if (userAcknowledgment) {
-      setUserAcknowledgmentError(false);
-    }
-  }, [userAcknowledgment]);
+    if (cacheJSON && !cacheStale) {
```
emlys: I have a general concern about caching the list of plugins. Imagining a scenario where a developer gets their plugin added to the registry, I think my first action would be to go try and install it into the workbench. (Or to tell colleagues that it's now available to install). Having it not appear for 24 hours could be pretty confusing and frustrating, especially if that caching behavior isn't transparent or controllable by the user.

If caching is necessary, I might expect the cache to automatically clear when I restart the workbench, rather than on a schedule. Or even when I close and reopen the modal... at that point maybe the cache isn't doing much.

workbench/src/renderer/components/PluginModal/index.jsx
```diff
@@ -28,124 +33,151 @@ export default function PluginModal(props) {
...
+        console.log('Using cached data');
+        setRegistryData(cacheJSON.data);
+        setFetchError(false);
+    } else {
+      console.log('Cache miss; fetching data...');
+      try {
+        // Fetch data from the Registry if not cached
+        const response = await fetch(registryMetadataURL);
```
emlys: This would be a great problem to have, but do we anticipate this URL and request needing to support paginated results if we had a much larger list of plugins?

Just a random thought, I don't think we need to implement that here yet.

emlys (changes requested):
Wow @megannissel, this represents a huge amount of work, and a big step forward for plugins! I don't have anything to add to @emilyanndavis's thorough review on the react components, but I added some more general thoughts, and a concern about caching.

workbench/src/renderer/components/PluginModal/PluginRegistryTab/index.jsx
megannissel: Frankly I think this was a combination of working on this piece really early on when I barely knew what I was doing, and initially styling these in a very different way. But you're absolutely right; this makes more sense as another set of `Tab`s. (And, good news, I have a much better understanding of what I'm doing now and can actually make it happen!)

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
megannissel: Thanks, Emily! The swap (while maintaining the existing styling) ended up being pretty straightforward to make

workbench/src/main/setupAddRemovePlugin.js
megannissel: Yep, the logging piece was very much a WIP when I opened the PR. Now that the Cloud Run function exists and has been plugged in, I've removed the debugging.

workbench/src/renderer/components/PluginModal/AdvancedSettingsTab/index.jsx
megannissel: Good call; seems redundant now

workbench/src/renderer/components/PluginModal/PluginRegistryTab/PluginDetailPane/index.jsx
megannissel: Fixed; thank you!

workbench/src/renderer/components/PluginModal/index.jsx
megannissel: I did briefly consider this, but it seems like we're quite a ways off from needing it. The Registry itself isn't currently set up to serve paginated metadata, either. I think we ought to keep it in mind (and probably open an issue on the topic), but I don't think we need to address it for v1.

workbench/src/main/setupAddRemovePlugin.js
megannissel: The Registry uses the value of `project.name` in the `pyproject.toml` as the unique identifier for each plugin when it comes to metadata generation and the subsequent creation of plugin pages on the site. (Note: This comment made me double-check that we were indeed enforcing uniqueness here for plugins listed on the Registry. Turns out, we weren't! But we're fixing that.)

For download stats logging, we definitely want to include this value, since that'll be the primary key used by the Registry site and the Workbench when displaying download counts. Since the Registry site cares not at all about `modelID` (and indeed doesn't ever access this value), I didn't figure it made sense to log it. But if there's value in tracking it for cross-referencing with usage stats, we could do so.

There's a parallel issue related to the `settingsStore` and needing to have a shared key that can be cross-referenced to determine if a plugin listed on the Registry is installed in the user's workbench. Since the Workbench was previously using the value of `tool.natcap.invest.package_name` in some places, I went with this key. But I'm now thinking that may have been a bad call and I ought to stick with using `project.name` for this as well.

workbench/src/main/setupAddRemovePlugin.js
megannissel: Your comment about logging made me take a step back. My goal here was to establish a shared key between the Registry and the `settingsStore` for identifying whether a plugin was installed already. But I think what I actually need to track across the board is the value of `project.name`, since that's what the Registry uses as its unique key. I don't know if we need to track `packageName` at all at this point.

megannissel: ...Although now I'm considering potential pitfalls with this approach as well. What if a user had a plugin installed from some source other than the Registry that happened to use the same `package.name` as a Registry plugin? The modal would show that Registry plugin as already installed, even though it isn't. I'm now wondering if it might be safer / more foolproof to use the `source` for cross-referencing, since the Registry does also track the git URL. What do you think, @emlys?

Issue #2722, opened 2026-08-27
Workbench: dropdown menu caret is tiny when option selected is invalid


claire-simpson:
## Description
When the selection for a dynamically-populated dropdown menu fails validation, the dropdown caret/chevron symbol is very tiny.

<img width="641" height="58" alt="Image" src="https://github.com/user-attachments/assets/1d19cc37-69ad-4a33-b1c7-a809960dd094" />

After input:

<img width="642" height="73" alt="Image" src="https://github.com/user-attachments/assets/99433a96-d436-489c-bbe2-2ff70945a0a0" />

## To replicate

1. Open the Urban Mental Health model
2. Before populating any inputs, the "Option" menu displays "Select an option...", which is an invalid choice. Observe the tiny caret.
3. After a valid model option is selected, the caret becomes larger/normal-sized.

You can also observe this behavior in Coastal Vulnerability:

<img width="1105" height="116" alt="Image" src="https://github.com/user-attachments/assets/e59037bb-02d9-4de0-bd20-50c8a3160a0a" />

## Expected behavior
The size of this dropdown arrow should be larger and should not change based on validation status.

emilyanndavis:
Good catch, @claire-simpson. We have custom styles overriding Bootstrap styles for dropdowns with no validation status and for dropdowns with the `is-valid` class, but not for dropdowns with the `is-invalid` class. Not sure how we never noticed this before; I wonder if something changed in Bootstrap 5? Anyway, there's a one-line fix for this; I'll see if I can sneak it into Emily's (unrelated, but already open) PR that will introduce some other CSS changes.

emilyanndavis:
Fixed in PR #2739.

Issue #2268, opened 2025-12-09
Workbench: use the latest model spec structures


davemfish:
When we redesigned the model spec, we avoided modifying the Workbench by translating the new spec back to the same structure the Workbench always used. It looks like we do this in `ModelSpec.to_json`:

```python
# rename 'inputs' to 'args' to stay consistent with the old api
spec_dict.pop('inputs')
spec_dict['args'] = {_input.id: _input for _input in self.inputs}
```

It would be nice to update the Workbench so it can use `inputs` directly.

davemfish:
There is also a lot of custom serialization done in this `to_json` method. Now that we're using pydantic models, we could define [Field Serializers](https://pydantic.dev/docs/validation/latest/concepts/serialization/#field-serializers) and perhaps replace this entire `to_json` method with those serializers and a call to `.model_dump()` or `.model_dump(mode='json')`

cc @emilyanndavis

PR #2739, opened 2026-09-10
Add support for input group labels

### LINKED ISSUE #1602: Add "group description label" Workbench UI input
Since we have groups of inputs in the Workbench, it can be helpful to "label" that group of inputs to make it more easily understandable by the user. The only way I can think to do this is by using a toggle input element. But there are times where I don't need a toggle, but just want to convey that these inputs are defining X Properties.

I think this could help in certain instances with #1601 . For instance, if there was a "Wind Price" description label, then toggling between the different input options might not be as confusing, since those inputs being hid and shown would stay put under that label.

One question is where this description should live. With the Plugins feature, we've created a `uiSpec` to go with `MODEL_SPEC`. Since this descriptive label means nothing to the core model, perhaps there is a way to flesh out use of a `uiSpec` and not add it to 'MODEL_SPEC.args'.

emlys:
## Description
Fixes #1602
image

On the invest side, added a new `spec.InputGroup` class which may be used instead of a plain list of strings to represent a group of inputs with an optional label. When the model spec is JSON serialized, the `input_field_order` is now returned as a list of dicts, not a list of lists. The `name` attribute (which can be left as an empty string) is then rendered in the workbench.

Like before, a dotted line visually separates input groups. If a group has a label, it will be displayed at the beginning of the dotted line. I think I like this layout generally, but open to feedback on how it looks.

## Checklist

### DIFF
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -218,6 +218,11 @@ def set_metadata_field_descriptions(field_specs, resource):
             LOGGER.debug(error)
 
 
+class InputGroup(BaseModel):
+    name: str = ''
+    keys: list[str]
+
+
 class ImmutableBaseModel(BaseModel):
     """BaseModel with frozen attributes."""
 
@@ -2299,18 +2304,26 @@ class ModelSpec(ImmutableBaseModel):
     Example: ``"https://github.com/natcap/invest-demo-plugin/blob/main/README.md"``
     """
 
-    input_field_order: list[list[str]]
-    """A list that specifies the order and grouping of model inputs.
-    Inputs will be displayed in the input form from top to bottom in the order
-    listed here. Sub-lists represent groups of inputs that will be visually
-    separated by a horizontal line. This improves UX by breaking up long lists
-    and visually grouping related inputs. If you do not wish to use groups,
-    all inputs may go in the same sub-list. It is a convention to begin with a
-    group of ``workspace_dir`` and ``results_suffix``. Each item in the
-    sub-lists must match the key of an ``Input`` in ``inputs``. The key of each
+    input_field_order: list[InputGroup | list[str]]
+    """A list that specifies the order and grouping of model inputs that will be
+    displayed in the workbench input form. This list may contain lists and/or
+    spec.InputGroups. Each sub-list or InputGroup represents a group of inputs
+    that will be visually separated by a horizontal line. This improves UX by
+    breaking up long lists and visually grouping related inputs. You can give a
+    group a label, which will be displayed in the workbench, by using an
+    InputGroup and setting the ``name`` property. Note that groups and their
+    labels only affect workbench rendering and have no effect
+
+    The InputGroup was added to allow labeling groups. Using plain lists is
+    still supported for backwards compatibility. If you do not wish to use
+    groups, all inputs may go in the same sub-list.
+
+    It is a convention to begin with a group of ``workspace_dir`` and
+    ``results_suffix``. Each item in each sub-list or InputGroup.keys list must
+    match the key of an ``Input`` in ``inputs``. The key of each
     ``Input`` must be included exactly once, unless it is hidden.
 
-    Example: ``[['workspace_dir', 'results_suffix'], ['foo'], ['bar', baz']]``
+    Example: ``[['workspace_dir', 'results_suffix'], InputGroup(name='Group A', keys=['bar', baz'])``
     """
 
     inputs: list[Input]
@@ -2366,6 +2379,8 @@ class ModelSpec(ImmutableBaseModel):
         or are marked as hidden."""
         found_keys = set()
         for group in self.input_field_order:
+            if isinstance(group, InputGroup):
+                group = group.keys
             for key in group:
                 if key in found_keys:
                     raise ValueError(
@@ -2432,6 +2447,20 @@ class ModelSpec(ImmutableBaseModel):
         spec_dict.pop('inputs')
         spec_dict['args'] = {_input.id: _input for _input in self.inputs}
         spec_dict['outputs'] = {_output.id: _output for _output in self.outputs}
+        spec_dict['input_field_order'] = []
+        # Encode the input field order as a list of dicts, where each dict
+        # contains a single item representing an input group. Each dict has two
+        # entries: 'name' which stores the group label (or '' if no label),
+        # and 'input_keys' and the value is the list of input ids in that group.
+        for input_group in self.input_field_order:
+            if isinstance(input_group, InputGroup):
+                spec_dict['input_field_order'].append({
+                    'name': input_group.name,
+                    'input_keys': input_group.keys})
+            else:  # is a list of keys
+                spec_dict['input_field_order'].append({
+                    'name': '',
+                    'input_keys': input_group})
         return json.dumps(spec_dict, default=fallback_serializer, ensure_ascii=False)
 
     def preprocess_inputs(self, input_values):
--- a/src/natcap/invest/spec.py
+++ b/src/natcap/invest/spec.py
@@ -221,11 +221,11 @@ def set_metadata_field_descriptions(field_specs, resource):
 class InputGroup(BaseModel):
     """Represent a group of input fields for display in the workbench."""
 
-    name: str = ''
+    label: str = ''
     """Optional label that will be displayed in the workbench above
        this group of inputs."""
 
-    input_keys: list[str]
+    input_ids: list[str]
     """List of model input ids that belong to this group. Each string must
        match the id of an Input in the model."""
 
@@ -2318,7 +2318,7 @@ class ModelSpec(ImmutableBaseModel):
     that will be visually separated by a horizontal line. This improves UX by
     breaking up long lists and visually grouping related inputs. You can give a
     group a label, which will be displayed in the workbench, by using an
-    InputGroup and setting the ``name`` property. Note that groups and their
+    InputGroup and setting the ``label`` property. Note that groups and their
     labels only affect workbench rendering and have no effect
 
     The InputGroup was added to allow labeling groups. Using plain lists is
@@ -2326,11 +2326,11 @@ class ModelSpec(ImmutableBaseModel):
     groups, all inputs may go in the same sub-list.
 
     It is a convention to begin with a group of ``workspace_dir`` and
-    ``results_suffix``. Each item in each sub-list or InputGroup.keys list must
-    match the key of an ``Input`` in ``inputs``. The key of each
+    ``results_suffix``. Each item in each sub-list or InputGroup.input_ids
+    list must match the key of an ``Input`` in ``inputs``. The key of each
     ``Input`` must be included exactly once, unless it is hidden.
 
-    Example: ``[['workspace_dir', 'results_suffix'], InputGroup(name='Group A', keys=['bar', baz'])``
+    Example: ``[['workspace_dir', 'results_suffix'], InputGroup(label='Group A', input_ids=['bar', baz'])``
     """
 
     inputs: list[Input]
@@ -2387,7 +2387,7 @@ class ModelSpec(ImmutableBaseModel):
         found_keys = set()
         for group in self.input_field_order:
             if isinstance(group, InputGroup):
-                group = group.input_keys
+                group = group.input_ids
             for key in group:
                 if key in found_keys:
                     raise ValueError(
@@ -2462,12 +2462,12 @@ class ModelSpec(ImmutableBaseModel):
         for input_group in self.input_field_order:
             if isinstance(input_group, InputGroup):
                 spec_dict['input_field_order'].append({
-                    'name': input_group.name,
-                    'input_keys': input_group.input_keys})
+                    'label': input_group.label,
+                    'input_ids': input_group.input_ids})
             else:  # is a list of keys
                 spec_dict['input_field_order'].append({
-                    'name': '',
-                    'input_keys': input_group})
+                    'label': '',
+                    'input_ids': input_group})
         return json.dumps(spec_dict, default=fallback_serializer, ensure_ascii=False)
 
     def preprocess_inputs(self, input_values):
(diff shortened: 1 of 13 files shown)


src/natcap/invest/spec.py
```diff
@@ -218,6 +218,18 @@ def set_metadata_field_descriptions(field_specs, resource):
             LOGGER.debug(error)
 
 
+class InputGroup(BaseModel):
+    """Represent a group of input fields for display in the workbench."""
+
+    label: str = ''
```
emilyanndavis, line 227: Might I suggest `group_label` instead of `label`? It might seem redundant in this context, but in the context of the Workbench, I think something more precise than `label` would help prevent potential confusion with HTML `<label>` elements (since a `<label>` would not be an appropriate element to render this text). (Along those lines, we could call it `legend`, but I worry that might bewilder model authors who aren't familiar with HTML… and there are additional elements that could be semantically appropriate—such as headings—so it's probably best to steer clear of naming a property after any particular HTML element.)

src/natcap/invest/spec.py
```diff
@@ -2299,18 +2311,26 @@ class ModelSpec(ImmutableBaseModel):
...
-    sub-lists must match the key of an ``Input`` in ``inputs``. The key of each
+    input_field_order: list[InputGroup | list[str]]
+    """A list that specifies the order and grouping of model inputs that will be
+    displayed in the workbench input form. This list may contain lists and/or
+    spec.InputGroups. Each sub-list or InputGroup represents a group of inputs
+    that will be visually separated by a horizontal line. This improves UX by
+    breaking up long lists and visually grouping related inputs. You can give a
+    group a label, which will be displayed in the workbench, by using an
+    InputGroup and setting the ``label`` property. Note that groups and their
+    labels only affect workbench rendering and have no effect
```
emilyanndavis: I can guess where you were going with this, but you seem to have trailed off here… 🙃

workbench/src/renderer/components/SetupTab/ArgsForm/index.jsx
```diff
@@ -155,11 +154,25 @@ class ArgsForm extends React.Component {
...
-          {groupItems}
-        </div>
+        <fieldset className={fieldsetClassName} key={k}>
+          {inputGroup.label &&
+            <legend className="arg-group-dotted-legend">{inputGroup.label}</legend>
+          }
+          <Form.Group className="arg-group">
+            {groupItems}
+          </Form.Group>
+        </fieldset>
```
emilyanndavis, line 165: Thank you for using fieldsets and legends to convey appropriate semantics here! 😀

emilyanndavis (changes requested):
Thanks, Emily! I think it will be great for the Workbench to be able to convey semantic groupings as well as visual groupings. I do have a request to call the `label` something a bit more precise (noted in a review comment), and (as you probably expected) I also have some styling requests/recommendations.

I've mocked up a couple of options. Neither of these has to be the final solution, but I think either one would fit the bill. Screenshots follow the descriptions.

1. Left-aligned legends: similar to what you've done, with the following modifications:
    - Larger text. When a text element functions as a heading for the content that follows, it's most intuitive for the heading text to be larger than the text that follows it.
    - Darker text. If a group label is important enough to be displayed in the UI, let's make sure there's sufficient color contrast for people to read it without straining.
    - Darker border. Color contrast is less crucial here than for the legend text, but I don't see a reason to let the border fade into the background.
    - All inputs are "indented". After making the legend text larger and darker, I think I understood why you had chosen a lighter color for the legend text—I needed a way to make the legends more obviously distinguishable from the input labels. Offsetting the left alignment by a handful of pixels achieves this.

2. Center-aligned legends: same as option 1, with the following differences:
    - Legends are center-aligned. Normally I avoid center alignment (general consensus is that it's harder to read than left-aligned text, and I think it usually just looks strange), but here I don't hate it. (Maybe the dotted border helps anchor the text so it doesn't seem like it's awkwardly floating in space?)
    - Inputs are not indented. When legends are center-aligned, they easily stand out from input labels—no need for an additional offset.

In both cases, forms without any group labels would remain unchanged (other than the borders being slightly thicker than before, as you've proposed here), and all of this can be done with CSS—no additional logic needed. (In fact, we can remove the styling-related logic you added to the JSX and accomplish what we need with CSS selectors alone.) Happy to share the CSS I've come up with in whatever way makes most sense!

#### Left-aligned legends
SDR input form with left-aligned legends

#### Center-aligned legends
SDR input form with center-aligned legends

#### No legends
AWY input form without legends

src/natcap/invest/spec.py
emlys: How about converting the name to `group_label` where it's JSON serialized? In the context of the python class `InputGroup`, I think calling the attribute `group_label` is too redundant.

emilyanndavis: I guess my thought was, if this class exists exclusively for the Workbench, why not cater to the Workbench? But if `group_label` is too annoying, then yes, we can certainly translate it elsewhere.

src/natcap/invest/spec.py
emlys: Fixed!

emlys:
Thanks for the detailed suggestions, @emilyanndavis. Of the two layouts you suggest, I like the left-aligned one more. If you already have specific CSS changes to propose, are they simple enough to go in Github suggestions in the PR? Or would it be easier to go over it on a call?

emilyanndavis:
> Of the two layouts you suggest, I like the left-aligned one more. If you already have specific CSS changes to propose, are they simple enough to go in Github suggestions in the PR? Or would it be easier to go over it on a call?

I'll see if I can suggest those changes here. But I'll also happily hop on a call if you have questions or would like a walk-through!

workbench/src/renderer/styles/style.css
emilyanndavis: These changes do the following:
1. Add 1rem of padding to the top of the form when the first fieldset contains a legend (i.e., when the first input group has a name).
2. Add 1.5rem of padding to the left side of each arg-group when _any_ fieldset contains a legend (i.e., when any input group has a name).
3. Add 1rem of padding to the bottom of the form in all cases (not necessarily related to this feature; it's just a subtle amount of blank space to help signal you've reached the end of the form).

```suggestion
  padding: 0 1.5rem 1rem;

  &:has(.arg-group-fieldset:first-of-type > legend) {
    padding-top: 1rem;
  }

  &:has(.arg-group-fieldset > legend) {
    .arg-group {
      padding-left: 1.5rem;
    }
  }
```

workbench/src/renderer/styles/style.css
```diff
@@ -632,16 +632,29 @@ exceed 100% of window.*/
...
+
+.arg-group-dotted-legend {
+  float: none;
+  width: auto;
+  padding: 0 0.75rem 0 0;
+  margin-bottom: 0;
+  font-size: 0.9rem;
+  font-weight: 600;
+  color: #6c757d;
+}
```
emilyanndavis: These changes do the following:
1. Apply styles to fieldsets & legends, based on what you'd suggested, with modifications to font size and colors (in this case both the text & border quietly inherit the main text color defined at the page level), as well as elimination of the fieldset's padding & margin rules (handling these with updated padding on the arg-group instead).
2. Rename `arg-group-dotted-fieldset` to `arg-group-fieldset`, to minimize confusion if we ever decide to switch to a non-dotted border style.
3. Use the `& > legend` selector nested inside the `arg-group-fieldset` rules, removing the need for a separate CSS class for legends in this context.

```suggestion
.arg-group-fieldset {
  border: none;

  &:not(:first-of-type),
  &:first-of-type:has(> legend) {
    border-top: 0.125rem dotted;
  }

  & > legend {
    float: none;
    width: auto;
    padding: 0 0.5rem 0 0;
    margin: 0;
    font-size: 1.125rem;
    font-weight: 600;
  }
}
```

workbench/src/renderer/styles/style.css
```diff
@@ -632,16 +632,29 @@ exceed 100% of window.*/
...
+
 .arg-group {
   display: flex;
   flex-direction: column;
   gap: 16px;
   margin: 0;
   padding: 1rem 0;
```
emilyanndavis: Updating the arg-group padding here saves us from the more finicky task of trying to adjust padding and/or margin on the fieldset.

```suggestion
  padding: 1.5rem 0;
```

workbench/src/renderer/styles/style.css
emilyanndavis: This last CSS change is entirely unrelated to this feature, but it lets us fix #2722 without having to create a whole new PR for a one-line change. 😃 

```suggestion
  .form-select.is-valid,
  .form-select.is-invalid {
```

workbench/src/renderer/components/SetupTab/ArgsForm/index.jsx
```diff
@@ -155,11 +154,25 @@ class ArgsForm extends React.Component {
...
+      if (k === 0 && !inputGroup.group_label) {
+        fieldsetClassName = "mt-3"
+      }
       formItems.push(
-        <div className="arg-group" key={k}>
-          {groupItems}
-        </div>
+        <fieldset className={fieldsetClassName} key={k}>
+          {inputGroup.group_label &&
+            <legend className="arg-group-dotted-legend">{inputGroup.group_label}</legend>
```
emilyanndavis: 1. No need for additional logic to add space above the first labeled input group—the CSS takes care of this for us.
2. Update fieldset class name from "arg-group-dotted-fieldset" to "arg-group-fieldset".
3. Remove class name from legend, since the CSS handles legend styling for us.

```suggestion
formItems.push(
        <fieldset className="arg-group-fieldset" key={k}>
          {inputGroup.group_label &&
            <legend>{inputGroup.group_label}</legend>
```

emlys: Github wouldn't let me apply this suggestion in the PR interface, but I committed the same changes.

src/natcap/invest/spec.py
```diff
@@ -2432,6 +2454,20 @@ def fallback_serializer(obj):
...
+        # and 'input_keys' and the value is the list of input ids in that group.
+        for input_group in self.input_field_order:
+            if isinstance(input_group, InputGroup):
+                spec_dict['input_field_order'].append({
+                    'group_label': input_group.label,
+                    'input_ids': input_group.input_ids})
+            else:  # is a list of keys
+                spec_dict['input_field_order'].append({
+                    'group_label': '',
+                    'input_ids': input_group})
```
emilyanndavis: Looks like the comment references `'name'` where the key is now called `'group_label'`.

Also, what are your thoughts on moving this logic to a [field serializer](https://pydantic.dev/docs/validation/latest/concepts/serialization/#field-serializers)? I don't have a super strong opinion in this case, but it's on my mind since I recently wrote a field serializer (on Dave's recommendation) for model input keywords to support Data Hub search.

workbench/src/renderer/components/SetupTab/ArgsForm/index.jsx
```diff
@@ -155,11 +154,21 @@ class ArgsForm extends React.Component {
           />
         );
       });
+      // Separate each group of input fields with a dotted line and label if
+      // applicable. Omit the dotted line above the first group if it has
+      // no label.
```
emilyanndavis: We can probably do without this comment now that we've moved the styling logic to CSS. (In hindsight, I probably should have removed it in my earlier suggestion!)

```suggestion
```

emilyanndavis (changes requested):
Thanks for the updates, @emlys! Just a couple of lingering suggestions.

src/natcap/invest/spec.py
emlys: Good point! Really, the whole `ModelSpec.to_json` method makes more sense as a pydantic model serializer. I made that change, and also added a model validator that converts any lists in the `input_field_order` to an `InputGroup` behind the scenes. And I moved this logic into a model serializer on `InputGroup`. So now that serialization happens automatically when we invoke it through `ModelSpec.model_dump_json()`!

emlys: I removed `ModelSpec.to_json`, which is technically a breaking change but I doubt anyone else is using that method. If we are really concerned about breaking things, we could keep it around as a wrapper of `model_dump_json`.

src/natcap/invest/spec.py
```diff
@@ -2454,21 +2471,11 @@ def fallback_serializer(obj):
...
-            else:  # is a list of keys
-                spec_dict['input_field_order'].append({
-                    'group_label': '',
-                    'input_ids': input_group})
-        return json.dumps(spec_dict, default=fallback_serializer, ensure_ascii=False)
+
+        # dump to json so that we can apply the fallback serializer,
+        # then convert back to a dictionary.
+        return json.loads(json.dumps(
+            spec_dict, default=fallback_serializer, ensure_ascii=False))
```
emilyanndavis, line 2478: My guess is we could skip these extra steps once we replace the fallback serializer with Pydantic field/model serializers, but there's a separate issue for that (#2268), so I'll leave it be!

emilyanndavis (changes requested):
Great updates! My only request is that we keep the `to_json` method around for now, on the off chance someone else _is_ using it. (You never know!) It could raise a `DeprecationWarning` to minimize surprise when we eventually remove it.

src/natcap/invest/spec.py
emlys: Yeah, I thought about that but decided to save it for the future. One difficulty is that some of these custom serializations (like for `pint.Unit`) are used across several different attributes of different input/output types. So using the fallback serializer is nice in that we can define that in just one place, vs. having to repeat it in a bunch of different model serializers.

emlys:
I restored the `to_json` method and added a deprecation warning onto it. Thanks!

src/natcap/invest/spec.py
emlys: The [annotated serializer pattern](https://pydantic.dev/docs/validation/latest/concepts/serialization/#field-serializers) would help reduce the overhead.

emilyanndavis (approved):
Thanks!

PR #2740, opened 2026-09-11
Keywords for Wind & Wave Energy Models

### LINKED ISSUE #2467: Annotate invest inputs with keywords

### LINKED ISSUE #2675: Create a mapping between invest keywords and keywords used on the DataHub
#2467 adds keywords to invest inputs with the intent of using those same keywords in the metadata of data on the DataHub. But there are is already data on the DataHub using slightly different keywords. We would like to update the metadata, but it may not happen right away.

To facilitate that update, or to support the linkage between the Workbench & the DataHub before we have the opportunity to update the metadata, we can make a mapping between keywords currently in use on the DataHub and the keywords they will ultimately be replaced with.

The dictionary of keywords used by InVEST is still subject to change though.

davemfish:
Also removed the `BIOPHYSICAL_TABLE` keyword that we used in a couple places but since decided we probably don't need it.

Fixes #2467 
Fixes #2675 

The RTD build failed due a reason that was already fixed on `main`, I just hadn't merged those changes into the target branch here until after I opened the PR, and it's not so convenient to re-trigger the RTD build from here.

## Checklist

### DIFF
--- a/src/natcap/invest/keywords.py
+++ b/src/natcap/invest/keywords.py
@@ -455,9 +455,17 @@ WAVE_ENERGY_CAPACITY = InvestKeyword(
     value='WAVE ENERGY CAPACITY',
     definition='The maximum energy production of a wave energy-capturing machine')
 
+WAVE_ENERGY_CAPITAL_COSTS = InvestKeyword(
+    value='WAVE ENERGY CAPITAL COSTS',
+    definition=(
+        'Economic costs of installing, connecting, operating, and'
+        ' maintaining a wave energy-capturing machine'))
+
 WAVE_ENERGY_PERFORMANCE = InvestKeyword(
     value='WAVE ENERGY PERFORMANCE',
-    definition='The energy production of a wave energy-capturing machine given sea state conditions')
+    definition=(
+        'The energy production of a wave energy-capturing machine'
+        ' given sea state conditions'))
 
 WAVE_ENERGY_PRODUCTION = GCMDKeyword(
     value='WAVE ENERGY PRODUCTION/USE',
--- a/src/natcap/invest/keywords.py
+++ b/src/natcap/invest/keywords.py
@@ -41,6 +41,10 @@ class InvestKeyword(Keyword):
     vocabulary: str = 'InVEST Keywords'
 
 
+ABOVEGROUND_BIOMASS_CARBON = InvestKeyword(
+    value='ABOVEGROUND BIOMASS CARBON',
+    definition='The pool of Carbon stored in above-ground living biomass')
+
 ADMINISTRATIVE_DIVISIONS = GCMDKeyword(
     value='ADMINISTRATIVE DIVISIONS',
     uuid='1ae304de-252c-45da-8dd8-df99a281e4f4',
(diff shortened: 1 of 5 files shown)


src/natcap/invest/keywords.py
```diff
@@ -449,16 +497,32 @@ class InvestKeyword(Keyword):
...
     value='WIND SPEED',
     uuid='a7ce84a3-8329-4eb7-b5de-72d2dea8c6bf',
     full_path='EARTH SCIENCE > OCEANS > OCEAN WINDS > SURFACE WINDS > WIND SPEED')
 
+WIND_TURBINE = InvestKeyword(
+    value='WIND TURBINE',
+    definition='A machine that converts wind into electricity.')
+
+
+
```
megannissel, line 525: I think we might have more newlines here than intended.

megannissel (approved):
Thanks, @davemfish, the new keywords look good to me! I'm going to approve this PR but hold off on merging it; I left a comment about some extra newlines at the end of `keywords.py` that you can feel free to address or ignore and then merge.

Issue #2727, opened 2026-09-01
Omit DataHub Query for certain inputs/models


davemfish:
As we add the feature to query the DataHub from the Workbench, it occurs to us there are a number of cases where it many not make sense to include this feature, at least at first.

### Models that do not yet have an AOI input
For example, the Carbon model. We disable queries until the user populates an AOI for reasons described in #2685 . And not all models have an AOI input until we complete #2267.

### Vector Inputs
There are so few vector datasets currently on the DataHub (2, I believe) that nearly every query from the Workbench on a vector input will come up empty. It may be friendlier to users to omit the search feature on vectors until we have more results available.

### Inputs without keywords
Some invest inputs have so few restrictions on what they represent that it is not really possible to describe them with keywords. An example are CSV inputs that contain references to other geospatial data. The geospatial datasets can often represent anything, so 1) it's hard to assign relevant keywords, and 2) it's unlikely someone would want to re-use someone else's CSV. Sharing the entire datastack is a much more likely use-case for these models.

Plus, DHAL queries don't really make sense if they don't include a list of tags.

Issue #2703, opened 2026-08-06
Workbench: integrate Data Hub search UI with DHAL


emilyanndavis:
Integrate search UI (#2702) with the DHAL to enable searching for real Data Hub datasets from the Workbench.

Issue #2685, opened 2026-08-04
Disable DataHub query until the user populates an AOI field


davemfish:
### From our DataHub / Workbench design doc:

>Spatial
InVEST users will want to find data for a specific geographic region. Often, this region can be defined by some other model input that a user has populated before doing their DH query. For example, if a user has populated several spatial inputs, the union of their bounding boxes can be used in a DH query. This allows discovery of layers that overlap with any of their existing input layers.
But what if the user has not populated any inputs yet before doing the query? 
...

We can disable the DH query feature until a user populates an AOI
* Prevents accidental global runs.
* Prevents the user from selecting a DH layer that would become invalid – due to no spatial overlap – after they load their local files.
* But not all models have an AOI (this may be changing https://github.com/natcap/invest/issues/2267)

### Designating the AOI
It's likely that some models will have an AOI input based on `natcap.invest.spec.AOI` and some will not, because an input with a different name serves that purpose (e.g. watersheds in Annual Water Yield). So we may need some new attribute of a `VectorInput` that designates it as the "AOI". Or a new subclass of `VectorInput`.

The Workbench will need to keep track of the validation status of that input in order to know whether the search feature should be enabled/disabled.

davemfish:
The AOI bounding box can also be used as the box for spatial queries to the DHAL. That is simpler than using the union of all the bounding boxes for all the inputs currently in the model form.

davemfish:
A number of models have a vector input that could theoretically be used as an "AOI", but the model does not currently use that input to constrain the extent of the analysis. For example, an optional polygon vector might be used to aggregate raster results at the end, but was not used clip rasters at the start. These "AOIs" would be sufficient to support querying the DataHub, but would not satisfy our requirement of preventing people from accidentally doing global model runs.
* Crop Production Percentile
* Crop Production Regression
* Forest Carbon Edge Effects
* HRA
* Stormwater
* Urban Nature Access

Other models that have vector inputs but they are not suitable as an "AOI" because they represent other things besides the full analysis area.
* DelineateIt
* Habitat Quality
* Pollination

And models with no AOI or vector inputs:
* Carbon
* Coastal Blue Carbon
* RouteDEM

emilyanndavis:
Thanks, @davemfish. In hindsight, I realize this was almost certainly documented when we started down this path; I just completely forgot to look for it.

PR #2743, opened 2026-09-12
Workbench: Data Hub search UI

### LINKED ISSUE #2686: Workbench DataHub queries for "sibling" datasets
Some invest inputs need to be closely associated with other inputs in order to make sense. The most common example is a LULC raster input and an associated biophysical table that must be based on the same landcover codes present in that raster.

The DataHub organizes these associated datasets into "collections". For example, the "ESA CCI LULC" collection includes ESA LULC rasters and CSV tables based on those same ESA landcover codes.

The DHAL `search_dataset` endpoint can take a "sibling" parameter with this "ESA CCI LULC" collection string in order to limit results to only datasets in that collection.

#### How will the Workbench know that a particular input has a sibling?
* We could use the `MODEL_SPEC` and add some attribute to indicate the relationship.
* We could create some other mapping of the relationships between inputs, perhaps stored in the Workbench.

#### If an input does have a sibling, how will the Workbench know the collection string value of that sibling?
* The collection string is part of the response data from the DHAL, but the Workbench will need to store that response somewhere that it can be checked later when a sibling input is being searched for. Similar to the validation state of an AOI (see #2685 ) there is some new state related to the model's setup form that the Workbench needs to track to facilitate DataHub queries.

### LINKED ISSUE #2702: Workbench: build UI for Data Hub search
Build UI components to support searching the Data Hub from the Workbench.

Includes the following:
- A search button alongside each model input field where searching the Data Hub is reasonable (i.e., raster, vector, CSV; but not workspace, file suffix, number, etc.).
- A modal that displays search query parameters before initiating a search, presents a spinner while waiting for a response, and finally renders search results.
- An option to select a search result, which populates the appropriate input field with the dataset URL.
- Helpful info when there are no search results.
- Error message and appropriate actions on server error.
- Error message and appropriate actions on attempt to search without first specifying an AOI.
- Mock search results and mock async behavior to support the needed UI states.

Does not include the following:
- Integration with the DHAL (see #2703).

emilyanndavis:
## Description
Resolves #2702, #2703, #2685.
Partially resolves #2686.

Specifically, it is now possible to search for Data Hub datasets from the Workbench UI (#2702, #2703). Some key details:
- Supported models include all core InVEST models that have an input that functions as an AOI (whether that's literally called "Area of Interest" or something else, such as "Watersheds", "Administrative Boundaries", etc.). If we want to deliberately block certain models from accessing the search feature for any reason (#2727), we can omit the `aoi_input_id` from the model spec.
- Supported datatypes include rasters and CSVs. Vectors are omitted for now since there are very few vectors currently on the Data Hub.
- The Workbench automatically forms a search query—including appropriate datatype, tags, extent, and (if relevant) collection names—from context.
- Since the search query uses the bounding box of the AOI input, a valid AOI must be specified in order to enable search.
- Search results are limited to a certain collection (#2686) under the following conditions:
  1. An LULC or a Biophysical Table that belongs to a specific `collection` has been selected via the Data Hub search feature.
  2. We are searching for the complement to the input selected under condition 1 (i.e., LULC selected and now searching for Biophysical Table, or Biophysical Table selected and now searching for LULC).
  3. The model is AWY, NDR, or SDR (because these models use exactly the same `id`s for their LULC and Biophysical Table inputs. Eventually all models will be supported via a more robust, likely modelspec-based, solution.) 

Many Workbench tests are failing due to TypeScript integration. Identifying the magic combination of configuration options is at the top of my to-do list!

There are also a handful of lower-stakes to-dos I am planning to address after this PR, during/after internal testing.

### No AOI, cannot search
dh_search-no_aoi-2026_09_11

### Query generated from context
dh_search-auto_query-2026_09_11

### Search results
dh_search-results-2026_09_11

### Biophysical Table search results, constrained to the "ESA CCI LULC" collection because an LULC from that collection has been selected
dh_search-results-collection_constrained-2026_09_11

### No results
dh_search-no_results-2026_09_11

### Error (mocked by using a local server configured to return an error, or by disconnecting from the internet)
dh_search-error-2026_09_11

## Checklist


src/natcap/invest/coastal_vulnerability/coastal_vulnerability.py
```diff
@@ -234,6 +234,7 @@ def get_vector_colnames(vector_path):
         ["population_raster_path", "population_radius"],
         ["slr_vector_path", "slr_field"]
     ],
+    aoi_input_id="aoi_vector_path",
```
davemfish, line 237: This model is an interesting case where some inputs (e.g. habitat distributions, wind/wave data) do not necessarily need to intersect with the AOI extent, as the AOI is just used to define the section of coastline to analyze. `aoi_extent + fetch distance` or `aoi_extent + habitat_search_radius` is what the model uses in practice to extract data from other layers. In 99% of cases I'm sure the AOI will be sufficient for searching the DataHub though. I'm not proposing any changes.

src/natcap/invest/delineateit/delineateit.py
```diff
@@ -35,6 +35,7 @@
         ["dem_path", "detect_pour_points", "outlet_vector_path", "skip_invalid_geometry"],
         ["snap_points", "flow_threshold", "snap_distance"]
     ],
+    aoi_input_id="outlet_vector_path",
```
davemfish: I think this input will commonly be a point dataset, which often might only include a single outlet point. In that case does it serve the purpose? It's bounding box will intersect with a DEM, so there will be DataHub results. But looking at the model, I think it always does the flow routing for the entire DEM. I don't think this model has any inputs that constrain the extent of analysis. Maybe we should omit this model for now and make a separate issue for supporting it?

src/natcap/invest/forest_carbon_edge_effect/forest_carbon_edge_effect.py
```diff
@@ -53,6 +53,7 @@
         ],
         ["aoi_vector_path"]
     ],
+    aoi_input_id="aoi_vector_path",
```
davemfish: This model also does not seem to use this vector to limit the processing extent, since it is an optional input. The same is probably true for the crop production models.

src/natcap/invest/habitat_quality/habitat_quality.py
```diff
@@ -40,6 +40,7 @@
         ["threats_table_path", "access_vector_path",
          "sensitivity_table_path", "half_saturation_constant"]
     ],
+    aoi_input_id="access_vector_path",
```
davemfish: This layer does not constrain the model's analysis extent. I'll make an issue for all these. I'll keep flagging them here as I go so we can keep track.

src/natcap/invest/hra/hra.py
```diff
@@ -142,6 +142,7 @@ def archive_for_datastack(self, value, datastack):
         ["n_overlapping_stressors"],
         ["visualize_outputs"]
     ],
+    aoi_input_id="aoi_vector_path",
```
davemfish: Does not constrain analysis extent.

src/natcap/invest/pollination/pollination.py
```diff
@@ -33,6 +33,7 @@
         ["landcover_raster_path", "landcover_biophysical_table_path"],
         ["guild_table_path", "farm_vector_path"]
     ],
+    aoi_input_id="farm_vector_path",
```
davemfish: Does not constrain the analysis extent.

src/natcap/invest/stormwater/stormwater.py
```diff
@@ -35,6 +35,7 @@
         ["adjust_retention_ratios", "retention_radius", "road_centerlines_path"],
         ["aggregate_areas_path", "replacement_cost"]
     ],
+    aoi_input_id="aggregate_areas_path",
```
davemfish: Does not constrain the analysis extent.

src/natcap/invest/urban_nature_access/urban_nature_access.py
```diff
@@ -73,6 +73,7 @@
         ],
         ["search_radius_mode", "decay_function", "search_radius"]
     ],
+    aoi_input_id="admin_boundaries_vector_path",
```
davemfish: Not all operations are constrained by this extent.

src/natcap/invest/spec.py
```diff
@@ -2591,6 +2603,19 @@ def check_inputs_in_field_order(self):
...
+    def check_aoi_input_id(self):
+        """Make sure ``aoi_input_id`` refers to a valid vector input."""
+        if self.aoi_input_id:
+            try:
+                aoi_input = self.get_input(self.aoi_input_id)
+            except KeyError:
+                raise KeyError(
+                    f'Invalid aoi_input_id. No input with id {self.aoi_input_id}.')
+            if not isinstance(aoi_input, VectorInput):
+                raise TypeError('aoi_input_id must refer to a VectorInput.')
```
davemfish: These exception types make total sense, but since this is a validator, [Pydantic suggests raising a `ValueError`](https://pydantic.dev/docs/validation/latest/concepts/validators/#raising-validation-errors) in order for the user to see a `PydanticValidationError`. Basically we're saying, "you chose a bad _value_ for `aoi_input_id`".

src/natcap/invest/spec.py
```diff
@@ -2691,6 +2716,8 @@ def fallback_serializer(obj):
 
         spec_dict = self.__dict__.copy()
         # rename 'inputs' to 'args' to stay consistent with the old api
+        # In addition, flatten each top-level input's ``keywords`` into a
+        # single list of strings, to facilitate Data Hub search.
```
davemfish: Is this comment still useful here now that serialization takes care of this?

workbench/src/renderer/components/SetupTab/ArgsForm/index.jsx
```diff
@@ -116,6 +127,100 @@ class ArgsForm extends React.Component {
...
+    this.setState({
+      ...this.state,
+      focusOnAoiRequested: false,
+      readyToFocusOnAoi: false,
+    });
+  };
+
+  getSiblingType = (argkey) => {
+    // These hard-coded values support AWY, NDR, and SDR.
+    // @TODO: add support for all search-enabled models (via model spec, probably).
```
davemfish, line 194: I know this is not intended to be the long-term solution. Just want to add that it's possible that these particular arg keys could theoretically be used in other models where they are _not_ part of a sibling relationship.

workbench/src/renderer/components/DataHubSearchModal/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+          return response.json();
+        })
+        .then(({ count, datasets }) => {
+          setNumSearchResults(count);
+          setSearchResults(datasets.map((d: DHALDataset) => transformDHALSearchResult(d)));
+          collapseAllCards();
+        })
+        .catch((error) => {
+          setSearchError(true);
+          console.error((error as Error).message);
```
davemfish: Maybe we should `logger.error` this so it makes it into the Workbench's logfile?

workbench/src/renderer/components/DataHubSearchModal/models.ts
```diff
@@ -0,0 +1,62 @@
...
+  collections: string[],
+  license: string,
+  author: string,
+  lastUpdated: string,
+  created: string,
+  datasetUrl: string,
+  webpageUrl: string,
+}
+
+export interface DHALDataset {
```
davemfish, line 16: If I understand, `DHALDataset` is the type returned by the DHAL, and `DataHubSearchResult` is the type the modal wants to render? They are very similar. Not every `DHALDataset` property will be displayed in the modal, but most will. I think we probably want to add `pixel_size` and `crs_units` to the `DataHubSearchResult` as well. Any consideration for using a single interface? Or extending one from the other?

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <dl className="search-params">
+      <dt>{t('Tags')}</dt>
+      <dd>{query.tags.join(', ')}</dd>
+      <dt>{t('Datatype')}</dt>
+      <dd>{query.datatype}</dd>
+      {
+        query.extent.length > 0 &&
+        <>
+          <dt>{t('Extent')}</dt>
+          <dd>{query.extent.join(', ')}</dd>
```
davemfish, line 18: It might be nice to tidy up the extent a bit. We generally don't need more than 6 points of precision on lon/lat coordinates -- that's already sub-meter scale at the equator! It would make the numbers easier to read, but we might also want to do that _before_ making the query. The DHAL layer has some bot-protection rules related to the length of query strings, and while we control those rules and can loosen them as-needed, shorter extent strings don't hurt.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/models.ts
```diff
@@ -0,0 +1,22 @@
+export interface DataHubSearchQuery {
+  tags: string[],
+  datatype: string,
+  extent: number[],
```
davemfish: Can we get a `// [xmin, ymin, xmax, ymax]` comment on extent please?

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
```diff
@@ -233,6 +239,71 @@ export default function ArgInput({
...
   }
 
+  // @TODO: ¿move search-related stuff to a separate module?
+  const [showSearchModal, setShowSearchModal] = useState(false);
+  const searchQuery = {
+    tags: argSpec.keywords || [],
+    datatype: argSpec.type,
+    extent: searchExtent,
+    collections: searchCollections,
+  };
```
davemfish, line 249: I'm guessing we can't use `DataHubSearchQuery` here without refactoring this module to typescript?

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/models.ts
```diff
@@ -0,0 +1,22 @@
...
+  collections: string[],
+}
+
+export class DHALSearchParams extends URLSearchParams {
+  constructor(workbenchQuery: DataHubSearchQuery) {
```
davemfish, line 9: If I understand correctly, this constructor is translating a query object defined in javascript to the `DataHubSearchQuery` interface? And the reason that query object was not declared as a `DataHubSearchQuery` is because it is defined in `ArgInput`, which doesn't yet support typescript? And if we didn't need to jump through those hoops we could probably skip this class and use
```javascript
const query = DataHubSearchQuery({...})
const params = new URLSearchParams(query)
```
?

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+            tags={query.tags}
+            datatype={query.datatype}
+            extent={query.extent}
+            collections={query.collections}
+          />
+        : <>
+            <div className="search-error">
+              <TbZoomCancel aria-label={t('Error')} className="error-icon" />
+              <span>
+                {t(`Before searching, you must specify a valid path for the following input:`)}
```
davemfish, line 41: Maybe it's a good idea if we tell people why they need to define this input first: the bounding box of this input will be used to filter results for datasets that intersect the bounding box.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <DataHubSearchParams
+        tags={query.tags}
+        datatype={query.datatype}
+        extent={query.extent}
+        collections={query.collections}
+      />
```
davemfish: I notice that the search params are displayed before and after search, but not during. Maybe we should leave them on the page while we display the spinner below them? I think reviewing the query terms while waiting for results could be nice and adds a bit of continuity through the sequence.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+      {
+        searchError
+        ? <>
+            <div className="search-error">
+              <TbZoomExclamation aria-label={t('Error')} className="error-icon" />
+              <span>
+                {t(`An error occurred. Please check your internet connection, then try again.
+                  If the problem persists, consider reporting it on the NatCap Community Forum.`)}
```
davemfish, line 104: Is there anything useful for us in the Workbench logs at this point? If so we could encourage people to upload their logfile when reporting the problem. We can use,
`import { handleClickFindLogfiles } from '../../menubar/handlers';` if so.

Or we perhaps we should consider rendering the error from the request so people can include it for us in their post.

davemfish (changes requested):
This looks great @emilyanndavis . I'm fully on board with the transition to Typescript. If the test configuration is a pain to figure out, perhaps we should seriously consider switching from `jest` to `vitest`. We're already using `vite` in our build pipeline and it supports Typescript with no extra configuration, which I believe would be true for the test environment as well. Migrating from jest to vitest would take some effort, but very doable. We would do that in a separate PR though I'm sure.

Once we get tests figured out, we'll probably want a few integration/unit tests for the new modal. That can also come in a future PR if you prefer.

Other than that I have mostly minor comments and questions. As we discussed, it probably makes sense to remove the AOI attribute from many models and them back incrementally as we address those models.

src/natcap/invest/coastal_vulnerability/coastal_vulnerability.py
davemfish: To clarify, I think this model _is ready_ to support `aoi_input_id`, unlike many of the others I commented on.

src/natcap/invest/spec.py
emilyanndavis: Ah, interesting, thanks!

src/natcap/invest/spec.py
emilyanndavis: Nope! I just forgot to remove it!

workbench/src/renderer/components/DataHubSearchModal/models.ts
emilyanndavis: They are similar, but I think it's worth defining two separate interfaces for a few reasons:
1. A few attribute names on the DHAL search result aren't completely intuitive. For example, `collection` actually holds an array of strings, not a single string. Calling it `collections` in the Workbench code makes it more immediately obvious how that bit of metadata should be handled. Similarly, in the Workbench context, I wouldn't be sure what a `source_catalog_url` is (even though I've no doubt it's very meaningful in CKAN-land), but I know what to do with a `webpageUrl`.
2. The DHAL search result doesn't have an `id`. For React purposes, it's convenient to refer to an obvious `id` property rather than having to generate one on the fly or point to some other attribute that may or may not truly be unique.
3. If/when the DHAL API changes in such a way that any of these attribute names change, we'll only have to update them once, in the models, rather than having to find and update all the references scattered across UI components.
4. It's nice to be able to use `camelCase` attribute names (rather than `snake_case`) when in JavaScript/TypeScript. Not the most important thing, I know, but it's a detail that can help keep one from getting confused on days when switching between programming languages is taxing.

I wouldn't want to extend one interface from the other, because a child interface retains all the attributes of its parent. In the case of renaming attributes (e.g., `source_catalog_url -> webpageUrl`), we'd end up with some redundant attributes (i.e., two names that map to the same value), and code could get confusing pretty quickly. It's best for the UI components to have a single, consistent name for each bit of metadata.

As for `pixel_size` and `crs_units`: yes, absolutely. I almost added these the other day, but I couldn't come up with any search results that had those fields populated, so I decided to wait until I could see some examples of exactly what those will look like so I'll know how to handle them in the UI.

emilyanndavis: All that said, I'm certainly open to renaming these interfaces to make it more immediately clear what they mean!

And/or: I could define `DataHubSearchResult` as a `class` whose constructor takes a `DHALDataset`, rather than using a standalone transformation function alongside two interfaces. I don't honestly remember why I decided to write that `transformDHALSearchResult` function instead of changing `DataHubSearchResult` to a `class`, but I don't think there's any reason it needs to be done that way. Do you have a preference?

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/index.tsx
emilyanndavis: I'd been meaning to ask about this, but I definitely wasn't thinking about the query string length angle! Thanks for the recommendation.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/models.ts
emilyanndavis: Definitely one of the most unusual call-and-response shoutouts I've heard, but yes!
🙌🏻 `[xmin, ymin, xmax, ymax]`!

workbench/src/renderer/components/SetupTab/ArgInput/index.jsx
emilyanndavis: Right. It has the shape of a `DataHubSearchQuery`, but we can't use a type annotation or benefit from type checking here since this is a `jsx` module, not a `tsx` module.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/models.ts
emilyanndavis: Not exactly. Since `DataHubSearchQuery` is an interface, not a class, there's no instantiation. Anything that has the shape defined by `DataHubSearchQuery` effectively _is_ a `DataHubSearchQuery` … even though only some modules (TS/TSX, not JS/JSX) are able to recognize it.

The `DHALSearchParams` is using a `DataHubSearchQuery` to create an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) with the specific params it needs. The `URLSearchParams` constructor takes no arguments, so we can't pass it that query object directly. But if we extend (i.e., subclass) `URLSearchParams`, we _can_ pass in the `DataHubSearchQuery` object and use it to create the needed params. We don't _need_ to use `URLSearchParams`, but the alternative would be creating the query string ourselves, and I'd rather not re-invent that particular (error-prone) wheel.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
emilyanndavis: Yeah, I like that idea. I guess I figured the actual search query wouldn't take long, so why show people anything they don't have time to read? But of course there are any number of reasons why the duration of a query will vary (including when the DHAL hasn't received a request recently and takes a few seconds to "wake up"). And if I do it right, the search params will appear constant, so even when the searching step is quick, you won't feel like you missed something. It'll probably make the markup simpler, too.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
emilyanndavis: Yes, good idea.

workbench/src/renderer/components/DataHubSearchModal/index.tsx
emilyanndavis: Yes, thanks for the recommendation!

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
emilyanndavis: I'm not sure rendering the error will be helpful since it's unlikely a user will be able to do anything to fix it. But I can certainly add more info to the logs. I'll have to think about whether a "find my logs" button seems right here. On one hand, people _should_ find instructions on the forum when they get there; on the other, repetition can be a good thing.

emilyanndavis:
## Updated screenshots

### "No AOI" error view (now has additional text explaining _why_ this input is required)
dh_search-no_aoi-2026_09_18

### Intro view (search params are now at the top, to match their location in subsequent views)
dh_search-intro-2026_09_18

### Searching view (now has search params)
dh_search-searching-2026_09_18

### Search results ("n results found" text is now a heading)
dh_search-results-2026_09_18

### "No results" view ("No results found" text is now a heading)
dh_search-no_results-2026_09_18

### Search error (now has a "Find my logs" button)
dh_search-search_error-2026_09_18

emilyanndavis:
Thanks for all the feedback, @davemfish! I believe I've addressed and/or responded to all your comments. I also posted some updated screenshots highlighting the changes.

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchModalViews/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+              <TbZoomCancel aria-label={t('Error')} className="error-icon" />
+              <p>
+                {t(`Before searching, you must specify a valid path for the following input:`)}
+                <strong className="aoi-input-name">{aoiInputName}</strong>
+              </p>
+            </div>
+            <p>
+              {t('When you search the Data Hub for spatial datasets (e.g., rasters), '
+                + 'search results will be filtered to include only datasets that intersect '
+                + `the bounding box of the ${aoiInputName} input.`)}
```
emilyanndavis: When I wrote this, it made sense to phrase it in terms of searching for spatial datasets (since the concept of a CSV _intersecting_ a bounding box doesn't make sense). But after confirming that a bounding box _can_ be used to filter CSV search results, I realize it's confusing to specifically mention spatial datasets here. I'll update this in the next round of changes!

emilyanndavis: Changed my mind—updating it now!

workbench/src/renderer/components/DataHubSearchModal/DataHubSearchParams/models.ts
davemfish: Okay, thanks for the explanation. I agree it's a good idea to use `URLSearchParams` to format the string for us. From the docs it does look like we can pass an object directly to it: `const params4 = new URLSearchParams({ foo: "1", bar: "2" });`

But it looks like there's still some manipulation we need to do for the `datatype` and `'sibling'`, so extending `URLSearchParams` makes sense to me.

HISTORY.rst
```diff
@@ -73,6 +73,14 @@ General
   input defines the output pixel size and projection.
   (`#2267 <https://github.com/natcap/invest/issues/2267>`_)
 
+
+Workbench
+=========
+* Some models now offer the option to search the Natural Capital Alliance Data
+  Hub for datasets you can use in InVEST without having to download them first.
```
davemfish, line 80: 🎉

davemfish (approved):
Thanks @emilyanndavis this looks great to me!

workbench/src/renderer/components/DataHubSearchModal/models.ts
davemfish: Thanks for the explanation; you make some really good points so I agree it will be nice to have the two separate interfaces. As for the function vs class, I generally reach for functions first as well and agree either way works just fine.

HISTORY.rst
emilyanndavis: Oops, you're right about `URLSearchParams`, thanks—I totally missed/forgot that constructor was overloaded. Still, in addition to the 'collections' -> 'sibling' and 'csv' -> 'table' translation, there's also the matter of needing multiple params with the same name (e.g., there are four params called `extent`, each with one of the four values that define the bounding box), so a plain object wouldn't work, and building the array of arrays to pass to the constructor could get messy fast!

Issue #2749, opened 2026-09-17
Automatically add more ModelSpec input attributes to the User Guide


claire-simpson:
## Background/Motivation
When generating the User's Guide, we currently only pull some of the info that is available in the ModelSpec, e.g., data type, `required`, `units`). There is other information available for each `Input` that would be helpful to automatically pull into the UG for a few reasons:

1. There are instances where we don't specify projection requirements for certain inputs anywhere in the UG, which is suboptimal. 
2. To better standardize our communication on the requirements/attributes of each input ( + if these helpful attributes are readily available, why not use them?)
3. In light of new `ModelSpec` attributes (related to DH-WB integration), to denote which inputs are the default projection target, default pixel size target, and "AOI"

Previous reasoning for not including projection/units was to keep the UG simple, however now that we'll be removing many of the strict projection requirements, we'll only need to add this info for a limited number of inputs, which should reduce information overload.

## Description
I propose adding, at minimum, `projected`, `projection_units`, and whether an input is the `default_projection_id`,  `default_pixelsize_id`, or `aoi`. We could also consider adding `data_type`.

PR #2751, opened 2026-09-18
Add data type, projection requirements, and default projection and pixel size to list of spatial inputs in User's Guide

claire-simpson:
## Description
Automatically add more of the attributes for spatial input data to the User's Guide.

Current Example:
**Data Needs**
**workspace directory** etc.
**precipitation** (raster, units: **mm/year**, _required_)

Proposed Changes in this PR:
**Data Needs**
**workspace directory** etc.
**precipitation** (raster [data type: **float**], units: **mm/year**, projected [projection units: **meter**], _required_, **default target projection**)

Note: I did not add any info about whether an input is the "AOI" for searching the datahub, as we might release this feature after the target projection and pixel size feature. It also might be worth discussing if its actually helpful to tag the AOI as "required to search the datahub" (or redundant to tag these vectors as "AOI") or if this would just be confusing to see in the UG.

Fixes #2749

## Checklist


src/natcap/invest/rst_generator.py
```diff
@@ -89,6 +90,20 @@ def describe_input(module_name, keys):
...
+        model_spec = importlib.import_module(module_name).MODEL_SPEC
+        attributes = []
+
+        if _input.id == model_spec.default_projection_id:
+            attributes.append(f'**{gettext('default projection input')}**')
+        if _input.id == model_spec.default_pixelsize_id:
+            attributes.append(f'**{gettext('default pixel size input')}**')
+
+        if attributes:
+            rst = rst.replace("): ", f', {", ".join(attributes)}): ')
```
claire-simpson: This whole thing is kinda clunky. The issue with documenting whether an input is the default projection or pixel size input in `describe_rst` is that `describe_rst` only has context about the individual `Input` and _not_ the ModelSpec, and since `default_projection_id` and `default_pixelsize_id` are top-level attributes of `ModelSpec`, this info isn't available to the input's `describe_rst` call. Another design would be to have `describe_rst` optionally take the ModelSpec as an argument, which is perhaps a cleaner design than manipulating the RST after it's generated, but would require changing all these `describe_rst` calls. Thoughts?

claire-simpson: I changed my mind about this and am now passing the additional `default projection` and `default pixel size` attributes to `describe_rst` to avoid direct RST manipulation after its been constructed. All `describe_rst` functions now accept an `additional_attributes` arg (which doesn't do anything unless the input is a vector or single band raster..)

claire-simpson:
I initially added `data type` but ended up removing it because it added clutter, doesn't seem all that helpful to the average user, and I don't _think_ this is something we enforce for float rasters at least

src/natcap/invest/spec.py
```diff
@@ -881,15 +881,14 @@ def validate(self, filepath: str):
             if projection_warning:
                 return projection_warning
 
-    def describe_rst(self):
+    def describe_rst(self, additional_attributes=None):
```
dcdenu4, line 884: Should we add `additional_attributes` to the docstring? And should we also add it to every instance of this function?

dcdenu4 (changes requested):
Thanks @claire-simpson . I just had one question for you.

